PR #24, opened 2020-01-07
built out MV doc strings system

codepope:
All the words are now in one place and indexed and keyed. This is the minimal viable docstrings version. While coding it I noted that flags and examples will need to be taken care of. Also probably code blocks.


michaeldwan:
This looks good. I'd like to have the strings injected as constants during the build before we release it though. I can do that in this branch. 

Could you do the flyctl rename in a separate PR?

codepope:
Not sure how to separate out the two parts. Make a fresh branch, do it and push that?

codepope:
Rather than injecting constants all over the place (which IME is a nightmare because the touched surface just gets bigger). I'd suggest moving to generating a single flyctldocstrings const file with a map to the handles which keeps the get mechanism in place. There's examples and flags still to implement and that's probably easiest done with the KeyStrings struct being passed about.

michaeldwan:
You can probably undo the source change in another commit here and do it again in a separate branch. Then we can test things like release without blocking the docs change.

You're right about constants everywhere being gross. I was thinking pretty much what you described but wasn't clear at all.

PR #69, opened 2020-02-27
Simplified completions - Always to stdout

codepope:
Simplified completions - made it -c or --completion - always outputs to stdout.

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -2,8 +2,9 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/superfly/flyctl/docstrings"
 	"os"
+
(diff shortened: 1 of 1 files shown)


michaeldwan:
@codepope how about we make this a new `completions` command?

codepope:
I'm very aware of the limited namespace we have to play with with commands @michaeldwan - completions is already in the c-space with certs and config.

flyctl util completions maybe?

michaeldwan:
Good point! How do you feel about merging this to fix your issue and improve later?

codepope:
I feel good... </james brown>

PR #70, opened 2020-02-28
Add better errors

Flaque:
# Overview

When a user tries to deploy, but does not have an access token, prompt them to signup:
```
No api access token available. Please login with 'flyctl auth signup' or 'flyctl auth login'
```

When a user tries to deploy, but doesn't have an app, prompt them to create one:
```
No app specified. Maybe create one with 'flyctl apps create [APPNAME]'
```


michaeldwan:
This is awesome, thanks for submitting!

michaeldwan:
This is awesome, thanks for submitting!

Issue #83, opened 2020-03-19
Documentation for "flyctl scale vm" could do with more details


simonw:
I'm trying to figure out how to scale my instance up from the default `micro-1x`.

It looks like I need to use `flyctl scale vm` - but the docs at https://fly.io/docs/flyctl/scale-vm/ don't specify how the `SIZE` parameter should be used.

At a guess I feed it one of the "type" values on https://fly.io/docs/pricing/ - but it would be great if the documentation page for that command was expanded to include some examples.

simonw:
If the documentation was in a public GitHub repo I'd be happy to open pull requests against it.

mrkurt:
Those docs are actually generated from this project: https://github.com/superfly/flyctl/blob/master/helpgen/flyctlhelp.toml#L247

Right now it references `flyctl platform vm-sizes`, which might be new since you submitted this issue, but putting some sizes in the help directly is a good idea.

codepope:
More details now in PR #90

PR #90, opened 2020-03-25
Added size names text

codepope:
Added size names text and tuned up the text


simonw:
This is exactly the information I was missing in #83 - thanks!

PR #109, opened 2020-04-16
Improve CLI Messages

alrs:
I tried running my first deploy with a bad `fly.toml`.

The message I got back was:
```
==> Validating app configuration
Error App configuration is not valid
```

This PR changes the `docker` package to return failed configs along with an error so that `cmd` can print the errors.

With these changes, the `flyctl` binary prints:
```
==> Validating app configuration
    ✘ services.0.ports.0.port: must be one of: [80, 443, 5000, 10000..10100]
    ✘ services.0.ports.1.port: must be one of: [80, 443, 5000, 10000..10100]
Error App configuration is not valid```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -71,6 +71,9 @@ func runDeploy(cc *CmdContext) error {
 
 	parsedCfg, err := op.ValidateConfig()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


michaeldwan:
Thanks, this is a lovely improvement.

PR #122, opened 2020-05-12
Scale command rework

codepope:
This PR retires the autoscale command and introduces

flyctl scale balanced [min=x] [max=y]
flyctl scale standard [min=x] [max=y]
Sets mode and parameters

flyctl scale set [min=x] [max=y]
Sets parameters on current mode

flyctl scale show
Shows overview of scale including vm size name (but not details)

retained:
flyctl scale vm (shows vm or sets vm if size name give)

----

Also included in PR are additional comments to reduce linter warnings and a makefile which supports "all" which seems to make for more reliable builds.


mrkurt:
This looks nice. 👍

cmd/scale.go
```diff
@@ -24,9 +29,105 @@ func newScaleCommand() *Command {
...
+	newcfg.BalanceRegions = &balanceRegions
+	newcfg.MinCount = &currentcfg.MinCount
+	newcfg.MaxCount = &currentcfg.MaxCount
+
+	kvargs := make(map[string]string)
+
+	for _, pair := range ctx.Args {
+		parts := strings.SplitN(pair, "=", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("Secrets must be provided as NAME=VALUE pairs (%s is invalid)", pair)
```
michaeldwan, line 76: Error message mentions secrets

cmd/scale.go
```diff
@@ -24,9 +29,105 @@ func newScaleCommand() *Command {
...
+			return errors.New("could not parse min count value")
+		}
+		minintval := int(minint64val)
+		newcfg.MinCount = &minintval
+		delete(kvargs, "min")
+	}
+
+	maxval, found := kvargs["max"]
+
+	if found {
```
michaeldwan, line 98: It looks like the `min` and `max` args are keywords rather than dynamic user input, should they be flags instead like `--min=123`? Flags have a simpler api and auto generated error messages and help output.

cmd/scale.go
```diff
@@ -24,9 +29,105 @@ func newScaleCommand() *Command {
...
+	if len(kvargs) != 0 {
+		unusedkeys := ""
+		for k := range kvargs {
+			if unusedkeys == "" {
+				unusedkeys = k
+			} else {
+				unusedkeys = unusedkeys + ", " + k
+			}
+		}
+		return errors.New("unrecognised parameters in command:" + unusedkeys)
```
michaeldwan, line 118: You can reduce this block to `strings.Join(unusedKeys, ", ")`

cmd/command.go
```diff
@@ -21,8 +21,10 @@ import (
 	"github.com/superfly/flyctl/terminal"
 )
 
+// CmdRunFn - Run function for commands which takes a command context
 type CmdRunFn func(*CmdContext) error
 
+// Command - Wrapper for a cobra command
```
michaeldwan, line 27: Nice improvement!

cmd/scale.go
codepope: Good spot. Fixed.

cmd/scale.go
codepope: They are by design keywords so we can add in more over time and aren't nailed down to exposing every one as part of the docs and help. Here I'm going for more compactness for the user over ease of coding.

cmd/scale.go
codepope: Except that the source for the unusedKeys strings is coming from the map keys and strings.Join doesn't take a map and the process of making an array beforehand is just as wordy. There's some trickery that can be done with reflect but its not pretty. I've gone for simple in this case.

cmd/command.go
codepope: I was just getting bugged by all the yellow in vscode's view. :)

Issue #134, opened 2020-05-28
Ability to restart monitoring required (and option to go to monitoring automatically)


codepope:
flyctl deploy, secrets set and other commands start background processes which can be monitored and in the case of deploy are. But once detached it isn't possible to reattach to the monitoring process. Suggest flyctl monitor which recreates a monitoring state and where commands can monitor, offer the ability to do and detach or do and monitor.

mrkurt:
#56 is a start!

codepope:
Merged and working for most deployment events.

PR #138, opened 2020-06-01
Monitor Command (and assorted fixups)

codepope:


codepope:
Fixes #134

codepope:
Problems with Deepsource noted. Look to fix in future.

PR #170, opened 2020-06-25
flyctl will block v0 deployment Fixes #169

### LINKED ISSUE #169: If an app fails to deploy first time, flyctl will block deployment
A first time app deploy fail will leave the v0 app at dead in current HEAD version.

Initial fix will be to ignore the blocking code if version is 0.

Longer term fix? Create a real paused state?

codepope:


codepope:
I want to hold this off till we get state/status unified.

michaeldwan:
@codepope I fixed up state last night so it returns `suspended` instead of `dead` while paused. It still says `dead` for a few seconds after resuming while new VMs are started, but that needs to be fixed separately.

codepope:
Dropping this request

PR #173, opened 2020-06-25
change overwrite protection to be proactive

### LINKED ISSUE #168: fly apps create can throw away deployments
If you already have a fly.toml file in your directory...

And you run `fly apps create...`

At the end you are prompted that you will overwrite the fly.toml file.

If you say no, the file is not overwritten but also the newly created app is left in place with no record of its creation.

Suggest moving the fly.toml check to the start of the process to avoid creating forgettable orphans.

codepope:
Fixes #168


michaeldwan:
this is nice

PR #174, opened 2020-06-25
Better ErrNoAuthToken message

simonw:
I got this error and couldn't remember how to login - I think it's more useful if it reminds you to use `flyctl auth login`.


michaeldwan:
Thanks!

PR #176, opened 2020-06-26
Set pause, resume, restart to use requireAppNameAsArg

codepope:


codepope:
Merging as is as all the initiators are in the mix for the command revision and the duplication will serve as a code smell.

PR #178, opened 2020-06-26
Fix build on systems where sh is not bash

joshtriplett:
When invoking `make` to build flyctl, I got the following error:

scripts/generate_docs.sh: 6: [[: not found

It looks like `scripts/generate_docs.sh` expects to run with bash, but the
Makefile invokes it using `sh scripts/generate_docs.sh`. Change the
Makefile to use `bash`, and also add a bash shebang to the script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,4 +10,4 @@ build: generate
 
 cmddocs: generate
 	@echo Running Docs Generation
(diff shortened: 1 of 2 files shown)


michaeldwan:
@joshtriplett nice fix, thanks!

PR #186, opened 2020-07-02
Make Betas Great Again - Prerel Support fixes #184

### LINKED ISSUE #184: add support for flyctl prerelease versions
- [ ] Make flyctl upgrade checker ignore the prerelease versions*

This has to happen first, with a release, so that people don't get accidentally prompted

- [ ] Make bumpversion support tagging with prerelease
- [ ] Make install.sh support prerelease as a env setting or flag so by default it gets release, unless told to.

codepope:
Fixes #184

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -36,6 +36,8 @@ archives:
       windows: Windows
       386: i386
       amd64: x86_64
+release:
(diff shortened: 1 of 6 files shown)


codepope:
Merging this will need a change to the goreleaser yml too.

PR #199, opened 2020-07-03
Setting a secret on a suspended deploy doesn't return without intervention

### LINKED ISSUE #180: Setting a secret on a suspended deploy doesn't return without intervention
If you set a secret on an deployed app which is suspended, fly sets the secret and then goes to redeploy the app, but because it's suspended, no deployment happens and the monitor stays silent.

(Considering, if app suspended, to warn after the secret is set and not to drop into monitoring deployment)

codepope:
Fixes #180


codepope:
This applies to set and unset secrets. It now returns with an error when the app is suspended. This is not ideal, but stops the cli just wedging waiting for a deployment that never comes. I considered detaching but that just gives the user the impression it's happening anyway.

PR #202, opened 2020-07-06
Handle prerel and normal releases safely

codepope:


codepope:
PR'd because want to check on secrets handling before committing.

PR #227, opened 2020-07-11
cmd: Function Pruning

alrs:
This removes unused functions from the `cmd` package.


alrs:
I suspect that DeepSource flagging function removals as a performance issue can be considered a flake.

codepope:
Merged and came back around to fix that Deepsource flagging. Thanks for the PR.

PR #238, opened 2020-07-15
Add a --local-only flag to the flyctl deploy command

### LINKED ISSUE #237: Add a --local-only flag to the flyctl deploy command
... or something along those lines. Inspired by --remote-only :)

Why?

Well, I've been trying a deploy from CircleCI. And getting some odd random build issues. Not sure if it is their service, but it is a standard Linux base so it should work ok. In their terminal, what seems to happen (without using the --remote-only flag) is that it installs flyctl, sees that it can't do a local build, and so tries a remote one. Which you'd normally want it to, I guess:

```
 ...
Saving to: ‘flyctl_0.0.137_Linux_x86_64.tar.gz’
flyctl_0.0.137_Linu 100%[===================>]  13.06M  --.-KB/s    in 0.1s    
2020-07-14 20:03:54 (131 MB/s) - ‘flyctl_0.0.137_Linux_x86_64.tar.gz’ saved [13696695/13696695]
flyctl
Deploying name
==> Validating App Configuration
--> Validating App Configuration done
Services
TCP 443 ⇢ 8080 
Deploy source directory '/home/circleci/build_dev'
Docker daemon unavailable, performing remote build...
 
Creating build context... ⣾Creating build context... ⣽Creating build conte ...
```

... and so on for loads and load of 'building messages'.

BUT after a while, _this_ shows up:
```
Flyctl was installed successfully to /usr/local/bin/flyctl
 
Run 'flyctl --help' to get started
 
0.0.137
 
printenv
 
Deploying name
 
==> Validating App Configuration
 
Building ⢿Building ⡿Building ⣟Building ⣯Building ⣷Building ⣾Building ⣽Building ⣻Building ⢿Building ⡿Building ⣟--> Validating App Configuration done
 
Services
 
TCP 443 ⇢ 8080 

Deploy source directory '/src'
 
Docker daemon available, performing local build...
 
==> Building with Dockerfile
 
Using Dockerfile: /src/Dockerfile
...
 ```

Sooo ... it says "Docker daemon available, performing local build...". Hmm. So CircleCI _does_ provide docker.

And so not sure what is happening behind the scenes at this point, as there seems to be a battle in the console between 'Building;' messages (by the hundred) which are presumably all from the remote build, in the background, and then all the normal local messages you get when running docker locally, like

```
--> Building with Dockerfile done
... etc ...
==> Creating Release
Release v13 created
Monitoring Deployment
 ```

And then once it deploys, CircleCI ends. But in the background, is that remote build still happening? So a deploy (n+1) is happening after CircleCI ends? Not clear. Or is the remote build abandoned once it sees the local build worked?

So I wondered if a _--local-only_ flag could avoid this weirdness by stopping flyctl even starting a remote build (despite the fact it thinks it needs to). Just a thought.

codepope:
Fixes #237

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -45,6 +45,10 @@ func newDeployCommand() *Command {
 		Name:        "remote-only",
 		Description: "Perform builds remotely without using the local docker daemon",
 	})
+	cmd.AddBoolFlag(BoolFlagOpts{
+		Name:        "local-only",
+		Description: "Only perform builds locally using the local docker daemon",
+	})
 	cmd.AddStringFlag(StringFlagOpts{
 		Name:        "strategy",
 		Description: "The strategy for replacing running instances. Options are canary, rolling, or immediate. Default is canary",
@@ -166,8 +170,7 @@ func runDeploy(commandContext *cmdctx.CmdContext) error {
 		//fmt.Printf("Deploy source directory '%s'\n", cc.WorkingDir)
 		commandContext.Statusf("flyctl", cmdctx.SINFO, "Deploy source directory '%s'\n", commandContext.WorkingDir)
 
-		if op.DockerAvailable() {
-			//fmt.Println("Docker daemon available, performing local build...")
+		if op.DockerAvailable() && !op.RemoteOnly() {
 			commandContext.Status("flyctl", cmdctx.SDETAIL, "Docker daemon available, performing local build...")
 
 			if commandContext.AppConfig.HasBuilder() {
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -207,8 +210,18 @@ func runDeploy(commandContext *cmdctx.CmdContext) error {
...
+			if !op.DockerAvailable() {
+				if op.LocalOnly() {
+					return fmt.Errorf("Docker daemon unavailable: Local-only set so cannot use to remote build")
+				}
+				commandContext.Status("flyctl", cmdctx.SINFO, "Docker daemon unavailable: Performing remote build...")
+			} else {
+				if op.RemoteOnly() {
+					commandContext.Status("flyctl", cmdctx.SINFO, "Remote-only set: performing remote build...")
+				} else {
+					commandContext.Status("flyctl", cmdctx.SINFO, "Docker daemon available: Still performing remote build...")
```
michaeldwan, line 222: I think this line is unreachable since docker available and remote only false would use the branch on line 173.

codepope: I pondered that and considered that it was possible if the remote/local flags lost their exclusion line in the deploy op. The fact that it doesn't flag unreachable leaves me tending to leave it there.

PR #264, opened 2020-08-12
Builtin Image

codepope:
Adds --image to the init command

Also adds image="" to the build section

At deploy time will do an image deploy using the tag *and* the config.

Also fixes up permissions on config file.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -127,14 +127,15 @@ func runDeploy(commandContext *cmdctx.CmdContext) error {
 	var image *docker.Image
 
 	configimageRef, _ := commandContext.Config.GetString("image")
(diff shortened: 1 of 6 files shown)


cmd/deploy.go
```diff
@@ -126,7 +126,18 @@ func runDeploy(commandContext *cmdctx.CmdContext) error {
 
 	var image *docker.Image
 
-	if imageRef, _ := commandContext.Config.GetString("image"); imageRef != "" {
+	configimageRef, _ := commandContext.Config.GetString("image")
+	appimageRef := commandContext.AppConfig.Build.Image
```
michaeldwan, line 130: `Build` might be a null pointer here if there's no config file or if there's no build section

codepope: On it.

PR #275, opened 2020-08-25
fix missing alloc restart counts

michaeldwan:


codepope (approved):
LGTM, structure is in place to take the data, so I'm assumng it was just zeroed through the CLI

PR #309, opened 2020-09-28
cmd/status: fix dropped error

alrs:
This fixes a dropped error in the `cmd` package.


alrs:
I have no opinion on DeepSource, other than the "bug" it is interested in has not much to do with this fix.

codepope:
Thanks Lars!

alrs:
That was quick, thanks!

Issue #75, opened 2020-03-08
Support for custom (public) docker registries


pinkiebell:
Adding support for other public docker registries would be super nice.
Trying to deploy images from `docker.pkg.github.com/...` currently returns `Error Unable to access image docker.pkg.github.com/<PACKAGE:TAG>: 401 Unauthorized`.

I guess the reason is that fly's API tries to fetch from docker hub?

Related #58

michaeldwan:
Correct we’re only using docker hub or our registry. We’ll add support for other registries soon!

pinkiebell:
Gentle ping. Since the `docker.pkg.github.com` requires authentication for public images, the new github container registry does not. Any plans on supporting the `ghcr.io` on fly.io?

I tried do deploy a test application via a public ghcr image but it hangs on the 
```
==> Optimizing Image
building fs... ⡿ 
```
step for ever.

PS:
🙌🍾 for this simplistic docker hosting platform. Great Job 👍
Do you guys consider making use of `Container on the Edge' as a marketing twist? 🤣

michaeldwan:
@pinkiebell thanks for the kind words 😊

We're going to support private registries but no timeframe yet. Do you mind sharing the image that's hanging?

pinkiebell:
> @pinkiebell thanks for the kind words 😊
> 
> We're going to support private registries but no timeframe yet. Do you mind sharing the image that's hanging?

<img width="870" alt="Screenshot 2020-10-01 at 21 14 39" src="https://user-images.githubusercontent.com/40266861/94853121-2f622c80-042b-11eb-9493-1dd68ed9987a.png">

michaeldwan:
@pinkiebell Looks like there's a bug in our code. That particular image targets several architectures and we weren't choosing the correct one so it was hanging. We'll get a fix out shortly. Otherwise public images on ghcr.io will work just fine.

michaeldwan:
@pinkiebell we just pushed a fix for image resolution -- ghcr.io/nutberry/artifacts/habitat:latest is working for me. Sorry about that!

pinkiebell:
> @pinkiebell we just pushed a fix for image resolution -- ghcr.io/nutberry/artifacts/habitat:latest is working for me. Sorry about that!

👌Perfect, works.  Btw, I just stumbled upon another bug:
I can't delete a Organization:

```
~ # flyctl orgs delete test
Error Field 'dnsZones' doesn't exist on type 'Organization'
```

[...]

PR #313, opened 2020-10-02
Fix api error when deleting an organization

michaeldwan:
fixes the issue mentioned in https://github.com/superfly/flyctl/issues/75#issuecomment-702433633


codepope (approved):
Looks good, nice deletions.

PR #332, opened 2020-11-25
Do not use black text for titles

fornwall:
Black does not work well for terminals with black backgrounds, and removing black makes the behaviour consistent with how presenter.go renders titles in [renderTable()](https://github.com/superfly/flyctl/blob/master/cmd/presenters/presenter.go#L49) and [renderFieldList()](https://github.com/superfly/flyctl/blob/master/cmd/presenters/presenter.go#L88).

Output from running `flyctl regions list` before change:
before

Output from running `flyctl regions list` after change:
after


michaeldwan:
Awesome, thanks for the fix! 

We added some credit to your fly account for contributing. 🙌

PR #335, opened 2020-12-02
Fixing up build.args etc for import and deploy

codepope:


cmd/deploy.go
```diff
@@ -144,6 +144,13 @@ func runDeploy(commandContext *cmdctx.CmdContext) error {
 	} else {
 		// no image specified, build one
 		buildArgs := map[string]string{}
+
+		if commandContext.AppConfig.Build.Args != nil {
```
michaeldwan, line 148: You might need to check that `commandContext.AppConfig.Build` isn't nil first. Try it out with an app without a `[build]` section

flyctl/app_config.go
```diff
@@ -25,7 +25,7 @@ const (
 type AppConfig struct {
 	AppName string
 	Build   *Build
-
+	// DockerBuildArgs map[string]string
```
michaeldwan, line 28: just nuke it 😉

michaeldwan (changes requested):
Looks good, but I _think_ you'll need a nil check

cmd/deploy.go
codepope: Ran it without a [build] and it worked, will add in the check tho. I suspect that the toml library fills in for ancestors... ie, build.args infers build as a section. So if there's a build.x, there will be an build to hold the .x in the tree. Our marshall to Toml code ends up putting in a build too which gets added to the .toml file.

So I think it's ok.

flyctl/app_config.go
codepope: Oh dash, nuked now. Must get into the habit of labelling kipple comments.

cmd/deploy.go
codepope: Well, I was so wrong there. Blows up when there's no build or build subsections at all. So... fix incoming.

PR #352, opened 2021-01-06
fly orgs list exits with unknown error

### LINKED ISSUE #351: fly orgs list exits with unknown error
No other diagnostics

Repeating the query on the playground returns server error and notices of deprecation.

codepope:
Fixes #351


codepope:
Just want to check that the backend changes were intended and this fly tweak is appropriate...

PR #358, opened 2021-01-14
context passed to checkErr fix

codepope:
Passes context which now has a Interactive field which is set by an isatty.isTerminal call on StdOut....


codepope:
Not needed for now.

PR #368, opened 2021-03-02
user-mode TCP/IP WireGuard for SSH connections

tqbf:
This is a thing we do now.


tqbf:
With `flyctl ssh console` I'm comfortable that this is sort of useable now; we'll get the exact commands right later. Still need to support `flyctl ssh exec ls` to just run `ls`.

PR #377, opened 2021-03-19
send [build] section in config with deploy

mrkurt:
This is a tiny change to send the original config contents with a deploy, instead of sending the `validateConfig` response.


mrkurt:
This is no longer needed.

PR #392, opened 2021-04-06
Fix instructions for installing a specific version

mtlynch:
The URL was incorrect, and the version tag in the README no longer matches flyctl's version tags.


michaeldwan:
Thank you!

PR #393, opened 2021-04-07
Remove debug(?) log line about fd pointers.

connyay:
Not terribly useful or helpful.


connyay:
this one also isn't very helpful https://github.com/superfly/flyctl/blob/23465faf/internal/build/imgsrc/dockerfile_builder.go#L186

tqbf:
Thank you for this!

PR #398, opened 2021-04-12
don't send everyone to singapore

tqbf:


connyay:
Just curious, was this removed due to a bug in TimeRegions? Or strange ping results that were leaning towards singapore for some reason?

I hope this isn't annoying. I am interested in this project and would like to follow along a bit if @superfly doesn't mind.

tqbf:
The TimeRegions thing used a weird thing that Kurt had deployed to do timing tests. It was already there in the flyctl codebase, and we didn't want people to have to explicitly pick which regions to WireGuard too if they didn't have to, so it was like, chocolate, meet peanut butter.

But it turned out that was pretty silly. Our API already has a notion of what the closest data center is --- it's how we pick which region to deploy the first instance of your apps in (when you haven't picked one). So instead of this weird dynamic pinging thing --- which was, as Kurt pointed out to me, sending everyone to Singapore for reasons I don't understand --- we just co-opted the logic we were using for app placement; now, if you make the GraphQL API call that allocates a new peer, and don't specify a region, we pick one for you.

connyay:
Awesome! I appreciate you taking the time :)

PR #402, opened 2021-04-25
Add Network field to CreateAppInput

mtlynch:
The `Network` field is an optional field in the GraphQL schema for `CreateAppInput`, but it's not present in the Go struct, so this change adds it.


mtlynch:
I'm using `github.com/superfly/flyctl/api` instead of rolling my own graphql client from scratch, though I'm guessing this is not an officially supported way for clients to integrate with fly.

I'm not sure how you want to expose the `network` field in your Go API, but I figure that as long as I'm adding fields to make it work in my app, I'd see if contributing the changes back upstream is useful to you.

michaeldwan:
Thank you!

PR #411, opened 2021-05-05
Remote builders over wireguard

michaeldwan:
- remote builder connects directly to dockerd via an on-demand wireguard tunnel, just like `ssh console`, bypassing the problematic proxy
- move `wireGuardCreate` & `wireGuardForOrg` from `cmd` to `internal/wireguard` so docker code can reach it
- automatically pick the closest gateway region and auto-generate a peer name instead of prompting if region or name are blank


jeromegn (approved):
This looks like it should work fine!

I'll defer to @tqbf for the wireguard changes, but they just look like a reorganization.

tqbf (approved):
This is much better than the previous organization and cmd/wireguard.go was definitely sprawling.

PR #414, opened 2021-05-11
flyname: cache os.Executable during init

azazeal:
This PR refactors the way the `flyname` package caches the executable's name while at the same time making `flyname.Name` safe for concurrent use.

It additionally eliminates the need for calling `Name` in [main](https://github.com/superfly/flyctl/blob/4652ebd3cb70634b173e242b449f366433f55ed1/main.go#L20) as the package will now panic during the initialization process.


michaeldwan:
Thank you!

PR #416, opened 2021-06-01
Add app runtime environment command

rugwirobaker:
This PR adds a new config subcommand `flyctl config env` that displays
an app's environment variables as seen by the app at runtime.

Example:
```sh
$ flyctl config env --app database
NAME           TYPE     VALUE
REPL_PASSWORD  Secret   REDACTED
SU_PASSWORD    Secret   REDACTED
PRIMARY_REGION Variable lhr
```

Notes:
* It redacts the values of secrets since they are not available throug the API.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/superfly/flyctl/cmd/presenters"
 	"github.com/superfly/flyctl/cmdctx"
 	"github.com/superfly/flyctl/internal/client"
 
@@ -29,6 +30,9 @@ func newConfigCommand(client *client.Client) *Command {
 
 	configValidateStrings := docstrings.Get("config.validate")
 	BuildCommandKS(cmd, runValidateConfig, configValidateStrings, client, requireSession, requireAppName)
+
+	configEnvStrings := docstrings.Get("config.env")
+	BuildCommandKS(cmd, runEnvConfig, configEnvStrings, client, requireSession, requireAppName)
 
 	return cmd
 }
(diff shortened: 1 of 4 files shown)


mrkurt:
For secrets, it would be better to show digests. So output like:

```text
Secrets
----------------------------
NAME                       DIGEST
REPL_PASSWORD   0f4ffc260149b14a7a12ef28165ecb35
SU_PASSWORD.      3164a7aa2afcacc868c15aec009ecf1f

Environment Variables
----------------------------
NAME                        VALUE
PRIMARY_REGION    lhr
```

Digests don't tell us much, but it's at least easy to see what changed. In fact, you could just use the exactly template from `fly secrets list` with this command to simplify things.

```
> fly secrets list
NAME          DIGEST                           DATE
REPL_PASSWORD 0f4ffc260149b14a7a12ef28165ecb35 2021-02-03T23:35:33Z
SU_PASSWORD   3164a7aa2afcacc868c15aec009ecf1f 2021-02-03T23:35:32Z
```

cmd/config.go
```diff
@@ -102,6 +106,30 @@ func runValidateConfig(commandContext *cmdctx.CmdContext) error {
...
+	if err != nil {
+		return err
+	}
+
+	cfg, err := ctx.Client.API().GetConfig(ctx.AppName)
+	if err != nil {
+		return err
+	}
+
+	vars := cfg.Definition["env"].(map[string]interface{})
```
michaeldwan, line 120: You should make sure `cfg.Definition` is not nil, then make sure it has an `env` key, then make sure the env key is a `map[string]string` to prevent a panic on bad data.

cmd/config.go
```diff
@@ -102,6 +106,30 @@ func runValidateConfig(commandContext *cmdctx.CmdContext) error {
...
+	// if !ok {
+	// 	return fmt.Errorf("can not parse environment variables")
+	// }
```
michaeldwan, line 123: drop the commented out code

michaeldwan (changes requested):
Great start!

rugwirobaker:
When using the Frender function how do you skip a section if there no values to be printed. For example if an does not have any secrets.

PR #418, opened 2021-06-01
Add app runtime environment command

rugwirobaker:
Renaming the branch(had a typo) removed the PR entirely.  

 This PR adds a new config subcommand `flyctl config env` that displays
    an app's runtime environment variables as seen by the app at runtime.
    
    Example:
    ```sh
    $ flyctl config env --app database
    NAME           TYPE     VALUE
    REPL_PASSWORD  Secret   REDACTED
    SU_PASSWORD    Secret   REDACTED
    PRIMARY_REGION Variable lhr
    ```
    
    Notes:
    * It redacts the values of secrets since they are not available through the API.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -112,14 +112,16 @@ func runEnvConfig(ctx *cmdctx.CmdContext) error {
 		return err
 	}
 
-	cfg, err := ctx.Client.API().GetConfig(ctx.AppName)
-	if err != nil {
-		return err
+	if len(secrets) > 0 {
+		err = ctx.Frender(cmdctx.PresenterOption{Presentable: &presenters.Secrets{Secrets: secrets},
+			Title: "Secrets",
+		})
+		if err != nil {
+			return err
+		}
 	}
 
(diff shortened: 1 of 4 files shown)


rugwirobaker:
I have made the changes you requested  @michaeldwan, @mrkurt:

    * Separate secrets from environment variables.
    * Double check `Definition` and `env` existence.

michaeldwan:
@rugwirobaker no worries about typos :)

I was replying to your message from the other PR:

> When using the Frender function how do you skip a section if there no values to be printed. For example if an does not have any secrets.

Honestly presenters are probably more trouble than they are worth. We have a helper that does roughly the same thing with less ceremony. Just declare a table and add rows with columns as needed. Here's an example: https://github.com/superfly/flyctl/blob/master/cmd/checks.go#L58-L64.

You can skip rendering by checking for `len(secrets) == 0`, printing a message, and returning

rugwirobaker:
> You can skip rendering by checking for `len(secrets) == 0`, printing a message, and returning

I will deal with this tomorrow.

jsierles:
:100:

PR #423, opened 2021-06-05
Add proxy command

### LINKED ISSUE #420: Proxy Command
This command opens a proxy from a local port to a vm port over WireGuard.

```
flyctl proxy <port[:port]>
```

Like other commands that operate on an app, this accepts an app name, either from `fly.toml` or the `--app` flag. Before connecting, the app should be fetched from the api to make sure the user has access and that it’s running. Once we know the app is accessible, use the app name to create the private network address, `app-name.internal`.

The command also requires the the local and remote ports as the first argument. Similar to Docker commands, you can specify port mapping in a long `5432:2345` or short `5432` format. In the short form, the same port is used on both ends. In the long form, the first port will be opened on the local host, the second will be the remote port.

You’ll open a WireGuard tunnel similar to what we do for [remote builders](https://github.com/superfly/flyctl/blob/master/internal/build/imgsrc/docker.go/#L188-L210). In that example, we verify the app, open a tunnel, then use the tunnel’s dial context to create an http client for the docker client.

rugwirobaker:
This PR adds a new command `flyctl proxy <port[:port]>` that proxies
from a local port to a vm port over WireGuard.

This PR is still very raw:

- [x] Needs help on context management and graceful exits. 
- [x] Forgoes instance selection(for now).

Closes #420


tqbf:
So, this is great, but you probably want to base this PR on the `wireguard-agent` branch, and move some of the logic from the command to `pkg/agent/agent.go` (and `client.go`).

The reason for that is, you can only run one command that uses a given WireGuard connection at a time; if you run this proxy command right now, and, while the proxy is running, try to run `flyctl ssh console`, the `ssh` command will steal the WireGuard connection from the proxy, which will abruptly stop working.

The agent solves this problem by spawning a background process (you don't have to worry about how; you can just add new commands to the agent) that maintains a single WireGuard connection, but lets multiple runs of `flyctl` use that connection.

tqbf:
So, you'd want to add a function like this:

https://github.com/superfly/flyctl/blob/9d518cbfb7bbfca1e1a1e88ee4a7071457092007/pkg/agent/agent.go#L198

(notice that there's a helper, `tunnelFor`, which looks up the tunnel for a given WireGuard connection; you don't have to do the whole `StateForOrg` dance)

You want to add your helper to this map here:

https://github.com/superfly/flyctl/blob/9d518cbfb7bbfca1e1a1e88ee4a7071457092007/pkg/agent/agent.go#L55

... and now it's a part of the agent protocol.

Finally, you want to add it to the agent interface, which a function in `client.go` like this:

https://github.com/superfly/flyctl/blob/9d518cbfb7bbfca1e1a1e88ee4a7071457092007/pkg/agent/client.go#L125

Then, in `flyctl/cmd`, your command handler can call `EstablishFlyAgent` to get a `*Client` that it can use to talk to the agent (that function will spawn a new agent process if it's not already running; to debug your agent code, you can run it explicitly by running `flyctl agent daemon-start`; `EstablishFlyAgent` will talk to that agent if it's running.)

tqbf:
*Alternatively*: you can call `Client.DialContext` (see `client.go`) to get a function you can use to connect over WireGuard, using the agent, without modifying the agent protocol --- your proxy will still run entirely as a `flyctl` command, in the main `flyctl` process, but it'll use the agent to get new connections.

mrkurt:
@rugwirobaker how is this going? We have a couple of folks that might benefit it, so we can make a nice post when it's ready to merge.

rugwirobaker:
@tqbf @mrkurt  I'm about to move the proxy into the agent.

rugwirobaker:
@mrkurt @tqbf @michaeldwan I have moved the proxy inside the agent but we still have to implement it for no-agent and add a command to stop it when it's running in the background.

rugwirobaker:
Note that the proxy keeps serving existing TCP connection after `flyctl proxy stop` until the client stops them.

michaeldwan (changes requested):
@rugwirobaker I just tried to run this and there's a compiler from changes to master.

I do think it would be wise to do as Thomas [suggested](https://github.com/superfly/flyctl/pull/423#issuecomment-855334384) and use a connection from the agent (just like the nats client) but run the actual proxy in the flyctl command. 

Running a proxy in the agent is a much more complicated ux since we need a stop command and some form of status so you can see what is actually being proxied, which this PR doesn't have yet. It also doesn't provide an easy way to see if the proxy is working since logs are hidden. We'd also need two implementations for the no-agent case. We're still working through reliability issues in the agent and I'd hate to add more to the pile. 

We can avoid most of that complexity by running the proxy in the flyctl command. When `flyctl proxy` starts, we get a connection from the agent and start the proxy. Output can be printed right to STDOUT so it's clear if it's working and what it's doing. Then when the command gets a SIGINT the proxy shuts down. 

Thoughts?

helpgen/flyctlhelp.toml
```diff
@@ -1025,3 +1025,8 @@ usage = "kill <id>"
 longHelp = "Remove a Fly machine"
 shortHelp = "Remove a Fly machine"
 usage = "remove <id>"
+
+[proxy]
+usage     = "proxy <local:remote>"
+shortHelp = "Commands creates a proxy to a fly app"
```
michaeldwan, line 1031: This message isn't clear. How about "Proxy connections to a fly app"?

pkg/agent/agent.go
```diff
@@ -31,6 +31,7 @@ var (
 type Server struct {
 	listener      *net.UnixListener
 	tunnels       map[string]*wg.Tunnel
+	proxy         net.Listener
```
michaeldwan, line 34: is this used?

michaeldwan (changes requested):
This looks and works great! Once the comments are addressed we can ship it.

michaeldwan:
@rugwirobaker almost 😄

Get rid of the `"Commands"` prefix on the help text string here since this one doesn't have any subcommands: https://github.com/superfly/flyctl/pull/423/commits/85de633bba900bad92d1f0627789003f38e79b58#diff-cd1f05056de04fa9945ef72198cde82f3c3d5132915b724b062f8a5ab86602a9R1031-R1032

michaeldwan (approved):
🎉

Issue #415, opened 2021-05-27
"fly ssh establish <arg>" errors


mrkurt:
```
$ fly ssh establish myorg
Oops, something went wrong! Could you try that again?
```

Works fine with no argument!

PR #439, opened 2021-06-21
fly ssh establish <arg> errors #415

bkochendorfer:
If we want to override and provide the organization name the length of args will be `2` instead of `1`.

Solves https://github.com/superfly/flyctl/issues/415

```bash
user@fun python-hellofly-flask]$ ~/fly-fun/flyctl/bin/flyctl ssh establish override
Error Could not resolve Organization
[user@fun python-hellofly-flask]$ ~/fly-fun/flyctl/bin/flyctl ssh establish personal override
Establishing SSH CA cert for organization personal
New organization root certificate:
ssh-ed25519-cert-v01@openssh.com AAAA....
.....
```

### DIFF
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -181,7 +181,7 @@ func runSSHEstablish(ctx *cmdctx.CmdContext) error {
 	}
 
 	override := false
-	if len(ctx.Args) >= 1 && ctx.Args[1] == "override" {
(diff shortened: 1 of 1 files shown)


tqbf:
Thanks for this!

rugwirobaker:
Shouldn't be merged @tqbf ?

Issue #446, opened 2021-07-03
Check for wireguard org state continues to pass after removing all wireguard connections


marcus-crane:
**Describe the bug**
One possible state for the user to get into (I have never gotten myself into this state accidentally *cough*) is one where they've deleted all of their wireguard connections. Most functionality continues to work as `~/.fly/config.yml` contains a valid access token that can be used to query the graphql endpoint.

However, if a user then tries to run `fly ssh console`, well, here's an image. Starting a new session shouldn't matter so much but the top session is the first one and the bottom is the second/newer.

![Screen Shot 2021-07-03 at 3 47 09 PM](https://user-images.githubusercontent.com/14816406/124342291-00c1ee00-dc17-11eb-98f3-1bcb9de02484.png)

What appears to be going on here is that the call to [StateForOrg](https://github.com/superfly/flyctl/blob/539d4cff129a083c8482ad3c1a4b10b290faae45/internal/wireguard/wg.go#L51) sees that there is a syntactically valid configuration and so, it just simply returns the existing config. Of course, what it should probably do is verify that the connection still exists.

Deleting `~/.fly/config.yml` fixes this issue pretty easily (as the next call to StateForOrg regenerates the `wire_guard_state` block) but that also nukes your access token and forces you to log back in so just deleting the wireguard state block is quicker.

I don't imagine this is related but I've also noticed the initial connection after creating a new peer is a little flaky?

```
fly ssh console -a gunslinger
Creating WireGuard peer "interactive-amaterasu-marcus-utf9k-net-601" in region "ord" for organization personal
Connecting to gunslinger.internal...⢿ Error connect to SSH server: dial: lookup gunslinger.internal. on fdaa:0:2d9f::3: read udp [fdaa:0:2d9f:a7b:bea:0:a:200]:16722: i/o timeout
 fly ssh console -a gunslinger
Connecting to gunslinger.internal... complete
```

It usually takes me two or three connections before the connection establishes so I imagine there's some sync delay between talking with graphql and it propogating whereever it needs to.

**Add `fly.toml`**
Doesn't apply as I'm running with the `-a` flag

**Command output:**
Maybe a bit redundant after the explainer 😛

marcus-crane:
Submitted an attempt at fixing this bug

michaeldwan:
Fixed in #483

PR #447, opened 2021-07-06
Add check to see if saved wireguard peer still exists

marcus-crane:
Closes: #446 

While I'm by no means a Go expert, I've done a bit during the last year and the issue seems small enough to fix myself.

As I understand it, `StateForOrg` is currently only called upon establishing a connection to a builder instance and upon trying to establish an SSH console. Both of which are instances where the primary action will already fail if the user isn't connected to the internet in the first place since my first concern was that this might add unnecessary overhead to all commands.

Some ways you might want to improve this:

* Add support to query a singular Wireguard Peer via GraphQL. Currently this pulls all peers and as you can imagine, fetching hundreds or thousands just to look up one isn't very fun.
* Add a Peer ID since as far as I can tell, there isn't a way to uniquely identify a peer? I chose the Peer IP since it seems like the most unique thing. The name could be arbitrarily changed in future so that's probably not good to use. I can also imagine deleting a Wireguard Peer and then waking up after 1000 years to find the Peer IP has been reassigned to my org but now the saved keys don't match the connection. Actually, perhaps checking the pubkey would be more ideal.
* Perhaps this should be checked before trying to connect in the first place? It seems the "connecting" message renders and then any checks happen in the middle which causes some wonky output as seen in the attached screenshots.

Beyond that, any feedback is appreciated. I didn't actually know Go has GOTOs as a feature so that was cool to find out.

The attached screenshots are how the affected commands perform after and before

Screen Shot 2021-07-06 at 7 59 47 PM
Screen Shot 2021-07-06 at 8 18 17 PM


marcus-crane:
@michaeldwan @tqbf Any chance of a review? 👀

michaeldwan:
I'm sorry, I don't know how we missed this. We ended up doing something similar in #483 but for all peers at once with the help of a new api call. We needed to validate peers more frequently so the agent can disconnect when one is deleted to fix #467 and #464.

marcus-crane:
Oh, that's no problem. I did raise it a few times and Thomas said he had it on his list but admittedly, even I forgot about this PR! Glad it's fixed now

PR #471, opened 2021-07-22
Postgres connect

davissp14:
Postgres connect allows you to quickly connect to a specified Postgres cluster.  

```
Connect to a postgres cluster

Usage:
  flyctl postgres connect [flags]

Flags:
  -a, --app string        App name to operate on
  -c, --config string     Path to an app config file or directory containing one (default "./fly.toml")
      --database string   The postgres database to connect to
  -h, --help              help for connect
      --password string   The postgres user password
      --user string       The postgres user to connect with

Global Flags:
      --verbose               verbose output
 ```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/gofrs/flock v0.7.3
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/hashicorp/go-multierror v1.1.1
+	github.com/hashicorp/go-version v1.3.0 // indirect
 	github.com/heroku/heroku-go/v5 v5.4.0
 	github.com/inancgumus/screen v0.0.0-20190314163918-06e984b86ed3
 	github.com/jpillora/backoff v1.0.0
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by rugwirobaker** [comment]
May I suggest a `user` flag? I may not want always connect with user `postgres`

**Comment by rugwirobaker** [comment]
Never mind already exists

**Comment by jsierles** [comment]
Should this be revisited or dropped in favor of new work?

**Comment by davissp14** [comment]
Now that SSH is more reliable, I should be able to add this pretty quickly.  I have it working for machines, so just a matter of creating a binary within the postgres-ha/postgres-standalone repos.

**Comment by davissp14** [comment]
This should be ready to go. 

I also added image version verification, which may be worth taking a glance at: https://github.com/superfly/flyctl/blob/864f5f5d7cb2d30112bcb325e12e46d1342c7113/cmd/postgres.go#L620-L653

Screen Shot 2021-12-09 at 11 37 53 PM

File: go.mod [hunk 1]
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/gofrs/flock v0.7.3
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/hashicorp/go-multierror v1.1.1
+	github.com/hashicorp/go-version v1.3.0 // indirect
```
rugwirobaker, line 21: This should be direct

rugwirobaker, line 21: This should be direct

rugwirobaker (changes requested):

PR #483, opened 2021-08-10
prune invalid wireguard peers from the config

### LINKED ISSUE #467: Deleting some WireGuard peers permanently confuses the agent
If `flyctl` has created a WireGuard connection for SSH (or whatever), on the fly, that peer will show up in the `flyctl wireguard list` list. That's good.

But if a user deletes that WireGuard connection, we delete it serverside, but not from `~/.fly/config`. The agent will think the connection is still valid, even though it isn't. 

We should make `flyctl wireguard delete` also delete the matching entry from `~/.fly/config`. But this only addresses the most obvious version of this problem, because if you delete that peer from a different host, `flyctl` will never know. 

So the other thing `flyctl` should probably do is, when a WireGuard connection times out, pull the list of WireGuard peers from the API and make sure it's still valid.

michaeldwan:
Fixes #467

### DIFF
--- a/api/resource_wireguard.go
+++ b/api/resource_wireguard.go
@@ -169,3 +169,24 @@ func (c *Client) ClosestWireguardGatewayRegion() (*Region, error) {
 
 	return data.NearestRegion, nil
 }
+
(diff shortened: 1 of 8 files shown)


tqbf (approved):
this all looks totally sane

PR #484, opened 2021-08-10
Use the agent for WireGuard tunnels to remote builders

michaeldwan:


tqbf (approved):
This is good; the move of Establish from cmd/ to agent/ is good. 

Is there a reason we use errors.Wrapf instead of just %w? (It doesn't matter for this PR).

michaeldwan:
Wrap preserves the original error which I was passing to `errors.Is` to check for a timeout in some uncommitted code. I just use it by habit, but if we aren't unwrapping the error or checking it `fmt.Errorf` is fine.

PR #485, opened 2021-08-10
[WIP] Close invalid tunnels without restarting agent

michaeldwan:
@rugwirobaker This stubs out what (_I think_) we can do to remove invalid tunnels from the agent without restarting. When we detect the WireGuard state has changed in the handler, run `validateTunnels` on the server to close any invalid tunnels. New tunnels will be created as needed after that.


michaeldwan:
Moved to #487

PR #487, opened 2021-08-11
Detect WireGuard peer changes in the agent

### LINKED ISSUE #488: The agent should detect peers removed without changing the config file
#483 added support for pruning invalid peers from the config, and fixing #464 will make sure the agent detects config file changes. There's still the case where peers are removed without changing the config file, such as from the API. To handle that case, the agent should periodically validate peers, maybe every 1-5 minutes. A simple/naive way to do this might be a goroutine that runs [`wireguard.PruneInvalidPeers`](https://github.com/superfly/flyctl/blob/master/internal/wireguard/wg.go#L154) (slow, removes peers from the config after checking the API) and [`server.validateTunnels`](https://github.com/superfly/flyctl/pull/485/commits/18216494d4302ba6d41a499b8ca519ef4cfa31c3#diff-18bf746206c8ac217eb16ffb0cf91a6e676e05a7b517dc70aef0260d0871028fR396) (fast, closes tunnels not in the config) on a ticker.

rugwirobaker:
- [x] cache last wireguard state change on agent start up.`
- [x] check for any change on every `agent.handle` call.
- [x] what if `config.yml` has been manually modified?
 
Fixes #488

### DIFF
--- a/flyctl/config.go
+++ b/flyctl/config.go
@@ -15,8 +15,8 @@ const (
 	ConfigInstaller       = "installer"
 	BuildKitNodeID        = "buildkit_node_id"
 
-	ConfigWireGuardState          = "wire_guard_state"
-	ConfigWireGuardStateTimestamp = "wire_guard_state_timestamp"
+	ConfigWireGuardState = "wire_guard_state"
+	// ConfigWireGuardStateTimestamp = "wire_guard_state_timestamp"
 
 	ConfigRegistryHost = "registry_host"
 )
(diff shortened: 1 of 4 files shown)


michaeldwan (changes requested):
This is working great, nice job. Could you remove the commented out parts? After that it's good to merge.

michaeldwan:
Awesome thanks!

michaeldwan (approved):
This is working great. It's cleanly removing peers deleted from the api and from the config. Good work!

PR #491, opened 2021-08-13
Dns timeout fixes

### LINKED ISSUE #472: First WireGuard+SSH Connection Lag Isn't Handled
When you connect to an app with `flyctl ssh` for the first time, `flyctl` needs to ask the API to provision a new WireGuard account. This happens quickly in the API, but it can be somewhat slow (right now, ~20-30 seconds) to propagate to the actual gateways. 

The UX for when this happens is very bad: the first (and sometimes second) time you connect, you get a DNS timeout (because the WireGuard connection isn't live), with no explanation. I'm sure it looks plain broken to a lot of users.

There's always going to be _some_ lag in setting up a new WireGuard connection, so `flyctl` should account for this. My thought is that `flyctl` should try pinging the DNS server when it sets up a tunnel, and not proceed until it gets a response, and print a status line for the user so they know we're waiting for config to propagate.

This should only be happening the first time you use a WireGuard connection. Unfortunately, for the business, that's probably the most important WireGuard connection users make, first impressions and all.

michaeldwan:
- switch resolver to tcp to fix probe responses
- add context to agent client functions, handle cancellation and timeouts
- remove retries in client probe function, return `ErrProbeFailed` so callers can retry
- added `WaitForTunnel` function to client which will keep probing until it gets a valid response or cancellation error
- remote builder connections now wait on a tunnel for 4 minutes in parallel before attempting to connect with Docker
- validate peers in `agent.Establish` before connecting to agent

Fixes #472

### DIFF
--- a/cmd/fly_agent.go
+++ b/cmd/fly_agent.go
@@ -65,7 +65,7 @@ func runFlyAgentStart(cc *cmdctx.CmdContext) error {
 		c.Kill(ctx)
 	}
 
-	_, err = agent.Establish(ctx, api, true)
+	_, err = agent.Establish(ctx, api)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "can't start agent: %s", err)
 	}
--- a/cmd/fly_agent.go
+++ b/cmd/fly_agent.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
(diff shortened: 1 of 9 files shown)


tqbf (approved):
This all looks good to me.

PR #492, opened 2021-08-14
Add distinct name for WireGuard connections generated by the agent

### LINKED ISSUE #465: flyctl ssh console WireGuard connections should have distinct names
`StateForOrg` is the function in `internal/wireguard` that generates a new WireGuard connection. It takes a "name" parameter, but if you leave it blank, it generates a sane name, like "interactive-YOUR_MACHINE_NAME-482".

This is fine, except that it means that the names people give to host-native WireGuard connections (via `flyctl wireguard create`) are indistinguishable from the ones we generate on the fly to do SSH.

We should pass a distinctive name to the connections the agent creates (they'll still need to be random or unique in some way) to make these problems easier to debug.

amalkh5:
Fixes #465 

it will add the word "agent" to WireGuard connections generated by the agent, for example:
``` shell 
#  host-native WireGuard connections
$ interactive-amals-MacBook-amalkms5-gmail-com-724    

# agent WireGuard connections
$ agent-interactive-amals-MacBook-amalkms5-gmail-com-714
```

### DIFF
--- a/internal/wireguard/wg.go
+++ b/internal/wireguard/wg.go
@@ -24,6 +24,23 @@ var (
 	cleanDNSPattern = regexp.MustCompile(`[^a-zA-Z0-9\\-]`)
 )
 
+func gernateConnectionnName(apiClient *api.Client, ctx context.Context) (string, error) {
+	user, err := apiClient.GetCurrentUser(ctx)
+	if err != nil {
+		return "", err
+	}
+	emailSlug := cleanDNSPattern.ReplaceAllString(user.Email, "-")
+
+	host, err := os.Hostname()
+	if err != nil {
+		return "", err
+	}
+	hostSlug := cleanDNSPattern.ReplaceAllString(strings.Split(host, ".")[0], "-")
+
+	name := fmt.Sprintf("%s-%s-%d", hostSlug, emailSlug, badrand.Intn(1000)) // skipcq: GSC-G404
+	return name, nil
+}
+
 func StateForOrg(apiClient *api.Client, org *api.Organization, regionCode string, name string) (*wg.WireGuardState, error) {
 	state, err := getWireGuardStateForOrg(org.Slug)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


amalkh5:
As for the failed checks, it was regreding existing code and I do not know if I should change it or keep it?

rugwirobaker (changes requested):
Sorry for the late review. Everything looks except course the little deep source complaints. You can change those.

amalkh5:
@rugwirobaker sorry I clicked on review by miskate, I will let you know when I fix the issues

amalkh5:
@rugwirobaker 
I fixed all things suggested in deepsource except the one where it requests to change `math/rand` to `crypto/rand` because I [read](https://yourbasic.org/golang/crypto-rand-int/) that using `crypto/rand. Reader` returns an error if the underlying system call fails.
 For instance, if it can't read /dev/urandom on a Unix system, or if CryptAcquireContext fails on a Windows system.

What do you think?

jsierles:
Hey, would you be interested in finishing this up?

amalkh5:
@jsierles sure, I will resolve the conflict and push it.

amalkh5:
@jsierles  Fixed, I removed "interactive" and added the word "agent" to WireGuard connections generated and by the agent, for example:


```
# host-native WireGuard connections
$ interactive-amals-MacBook-xxx-amalxxxx-gmail-com-860    

# agent WireGuard connections
$ agent-amals-MacBook-xxx-amalxxxx-gmail-com-534
```

internal/wireguard/wg.go
```diff
@@ -24,6 +24,23 @@ var (
 	cleanDNSPattern = regexp.MustCompile(`[^a-zA-Z0-9\\-]`)
 )
 
+func gernateConnectionnName(apiClient *api.Client, ctx context.Context) (string, error) {
```
jsierles, line 27: This should be named correctly, perhaps with a better name like `generatePeerName`

jsierles: Also, the context should be the first argument passed in.

jsierles (changes requested):

amalkh5:
@jsierles fixed

jsierles:
Thanks! Actually we need to keep the name `interactive-`. Can you reverse it so we have something like `interactive-agent-*`?

jsierles (commented):
Looks good except for the comment about keeping the `interactive-*` prefix.

PR #493, opened 2021-08-17
remove redundant defaults in flag descriptions

technillogue:
`flyctl volumes --help` lists defaults twice:

Create new volume for app. --region flag must be included to specify
region the volume exists in. --size flag is optional, defaults to 10,
sets the size as the number of gigabytes the volume will consume.

Usage:
  flyctl volumes create <volumename> [flags]

Flags:
  -a, --app string      App name to operate on
  -c, --config string   Path to an app config file or directory containing one (default "./fly.toml")
      --encrypted       Encrypt the volume **(default: true)** (default true)
  -h, --help            help for create
      --region string   Set region for new volume
      --size int        Size of volume in gigabytes**, default 10GB** (default 10)

Global Flags:
  -v, --verbose               verbose output

### DIFF
--- a/cmd/volumes.go
+++ b/cmd/volumes.go
@@ -35,13 +35,13 @@ func newVolumesCommand(client *client.Client) *Command {
 
 	createCmd.AddIntFlag(IntFlagOpts{
 		Name:        "size",
-		Description: "Size of volume in gigabytes, default 10GB",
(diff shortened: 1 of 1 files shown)


michaeldwan (approved):
Thank you!

PR #496, opened 2021-08-17
Improve agent error handling

michaeldwan:
This is a little refactor of the client code to better catch and propagate errors to the caller. The functionality is the same, but `Client` now references an interface which is conditionally set to be an agent client or non-agent client depending on build flags. `Client` handles errors and adds functions to test a tunnel or host and leaves the rest to the individual implementations. 

highlights:
- refactor agent & non-agent clients so shared logic & error handling is in `agent.Client` and the platform dependent code is behind an interface
- map string errors (eg `i/o timeout` or `no such host`) to types to make detecting tunnel and resolution faults easier. Propagate stack and context for everything else
- new `WaitForTunnel` and `WaitForHost` functions on client that blocks until the tunnel is healthy and the host is reachable
- call `WaitForTunnel` and `WaitForHost` before connecting to a vm for ssh or remote build
- move sentry error capturing to the flyctl side of the client since errors happen regularly (and are handled with the new error types). Should get better errors and stack traces in sentry with wrapped errors

### DIFF
--- a/api/resource_remote_builders.go
+++ b/api/resource_remote_builders.go
@@ -7,6 +7,9 @@ func (client *Client) EnsureRemoteBuilderForApp(appName string) (string, *App, e
 				url,
 				app {
 					name
+					organization {
+						slug
+					}
 				}
 			}
 		}
--- a/cmd/fly_agent.go
+++ b/cmd/fly_agent.go
@@ -44,7 +44,7 @@ func newAgentCommand(client *client.Client) *Command {
 }
 
 func runFlyAgentDaemonStart(ctx *cmdctx.CmdContext) error {
-	agent, err := agent.DefaultServer(ctx)
+	agent, err := agent.DefaultServer(ctx.Client.API())
 	if err != nil {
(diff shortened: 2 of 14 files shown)


rugwirobaker (commented):
Looks good to go. Great work.

rugwirobaker (approved):
Looks good to go🎉

michaeldwan:
@tqbf since you last looked...
- I tested on Windows and found a LOT of issues, mostly because [custom dialers on net.Resolver don't work on windows ](https://github.com/golang/go/issues/33086)
- I fixed those issues with 2 new functions on tunnel `LookupTXT` and `LookupAAAA` which uses [meikg/dns](https://github.com/miekg/dns). probe and resolve now use those functions instead of the tunnel's resolver
- swapped the DNS lookup in `fetchInstances` with the new functions to fix instance selection on Windows
- fixed a few issues with error propagation and cancellation that I found testing on Mac and some windows specific.
- added better output to `ssh console` when waiting on the tunnel or the host

PR #499, opened 2021-08-18
Update dns api types

### LINKED ISSUE #497: Update DNS graphql queries with new names
The API is using `DNS` instead of `Dns` now. Update the type names in queries to match.

For example, [this query](https://github.com/superfly/flyctl/blob/master/api/resource_dns.go#L44) should use the `ExportDNSZoneInput ` input type instead of `ExportDnsZoneInput `. Probably a bunch more.

rugwirobaker:
Fixes #497﻿


michaeldwan (approved):
thank you!

PR #503, opened 2021-08-20
Display os and arch in build output

michaeldwan:
Changed the build output to help debugging, particularly on M1

```
==> Building image with Docker (darwin arm64)
```

### DIFF
--- a/internal/build/imgsrc/buildpacks_builder.go
+++ b/internal/build/imgsrc/buildpacks_builder.go
@@ -50,7 +50,7 @@ func (s *buildpacksBuilder) Run(ctx context.Context, dockerFactory *dockerClient
 		terminal.Debug("failed to fetch docker server info %s", err)
 	}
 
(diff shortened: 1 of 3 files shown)


rugwirobaker (commented):
Shouldn't this display the architecture of the machine docker is running on? Is it by design that it displays my local architecture even though I'm using the remote builder? 
```sh
 ./flyctl deploy --remote-only
Deploying flyscale
==> Validating app configuration
--> Validating app configuration done
Remote builder fly-builder-summer-surf-4014 ready
==> Creating build context
--> Creating build context done
==> Building image with Docker (darwin arm64)
Sending build context to Docker daemon  1.049MB
```

michaeldwan:
I just needed a way to see what flyctl was running on, but I like your suggestion better. Would you be up a new PR that prints the server version, os, and arch from docker?

Maybe something like this?

```
==> Building image with Docker
docker host: 20.10.8 linux x86_64
```

rugwirobaker (approved):

PR #507, opened 2021-08-25
Create volume/pg from snapshot

rugwirobaker:
Backups are restored by creating a new volume with a snapshot id. 
Postgres clusters are restored by creating a new cluster with a snapshot id. 
This PR add a snapshot id flag to flyctl volumes create and flyctl pg create.

The flag value just needs to be added to the mutation input if present. 
﻿


jsierles:
Closing this in favor of a PR that works on the new CLI volumes command.

PR #513, opened 2021-08-30
add flag to auto-generate the app name when running apps create

### LINKED ISSUE #510: Flag for auto generating the app name
**Describe the bug**
As discussed in the [Forum](https://community.fly.io/t/flyctl-apps-create-name-auto/2347), for integration purposes, it would be useful to have a flag to auto-generate the application name when running `flyctl apps create`, to bypass the confirmation prompt. Example:

```
fly apps create --name-auto
```

Alternatively, the auto-generated name can be the default option, so if you don't pass the app name as a parameter, it will be auto generated.

amalkh5:
Fixes #510 

Add flag to auto-generate the app name when running `flyctl apps create`

```
$ flyctl  apps   create  --generatename
***An existing configuration file has been found.
? Overwrite file '/Users/fly.toml' Yes

? Select organization: org2 (org2)

? Select builder: go
    Go Builtin
Builtins use port 8080
New app created
  Name         = cold-haze-2358
  Organization = org2
  Version      = 0
  Status       =
  Hostname     = <empty>

App will initially deploy to cdg (Paris, France) region

Wrote config file fly.toml
```

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -57,6 +57,11 @@ func newAppsCommand(client *client.Client) *Command {
 		Description: "Never write a fly.toml file",
 	})
 
+	create.AddBoolFlag(BoolFlagOpts{
+		Name:        "generatename",
+		Description: "Always generate a name for the app",
+	})
+
 	appsDestroyStrings := docstrings.Get("apps.destroy")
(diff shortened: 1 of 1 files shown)


jsierles:
Thanks for this! It was ultimately implemented in #541 where we gave it a better name of `generate-name`.

PR #519, opened 2021-09-06
Stream logs from nats

rugwirobaker:
Try and stream log from nats and fallback to polling case of an error.

- [] Streaming from nats  currently blocks forever.
- [] Need to make a better unified format.
- [] Handle errors


mrkurt:
1. this will be amazing
2. will you make it require a flag like `--stream` to use at first? It'll be nice to get it in a release (when it's close) to let people try it vs current log polling.

rugwirobaker:
> 1. this will be amazing
> 2. will you make it require a flag like `--stream` to use at first? It'll be nice to get it in a release (when it's close) to let people try it vs current log polling.

Will do

cmd/logs.go
```diff
@@ -27,12 +29,39 @@ func newLogsCommand(client *client.Client) *Command {
...
-		VMID:       ctx.Config.GetString("instance"),
-		RegionCode: ctx.Config.GetString("region"),
-	})
+func runLogs(cc *cmdctx.CmdContext) error {
+	ctx := createCancellableContext()
+
+	client := cc.Client.API()
+
+	opts := &logs.LogOptions{
+		AppName:    cc.Config.GetString("app"),
```
jeromegn, line 38: You might want to use `cc.AppName` instead, since the app name might be coming from `fly.toml`.

pkg/logs/nats.go
```diff
@@ -0,0 +1,174 @@
...
+	}()
+
+	return out
+}
+
+func (s *natsLogStream) Err() error {
+	return s.err
+}
+
+func newNatsClient(ctx context.Context, dialer agent.Dialer, app *api.App) (*nats.Conn, error) {
```
michaeldwan, line 120: Just opened a PR that'll let you get the peerip without parsing the config, which will likely have race conditions with the agent creating peers but this query coming from flyctl

https://github.com/superfly/flyctl/pull/525

pkg/logs/nats.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+type natsDialer struct {
+	agent.Dialer
+	ctx context.Context
+}
+
+func (d *natsDialer) Dial(network, address string) (net.Conn, error) {
+	return d.Dialer.DialContext(d.ctx, network, address)
+}
```
michaeldwan, line 161: Just curious, do you know if this works without creating a custom dialer with a context? I _think_ it would work with the normal dialer since context cancellations would happen higher up causing the cli process to end.

cmd/monitor.go
```diff
@@ -112,8 +112,8 @@ func monitorDeployment(ctx context.Context, commandContext *cmdctx.CmdContext) e
 				}
 
 				commandContext.Status("monitor", cmdctx.STITLE, "Recent Logs")
-				logPresenter := presenters.LogPresenter{HideAllocID: true, HideRegion: true, RemoveNewlines: true}
-				logPresenter.FPrint(commandContext.Out, commandContext.OutputJSON(), alloc.RecentLogs)
+				// logPresenter := presenters.LogPresenter{HideAllocID: true, HideRegion: true, RemoveNewlines: true}
+				// logPresenter.FPrint(commandContext.Out, commandContext.OutputJSON(), alloc.RecentLogs)
```
michaeldwan, line 116: remove commented out lines when you're done 🧹

michaeldwan (changes requested):

pkg/logs/nats.go
rugwirobaker: I'm gonna try that

michaeldwan:
Oh forgot to mention in the review, timestamps don't have the same number of decimal points so the output shifts around a bit. We don't need nanosecond precision so let's truncate that to milliseconds in the formatted output. I made a todo in the PR body for this

rugwirobaker:
I was wondering where that shift was coming from

pkg/logs/nats.go
rugwirobaker: @michaeldwan the `nats.SetCustomDialer()` option only accepts a Dialer that does not take a context.

michaeldwan: okay then! :)

michaeldwan:
@rugwirobaker this looks good, there's 2 issues we need to figure out before releasing

1. I'm seeing an authentication error today, might need to ask @steveberryman if he knows what's up
```
error creating nats connection: could not connect to nats: nats: Authorization Violation
```

2. the current behavior shows the most recent ~25 lines while the nats stream only shows new logs which gives no feedback if the app isn't busy. Perhaps we can have it start the polling logger while the nats logger is connecting, and if successful stop the polling logger and let the nats stream take over. We don't need to spend time blending the logs, just showing something from before the nats stream starts is good enough.

Aside from those issues this is looking fantastic.

rugwirobaker:
@michaeldwan I'm seeing the same error too.

michaeldwan:
@rugwirobaker there was discussion in Slack about this but I can't find it now. Ask Steve in the morning, he'll know what's up 😄

michaeldwan (approved):
This is awesome, nice work!

PR #526, opened 2021-09-10
Stop agent on logout

### LINKED ISSUE #517: Stop the agent when signing out
The agent will continue to run with the previous credentials cached in memory.

rugwirobaker:
Closes #517

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -13,7 +12,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/superfly/flyctl/docstrings"
 	"github.com/superfly/flyctl/internal/client"
-	"github.com/superfly/flyctl/internal/flyerr"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/briandowns/spinner"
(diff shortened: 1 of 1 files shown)


cmd/auth.go
```diff
@@ -214,7 +217,31 @@ func runInteractiveLogin(ctx *cmdctx.CmdContext) error {
...
+	agentclient, err := agent.Establish(ctx, cc.Client.API())
+	if err != nil {
+		captureError(err)
+		return errors.Wrap(err, "can't establish agent")
+	}
+
+	if err := agentclient.Kill(ctx); err != nil {
+		captureError(err)
+		return errors.Wrap(err, "can't kill agent")
+	}
```
michaeldwan, line 243: `agent.Establish` would _start_ an agent if it isn't already running. There's some code to stop a running agent based on the pid file we write here: https://github.com/superfly/flyctl/blob/launch-redwood/pkg/agent/agent.go/#L127-L138

It's okay to ignore these errors since it's a cleanup process.

cmd/auth.go
```diff
@@ -214,7 +217,31 @@ func runInteractiveLogin(ctx *cmdctx.CmdContext) error {
 	return flyctl.SaveConfig()
 }
 
-func runLogout(ctx *cmdctx.CmdContext) error {
+func runLogout(cc *cmdctx.CmdContext) error {
```
michaeldwan, line 220: 🙌

michaeldwan (changes requested):

cmd/auth.go
rugwirobaker: Makes sense

michaeldwan (approved):

PR #527, opened 2021-09-15
Fly launch: generate fly.toml and a Dockerfile for RedwoodJS apps

jsierles:
RedwoodJS has some special deployment requirements that, for now, are most easily satisfied with a Dockerfile. It also benefits from having `statics` preconfigured since the entire clientside app is static.

This PR adds:

*   A way to generate files from templates embedded in `flyctl`
*   A way to add statics to `fly.toml` via `fly launch`
*   A Redwood-specific app detector

TODO:

*   \[x\] Fix generated TOML to format with underscores

### DIFF
--- a/internal/sourcecode/scan.go
+++ b/internal/sourcecode/scan.go
@@ -1,11 +1,17 @@
 package sourcecode
 
 import (
+	"embed"
+	"io/fs"
 	"os"
 	"path/filepath"
 
+	"github.com/pkg/errors"
 	"github.com/superfly/flyctl/helpers"
 )
+
+//go:embed templates/**
+var content embed.FS
 
 type SourceInfo struct {
 	Family         string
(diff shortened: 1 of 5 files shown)


jsierles:
@michaeldwan The code that adds statics here generates TOML without the underscores.

```
[[statics]]
  GuestPath = "/app/public"
  UrlPrefix = "/"
```

jsierles:
OK, the encoding issue was fixed.

internal/sourcecode/scan.go
```diff
@@ -134,7 +169,68 @@ func configureElixir(sourceDir string) (*SourceInfo, error) {
...
+
+	s := &SourceInfo{
+		Family: "Redwood",
+		Files:  templates("templates/redwood"),
+		Port:   8911,
+		Env: map[string]string{
+			"PORT": "8911",
+		},
+		Statics: []Static{
+			0: {
```
michaeldwan, line 194: don't need the array index with slice literals

michaeldwan (approved):
looks good

PR #529, opened 2021-09-17
Update static and hugo-static builtin builders

ananthb:
Use ananthb/thttpd-container instead of PierreZ/goStatic for a super minimal HTTP server container. thttpd static binary weighs in at around 180KiB. Removed httpsonly option and added maxage option.

Added hugoversion option to choose the hugo image that will be used. This allows choosing the extended edition by appending -ext to the version number.


jsierles:
Thanks for the PR. We're going to deprecate these builtins - along with `fly init` - in favor of new behavior allowing `fly launch` to write templates, like a Dockerfile. Once that is prepared, maybe we can revisit this?

ananthb:
Yeah makes sense @jsierles. It took me a while to figure out what container image was being used when I used a builtin. Having a template written to disk will be much more explicit.

jsierles:
As discussed, this will be revisited as soon as #530 is merged.

PR #530, opened 2021-09-20
Remove 'fly init' and make 'fly apps create' only register a new application

jsierles:
`fly launch` should be solely responsible for identifying and generating configuration for existing apps.

We may also want an option to create a 'blank' app without a template.

A docs PR should accompany the release of this PR.


vpn12out:
Why removed `fly init` ? It has become so much difficult to deploy app. With  `fly init` apps can be deployed easily because it auto creates `fly.toml` along with allowing us to select buildpacks. It was a great feature, please bring it back.

michaeldwan:
We removed it because it sat awkwardly between `flyctl launch` and `flyctl apps create`. If you have an existing app you want to deploy, `launch` should do the trick. Otherwise you can create an app and manually configure with `apps create`. Let us know if there's a feature `init` had that isn't covered by one of those and we'll try to get it sorted out.

vpn12out:
Ok, tell me how to select buildpacks ? As previously I used to select buildpacks using `fly init`

michaeldwan:
Launch will configure buildpacks automatically for several common frameworks and languages. What type of app are you trying to deploy?

vpn12:
Mostly I deploy nodejs and python apps. Launch automatically configures bulidpacks but it will better if we can select bulidpacks like in `fly init`. This was one of the features which has not been covered in `launch` and `app create`. So, pls add this feature or bring back `fly init`

PR #534, opened 2021-09-21
Add a Dockerfile-based Deno launcher with fly.toml processes and grep support

jsierles:
This PR adds a new Deno app launcher with a Dockerfile. The Deno launcher required a few new scanner features. They are:

*   Matching strings present in a given set of files in the source directory, since Deno apps have no filename conventions yet
*   Adding 'processes' to `fly.toml` (web, worker, etc), since developers should have a path to override the run command to point to their JS entrypoint

### DIFF
--- a/cmd/launch.go
+++ b/cmd/launch.go
@@ -179,6 +179,10 @@ func runLaunch(cmdctx *cmdctx.CmdContext) error {
 		if len(srcInfo.Statics) > 0 {
 			appConfig.SetStatics(srcInfo.Statics)
 		}
+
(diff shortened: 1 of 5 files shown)


rugwirobaker:
@jsierles  Everything in the build process works but you get a 502 undocumented error when you try to browse the app. Does it have anything to do with the processes?

jsierles:
I just tried this and the app opened fine. Will find you on Slack to discuss!

rugwirobaker (approved):

Issue #430, opened 2021-06-10
Add setting to always use remote builder, even when docker is running


mrkurt:
Docker can be problematic on m1 Macs, and there are some cases where it makes sense to have Docker running locally but use remote builds for Fly.

`fly deploy --remote-only` works just fine, but it would be nice to be able to configure a system to default to `--remote-only`. Can we have a `FLY_BUILD_REMOTE_ONLY` environment variable for this?

jsierles:
Why not make remote the default?

mik3y:
Other than the M1 issue, I've found a couple situations where I always prefer remote build: e.g. working from a plane, rural/crappy internet, etc.

Remote is preferable here because of Docker Stuff: while I might only have a few kb of code (docker context), the build itself might suck down gigs of layers to get there.

I'd be happy to submit a patch for this - could use a pointer or preference on approach, if it's a viable idea. Is exposing an env var the way to go..? (Is there an app-unspecific `.flyctrlrc` where this kind of preference could fit the bill..?)

kzys:
Resolving? It is default now.

PR #536, opened 2021-09-26
Prevent local builds on Apple M1 devices

### LINKED ISSUE #532: Force remote builder on m1/arm systems
Docker can't reliably build x86_64 images on arm systems because [qemu segfaults](https://community.fly.io/t/docker-build-deploy-fail-on-apple-m1/2590). We should not use the local Docker daemon to build on ARM systems.

jsierles:
Locals Docker builds aren't well supported on M1 devices, so don't allow them for now. Fixes #532.


rugwirobaker:
I believe this closes #430

PR #548, opened 2021-09-29
Separate API module from rest of flyctl

jeromegn:
This allows importing a pretty lightweight `github.com/superfly/flyctl/api` module in other projects.


gedw99:
This is great . Thanks

PR #551, opened 2021-09-30
Do not check for updates before starting agent daemon

### LINKED ISSUE #547: Make sure flyctl isn't checking for an update when launching the daemon
We've been seeing intermittent timeouts when starting the agent. It _should_ be fast since there's not network IO. But flyctl does check for updates periodically on start which might be slow in some cases (we need to make it not slow). It's possible an update check could slow the agent. We just need to make sure it's not happening. There's an env var that prevents it we can set when launching the agent process.

rugwirobaker:
Closes #547﻿

### DIFF
--- a/main.go
+++ b/main.go
@@ -157,6 +157,7 @@ func checkForUpdate(currentVersion semver.Version) (*update.Release, error) {
 	if !shouldCheckForUpdate() {
 		return nil, nil
 	}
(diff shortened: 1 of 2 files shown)


michaeldwan (approved):
Thanks!

rugwirobaker:
@michaeldwan  th only network call we are doing before starting the agent now is to valide the peers.

PR #566, opened 2021-10-15
POC WireGuard over WebSockets

tqbf:
This is pretty unserious, POC-only code. 

There's a semi-folkloric belief that some people are having trouble connecting to WireGuard because of UDP filters or stateful NATs or something like that. We could, in theory, detect endpoints that have trouble making UDP WG sessions, and then promote --- err, demote --- them to a WebSockets proxy for same.


pkg/wg/tunnel.go
```diff
@@ -25,6 +32,118 @@ type Tunnel struct {
...
+
+		for {
+			if ctx.Err() != nil {
+				return
+			}
+
+			var lbuf [4]byte
+			if !doRead(c, lbuf[:]) {
+				// we're broken here, kill the whole thing
+				continue
```
connyay, line 119: comment is confusing. was this meant to be a return?

tqbf: This is just code I'm goofing around with, if we ever do anything with it it'll look much different.

PR #567, opened 2021-10-16
main cleanup

azazeal:
This PR cleans up a few things, by:

- introducing a couple of new internal packages (`sentry` & `env`).
- introducing a top-level context which cancels on `os.Interrupt` & `os.Kill`.
- ensuring said context is passed to commands via `ExecuteContextC`.
- refactoring `checkForUpdate` to `update.PromptFor`.
- renaming `client.NewClient` to `New`.

### DIFF
--- a/doc/main.go
+++ b/doc/main.go
@@ -22,7 +22,7 @@ import (
 )
 
 func main() {
-	cc := client.NewClient()
+	cc := client.New()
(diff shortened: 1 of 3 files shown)


michaeldwan (approved):
This looks great! We can merge and do a prerelease so it goes through our integration tests.

PR #569, opened 2021-10-18
Support 'fly launch' with a Dockerfile for Remix apps

jsierles:
This PR adds scanner support for `fly launch` for Remix apps. To do this correctly, the PR introduces two new keys.

`SourceInfo.Docs` will be printed out at the end of a successful launch step. It's useful here since `fly deploy` won't work on an app that requires build arguments.

`SourceInfo.SkipDeploy` prevents `fly deploy` from running for the same reason.

### DIFF
--- a/cmd/launch.go
+++ b/cmd/launch.go
@@ -253,9 +253,15 @@ func runLaunch(cmdctx *cmdctx.CmdContext) error {
 		return nil
 	}
 
-	fmt.Println("Your app is ready. Deploy with `flyctl deploy`")
+	if srcInfo.SkipDeploy && srcInfo.Docs == "" {
+		fmt.Println("Your app is ready. Deploy with `flyctl deploy`")
+	} else if srcInfo.Docs != "" {
+		fmt.Println(srcInfo.Docs)
+	}
 
-	if !cmdctx.Config.GetBool("no-deploy") && (cmdctx.Config.GetBool("now") || confirm("Would you like to deploy now?")) {
+	if !cmdctx.Config.GetBool("no-deploy") &&
+		!srcInfo.SkipDeploy &&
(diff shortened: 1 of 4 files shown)


internal/sourcecode/scan.go
```diff
@@ -12,7 +13,7 @@ import (
 	"github.com/superfly/flyctl/helpers"
 )
 
-//go:embed templates/**
+//go:embed templates/** templates/**/.dockerignore
```
michaeldwan, line 16: the first `templates/**` should recursively glob everything, so the .dockerignore pattern might not be needed.

internal/sourcecode/scan.go
```diff
@@ -267,6 +293,8 @@ func templates(name string) (files []SourceFile) {
 			return nil
 		}
 
+		fmt.Println(path)
```
michaeldwan, line 296: is this a leftover from debugging?

cmd/launch.go
```diff
@@ -253,9 +253,15 @@ func runLaunch(cmdctx *cmdctx.CmdContext) error {
 		return nil
 	}
 
-	fmt.Println("Your app is ready. Deploy with `flyctl deploy`")
+	if srcInfo.SkipDeploy && srcInfo.Docs == "" {
+		fmt.Println("Your app is ready. Deploy with `flyctl deploy`")
+	} else if srcInfo.Docs != "" {
```
michaeldwan, line 258: This comment was always printed because people kept running `flyctl launch` for subsequent deploys. The issue seemed to go away after telling them to deploy with `flyctl deploy` then offering to do it now with a prompt, rather than making it seem like `flyctl launch` was doing something special to deploy. I don't love that solution, but we should probably keep it unless you have a better idea.

PR #570, opened 2021-10-19
Add dockefile key to app config builder section

### LINKED ISSUE #564: Support `dockerfile` in fly.toml
We should be able to specify the target Dockerfile and Docker build target in `fly.toml`. This is a prerequisite for future plans to fetch canonical deployment templates and place them into a `.fly` directory. These options are currently only supported via `deploy --dockerfile` and `deploy --build-target`.

I believe ene catch is that we would need to extend the API to support these entries, or extract them before sending the config to the backend.

rugwirobaker:
```toml
[build]
  dockerfile = "/path_to/dockefile"
```

Closes #564


rugwirobaker:
> This looks good. You just need to wire it up so it's used in the deploy. Here's where the deploy command handles a Dockerfile from a flag. You can extend that to check the config file as well.

@michaeldwan do you mean in the deploy command or the deploy section of the launch command?

PR #572, opened 2021-10-21
initial implementation of the new CLI

azazeal:
This PR significantly refactors the way `flyctl` reads its configuration and runs the CLI.


michaeldwan (approved):
The code looks sensible and it's been working for normal use locally. Let's do a prerelease and have more folks test it out.

Random thoughts:

- I don't think we should have debug logs on by default in dev mode since it obscures the experience users see, especially for interactive or long-running sequences.
- Why are commands under `internal/cli/internal` instead of just `internal/cli`? If I understand the module import rules, the first `internal` prevents unexpected external use so I'm not sure what the second does. 
- the update check is getting ran in dev and printing this:
  ```
  Update available 0.0.0-1635791221+dev -> v0.0.250.
  Run "flyctl version update" to upgrade.
  ```

PR #576, opened 2021-10-27
Recipes

davissp14:
This is very much a WIP

## Recipes

The idea behind Recipes is to give users the power and flexibility to orchestrate multiple API calls under a single command.  As of right now, almost all of our orchestration is handled internally and we think that needs to change.  
   

**Near-term goals**

Introduce and expose some of the power and flexibility that Recipes/Machines can offer.  Users will be able to fork flyctl and add their own Recipes as well as modify existing Recipes to accommodate their specific needs.  Users can then PR new Recipes or tweaks that they have made if they feel they would be valuable to the community. 

**Long-term** 

Depending on how Recipes are adopted, I think it would be interesting to create a framework that would encourage users to develop and maintain Recipes outside of Flyctl.  One thought would be to maintain a manifest of official and unofficial Recipes that the community can query and leverage.


jsierles:
Closing in preparation for moving things out of `internal/cli/internal`.

PR #582, opened 2021-10-28
Make sure ENV in the config is not cleared by setting `-e`

### LINKED ISSUE #560: Specifying env vars with 'fly deploy -e' wipes out env vars set in fly.toml
Setting env vars with `-e` should merge with those set on `fly.toml`, overriding only those with the same keys. Now, all env vars will be wiped out and replaced.

sudhirj:
Made ENV vars a property of the AppConfig, and made sure SetEnvVariable and SetEnvVariables don't clear out the entire set of vars.

Fixes #560

### DIFF
--- a/flyctl/app_config.go
+++ b/flyctl/app_config.go
@@ -57,7 +57,7 @@ func LoadAppConfig(configFile string) (*AppConfig, error) {
 		return nil, err
 	}
 
-	appConfig := *NewAppConfig()
+	appConfig := NewAppConfig()
 
 	file, err := os.Open(fullConfigFilePath)
 	if err != nil {
@@ -72,7 +72,7 @@ func LoadAppConfig(configFile string) (*AppConfig, error) {
 		return nil, errors.New("Unsupported config file format")
 	}
 
-	return &appConfig, err
+	return appConfig, err
 }
 
 func (ac *AppConfig) HasDefinition() bool {
--- a/flyctl/app_config.go
+++ b/flyctl/app_config.go
@@ -25,6 +25,15 @@ const (
 
 type Environment map[string]string
 
+func (e Environment) Set(key, value string) { e[key] = value }
+func (e Environment) Get(key string) string { return e[key] }
+func (e Environment) Unset(key string)      { delete(e, key) }
+func (e Environment) MultiSet(keyValues map[string]string) {
+	for k, v := range keyValues {
+		e[k] = v
+	}
+}
+
 type AppConfig struct {
 	AppName    string
 	Build      *Build
@@ -135,7 +144,7 @@ func (ac *AppConfig) unmarshalNativeMap(data map[string]interface{}) error {
 	}
 	if env, ok := data["env"].(map[string]interface{}); ok {
 		for k, v := range env {
-			ac.Env[k] = fmt.Sprint(v)
+			ac.Env.Set(k, fmt.Sprint(v))
 		}
 	}
 	delete(data, "app")
@@ -321,13 +330,11 @@ func (ac *AppConfig) GetInternalPort() (int, error) {
 }
 
 func (ac *AppConfig) SetEnvVariables(vals map[string]string) {
-	for k, v := range vals {
-		ac.SetEnvVariable(k, v)
-	}
+	ac.Env.MultiSet(vals)
 }
 
 func (ac *AppConfig) SetEnvVariable(name, value string) {
-	ac.Env[name] = value
+	ac.Env.Set(name, value)
 }
 
 func (ac *AppConfig) SetProcess(name, value string) {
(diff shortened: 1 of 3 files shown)


flyctl/app_config.go
```diff
@@ -27,6 +27,7 @@ type AppConfig struct {
 	AppName    string
 	Build      *Build
 	Definition map[string]interface{}
+	Env        map[string]string
```
azazeal, line 30: I think there might be benefit to declaring a new type, `Environment`, on which the rest of the functionality of this PR should be implemented on. `Build` works similarly.

flyctl/app_config.go
```diff
@@ -54,9 +56,7 @@ func LoadAppConfig(configFile string) (*AppConfig, error) {
 		return nil, err
 	}
 
-	appConfig := AppConfig{
-		Definition: map[string]interface{}{},
-	}
+	appConfig := *NewAppConfig()
```
azazeal, line 59: Why are we dereferencing here?

flyctl/app_config_test.go
```diff
@@ -54,3 +54,21 @@ func TestLoadTOMLAppConfigWithServices(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, p.Definition, rawData)
 }
+
+func TestLoadTOMLAppConfigWithEnvVars(t *testing.T) {
+	path := "./testdata/env-vars.toml"
+	p, err := LoadAppConfig(path)
+	assert.NoError(t, err)
```
azazeal, line 61: This should probably be `require.NoError(t, err)` as the rest of the test is pointless unless the configuration is loaded.

flyctl/app_config_test.go
```diff
@@ -54,3 +54,21 @@ func TestLoadTOMLAppConfigWithServices(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, p.Definition, rawData)
 }
+
+func TestLoadTOMLAppConfigWithEnvVars(t *testing.T) {
+	path := "./testdata/env-vars.toml"
```
azazeal, line 59: Maybe this should be a constant `const path = "testdata/env-vars.toml"`.

flyctl/app_config.go
sudhirj: That sounds good.

flyctl/app_config.go
sudhirj: The method returns a pointer by default, which we actually don't want in this context. We should probably get the method to return a struct by default and let the caller take the address if they want.

flyctl/app_config_test.go
sudhirj: Oh, does assert not fail immediately? Will change.

flyctl/app_config.go
azazeal: I think this is part of [`func LoadAppConfig(configFile string) (*AppConfig, error) {`](https://github.com/superfly/flyctl/blob/89efffb440a3512f80944c30cb1832ff82f0acb7/flyctl/app_config.go#L53) which returns a pointer, no? Sorry if I'm missing something obvious!

azazeal: Hi there again - let me clarify: I understand that in this particular context we used to do the following:

```
appConfig := AppConfig{
	Definition: map[string]interface{}{},
}
```

Changing from this to `appConfig := NewAppConfig()` (with `appConfig` ending up being a pointer) would not affect the rest of the code.

Sorry for the confusion!

sudhirj: The last line (the return) was taking the address again – I've removed that now and made this a simple call to `NewAppConfig()`.

flyctl/app_config_test.go
sudhirj: Do you think I should standardise all the rest of the test as well? seems like a good idea.

flyctl/app_config.go
sudhirj: Actually that seems like it'll add more code for not much benefit — will need to add a `Get`, `Set` and a `Range` (all three are used) methods, or just expose the map again on the `Env` struct. Checking if see if there's any other benefit to pulling this up, or if we can just use one of the stdlib maps like Values.

sudhirj: Sorry, misunderstood. I've added a new type called `Environment` which is just `map[string]string`.

jsierles:
How do we feel about merging?

rugwirobaker:
LGTM!!

diwu1989:
Guys, does this really work?
I just upgraded my flyctl today and did a deploy, and all env variables in fly.toml got wiped out!

michaeldwan:
@diwu1989 there's an issue, we'll revert and cut a new release

PR #588, opened 2021-10-29
quicker update.PromptFor

azazeal:
internal/update: ensure PromptFor doesn't take more than 128ms


azazeal:
Due to #572 this is no longer required.

PR #590, opened 2021-11-02
implementation of new apps command

azazeal:
This PR implements the new apps command.


rugwirobaker (approved):
LGTM. Should remove [orgKey](https://github.com/superfly/flyctl/blob/bfc273710be258200e87fcfad49e7d5db4bf6027/internal/cli/internal/state/state.go#L16) since it's no longer used.

azazeal:
> LGTM. Should remove [orgKey](https://github.com/superfly/flyctl/blob/bfc273710be258200e87fcfad49e7d5db4bf6027/internal/cli/internal/state/state.go#L16) since it's no longer used.

Thanks for spotting this!

PR #599, opened 2021-11-10
Add python scanner

rugwirobaker:
This python scanner supports pip, pipenv and conda and so far needs a procfile to work.
﻿

### DIFF
--- a/internal/sourcecode/scan.go
+++ b/internal/sourcecode/scan.go
@@ -49,6 +49,7 @@ func Scan(sourceDir string) (*SourceInfo, error) {
 		configureRuby,
 		configureGo,
 		configureElixir,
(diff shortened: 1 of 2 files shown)


jsierles:
If it needs a Procfile, we probably should document this. There is a new `[DeployDocs](https://github.com/superfly/flyctl/blob/master/internal/sourcecode/scan.go#L29)` that will get printed out after launch. That could contain the necessary docs, which I think we'd want to source from our example repo. We should also make sure our docs and sample repos align with this command.

Might also be good to add a commented-out Procfile as a template for generic python apps. For Django, maybe we would want to supply one that works out of the box?

jsierles (approved):

PR #603, opened 2021-11-10
Install shell completions as part of the `brew install`

rouge8:
Logic copied from the `miniserve` formula: https://github.com/Homebrew/homebrew-core/blob/064e4491fecd7c3ea198700b92d0bd9a2feea69f/Formula/miniserve.rb#L22-L27

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -93,3 +93,10 @@ brews:
     install: |
       bin.install "flyctl"
       bin.install_symlink "flyctl" => "fly"
+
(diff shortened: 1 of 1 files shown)


rouge8:
@michaeldwan do you have a chance to look at this?

jsierles:
Is there a way to test this locally?

rouge8:
It's a standard snippet used in many formula, but you could probably test locally by running `brew edit flyctl`, adding the completion generation, and then `brew reinstall flyctl`.

PR #607, opened 2021-11-12
Support specifying build_target in fly.toml

### LINKED ISSUE #606: Support Docker build targets in fly.toml
We should allow specifying the Docker build target in `fly.toml`, just like we do as an argument: `--build-target`.

jsierles:
Fixes #606.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -181,7 +181,7 @@ func runDeploy(cmdCtx *cmdctx.CmdContext) error {
 		if dockerBuildTarget := cmdCtx.Config.GetString("build-target"); dockerBuildTarget != "" {
 			opts.Target = dockerBuildTarget
 		} else if dockerBuildTarget := cmdCtx.AppConfig.DockerBuildTarget(); dockerBuildTarget != "" {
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -178,6 +178,12 @@ func runDeploy(cmdCtx *cmdctx.CmdContext) error {
 			opts.DockerfilePath = filepath.Join(filepath.Dir(cmdCtx.ConfigFile), dockerfilePath)
 		}
 
+		if dockerBuildTarget := cmdCtx.Config.GetString("build-target"); dockerBuildTarget != "" {
+			opts.Target = dockerBuildTarget
+		} else if dockerBuildTarget := cmdCtx.AppConfig.DockerBuildTarget(); dockerBuildTarget != "" {
+			opts.Target = filepath.Join(filepath.Dir(cmdCtx.ConfigFile), dockerBuildTarget)
```
michaeldwan, line 184: Should this be the raw value from the config rather than joined with the config path?

jsierles: Doh!

michaeldwan: Looks good!

PR #617, opened 2021-11-25
implementation of file-based mutexes

azazeal:
This PR implements the `filemu` package via which one can now acquire file-based locks (exclusive and shared). It additionally refactors the `cache` package to demonstrate usage.


rugwirobaker (approved):
This really nice🔥

azazeal:
> This really nice:fire: 

Thank you, but I's also baked into #601 so this will be closed :).

azazeal:
Baked into #601

PR #624, opened 2021-11-29
Use machine-based remote builders exclusively

jsierles:
This is a step towards deprecating application-based remote builders.

### DIFF
--- a/api/resource_remote_builders.go
+++ b/api/resource_remote_builders.go
@@ -2,7 +2,7 @@ package api
 
 import "context"
 
-func (client *Client) EnsureRemoteBuilder(ctx context.Context, orgID string, appName string) (*Machine, *App, error) {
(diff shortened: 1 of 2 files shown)


api/resource_remote_builders.go
```diff
@@ -2,7 +2,7 @@ package api
 
 import "context"
 
-func (client *Client) EnsureMachineRemoteBuilderForApp(ctx context.Context, appName string) (*Machine, *App, error) {
+func (client *Client) EnsureRemoteBuilder(ctx context.Context, orgID string, appName string) (*Machine, *App, error) {
```
azazeal, line 5: since `orgID` and `appName` are both strings, you can compress their declaration like so: `orgID, appName string`.

azazeal (changes requested):

azazeal (approved):
LGTM

PR #627, opened 2021-12-02
Improve app name selection

rugwirobaker:
* Add `--name` flag
* Prompt user for name if `--name` is not set with heroku app name as the default﻿

Example: [asciicast](https://asciinema.org/a/2gLAB6gi4tjc4iwqHWOeklnMW)


jsierles (approved):
Nice improvement! Also think standard launches could feel nicer with this style of app name selection.

PR #633, opened 2021-12-04
Added a flag to the launch that creates app without prompting if you want to use the existing config

LukeShay:
# What

Added the flag `--copy-config` that will use the current config to launch a new app without prompting.

# Why

Simplifies launching a new app from CI without user intervention


jsierles:
Thanks - you read my mind.

PR #635, opened 2021-12-05
Add a 'launch --from' option to launch from a remote fly.toml

jsierles:
This is useful to launch apps like [Fly's official Redis image](https://github.com/fly-apps/redis/blob/main/fly.toml) along with a volume. This will only work with `fly.toml` that specify a Docker image source. Example:

`fly launch --from https://raw.githubusercontent.com/fly-apps/redis/main/fly.toml`

This should, at a minimum:

* Validate the TOML
* Ask to create a volume if one is required
* Ask to set required secrets (no way to define this yet in fly.toml)
* Perform the deployment


jsierles:
This was a noble experiment, but should be implemented under the new command chain once launch is ported and refactored.

PR #641, opened 2021-12-07
Add django launcher

rugwirobaker:
* Simple dockertfile(WIP)﻿


rugwirobaker:
Creating not just `DATABASE_URL` but  `DATABASE_HOST` `DATABASE_USER`, `DATABASE_PASSWORD` when attaching Postgres database would reduce the number of modifications we have to make too `settings.py`.

jsierles:
Going to merge this for now, as it's better to having something than nothing! We can work on it iteratively if we need to augment `settings.py`.

PR #658, opened 2021-12-10
Manual entry must be set in order to use new attach logic

davissp14:


jsierles:
I believe this is irrelevant with the new attach method.

PR #659, opened 2021-12-11
Add `--availability` flag, ability to bypass interactive prompts on `pg create`

### LINKED ISSUE #645: flyctl postgres create <flags>
**Describe the bug**
Tell us what broke, what operating system you're using, etc.
When you attempt to run `flyctl postgres create <flags>` there is no flag to bypassing the configuration prompt. I have attempted to review the `postgres.go` file and it appears that on `line 214` there is a condition for a FALSE `customConfig` flag. Then, on `line 236` there is a FALSE condition. The `else` condition on `line 271` never appears to actually get executed and, as a result, there is no way to bypass the prompt for configuration. I have tried all the permutations of the `flyctl postgres create <flags>` to try and bypass the prompt but it does not seem to work.

I think we need a `--config`, `--conf`, `-c`, and/or `--configuration` for bypassing this prompt in some way.

sammccord:
closes #645

Added `--availability` flag with possible values `high` and `standalone`. Will prompt if neither value, nil, or some other string.

This API might be a little wordy, so open to suggestions. I also considered something like `--development`, but I believe this is more explicit.

<3 fly keep up the good work!

### DIFF
--- a/cmd/input.go
+++ b/cmd/input.go
@@ -203,6 +203,35 @@ func inputAppName(defaultName string, autoGenerate bool) (name string, err error
 	return name, nil
 }
 
+const (
+	highAvailability       = "high"
+	standaloneAvailability = "standalone"
+)
+
+func postgresClusteringOptionsInput(availability string) (PostgresClusterOption, error) {
+	if availability == highAvailability {
+		return highlyAvailablePostgres(), nil
+	}
+	if availability == standaloneAvailability {
+		return standalonePostgres(), nil
+	}
+	selected := 0
+	options := []string{}
+	for _, opt := range postgresClusteringOptions() {
+		options = append(options, opt.Name)
+	}
+	prompt := &survey.Select{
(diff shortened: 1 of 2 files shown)


davissp14 (changes requested):
First off, thanks for the PR!

Looks like there's an issue when the configuration option is chosen interactively.  Instead of auto-resolving configuration based on the option selected, it's currently re-prompting for each of the individual options.
 
Screen Shot 2021-12-16 at 3 19 07 PM

sammccord:
> 
> Looks like there's an issue when the configuration option is chosen interactively. Instead of auto-resolving configuration based on the option selected, it's currently re-prompting for each of the individual options.
> 

Right! Thanks for catching that. Should be good now.

davissp14 (approved):
Sorry for the delayed review!  Tested this out and things look good, thanks a ton!

PR #666, opened 2021-12-17
Turboku with standalone postgres

rugwirobaker:
* Adds new flag `--with-postgres` which false by default.

Maybe do:

* Check the postgres addon version
* Migrate the user's data to the new database'


jsierles:
We've discussed that we won't do this for the first release of Turboku, so closing for now.

PR #667, opened 2021-12-17
Prototype: Launch a Redis instance using machines

jsierles:
This PR starts the work towards implementing launchers solely via the Machines API. The main goal is not to get a functional Redis launcher. Instead, we're laying the groundwork for building launcher commands in the new flyctl command structure.

Some smaller things implemented here:

* Reworking of org and region selection
* Env var support for selecting a region
* Hiding experimental commands behind the `DEV` env var


jsierles:
Merging this to pull in the useful helper updates for new commands.

PR #670, opened 2021-12-18
Port 'open' to the new command structure

rugwirobaker:


internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,66 @@
+package open
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/skratchdot/open-golang/open"
+	"github.com/spf13/cobra"
```
azazeal, line 8: typically one separates 3rd party imports before their own, as per the rest of the docs.

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,66 @@
...
+
+func run(ctx context.Context) (err error) {
+	var (
+		path = "/"
+
+		appName = app.NameFromContext(ctx)
+	)
+
+	if flag.Len(ctx) > 1 {
+		return fmt.Errorf("too many arguments - only one path argument allowed")
```
azazeal, line 44: why `fmt.Errorf` instead of `errors.New`? additionally, why don't you just set the maximum amount of accepted commands (NArgs) in `New`?

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	if flag.Len(ctx) > 0 {
+		path = flag.FirstArg(ctx)
```
azazeal, line 48: `flag.FirstArg` reports "" in case there's no arg, I'm not sure this if is required.

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	app, err := client.FromContext(ctx).API().GetApp(ctx, appName)
+	if err != nil {
+		return
```
azazeal, line 53: perhaps we should return a better error here, like `fmt.Errorf("failed retrieving app: %w", err)`

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	if !app.Deployed {
+		fmt.Println(`App has not been deployed yet. Try running "` + buildinfo.Name() + ` deploy --image flyio/hellofly"`)
```
azazeal, line 57: if the app hasn't been deployed then this is an error. `return fmt.Errorf(...)` instead of `fmt.Println`, please.

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil
+	}
+
+	appUrl := "http://" + app.Hostname + path
+	fmt.Println("Opening", appUrl)
```
azazeal, line 62: we don't use `fmt` to print to `stdout` any more. You should retrieve the io streams from `iostreams.FromContext(ctx)` and then `fmt.Fprintln(...)` so that you end up writing to the appropriate io stream.

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,66 @@
...
+
+	return open.Run(appUrl)
```
azazeal, line 64: instead of `return open.Run(appUrl)` maybe we should be decorating the error with `fmt.Errorf("failed opening URL %s: %w", url, err)`

internal/cli/internal/flag/context.go
```diff
@@ -24,6 +24,11 @@ func Args(ctx context.Context) []string {
 	return FromContext(ctx).Args()
 }
 
+// Len returns the number of flags in ctx.
+func Len(ctx context.Context) int {
```
azazeal, line 28: We don't need this, we already have the `Args` function.

azazeal (changes requested):

internal/cli/internal/command/open/open.go
rugwirobaker: Do we keep the "Try running..." section of the error?

azazeal: Sure, I don't see why not (assuming the command still works).

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,69 @@
...
+
+	path = flag.FirstArg(ctx)
+
+	app, err := client.FromContext(ctx).API().GetApp(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("failed retrieving app: %w", err)
+	}
+
+	if !app.Deployed {
+		return fmt.Errorf(`app has not been deployed yet. Try running "` + buildinfo.Name() + ` deploy --image flyio/hellofly"`)
```
azazeal, line 55: instead of 
```go
return fmt.Errorf(`app has not been deployed yet. Try running "` + buildinfo.Name() + ` deploy --image flyio/hellofly"`)
```

try 

```go
return fmt.Errorf(`app has not been deployed yet. Try running "%s" deploy --image flyio/hellofly"`, buildinfo.Name())
```

yes?

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	appUrl := "http://" + app.Hostname + path
+
+	iostream := iostreams.FromContext(ctx)
+
+	fmt.Fprintf(iostream.Out, "Opening %s\n", appUrl)
```
azazeal, line 62: perhaps after opening use `...` to denote something's happening elsewhere.

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/skratchdot/open-golang/open"
+	"github.com/spf13/cobra"
+
+	"github.com/superfly/flyctl/internal/buildinfo"
+	"github.com/superfly/flyctl/internal/cli/internal/app"
+	"github.com/superfly/flyctl/internal/cli/internal/command"
+	"github.com/superfly/flyctl/internal/cli/internal/flag"
+	"github.com/superfly/flyctl/internal/client"
+	"github.com/superfly/flyctl/pkg/iostreams"
```
azazeal, line 15: I'm so sorry I didn't spot this: `pkg` imports should be listed after 3rd party ones, in their own section.

You can see an example of this [here](https://github.com/superfly/flyctl/blob/c98d6cadd9d491abfcf7e8f3060d79dee19ad82b/internal/cli/internal/command/apps/create.go#L3)

azazeal (changes requested):

internal/cli/internal/command/open/open.go
rugwirobaker: It doesn't have work if you haven't run the flyctl launch. Plus it doesn't even make sense to suggest deploying the `flyio/hellofly` image when the user might be working with something different

rugwirobaker: It doesn't have work if you haven't run the flyctl launch. Plus it doesn't even make sense to suggest deploying the `flyio/hellofly` image when the user might be working with something different

azazeal: > It doesn't have work if you haven't run the flyctl launch. Plus it doesn't even make sense to suggest deploying the `flyio/hellofly` image when the user might be working with something different. `Try running flyctl deploy` makes more sense to me as a user.

Then maybe prompt the user to launch the application first instead? We don't need to be specific, just give them a rough idea why this isn't working, imo.

azazeal: something like 

> "app has not been deployed yet. Please try deploying your app first"

azazeal (approved):
LGTM; please test locally before cutting a release

azazeal (changes requested):
Sorry - missed something; this won't work due to the `[PATH]` argument. This doesn't have the required interaction with the way `RequireAppName` works. Let's get over this 1-1 offline

rugwirobaker:
Now we can `flyctl apps open` too.

internal/cli/internal/command/open/open.go
```diff
@@ -0,0 +1,67 @@
+package open
```
azazeal, line 1: This `open.go` file essentially duplicates the contents of `apps/move.go`. There's no reason to duplicate this, please do check what we've done for [`move.go`](https://github.com/superfly/flyctl/blob/c98d6cadd9d491abfcf7e8f3060d79dee19ad82b/internal/cli/internal/command/move/move.go). In your case, both the code for the constructor as well as the runner are the same so you can minimize the duplicated code even further.

azazeal (changes requested):

PR #673, opened 2021-12-21
Move logs command to the new command format

jsierles:
A few doubts to resolve:

* How should the log presenter be ported to the new system? This presenter is used by three different packages currently.
* What should we do with `terminal.Debug`?


azazeal:
A few notes on this:

- `terminal.Debug` is going away so don't use it. If you need to display debug messages use the `logger` package.
- presenters are also going away so we'll have to find a way to minimize their footprint

PR #701, opened 2021-12-31
fix: resolve dockerfile path

azazeal:


azazeal:
Closed in favor of #700.

PR #710, opened 2022-01-05
Avoid bubbling errors from intentional closure of release command logs

jsierles:
During deployment, we intentionally shut down log tailing after 3 seconds of inactivity. The resulting error was bubbling up and manifesting with an early exit `127` and an aborted deployment. This PR ensures the parent context error is ignored.


azazeal (approved):
LGTM

PR #721, opened 2022-01-07
Refactoring dig commmand

rugwirobaker:
* add roundtrip func﻿


internal/cli/internal/command/command.go
```diff
@@ -408,21 +408,11 @@ func RequireAppName(ctx context.Context) (context.Context, error) {
...
-	name := flag.GetApp(ctx)
-	if name == "" {
-		// if there's no flag present, first consult with the environment
-		if name = env.First("FLY_APP"); name == "" {
-			// and then with the config file (if any)
-			if cfg := app.ConfigFromContext(ctx); cfg != nil {
-				name = cfg.AppName
-			}
-		}
+	if err != nil && err == errRequireAppName {
```
azazeal, line 413: use `errors.Is` here, please.

internal/cli/internal/command/command.go
```diff
@@ -408,21 +408,11 @@ func RequireAppName(ctx context.Context) (context.Context, error) {
...
+		err = nil
 	}
 
-	return app.WithName(ctx, name), nil
+	return ctx, err
```
azazeal, line 417: I think this may be introducing a bug.

azazeal (commented):
a few starting thoughts

internal/cli/internal/command/command.go
```diff
@@ -408,21 +408,15 @@ func RequireAppName(ctx context.Context) (context.Context, error) {
...
 
-	name := flag.GetApp(ctx)
-	if name == "" {
-		// if there's no flag present, first consult with the environment
-		if name = env.First("FLY_APP"); name == "" {
-			// and then with the config file (if any)
-			if cfg := app.ConfigFromContext(ctx); cfg != nil {
-				name = cfg.AppName
-			}
+	if err != nil {
```
azazeal, line 413: `errors.Is` understands nils, you don't need to nest your ifs here.

internal/cli/internal/command/command.go
```diff
@@ -408,21 +408,15 @@ func RequireAppName(ctx context.Context) (context.Context, error) {
...
+		if errors.Is(err, errRequireAppName) {
+			return ctx, nil
```
azazeal, line 415: if you get an error you should set the name to an empty string. here you're returning a context which doesn't carry a name so callers of `app.NameFromContext` will crash.

internal/cli/internal/command/dig/dig.go
```diff
@@ -238,10 +194,54 @@ func run(ctx context.Context) error {
...
 }
 
-// ResolverForOrg takes a connection to the wireguard agent and an organization
+// roundTrip a DNS request across a "TCP" socket; we'd just use miekg/dns's Client, but I don't think it promises to
+// work over our weird UDS TCP proxy.
+func roundTrip(conn net.Conn, m *dns.Msg) (*dns.Msg, error) {
+	m.Id = dns.Id()
+	m.Compress = true
+
+	m.Id = dns.Id()
```
azazeal, line 203: why you set the ID and compress twice?

azazeal (changes requested):
this can be cleaned up a bit further

internal/cli/internal/command/dig/dig.go
rugwirobaker: Copy pasted code🤦🏽‍♂️

internal/cli/internal/command/command.go
rugwirobaker: You are right I should have tested that path

internal/cli/internal/command/dig/dig.go
```diff
@@ -238,10 +194,51 @@ func run(ctx context.Context) error {
...
-// ResolverForOrg takes a connection to the wireguard agent and an organization
+// roundTrip a DNS request across a "TCP" socket; we'd just use miekg/dns's Client, but I don't think it promises to
+// work over our weird UDS TCP proxy.
+func roundTrip(conn net.Conn, m *dns.Msg) (*dns.Msg, error) {
+	m.Id = dns.Id()
+	m.Compress = true
+
+	buf, err := m.Pack()
+	if err != nil {
+		return nil, fmt.Errorf("dns round trip: %w", err)
```
azazeal, line 205: this isn't a dns round trip error; I wouldn't decorate my errors in a helper function like `roundTrip` but at the caller.

azazeal (changes requested):

internal/cli/internal/command/dig/dig.go
rugwirobaker: @azazeal  Yeah indeed. But if i's, not a round trip error, what decoration would give it in the main function?

internal/cli/internal/command/dig/dig.go
```diff
@@ -226,7 +182,7 @@ func run(ctx context.Context) error {
 	case "TXT-NATIVE":
 		txts, err := r.LookupTXT(ctx, name)
 		if err != nil {
-			return FixNameError(err, ns)
+			return fixNameError(err, ns)
 		}
 
 		fmt.Printf("%s\n", strings.Join(txts, ""))
```
azazeal, line 188: please use `iostreams`

internal/cli/internal/command/dig/dig.go
azazeal: just don't decorate them and bubble them up. all of these are errors due to writing to the connection - it'll be obvious where they're coming from

azazeal (changes requested):

azazeal (approved):

PR #725, opened 2022-01-11
Redefine PG Standalone setup

davissp14:
This PR redefines the standalone PG setup and allows the `initial cluster size` to be tunable on provision.

Single node, dev configurations will run Stolon.  This is not ideal, but it drastically simplifies the UX as well as the horizontal scaling story.  We may re-introduce a true standalone configuration later on.  

cc:// @jsierles


jsierles (approved):
Looks good to me. Hopefully soon we can tackle porting the launch command to the new format.

PR #732, opened 2022-01-13
Pg launch

davissp14:
Exposes Postgres on Machines:  `DEV=1 fly services postgres launch`

### DIFF
--- a/api/resource_images.go
+++ b/api/resource_images.go
@@ -35,3 +35,20 @@ func (client *Client) GetImageInfo(ctx context.Context, appName string) (*App, e
 
 	return &data.App, nil
 }
+
+func (client *Client) GetLatestImageTag(ctx context.Context, repository string) (string, error) {
+	query := `
+		query($repository: String!) {
+			latestImageTag(repository: $repository) 
+		}
+	`
+	req := client.NewRequest(query)
+	req.Var("repository", repository)
+
+	data, err := client.RunWithContext(ctx, req)
+	if err != nil {
+		return "", err
+	}
+
+	return data.LatestImageTag, nil
+}
--- a/api/resource_postgres.go
+++ b/api/resource_postgres.go
@@ -186,7 +186,7 @@ func (client *Client) ListPostgresUsers(ctx context.Context, appName string) ([]
 }
 
 func (client *Client) EnablePostgresConsul(ctx context.Context, appName string) (*PostgresEnableConsulPayload, error) {
-	query := `
+	const query = `
 		mutation($appName: String!) {
 			enablePostgresConsul(input: {appId: $appName}) {
 				consulUrl
--- a/api/resource_postgres.go
+++ b/api/resource_postgres.go
@@ -185,6 +185,25 @@ func (client *Client) ListPostgresUsers(ctx context.Context, appName string) ([]
 	return *data.App.PostgresAppRole.Users, nil
 }
 
+func (client *Client) EnablePostgresConsul(ctx context.Context, appName string) (*PostgresEnableConsulPayload, error) {
+	query := `
+		mutation($appName: String!) {
+			enablePostgresConsul(input: {appId: $appName}) {
+				consulUrl
+			}
+		}
+	`
+	req := client.NewRequest(query)
+	req.Var("appName", appName)
+
+	data, err := client.RunWithContext(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	return data.EnablePostgresConsul, nil
+}
+
 // func (client *Client) CreatePostgresDatabase(name string) (*PostgresClusterUser, error) {
 // 	query := `
 // 		mutation($input: CreatePostgresClusterUserInput!) {
--- a/internal/cli/internal/command/services/postgres/launch.go
+++ b/internal/cli/internal/command/services/postgres/launch.go
@@ -72,7 +72,7 @@ func runLaunch(ctx context.Context) (err error) {
 	vmSize := flag.GetString(ctx, "vm-size")
 	snapshotId := flag.GetString(ctx, "snapshot-id")
 
-	// TODO - Resolve latest version from flyctl.
+	// TODO - Resolve latest version from graphql
 	imageRef := "flyio/postgres:14.1"
 
 	consulUrl := flag.GetString(ctx, "consul-url")
--- a/internal/cli/internal/command/services/postgres/launch.go
+++ b/internal/cli/internal/command/services/postgres/launch.go
@@ -36,6 +36,7 @@ func newLaunch() (cmd *cobra.Command) {
 		flag.Org(),
 		flag.String{Name: "password", Shorthand: "p", Description: "The superuser password. The password will be generated for you if you leave this blank"},
 		flag.String{Name: "vm-size", Description: "the size of the VM", Default: "shared-cpu-1x"},
+		flag.String{Name: "consul-url", Description: "Opt into using an existing consul as the backend store by specifying the target consul url."},
 		flag.Int{Name: "volume-size", Description: "The volume size in GB", Default: 10},
 		flag.Int{Name: "initial-cluster-size", Description: "Initial cluster size", Default: 2},
 		flag.String{Name: "snapshot-id", Description: "Creates the volume with the contents of the snapshot"},
@@ -50,22 +51,21 @@ type PostgresLaunch struct {
 }
 
 type PostgresProvisionConfig struct {
-	AppName      string
-	ConsulUrl    string
-	Count        int
-	EtcdUrl      string
-	ImageRef     string
-	Organization *api.Organization
-	Password     string
-	Region       string
-	SnapshotId   string
-	VolumeSize   int
-	VMSize       string
+	AppName            string
+	ConsulUrl          string
+	ImageRef           string
+	InitialClusterSize int
+	Organization       *api.Organization
+	Password           string
+	Region             string
+	SnapshotId         string
+	VolumeSize         int
+	VMSize             string
 }
 
 func runLaunch(ctx context.Context) (err error) {
 	name := flag.GetString(ctx, "name")
-	count := flag.GetInt(ctx, "count")
+	initialClusterSize := flag.GetInt(ctx, "initial-cluster-size")
 	region := flag.GetString(ctx, "region")
 	password := flag.GetString(ctx, "password")
 	volumeSize := flag.GetInt(ctx, "volume-size")
@@ -73,7 +73,9 @@ func runLaunch(ctx context.Context) (err error) {
 	snapshotId := flag.GetString(ctx, "snapshot-id")
 
 	// TODO - Resolve latest version from flyctl.
-	imageRef := "flyio/postgres"
+	imageRef := "flyio/postgres:14.1"
+
+	consulUrl := flag.GetString(ctx, "consul-url")
 
 	var org *api.Organization
 	if org, err = prompt.Org(ctx, nil); err != nil {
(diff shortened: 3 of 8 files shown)


internal/cli/internal/command/services/postgres/postgres.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/superfly/flyctl/internal/cli/internal/command"
+)
+
+// New initializes and returns a new postgres Command.
+func New() (cmd *cobra.Command) {
+	const (
+		// TODO: document command
+		long = `
```
azazeal, line 14: don't we need docs here?

internal/cli/internal/command/services/postgres/util.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"github.com/superfly/flyctl/internal/client"
+)
+
+func WaitForMachineState(ctx context.Context, client *client.Client, appId, machineId, state string) error {
```
azazeal, line 12: if there's no other caller for this, please don't export it.

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/superfly/flyctl/internal/cli/internal/command"
+	"github.com/superfly/flyctl/internal/cli/internal/flag"
+	"github.com/superfly/flyctl/internal/cli/internal/prompt"
+	"github.com/superfly/flyctl/internal/client"
+)
+
+func newLaunch() (cmd *cobra.Command) {
+	const (
+		// TODO: document command
+		long = `
```
azazeal, line 21: no dox?

internal/cli/internal/command/services/postgres/util.go
```diff
@@ -0,0 +1,46 @@
...
+	fmt.Printf("Waiting for machine %q to reach a healthy state.\n", machineId)
+	timeout := time.After(2 * time.Minute)
+	tick := time.Tick(1 * time.Second)
+	for {
```
azazeal, line 16: I think a `for ctx.Err() == nil` loop in combination with `pause.For(ctx, 5*time.second)` would work better for you here, no? ([pause](https://github.com/azazeal/pause))

azazeal (changes requested):
good work overall, a few slight mods and we'll get it merged.

internal/cli/internal/command/services/postgres/postgres.go
davissp14: There's going to be a lot of work going into this over the next two weeks and given it's hidden behind the Dev flag I'm not super concerned about the docs quite yet.

internal/cli/internal/command/services/postgres/util.go
davissp14: It's unclear what the difference would be exactly.  Is there a clear benefit to using pause vs. ticks?

azazeal: `pause` would return once `ctx` is canceled or deadlined. `time.Sleep` would end up waiting for 5seconds when the user hits Ctrl+C (as an example).

internal/cli/internal/command/services/postgres/util.go
azazeal: Oops, I spotted another thing. In Go we typically use [MixedCaps](https://tip.golang.org/doc/effective_go#mixed-caps) so unfortunately `machineId` should be `machineID`.

internal/cli/internal/command/services/postgres/util.go
```diff
@@ -0,0 +1,46 @@
...
+				}
+			}
+			time.Sleep(5 * time.Second)
+		}
+	}
+}
+
+// encodeCommand will base64 encode a command string so it can be passed
+// in with  exec.Command.
+func EncodeCommand(command string) string {
```
azazeal: is anyone else calling this?

azazeal (commented):
some further comments

internal/cli/internal/command/services/postgres/util.go
davissp14: haha, all good.  I'll update.

internal/cli/internal/command/services/postgres/util.go
davissp14: I'd still need a top-level timeout though, right?  Is there a better way to handle this?

```golang
ctx, cancel := context.WithTimeout(parent, time.Duration(time.Minute*2))
defer cancel()

go func(ctx context.Context) {
    for ctx.Err() == nil {
	  pause.For(ctx, time.Second)
  
	  machines, err := client.API().ListMachines(ctx, appID, "")
	  if err != nil {
		  logger.Debugf("failed retrieving machines: %v", err)
  
		  continue
	  }
  
	  for _, machine := range machines {
		  if machine.ID == machineID {
			  if machine.State == state {
				  cancel()
			  }
  
			  fmt.Fprintf(io.Out, "Machine state: %q, wanted: %q\n", machine.State, state)
		  }
	  }
    }
}(ctx)

select {
    case <-ctx.Done():
        if ctx.Err() == context.DeadlineExceeded {
          return ctx.Err()
        }

       return ctx.Err()
}
```

api/resource_postgres.go
```diff
@@ -185,6 +185,25 @@ func (client *Client) ListPostgresUsers(ctx context.Context, appName string) ([]
 	return *data.App.PostgresAppRole.Users, nil
 }
 
+func (client *Client) EnablePostgresConsul(ctx context.Context, appName string) (*PostgresEnableConsulPayload, error) {
+	query := `
```
azazeal, line 189: this should probably be a `const`ant, no?

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,308 @@
...
+		if err != nil {
+			return err
+		}
+
+		if err = waitForMachineState(ctx, p.client, p.config.AppName, machine.ID, "started"); err != nil {
+			return err
+		}
+	}
+
+	fmt.Fprintf(io.Out, "Provision complete\n\n")
```
azazeal, line 192: You're not formatting anything here, use `fmt.Println` here instead please.

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,308 @@
...
+
+	connStr := fmt.Sprintf("postgres://postgres:%s@%s.internal:5432\n", secrets["OPERATOR_PASSWORD"], p.config.AppName)
+
+	fmt.Fprintf(io.Out, "Any app within the %s organization can connect to this Postgres using the following credentials:\n", p.config.Organization.Name)
```
azazeal, line 196: this is a block of many consecutive writes to the same output (`io.Out`). Ideally, what we'd want is for all the output to be written to a buffer first and then flushed to `io.Out`. This would keep output consistent in case, as an example, another goroutine tries to write to `os.Out` while we do.

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,308 @@
...
+	fmt.Fprintf(io.Out, "  Hostname:    %s.internal\n", p.config.AppName)
+	fmt.Fprintf(io.Out, "  Proxy port:  5432\n")
+	fmt.Fprintf(io.Out, "  Postgres port:  5433\n")
+	fmt.Fprintf(io.Out, "  Connection string:  %s\n", connStr)
+	fmt.Fprintf(io.Out, "Save your credentials in a secure place, you won't be able to see them again!\n")
+
+	return nil
+}
+
+func (p *Launch) configurePostgres() (api.MachineConfig, error) {
```
azazeal, line 208: `xPostgres` & `postgresX` suffer from the same problem: we're inside the `postgres` package. a simple `configure` would suffice here would do the trick just fine, I think.

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,308 @@
...
+	machineConfig.Config["size"] = p.config.VMSize
+	machineConfig.Config["image"] = p.config.ImageRef
+	machineConfig.Config["restart"] = map[string]string{
+		"policy": "no",
+	}
+
+	// Set mounts
+	var volumeHash string
+	var err error
+	if volumeHash, err = helpers.RandString(5); err != nil {
```
azazeal, line 226: you can just put `volumeHash, err := ...` on its own line here instead of using var twice.

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,308 @@
...
+		"path":      "/data",
+	})
+	machineConfig.Config["mounts"] = mounts
+
+	return api.MachineConfig(machineConfig.Config), nil
+}
+
+func (p *Launch) createApp(ctx context.Context) (*api.App, error) {
+
+	fmt.Println("Creating app...")
```
azazeal, line 244: I think you missed this call to `fmt.Println`.

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,308 @@
...
+		PreferredRegion: &p.config.Region,
+		Runtime:         "FIRECRACKER",
+		AppRoleID:       "postgres_cluster",
+	}
+
+	return p.client.API().CreateApp(ctx, appInput)
+}
+
+func (p *Launch) setSecrets(ctx context.Context) (map[string]string, error) {
+	fmt.Println("Setting secrets...")
```
azazeal, line 257: `fmt.Println` is not something we should be calling.

internal/cli/internal/command/services/postgres/launch.go
```diff
@@ -0,0 +1,308 @@
...
+
+	var suPassword, replPassword, opPassword string
+	var err error
+
+	if suPassword, err = helpers.RandString(15); err != nil {
```
azazeal, line 262:
```go
suPassword, err := ...
if err != nil
```

is just fine and more idiomatic Go

internal/cli/internal/command/services/postgres/util.go
```diff
@@ -0,0 +1,51 @@
+package postgres
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"github.com/azazeal/pause"
```
azazeal, line 9: we generally try to add an empty line between 3rd party imports and local ones. Your editor should be doing this automatically, if not ping me in Slack and we'll see what's up with it.

internal/cli/internal/command/services/postgres/util.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/superfly/flyctl/internal/client"
+	"github.com/superfly/flyctl/internal/logger"
+	"github.com/superfly/flyctl/pkg/iostreams"
+)
+
+func waitForMachineState(parent context.Context, client *client.Client, appID, machineID, state string) error {
+
```
azazeal, line 16: I've noticed a lot of empty lines between function declaration and first line of body - this isn't something one sees in Go codebases.

internal/cli/internal/command/services/postgres/util.go
```diff
@@ -0,0 +1,51 @@
...
+	io := iostreams.FromContext(parent)
+	logger := logger.FromContext(parent)
+
+	ctx, cancel := context.WithTimeout(parent, time.Duration(time.Minute*2))
+	defer cancel()
+
+	for ctx.Err() == nil {
```
azazeal, line 23: this block reads much better now!

api/resource_images.go
```diff
@@ -35,3 +35,20 @@ func (client *Client) GetImageInfo(ctx context.Context, appName string) (*App, e
 
 	return &data.App, nil
 }
+
+func (client *Client) GetLatestImageTag(ctx context.Context, repository string) (string, error) {
+	query := `
```
azazeal, line 40: this should be a constant

azazeal (changes requested):
A few small things here and there - we can merge once you amend!

api/resource_postgres.go
davissp14: I am just following the pattern this file and others are using.

azazeal: It's a legacy thing and I get it. Not going to insist on this

davissp14:
Thank you very much for the reviews!

PR #743, opened 2022-01-19
Pg connect on machines

davissp14:
Exposes the quick connect command: `DEV=1 fly services postgres connect`. 

Screen Shot 2022-01-19 at 4 56 11 PM

cc:// @azazeal  One thing i'm unsure about here is the ssh package.  Would love your thoughts.

### DIFF
--- a/internal/cli/internal/command/services/postgres/connect.go
+++ b/internal/cli/internal/command/services/postgres/connect.go
@@ -10,7 +10,6 @@ import (
 	"github.com/superfly/flyctl/internal/cli/internal/app"
 	"github.com/superfly/flyctl/internal/cli/internal/command"
 	"github.com/superfly/flyctl/internal/cli/internal/command/ssh"
-
 	"github.com/superfly/flyctl/internal/cli/internal/flag"
 	"github.com/superfly/flyctl/internal/client"
 	"github.com/superfly/flyctl/pkg/agent"
@@ -72,7 +71,7 @@ func runConnect(ctx context.Context) error {
 
 	dialer, err := agentclient.Dialer(ctx, &app.Organization)
 	if err != nil {
-		return fmt.Errorf("failed to build tunnel for %s. %v", app.Organization.Slug, err)
+		return fmt.Errorf("failed to build tunnel for %s: %v", app.Organization.Slug, err)
 	}
 
 	database := flag.GetString(ctx, "database")
--- a/internal/cli/internal/command/services/postgres/connect.go
+++ b/internal/cli/internal/command/services/postgres/connect.go
@@ -17,9 +17,8 @@ import (
 
 func newConnect() (cmd *cobra.Command) {
 	const (
-		long = `
-			Connect to the Postgres console
-		`
+		long = `Connect to the Postgres console
+`
 		short = "Connect to the Postgres console"
 		usage = "connect"
 	)
--- a/internal/cli/internal/command/services/postgres/connect.go
+++ b/internal/cli/internal/command/services/postgres/connect.go
@@ -5,9 +5,9 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
+	"github.com/superfly/flyctl/internal/cli/internal/app"
 	"github.com/superfly/flyctl/internal/cli/internal/command"
 	"github.com/superfly/flyctl/internal/cli/internal/command/ssh"
 
@@ -22,15 +22,17 @@ func newConnect() (cmd *cobra.Command) {
 			Connect to the Postgres console
 		`
 		short = "Connect to the Postgres console"
-		usage = "connect [APPNAME]"
+		usage = "connect"
 	)
 
 	cmd = command.New(usage, short, long, runConnect,
 		command.RequireSession,
+		command.RequireAppName,
 	)
 	cmd.Args = cobra.MaximumNArgs(1)
 
 	flag.Add(cmd,
+		flag.App(),
 		flag.String{
 			Name:        "database",
 			Shorthand:   "d",
@@ -54,18 +56,17 @@ func newConnect() (cmd *cobra.Command) {
 }
 
 func runConnect(ctx context.Context) error {
+	appName := app.NameFromContext(ctx)
 	client := client.FromContext(ctx).API()
-
-	appName := flag.FirstArg(ctx)
 
 	app, err := client.GetApp(ctx, appName)
 	if err != nil {
-		return fmt.Errorf("get app: %w", err)
+		return fmt.Errorf("failed retrieving app %s: %w", appName, err)
 	}
 
 	agentclient, err := agent.Establish(ctx, client)
 	if err != nil {
-		return errors.Wrap(err, "can't establish agent")
+		return fmt.Errorf("failed to establish agent: %w", err)
 	}
 
 	dialer, err := agentclient.Dialer(ctx, &app.Organization)
--- a/internal/cli/internal/command/services/postgres/connect.go
+++ b/internal/cli/internal/command/services/postgres/connect.go
@@ -18,11 +18,10 @@ import (
 
 func newConnect() (cmd *cobra.Command) {
 	const (
-		// TODO: document command
 		long = `
-			Connect to postgres instance.
+			Connect to the Postgres console
 		`
-		short = "Establishs a session with Postgres"
+		short = "Connect to the Postgres console"
 		usage = "connect [APPNAME]"
 	)
 
@@ -91,5 +90,4 @@ func runConnect(ctx context.Context) error {
 		Stdout: os.Stdout,
 		Stderr: os.Stderr,
 	}, addr)
-
 }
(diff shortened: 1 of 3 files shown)


internal/cli/internal/command/services/postgres/connect.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/superfly/flyctl/internal/cli/internal/flag"
+	"github.com/superfly/flyctl/internal/client"
+	"github.com/superfly/flyctl/pkg/agent"
+)
+
+func newConnect() (cmd *cobra.Command) {
+	const (
+		long = `
+			Connect to the Postgres console
```
azazeal, line 22: this should't be tabbed; it'll be indented when printed to the output. [Here](https://github.com/superfly/flyctl/blob/79ff1a15d3d3eef7b15b27a084fb84ad6679cbe0/internal/cli/internal/command/apps/apps.go#L13) is a correct example

internal/cli/internal/command/services/postgres/connect.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+	)
+
+	return
+}
+
+func runConnect(ctx context.Context) error {
+	client := client.FromContext(ctx).API()
+
+	appName := flag.FirstArg(ctx)
```
azazeal, line 59: If you have a command which expects an app name, we've built helpers for that. You can check how `apps open` behaves, as an example, [here](https://github.com/superfly/flyctl/blob/79ff1a15d3d3eef7b15b27a084fb84ad6679cbe0/internal/cli/internal/command/apps/open.go).

internal/cli/internal/command/services/postgres/connect.go
```diff
@@ -0,0 +1,93 @@
...
+
+	app, err := client.GetApp(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("get app: %w", err)
```
azazeal, line 63: we should probably wrap this common error into its own class. for now please decorate as `apps open` [does](https://github.com/superfly/flyctl/blob/79ff1a15d3d3eef7b15b27a084fb84ad6679cbe0/internal/cli/internal/command/apps/open.go#L52):

```go
if err != nil {
	return fmt.Errorf("failed retrieving app %s: %w", appName, err)
}
```

internal/cli/internal/command/services/postgres/connect.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	agentclient, err := agent.Establish(ctx, client)
+	if err != nil {
+		return errors.Wrap(err, "can't establish agent")
```
azazeal, line 68: we don't use `errors.Wrap` but `fmt.Errorf` with `%w`

internal/cli/internal/command/services/postgres/connect.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	dialer, err := agentclient.Dialer(ctx, &app.Organization)
+	if err != nil {
+		return fmt.Errorf("failed to build tunnel for %s. %v", app.Organization.Slug, err)
```
azazeal, line 73: I think you added a `.` instead of a `:` before `%w`

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
+package ssh
+
```
azazeal, line 2: separate the imports please

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/helpers"
+	"github.com/superfly/flyctl/internal/client"
+	"github.com/superfly/flyctl/pkg/agent"
+	"github.com/superfly/flyctl/pkg/ssh"
+	"github.com/superfly/flyctl/terminal"
+)
+
+func spin(in, out string) context.CancelFunc {
+	ctx, cancel := context.WithCancel(context.Background())
```
azazeal, line 27: we don't use `context.Background`, in the new CLI we carry over the same context so that we can pickup cancellation.

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+		s.Start()
+		defer s.Stop()
+
+		<-ctx.Done()
+	}()
+
+	return cancel
+}
+
+type SSHParams struct {
```
azazeal, line 48: why are you exporting these params and prefixing them with `SSH`?

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+	Dialer         agent.Dialer
+	Cmd            string
+	Stdin          io.Reader
+	Stdout         io.WriteCloser
+	Stderr         io.WriteCloser
+	DisableSpinner bool
+}
+
+func SSHConnect(p *SSHParams, addr string) error {
+	terminal.Debugf("Fetching certificate for %s\n", addr)
```
azazeal, line 61: we don't use the `terminal` package anymore, `logger.FromContext` is what you're looking for.

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+
+	cert, err := singleUseSSHCertificate(p.Ctx, p.Org)
+	if err != nil {
+		return fmt.Errorf("create ssh certificate: %w (if you haven't created a key for your org yet, try `flyctl ssh establish`)", err)
```
azazeal, line 65: we also support custom error types which implement `Description() string` for more elaborate error printing. Ping me on slack and I'll guide you through that.

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	pk, err := parsePrivateKey(cert.Key)
+	if err != nil {
+		return errors.Wrap(err, "parse ssh certificate")
```
azazeal, line 70: we don't use `errors.Wrap`, please :)

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	if err := sshClient.Connect(context.Background()); err != nil {
+		return errors.Wrap(err, "error connecting to SSH server")
+	}
+	defer sshClient.Close()
+
+	terminal.Debugf("Connection completed.\n", addr)
+
+	if !p.DisableSpinner {
```
azazeal, line 101: I think we need to work on the spinners a bit more

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+
+	if err := sshClient.Shell(context.Background(), term, p.Cmd); err != nil {
+		return errors.Wrap(err, "ssh shell")
+	}
+
+	return nil
+}
+
+// // stolen from `mikesmitty`, thanks, you are a mikesmitty and a scholar
+func MarshalED25519PrivateKey(key ed25519.PrivateKey, comment string) []byte {
```
azazeal, line 120: I don't believe you should be exporting this

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+	magic := append([]byte("openssh-key-v1"), 0)
```
azazeal, line 121: this can easily be `magic := []byte("openssh-key-v1\x00")`

internal/cli/internal/command/ssh/ssh_terminal.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	hours := 1
+	return client.IssueSSHCertificate(ctx, org, user.Email, nil, &hours)
+}
+
+func parsePrivateKey(key64 string) (ed25519.PrivateKey, error) {
+	pkeys, err := base64.StdEncoding.DecodeString(key64)
+	if err != nil {
+		return nil, fmt.Errorf("API error: can't parse API-provided private key: %w", err)
```
azazeal, line 201: this isn't an API error and you're inside a helper function, you shouldn't be decorating the error here but at the caller.

azazeal (changes requested):

internal/cli/internal/command/ssh/ssh_terminal.go
davissp14: Got it, this entire file was copy and pasted over from the `cmd` package.

internal/cli/internal/command/ssh/ssh_terminal.go
davissp14: This was just a copy and paste from the cmd package to help get me unblocked.

davissp14:
Appreciate the review!  The connect file should be good to go, i added a TODO for the ssh_terminal file for the time being.  I just moved this over to unblock myself.

Issue #669, opened 2021-12-17
Improve `fly checks list` formatting.


davissp14:
The health checks listed via `fly checks list --app <postgres-app>` are not easy to read.  It would be awesome if we could make these more presentable and easier to read. 

<img width="820" alt="Screen Shot 2021-12-17 at 4 28 02 PM" src="https://user-images.githubusercontent.com/423038/146615214-912f82f4-8d5d-4856-85f7-8d5f688eb659.png">

redjonzaci:
@davissp14 @jsierles can we close this since @shortdiv's work is merged?

PR #766, opened 2022-01-25
fix: Update formatting for output in checks list

shortdiv:
Adds better formatting for the `fly checks list --app` command

Go playground for reference: https://go.dev/play/p/ch7rYe5h7GE

Addresses #669

### DIFF
--- a/cmd/checks.go
+++ b/cmd/checks.go
@@ -254,7 +254,7 @@ func runAppCheckList(cmdCtx *cmdctx.CmdContext) error {
 		nameFilter = api.StringPointer(val)
 	}
 
-	checks, err := cmdCtx.Client.API().GetAppHealthChecks(ctx, cmdCtx.AppName, nameFilter, nil, api.BoolPointer(true))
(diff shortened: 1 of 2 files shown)


mrkurt:
If you can think of a quick way to render the line breaks in the status output, that'd be amazing. This is good stuff that I'm guessing we'll use in machines too, at some point.

shortdiv:
Do you mean the status like "passing" or "failed", does it need a line break?

mrkurt:
:derp: I meant the `Output` field.

michaeldwan:
@shortdiv nice work! There's an option [on the check output field](https://github.com/superfly/flyctl/blob/master/cmd/checks.go/#L245) that compacts the result, it's true right now. Setting that to false will _probably_ do what you want. It might look garbled if there's a bunch of leading whitespace from indention, but that's not what we're optimizing for here.

jsierles:
Is this ready for merge?

shortdiv:
I'd say yes

PR #783, opened 2022-01-28
Add a Rails launcher which can launch a Postgres database

jsierles:
This launcher sets Rails app up for deployment using the Heroku buildpack. It also sets `RAILS_MASTER_KEY` and offers launching a Postgres database if the `postgresql` adapter is being used in `database.yml`. 

This PR also updates the inline Postgres launcher to allow developers to select the cluster size.

### DIFF
--- a/cmd/launch.go
+++ b/cmd/launch.go
@@ -323,6 +323,8 @@ func runLaunch(cmdCtx *cmdctx.CmdContext) error {
 					fmt.Errorf("Could not generate random string: %w", err)
 				}
 
+			} else if secret.Value != "" {
+				val = secret.Value
 			} else {
 				prompt := fmt.Sprintf("Set secret %s:", secret.Key)
 
--- a/cmd/launch.go
+++ b/cmd/launch.go
@@ -430,33 +430,33 @@ func runLaunch(cmdCtx *cmdctx.CmdContext) error {
 
 		clusterAppName := app.Name + "-db"
 
-		// Create a standalone Postgres in the same region as the app and organization
-		clusterInput := api.CreatePostgresClusterInput{
-			OrganizationID: org.ID,
-			Name:           clusterAppName,
-			Region:         api.StringPointer(region.Code),
-			ImageRef:       api.StringPointer("flyio/postgres"),
-			Count:          api.IntPointer(1),
-		}
-		payload, err := runApiCreatePostgresCluster(cmdCtx, org.Slug, &clusterInput)
+		cmdCtx.Config.Set("name", clusterAppName)
+		cmdCtx.Config.Set("region", region.Code)
+		cmdCtx.Config.Set("organization", org.Slug)
+
+		err = runCreatePostgresCluster(cmdCtx)
 
 		if err != nil {
 			err = fmt.Errorf("failed creating the Postgres cluster %s: %w", clusterAppName, err)
 			return err
 		}
 
-		// Reset the app name here beacuse AttachPostgresCluster sets it on the cmdCtx :/
-		cmdCtx.AppName = app.ID
-
 		cmdCtx.Config.Set("postgres-app", clusterAppName)
+
+		// Reset the app name here beacuse runCreatePostgresCluster overrides it
+		cmdCtx.AppName = app.ID
 		err = runAttachPostgresCluster(cmdCtx)
 
+		// Reset the app name here beacuse AttachPostgresCluster overrides it
+		cmdCtx.AppName = app.ID
+
 		if err != nil {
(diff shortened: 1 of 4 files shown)


bradgessler:
I tried this on my Rails project and the deploy failed during the release command:

```sh
==> Release command
Command: bundle exec rails db:migrate
	 Starting instance
	 Configuring virtual machine
	 Pulling container image
	 Unpacking image
	 Configuring firecracker
	 Preparing to run: `launcher bundle exec rails db:migrate` as heroku
	 2022/02/02 07:46:09 listening on [fdaa:0:386b:a7b:2295:7aa8:1a25:2]:22 (DNS: [fdaa::3]:53)
	 LoadError: Error loading the 'sqlite3' Active Record adapter. Missing a gem it depends on? sqlite3 is not part of the bundle. Add it to your Gemfile.
	 Gem::LoadError: sqlite3 is not part of the bundle. Add it to your Gemfile.
	 Tasks: TOP => db:migrate => db:load_config
	 Starting clean up.
Error Release command failed, deployment aborted
```

I think the source of the issue is the assumption made at: https://github.com/superfly/flyctl/pull/783/files#diff-57a9edf46bed4dfc6b1667fc93972e008bcb0de09d35f0434bcfdd31b268da35R220-R225

My database.yml file is untouched from Rails default:

```yml
# SQLite. Versions 3.8.0 and up are supported.
#   gem install sqlite3
#
#   Ensure the SQLite 3 gem is defined in your Gemfile
#   gem "sqlite3"
#
default: &default
  adapter: sqlite3
  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
  timeout: 5000

development:
  <<: *default
  database: db/development.sqlite3

# Warning: The database defined as "test" will be erased and
# re-generated from your development database when you run "rake".
# Do not set this db to the same as development or production.
test:
  <<: *default
  database: db/test.sqlite3

production:
  <<: *default
  database: db/production.sqlite3
```

On Heroku (and many other "12 factor" production environments that mimic Heroku) the `config/database.yml` file is essentially ignored and they automatically set a `DATABASE_URL` env var that points to a pg database.

If you want to support both sqlite3 and pg, you'd be better off asking the user what db engine they'd like to run instead of inferring from the `config/database.yml` file.

To go on a bit of a tangent, there are advantages to your operations by choosing sqlite3 as the default. On Heroku you always have to run `bundle add pg` and then move the `sqlite3` under `development, test` or remove it all together. There's a lot of support articles on their website about this, so I suspect it trips beginners up. That said, once you understand Heroku runs `pg` for their apps, its much easier getting them running.

If you don't want to offer a choice, I'd probably recommend setting the `DATABASE_URL` to a pg database and verifying the `pg` gem is present in the Gemfile.

bradgessler:
A separate issue: it looks like https://github.com/superfly/flyctl/pull/783/files#diff-57a9edf46bed4dfc6b1667fc93972e008bcb0de09d35f0434bcfdd31b268da35R218 is run once during the first deployment of the rails application. Assuming that's true, I believe you want to run `rake db:schema:load`. That loads the contents of the `config/db/schema.rb` file into the database as opposed to running migrations in serial.

Subsequent deployments would then run `rake db:migrate`.

jsierles:
Thanks for the report! Indeed this is not ideal.

So we check for `postgres` in `database.yml` to decide whether to ask about Postgres. But I think it's better just to always offer Postgres and try to add the pg gem for you.

We can support sqlite in production, but prefer to push people towards Postgres given it's more flexible. With SQLite you're stuck with one instance and that's not obvious to newcomers.

bradgessler:
Good call on NOT offering sqlite3 as an option - that would probably just confuse newcomers, especially as they collect more data and wonder why things "are acting weird".

Something to consider re: adding `pg` to the Gemfile. Doing that without asking or informing newcomers might do weird things to the difference between their dev/test sqlite3 envs and their production pg env.

The good news is that there's a `bundle add pg --group production` command you can use if you go down that route that means the go code you write won't have to reason about the Gemfile. I'm not sure how familiar you are with Ruby, so forgive me if you already know this, but check out the documentation for that command. I barely see the Ruby community using it in their README files.

Another thing you should consider that Heroku does NOT do automatically is adding the linux platform to `Gemfile.lock`. If you develop a Rails application on Windows or macOS, the deploy always breaks and Heroku tells you in an error message to run the bundle command that adds the Linux platform to the lock file. This is probably another one of those things you could just do for the file via this command:

```
bundle lock --add-platform x86_64-linux
```

If I think of any other defaults that will make this a more pleasant out-of-the-box experience, I'll drop it here.

Thanks so much for the work you're doing on this!

bradgessler:
Oh also, I agree with your approach to migrations since you clarified those migrations are run per-deploy. Commenting it out when the application gets bigger is 👍.

PR #787, opened 2022-01-31
slugify hostname when using for interactive wireguard tunnel

jipperinbham:
Currently, if a user has their hostname set to anything that is not DNS safe it will error out when trying to create the wireguard tunnel:

```
error establishing wireguard connection for personal organization: name must consist solely of letters, numbers, and the dash character
```

The following change utilizes the same regular expression for slugifying the email for the return call to `os.Hostname()`.

### DIFF
--- a/internal/wireguard/wg.go
+++ b/internal/wireguard/wg.go
@@ -20,6 +20,10 @@ import (
 	"golang.org/x/crypto/curve25519"
 )
 
+var (
+	cleanDNSPattern = regexp.MustCompile(`[^a-zA-Z0-9\\-]`)
+)
+
 func StateForOrg(apiClient *api.Client, org *api.Organization, regionCode string, name string) (*wg.WireGuardState, error) {
 	state, err := getWireGuardStateForOrg(org.Slug)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/wireguard/wg.go
```diff
@@ -55,17 +59,20 @@ func Create(apiClient *api.Client, org *api.Organization, regionCode, name strin
...
 			return nil, err
 		}
-		host, _ := os.Hostname()
+		emailSlug := cleanDNSPattern.ReplaceAllString(user.Email, "-")
+
+		host, err := os.Hostname()
+		if err != nil {
+			return nil, err
+		}
+		hostSlug := cleanDNSPattern.ReplaceAllString(strings.Split(host, ".")[0], "-")
```
azazeal, line 68: I think `host` can be an empty string - would you mind taking the first part of `host` only if `host` isn't empty?

azazeal (changes requested):

internal/wireguard/wg.go
azazeal: No need, the next version of the CLI ensures a hostname is present. For now, and just for _not_ having the possibility of a panic in `strings.Split(host, ".")[0]` just assign to hostSlug only if `host` isn't empty.

jipperinbham: `strings.Split` would not panic on an empty string, [example](https://go.dev/play/p/J-gNDibfcZM)

azazeal: Sorry, meant the `[0]` part but obviously you still get back a non-empty slice (containing at least 1 element). Sorry for the brain outage!

azazeal (approved):
LGTM - sorry for the fuss.

PR #794, opened 2022-02-01
Exposing restart command for PG on machines

davissp14:
Exposes `DEV=1 flyctl services postgres restart --app <app-name>`

Under the hood, this will run `pg_ctl -m fast -D /path/to/postgres restart` against each Postgres instance.

### DIFF
--- a/internal/cli/internal/command/services/postgres/attach.go
+++ b/internal/cli/internal/command/services/postgres/attach.go
@@ -15,7 +15,6 @@ import (
 	"github.com/superfly/flyctl/internal/cli/internal/flag"
 	"github.com/superfly/flyctl/internal/cli/internal/prompt"
 	"github.com/superfly/flyctl/internal/client"
-	"github.com/superfly/flyctl/pkg/agent"
 	"github.com/superfly/flyctl/pkg/iostreams"
 )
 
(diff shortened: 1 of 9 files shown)


jsierles:
Nice! What happens if this command fails in the middle of a restart?

davissp14:
> What happens if this command fails in the middle of a restart?

In the event of a failed restart I am currently exiting.  Under-the-hood the restart process is calling `pg_ctl  -D /data/postgres restart`, so while there's certainly opportunities for failure, it should be pretty reliable.   

I think the more interesting failure case is with the failover process.  There are some conditions where this expected to fail, the obvious failure case is when there's no replicas eligible for promotion.  In this case, we have one of two options:

1. We exit the process completely. 
2. We treat the failover process as "best-effort" and restart the primary anyways. ( restart's are pretty fast when issued via pg_ctl ).

jsierles:
I was referring more to network conditions - this does seem like a legit concern to me. If someone is on a spotty connection and tries to run an operation like this, a cluster could be left half upgraded. A user may not understand that the operation is not orchestrated by our backend. In the worst case, would a second attempt function as expected?

davissp14:
This is not a huge problem with this particular operation as it's just issuing restarts, not upgrades.  Worst case, it could be re-run with minimal disruption.

> I was referring more to network conditions - this does seem like a legit concern to me. If someone is on a spotty connection and tries to run an operation like this, a cluster could be left half upgraded

Yeah, this is a bigger issue with flyctl based orchestration and unfortunately not an issue that will be solved by this PR.

jsierles:
Yeah, cool. Not least, I asked to clarify my own thoughts on launchers - as long as commands can be rerun, we're in a good enough place!

PR #800, opened 2022-02-02
feat: Add status command to machines

shortdiv:
This PR copies the status command from VM and moves it over the machines, since we haven't yet discussed what we want the output for machine status to be, this is an exact replica of what `fly vm status <image-id>` output

terminal screen with output

### DIFF
--- a/docstrings/gen.go
+++ b/docstrings/gen.go
@@ -485,6 +485,10 @@ to all instances running in a specific region using the --region/-r flag.`,
 		return KeyStrings{"stop <id>", "Stop a Fly machine",
 			`Stop a Fly machine`,
 		}
+	case "machine.status":
+		return KeyStrings{"status <id>", "Get status on a fly machine",
+			`Get status for a Fly machine`,
+		}
 	case "monitor":
 		return KeyStrings{"monitor", "Monitor deployments",
 			`Monitor application deployments and other activities. Use --verbose/-v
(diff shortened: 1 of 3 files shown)


shortdiv:
Not sure why tests are failing, I added the command to the docs file which I believe is where the tests are pulling from

docstrings/gen.go
```diff
@@ -485,6 +485,10 @@ to all instances running in a specific region using the --region/-r flag.`,
 		return KeyStrings{"stop <id>", "Stop a Fly machine",
 			`Stop a Fly machine`,
 		}
+	case "machine.status":
```
rugwirobaker, line 488: make test is deleting your changes to this file cause it's generated from this file:
https://github.com/superfly/flyctl/blob/b117369194768c80f0d31c4303518fcf296b36d5/helpgen/flyctlhelp.toml#L1009-L1040

shortdiv:
🙌

jsierles:
What do we think about trying to pull these commands under the new CLI format?

rugwirobaker:
@jsierles I would be happy about it. After working in the new format it feels slow to work in this one, especially when it comes to rendering things

rugwirobaker:
@shortdiv once this merged this should help you get  a machine status: https://github.com/superfly/flyctl/pull/806

shortdiv:
Yeah I think we should merge this and then work on those updates later @jsierles

PR #803, opened 2022-02-03
attempt to guess $DOCKER_HOST unless correctly specified

azazeal:
This PR refactors the boot process of the CLI so that it attempts to guess the actual value partially defined `$DOCKER_HOST` values should have.


rugwirobaker (approved):
Works for both my usecases where I have specified the protocol and when I have not.

PR #812, opened 2022-02-07
flyctl machine status

rugwirobaker:
`flyctl machine status` so far:
Screen Shot 2022-02-07 at 16 09 27


shortdiv (approved):
Looks good to me!

rugwirobaker:
Update: @shortdiv what do you think of a diff field here?
Screen Shot 2022-02-10 at 18 13 51

shortdiv:
I don't think the diff object really makes sense, and it's apparently difficult to retrieve from the database into graphql. Instead of showing diff, we should show machine event types, from start/exit/destroy i.e.`exit_code`, `oom_killed` and `requested_stop`, you can see available fields here: https://github.com/superfly/web/blob/master/app/graphql/types/machine_event_exit.rb

cmd/presenters/machine_events.go
```diff
@@ -18,18 +19,26 @@ func (p *MachineEvents) APIStruct() interface{} {
...
+		if event.ExitCode != nil {
+			exitCode = fmt.Sprintf("%d", *event.ExitCode)
+		}
+
 		out = append(out, map[string]string{
+			"Id":        event.ID,
 			"Kind":      event.Kind,
 			"Timestamp": event.Timestamp.Format(time.RFC3339),
-			// "Message":   event.Message,
+			`Exit Code`: exitCode,
```
shortdiv, line 41: [nit] Same as above, consistency on using "" vs ``for strings

api/types.go
```diff
@@ -1231,7 +1232,43 @@ type DeleteOrganizationMembershipPayload struct {
...
+	Timestamp time.Time
+	ExitCode  *int
+	OOMKilled bool
+	// Body      map[string]interface{}
+}
+
+func (e *MachineEvent) UnmarshalJSON(data []byte) error {
+	raw := map[string]interface{}{}
+
+	if err := json.Unmarshal(data, &raw); err != nil {
```
shortdiv, line 1246: I think unmarshalling the data directly into the struct would save a lot of trouble when it comes to type assertions, something like:

```
var raw MachineEvent
if err := json.Unmarshal(data, &raw); err != nil {
  return err
}
```

Which will then allow you to just directly check for data like raw.ID etc

api/types.go
```diff
@@ -1231,7 +1232,43 @@ type DeleteOrganizationMembershipPayload struct {
...
+		return err
+	}
+
+	e.ID = raw["id"].(string)
+	delete(raw, "id")
```
shortdiv, line 1251: [nit] What's the reason for deleting data from the map?

api/types.go
rugwirobaker: MachineEvent takes more than one form such as `MachineEventStart`, `MachineEventExit`... and they don't share all the fields and so I must be more generic and manually check if the fields exists before trying to set them.

api/types.go
rugwirobaker: Nothing special, just a way of keeping only the data that has not yet been set unto a field.

shortdiv: Ah ok, I think this is probably stylistic, but I don't see the point of it if the object gets garbage collected anyway once the function returns

api/types.go
shortdiv: On second thought I think a case switch could work, something like this: https://play.golang.com/p/49xJe_HysXb

api/types.go
```diff
@@ -1235,40 +1235,35 @@ type MachineEvent struct {
...
 func (e *MachineEvent) UnmarshalJSON(data []byte) error {
-	raw := map[string]interface{}{}
+	var genericMachineEvent MachineEvent
+	var machineStop MachineEventStop
 
-	if err := json.Unmarshal(data, &raw); err != nil {
+	if err := json.Unmarshal(data, &genericMachineEvent); err != nil {
 		return err
 	}
+	machineBodyinBytes, _ := json.Marshal(genericMachineEvent.Metadata)
```
rugwirobaker, line 1253: This right here causes a stack overflow:
```bash
 go run . machine status 71d5c016 --app <app-name>
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0x14020700550 stack=[0x14020700000, 0x14040700000]
fatal error: stack overflow

```

PR #822, opened 2022-02-09
Retry SSH connections in case they fail due to a delayed root key installation

### LINKED ISSUE #807: SSH connections can fail when racing against delayed distribution of SSH root keys
Recently, a backend change was made that creates a new SSH root key during an org's first attempt to use SSH. The key needs to be published over Consul, which can be delayed, so we've started seeing exceptions like this:

https://sentry.io/organizations/flyio/issues/2782030248

To fix this, we might want to check DNS for the existence of the `TXT _orgcert.internal` record before attempting the connection.

jsierles:
May fix #807


jsierles:
Closing in favor of a DNS-based approach.

PR #825, opened 2022-02-09
Run pg specific commands only on confirmed pg apps

### LINKED ISSUE #818: Listing postgres databases fails when pointing to a non-postgres Fly app
See #597 for details.

Flyctl should check if the target app is a postgres instance/cluster before trying to run this command.

rugwirobaker:
* `flyctl pg users list`
* `flyctl pg db list`

Fixes #818

### DIFF
--- a/cmd/postgres.go
+++ b/cmd/postgres.go
@@ -679,8 +679,7 @@ func runListPostgresDatabases(cmdCtx *cmdctx.CmdContext) error {
 		return fmt.Errorf("get app: %w", err)
 	}
 
-	// check if app is a postgres app
-	if app.ImageDetails.Repository != "flyio/postgres" && app.ImageDetails.Repository != "flyio/postgres-standalone" {
+	if !isPostgresApp(&app.ImageDetails) {
 		return fmt.Errorf("%s is not a postgres app", cmdCtx.AppName)
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/postgres.go
```diff
@@ -679,6 +679,11 @@ func runListPostgresDatabases(cmdCtx *cmdctx.CmdContext) error {
 		return fmt.Errorf("get app: %w", err)
 	}
 
+	// check if app is a postgres app
+	if app.ImageDetails.Repository != "flyio/postgres" && app.ImageDetails.Repository != "flyio/postgres-standalone" {
```
jsierles, line 683: We probably shouldn't duplicate this code, especially with hardcoded values. Can this be moved to a function?

docstrings/gen.go
```diff
@@ -481,14 +481,14 @@ to all instances running in a specific region using the --region/-r flag.`,
 		return KeyStrings{"start <id>", "Start a Fly machine",
 			`Start a Fly machine`,
 		}
+	case "machine.status":
```
jsierles, line 484: These changes should be coming from master.

jsierles (changes requested):

jsierles (approved):

PR #832, opened 2022-02-10
Handle cancelation while establishing ssh connection

### LINKED ISSUE #830: Can't cancel `ssh console` while waiting on host
SIGINT isn't canceling an `ssh console` connection attempt.

```
$ fly ssh console -a app-name
Connecting to top1.nearest.of. app-name.internal...⣯^C
Connecting to top1.nearest.of. app-name.internal...⣷^C
Connecting to top1.nearest.of. app-name.internal...⣽^C
Connecting to top1.nearest.of. app-name.internal...⣽^C
```

michaeldwan:
Fixes #830 

Possibly relevant issue https://github.com/golang/go/issues/20288

Seems like cancelation isn't being handled in `ssh.NewClientConn`. Comments on that issue indicate that cancelation will happen if the `net.Conn` being passed in is correctly closed on cancel. This is _likely_ not happening as needed due to the way we pass in a connection from the agent's dialer. 

Handling cancelation independently solves the issue for now.

### DIFF
--- a/pkg/ssh/client.go
+++ b/pkg/ssh/client.go
@@ -32,6 +32,12 @@ func (c *Client) Close() error {
 	return nil
 }
 
+type connResp struct {
+	err    error
+	conn   ssh.Conn
+	client *ssh.Client
+}
+
 func (c *Client) Connect(ctx context.Context) error {
 	pubKey, _, _, _, err := ssh.ParseAuthorizedKey([]byte(c.Certificate))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/ssh/client.go
```diff
@@ -67,13 +73,34 @@ func (c *Client) Connect(ctx context.Context) error {
...
-	conn, chans, reqs, err := ssh.NewClientConn(tcpConn, tcpConn.RemoteAddr().String(), conf)
-	if err != nil {
-		return err
-	}
+	respCh := make(chan connResp)
 
-	c.conn, c.client = conn, ssh.NewClient(conn, chans, reqs)
-	return nil
+	// ssh.NewClientConn doesn't take a context, so we need to handle cancelation on our end
+	go func() {
```
azazeal, line 79: it'd be far simpler instead of launching this in the background, and since you already have a reference to the underlying connection, to just close it on context cancellation instead of this.

pkg/ssh/client.go
```diff
@@ -67,13 +73,34 @@ func (c *Client) Connect(ctx context.Context) error {
...
+			respCh <- connResp{err: err}
+			return
+		}
+
+		client := ssh.NewClient(conn, chans, reqs)
+
+		respCh <- connResp{nil, conn, client}
+	}()
+
+	for {
```
azazeal, line 91: why are you looping?

PR #835, opened 2022-02-14
Update type of $appName

shortdiv:
This fixes the type issue `Error failed launching postgres: Type mismatch on variable $appName and argument appId (String! / ID)` 

See https://flyio.slack.com/archives/C02PD4DUL2J/p1644863139676659 for context


davissp14 (approved):
Thank you!

PR #836, opened 2022-02-15
Use nginx for RedwoodJS apps for serving assets and proxying API requests

jsierles:
This is required because Redwood needs all non-API requests to be rewritten to `index.html`. It has some upsides, like getting cache headers, and maybe lightweight HTTP caching out of the box if desired.


jsierles:
Superseded by #853

PR #837, opened 2022-02-16
feat: add placeholder buildargs

### LINKED ISSUE #778: Inject placeholder secrets/envs into build vars
Folks get tripped up when builds fail because a secret or env var available at runtime isn't available at build time. For example: https://community.fly.io/t/ruby-version-mismatch/3815/4?u=michael

It's almost always a library throwing an exception when a required env var is missing, like `DATABASE_URL` or `SECRET_KEY_BASE`. We're not going to bring real secrets/env in for obvious reasons, but maybe we can add placeholders for the current secrets and env.  For example, for each secret and env var, make a build-arg with a value like `<placeholder-env>` or `<placeholder-secret>`. Dumb existence checks would "just work", and other errors would at least give some indication of what the problem is.

me-diru:
- fixes #778

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -96,8 +96,30 @@ func runDeploy(cmdCtx *cmdctx.CmdContext) error {
 			return err
 		}
 	} else {
+		buildArgs := make(map[string]string)
 
-		var buildArgs map[string]string
+		// Add placeholder build-args for secrets
+		secrets, err := cmdCtx.Client.API().GetAppSecrets(ctx, cmdCtx.AppName)
+		if err != nil {
+			return err
+		}
+
+		for _, secret := range secrets {
+			buildArgs[secret.Name] = "<placeholder-secret>"
+		}
+
+		// Add placeholder build-args for envs
+		var env map[string]interface{}
+
+		if rawEnv, ok := cmdCtx.AppConfig.Definition["env"]; ok {
+			if castEnv, ok := rawEnv.(map[string]interface{}); ok {
+				env = castEnv
+			}
+		}
+
+		for key := range env {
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -96,8 +96,30 @@ func runDeploy(cmdCtx *cmdctx.CmdContext) error {
 			return err
 		}
 	} else {
+		buildArgs := make(map[string]string)
 
-		var buildArgs map[string]string
+		// Add placeholder build-args for secrets
+		secrets, err := cmdCtx.Client.API().GetAppSecrets(ctx, cmdCtx.AppName)
```
me-diru, line 102: tested this one locally with `secrets set`

cmd/deploy.go
```diff
@@ -96,8 +96,30 @@ func runDeploy(cmdCtx *cmdctx.CmdContext) error {
...
+		var env map[string]interface{}
+
+		if rawEnv, ok := cmdCtx.AppConfig.Definition["env"]; ok {
+			if castEnv, ok := rawEnv.(map[string]interface{}); ok {
+				env = castEnv
+			}
+		}
+
+		for key := range env {
+			buildArgs[key] = "<placeholder-env>"
```
me-diru, line 121: tested by passing in fly.toml and `flyctl launch`

me-diru:
I think this is ready to go, lmk your thoughts :D @michaeldwan @rugwirobaker

jsierles (approved):

jsierles:
This needs to be implemented for the new deployment command: https://github.com/superfly/flyctl/blob/master/internal/command/deploy/deploy.go

PR #838, opened 2022-02-17
feat: remind to add alternate certificate

### LINKED ISSUE #736: Post-create instructions for "example.com" certificate creation
People almost always want both `example.com` and `www.example.com` when they add one or the other, but it's not obvious they need to manually configure both. We'd save them a lot of agony if we just told them that. Something like:

> You've just created a certificate for `example.com`, make sure you create another certificate for `www.example.com`

me-diru:
fixes #736

### DIFF
--- a/cmd/certificates.go
+++ b/cmd/certificates.go
@@ -17,6 +17,15 @@ import (
 	"golang.org/x/net/publicsuffix"
 )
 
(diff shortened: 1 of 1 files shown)


me-diru:
probably I think we should check if a certificate for the alternate domain is issued or not before the reminder

PR #844, opened 2022-02-22
Status support for machines

rugwirobaker:
Screen Shot 2022-02-22 at 17 50 37

### DIFF
--- a/api/resource_apps.go
+++ b/api/resource_apps.go
@@ -67,6 +67,7 @@ func (client *Client) GetApp(ctx context.Context, appName string) (*App, error) 
 				status
 				version
 				appUrl
+				platformVersion
 				config {
 					definition
 				}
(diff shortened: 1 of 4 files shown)


rugwirobaker:
Missing:
* Desired state
* Displaying the role alongside the state(from metadata?)
* Number of restarts
* Watching machines state transitions(stopped->started and everything in between).

api/resource_apps.go
```diff
@@ -95,6 +96,30 @@ func (client *Client) GetApp(ctx context.Context, appName string) (*App, error)
 					repository
 					version
 				}
+				machines{
```
jsierles, line 99: Appending this to the app query may end up being unwieldy compared to directly querying machines. But I think it's OK for now until we have a better strategy for GraphQL.

jsierles (approved):
Looks OK to me.

PR #853, opened 2022-03-01
fix: Use `rw-server` for RedwoodJS apps

callingmedic911:
Related: #836 

This PR replaces the usage of `rw-api-server` with `rw-server` (on port 8910) which serves both API and Web side of a Redwood app with fastify. `rw-server` is configured to redirect the unknown requests to index.html.

Internal discussion: https://redwoodjs.slack.com/archives/C01R3RYH4LW/p1646076378318309 Gist is: Running fastify for the web side may not be ideal for all cases but it keeps things simple as a start config. For advanced use cases, individual projects can configure ngnix.

Looping in: @dthyresson @dac09 @jsierles

### DIFF
--- a/internal/sourcecode/templates/redwood/Dockerfile
+++ b/internal/sourcecode/templates/redwood/Dockerfile
@@ -39,7 +39,5 @@ COPY --from=api_build /app/api/dist /app/api/dist
 COPY --from=api_build /app/api/db /app/api/db
 COPY --from=api_build /app/node_modules/.prisma /app/node_modules/.prisma
 
-EXPOSE 8910
-
 # Entrypoint to @redwoodjs/api-server binary
 CMD [ "yarn", "rw-server", "--port", "8910" ]
--- a/internal/sourcecode/templates/redwood/Dockerfile
+++ b/internal/sourcecode/templates/redwood/Dockerfile
@@ -39,7 +39,7 @@ COPY --from=api_build /app/api/dist /app/api/dist
 COPY --from=api_build /app/api/db /app/api/db
 COPY --from=api_build /app/node_modules/.prisma /app/node_modules/.prisma
 
-EXPOSE 8911
+EXPOSE 8910
 
 # Entrypoint to @redwoodjs/api-server binary
-CMD [ "yarn", "rw-api-server", "--port", "8911", "--rootPath", "/api" ]
+CMD [ "yarn", "rw-server", "--port", "8910" ]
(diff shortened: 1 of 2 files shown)


internal/sourcecode/templates/redwood/Dockerfile
```diff
@@ -39,7 +39,7 @@ COPY --from=api_build /app/api/dist /app/api/dist
 COPY --from=api_build /app/api/db /app/api/db
 COPY --from=api_build /app/node_modules/.prisma /app/node_modules/.prisma
 
-EXPOSE 8911
+EXPOSE 8910
```
jsierles, line 42: On Fly this line is not used at all, so we could get rid of it. It might be useful for other platforms, though.

internal/sourcecode/templates/redwood/Dockerfile
```diff
@@ -39,7 +39,7 @@ COPY --from=api_build /app/api/dist /app/api/dist
...
 
 # Entrypoint to @redwoodjs/api-server binary
-CMD [ "yarn", "rw-api-server", "--port", "8911", "--rootPath", "/api" ]
+CMD [ "yarn", "rw-server", "--port", "8910" ]
```
jsierles, line 45: Do we need any more CLI options for the rewrites to work?

jsierles (approved):
Anything else need doing here?

internal/sourcecode/templates/redwood/Dockerfile
callingmedic911: It should be able to figure out the root path from redwood.toml automatically:

https://github.com/redwoodjs/redwood/blob/902b36d4a3907e7e1a99930769c539b15d3c3b6f/packages/api-server/src/cliHandlers.ts#L89

internal/sourcecode/templates/redwood/Dockerfile
callingmedic911: Considering it's generated from flyctl launch command, it makes sense to keep it minimal then. Removed this line.

callingmedic911:
> Anything else need doing here?

I was thinking about what @dac09 said, even though this should give a good starting point for Redwood project, should we document someplace that it's a good idea to deploy web and API side separately? Maybe just a comment. What do you think?

jsierles:
For now I think we're good. I don't want to encourage deploying API and web separately on Fly, as it adds complexity and cost. We'll get to a place where that could be handled automatically if someone wants it.

I'll give this a Spin and if it looks good, merge it. I want to see if we get cache-control headers on assets, which would be desirable for browser cache or running through a CDN.

PR #859, opened 2022-03-08
Adding flaps endpoints for machines

davissp14:


internal/cli/internal/command/machines/flaps.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return &FlapsClient{
+		ctx:       ctx,
+		app:       app,
+		peerIP:    resolvePeerIP(dialer.State().Peer.Peerip),
+		authToken: flyctl.GetAPIToken(),
+	}, nil
+}
+
+func (f *FlapsClient) Launch(builder api.LaunchMachineInput) (*Machine, error) {
```
jsierles, line 55: Should we be passing the context down to these methods, and to the client? For cancelability, and also to use iostreams instead of printing directly to stdout.

jsierles (commented):
Nice one! Just left a comment.

internal/cli/internal/command/machines/flaps.go
jsierles: Also, might we derive value from the flaps package existing on its own in `pkg` instead of inside the command structure?

PR #860, opened 2022-03-09
Deprecate the old top-level list command

jsierles:
This command overlaps with `fly apps list` and `fly orgs list`.


jsierles:
Closing in favor of simply hiding commands.

PR #862, opened 2022-03-10
Port machines to new cli structure

rugwirobaker:
- Port machine list
- Port machine stop
- Port machine start
- Port machine kill
- Port machine clone
- Port machine remove

### DIFF
--- a/internal/cli/internal/command/machine/clone.go
+++ b/internal/cli/internal/command/machine/clone.go
@@ -1,0 +1,51 @@
+package machine
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/internal/cli/internal/command"
+	"github.com/superfly/flyctl/internal/cli/internal/flag"
+)
+
+func newClone() *cobra.Command {
+	const (
+		short = "Clone a machine"
+		long  = short + "\n"
+
+		usage = "clone"
+	)
+
+	cmd := command.New(usage, short, long, runMachineClone,
+		command.RequireSession,
+		command.RequireAppName,
+	)
+
+	cmd.Args = cobra.NoArgs
+
+	flag.Add(
+		cmd,
+		flag.String{
+			Name:        "region",
+			Shorthand:   "r",
+			Description: "Target region",
+		},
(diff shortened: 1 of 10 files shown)


shortdiv (commented):
Looks good so far, good work 👍

internal/cli/internal/command/machine/clone.go
```diff
@@ -0,0 +1,154 @@
...
+	)
+
+	cmd.Args = cobra.ExactArgs(1)
+
+	flag.Add(
+		cmd,
+		flag.App(),
+		flag.AppConfig(),
+		flag.String{
+			Name:        "region",
```
jsierles, line 46: Region and organization flahs have extracted methods in `flag.go`. We should use these to keep our behavior and documentation consistent across commands.

internal/cli/internal/command/machine/clone.go
```diff
@@ -0,0 +1,154 @@
...
+			Shorthand:   "r",
+			Description: "Target region",
+		},
+		flag.String{
+			Name:        "name",
```
jsierles, line 51: Should this be called ID? I'm not clear on what the difference is between the machine alloc ID and its name.

internal/cli/internal/command/machine/list.go
```diff
@@ -0,0 +1,109 @@
...
+	)
+
+	state := flag.GetString(ctx, "state")
+	if flag.GetBool(ctx, "all") {
+		state = ""
+	}
+
+	machines, err := client.ListMachines(ctx, appName, state)
+	if err != nil {
+		return fmt.Errorf("could not get lisyt of machines: %w", err)
```
jsierles, line 70: typo `list`

jsierles (changes requested):
Looks good - just recommended a few simple changes.

internal/cli/internal/command/machine/clone.go
rugwirobaker: I can't tell either

shortdiv: I think since the goal of this PR is to port, we can keep this convention and update later

PR #870, opened 2022-03-15
Launch Rails apps with default production environment variables

jsierles:
`RAILS_ENV` and `RAILS_LOG_TO_STDOUT` should probably be set up front.

### DIFF
--- a/internal/sourcecode/scan.go
+++ b/internal/sourcecode/scan.go
@@ -202,6 +202,11 @@ func configureRails(sourceDir string) (*SourceInfo, error) {
 				Value: string(masterKey),
 			},
 		}
+
+		s.Env = map[string]string{
+			"RAILS_ENV":           "production",
+			"RAILS_LOG_TO_STDOUT": "true",
+		}
 	}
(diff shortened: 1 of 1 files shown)


jsierles:
Yeah - I think it's good we set it here anyway. In general, our goal is to start moving configuration like this closer to the user so they're aware of where and how to customize it.

internal/sourcecode/scan.go
```diff
@@ -202,6 +202,11 @@ func configureRails(sourceDir string) (*SourceInfo, error) {
 				Value: string(masterKey),
 			},
 		}
+
+		s.Env = map[string]string{
+			"RAILS_ENV":           "production",
+			"RAILS_LOG_TO_STDOUT": "true",
+		}
```
getaaron, line 209: @jsierles Why is this only included if `config/master.key` is set? They seem unrelated - should they be outside the conditional?

jsierles: Good catch! I pushed a fix.

PR #871, opened 2022-03-15
Revert "Add region selection when creating app"

jeromegn:
Reverts superfly/flyctl#868


jeromegn:
Without more context, this change seems unnecessary and is breaking continuous integration.

PR #876, opened 2022-03-22
feat: push --build-only image

### LINKED ISSUE #873: allow fly deploy --build-only to push an image
It is sometimes desirable to build an image in one step then deploy it in another. `fly deploy --build-only` doesn't push, though. So when you use it with a remote builder you can't get access to the image.

me-diru:
- fixes #873


me-diru:
There are few other places where `Publish` is disabled when --build-only Is passed, but the change done in this PR was enough to push images after build is done.

jsierles:
I think it would make sense to make this optional rather than the default.

PR #879, opened 2022-03-24
Add flaps to start and stop machine

shortdiv:
Add machine start and stop command to use flaps

Machine Start:
`go run . machine start 024d8941b45879 --app testingtestingtesting`

Machine Stop:
`go run . machine stop 024d8941b45879 --app testingtestingtesting`

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -1194,6 +1194,17 @@ type Machine struct {
 	CreatedAt time.Time
 }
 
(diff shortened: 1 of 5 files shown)


jipperinbham (approved):
We could call `flapsClient.Wait` after the `start` command but that could be done in a follow-up PR to improve UX a bit.

PR #893, opened 2022-03-30
Feature:Postgres commands through http

rugwirobaker:
- Add postgres http client
- List postgres resources through http

### DIFF
--- a/cmd/postgres.go
+++ b/cmd/postgres.go
@@ -21,6 +21,7 @@ import (
 	"github.com/superfly/flyctl/helpers"
 	"github.com/superfly/flyctl/internal/client"
 	"github.com/superfly/flyctl/pkg/agent"
+	"github.com/superfly/flyctl/pkg/flypg"
 )
 
 type PostgresConfiguration struct {
@@ -675,9 +676,9 @@ func runListPostgresDatabases(cmdCtx *cmdctx.CmdContext) error {
 		return fmt.Errorf("get app: %w", err)
 	}
 
-	if !isPostgresApp(&app.ImageDetails) {
-		return fmt.Errorf("%s is not a postgres app", cmdCtx.AppName)
-	}
+	// if !isPostgresApp(&app.ImageDetails) {
+	// 	return fmt.Errorf("%s is not a postgres app", cmdCtx.AppName)
+	// }
 
 	agentclient, err := agent.Establish(ctx, cmdCtx.Client.API())
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/postgres.go
```diff
@@ -675,9 +676,9 @@ func runListPostgresDatabases(cmdCtx *cmdctx.CmdContext) error {
 		return fmt.Errorf("get app: %w", err)
 	}
 
-	if !isPostgresApp(&app.ImageDetails) {
-		return fmt.Errorf("%s is not a postgres app", cmdCtx.AppName)
-	}
+	// if !isPostgresApp(&app.ImageDetails) {
```
davissp14, line 679: Are we checking this somewhere else?

cmd/postgres.go
```diff
@@ -733,24 +734,30 @@ func runListPostgresUsers(cmdCtx *cmdctx.CmdContext) error {
 
 	dialer, err := agentclient.Dialer(ctx, app.Organization.Slug)
 	if err != nil {
-		return fmt.Errorf("ssh: can't build tunnel for %s: %s\n", app.Organization.Slug, err)
+		return fmt.Errorf("ssh: can't build tunnel for %s: %s", app.Organization.Slug, err)
 	}
 
-	pgCmd := NewPostgresCmd(cmdCtx, app, dialer)
+	// pgCmd := NewPostgresCmd(cmdCtx, app, dialer)
```
davissp14, line 740: We can probably just remove the commented code.

pkg/flypg/http.go
```diff
@@ -0,0 +1,120 @@
...
+
+	req = req.WithContext(ctx)
+
+	res, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	if res.Status > "299" {
```
davissp14, line 52: What are the expected status codes?  Also, looks like you're comparing an `int` with a `string` here.

pkg/flypg/pg.go
```diff
@@ -0,0 +1,146 @@
...
+		return false, err
+	}
+
+	if out.Result.Username == name {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (c *Client) GrantAccess(ctx context.Context, dbName, userName string) error {
```
davissp14, line 120: This should probably be coupled with CreateUser.  Given the complexity surrounding user permissions, we shouldn't expose this as its own endpoint.

pkg/flypg/pg.go
```diff
@@ -0,0 +1,146 @@
...
+		Username: userName,
+	}
+
+	if err := c.Do(ctx, http.MethodPost, endpoint, in, nil); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Client) RevokeAccess(ctx context.Context, dbName, userName string) error {
```
davissp14, line 134: This should probably be coupled with DeleteUser.  Like GrantAccess, let's not expose this as its own endpoint.

davissp14 (changes requested):
Overall, looks great!  Just a few notes.

cmd/postgres.go
rugwirobaker: @davissp14 I only removed this temporary for testing, cause pg apps deployed with `fly deploy` fail to pass this checks

cmd/postgres.go
rugwirobaker: Yeah will do, Just wanted to get a quick reaction before going to far with it.

pkg/flypg/http.go
rugwirobaker: True I'll work on it.

pkg/flypg/pg.go
rugwirobaker: Will do

pkg/flypg/http.go
davissp14: Will we be working with any valid status codes, other than 200 and 201?

davissp14 (approved):
Great work!  We just need to ensure that the version is bumped in `postgres-ha` before this gets merged.

PR #913, opened 2022-04-09
Create NuxtJS launcher

lubien:
This will make launching a nuxtjs app a single command task.

Later on we can add ourselves to https://nuxtjs.org/deployments/

### DIFF
--- a/internal/sourcecode/templates/nuxtjs/Dockerfile
+++ b/internal/sourcecode/templates/nuxtjs/Dockerfile
@@ -1,0 +1,33 @@
+# Source: https://nuxtjs.org/deployments/koyeb#dockerize-your-application
+FROM node:lts as builder
+
+WORKDIR /app
+
+COPY . .
+
+RUN yarn install \
+  --prefer-offline \
+  --frozen-lockfile \
+  --non-interactive \
+  --production=false
+
+RUN yarn build
+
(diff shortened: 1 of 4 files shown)


internal/sourcecode/templates/nuxtjs/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+FROM node:lts
+
+WORKDIR /app
+
+COPY --from=builder /app  .
+
+ENV HOST 0.0.0.0
+ENV PORT 8080
+EXPOSE 80
```
jsierles, line 31: This line doesn't mean anything on Fly, so probably should be removed to avoid confusion.

jsierles (approved):
Looks good, just a minor comment. We should add a docs page to accompany this!

internal/sourcecode/templates/nuxtjs/Dockerfile
lubien: Thanks. Should be fine now

lubien: > We should add a docs page to accompany this!

Going for it

lubien: https://github.com/superfly/docs/pull/111

PR #915, opened 2022-04-12
Migrate postgres node operations from ssh to http

rugwirobaker:
- Migrate more postgres operations.
- flyct services potstgres now uses the flypg package

Works with https://github.com/fly-apps/postgres-ha/pull/68


davissp14 (approved):
Nice work!

PR #919, opened 2022-04-13
Switches to mislav's homebrew bump github action

jphenow:
Homebrew bump failed a few times recently, and it was pointed out that this action has some better functionality. This gives us a chance to wire up a bot user for homebrew operations as well.


michaeldwan (approved):
we can fiddle with the token this afternoon

PR #922, opened 2022-04-14
machine run UX improvements

jipperinbham:
Two UX changes:
- Currently, we attempt to build the image early in the process which can take some time and if a flag was defined incorrectly the users wasted time watching the image build. Now we'll process as much as possible before trying to build the image.
- If the app flag was specified but did not exist, we would return the error `Could not resolve App`. Now we'll prompt the user if they would like to first create the app with the name specified in the flag.


fasterthanlime:
Hurray, thanks for this!

PR #924, opened 2022-04-16
Check to make sure people are SSH'ing to internal addresses

tqbf:
it'll do for now

actually checking to make sure we're hitting an fdaa address is
especially annoying b/c of how the agent works

### DIFF
--- a/cmd/ssh_terminal.go
+++ b/cmd/ssh_terminal.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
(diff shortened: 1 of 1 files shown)


cmd/ssh_terminal.go
```diff
@@ -129,6 +130,10 @@ func runSSHConsole(cc *cmdctx.CmdContext) error {
 		addr = fmt.Sprintf("top1.nearest.of.%s.internal", cc.AppName)
 	}
 
+	if strings.HasSuffix(addr, "fly.dev") {
+		return fmt.Errorf("attempting to use `flyctl ssh` to connect to a public address (.fly.dev); try without the hostname")
```
jsierles, line 134: Are people trying to do this in the wild?

PR #925, opened 2022-04-16
Fix recognition of other apps than Django

### LINKED ISSUE #878: Flask App misrecognized as Django App by flyctl launch
**Describe the bug**

The flask app [https://github.com/fly-apps/python-hellofly-flask](https://github.com/fly-apps/python-hellofly-flask) from the tutorial [https://fly.io/docs/getting-started/python/](https://fly.io/docs/getting-started/python/) is recognized as Django app by `flyctl launch`. Therefore it uses the Docker template for Django, which has specific commands, that don't work with Flask.

* MacOS Monterey
* fly v0.0.306 darwin/amd64 Commit: b5101f2 BuildDate: 2022-03-15T18:17:35Z

Could it be that the `configureDjango` checker `checksPass` even if there is only `requirements.txt` and not `manage.py`?

[https://github.com/superfly/flyctl/blob/8460965950c02314a7d4d4510202b5423db44e19/internal/sourcecode/scan.go#L479](https://github.com/superfly/flyctl/blob/8460965950c02314a7d4d4510202b5423db44e19/internal/sourcecode/scan.go#L479)

staeff:
Classify apps as Django, if `manage.py` exists to fix #878. This is a workaround, because the core issue is, that . This change avoids the misclassification of other Python apps, that most likely have a `requirements.txt`, but pretty sure don't have a `manage.py`.

The underlying issue is, that `fileExists` when just the first file exist. 

https://github.com/superfly/flyctl/blob/8460965950c02314a7d4d4510202b5423db44e19/internal/sourcecode/helpers.go#L10

Generally assuming a Django App, when the two files above exist, is a good enough heuristic. But for other Python frameworks you might not find such a pattern. I was actually surprised, that you are not reading the contents of `requirements.txt` and look for the name of the framework, that you want to support.

### DIFF
--- a/internal/sourcecode/scan.go
+++ b/internal/sourcecode/scan.go
@@ -541,7 +541,7 @@ func configureNuxt(sourceDir string) (*SourceInfo, error) {
 
 // setup django with a postgres database
 func configureDjango(sourceDir string) (*SourceInfo, error) {
-	if !checksPass(sourceDir, fileExists("requirements.txt", "manage.py")) {
+	if !checksPass(sourceDir, fileExists("manage.py", "requirements.txt")) {
 		return nil, nil
 	}
 


jsierles:
Hey, sorry the PR was closed. We'll look into this generally over the next day or two. I agree that we need to clarify AND/OR for file existence checks here.

kellan:
This issue is still present in flyctl version "flyctl v0.0.325 darwin/arm64 Commit: da2b638 BuildDate: 2022-04-28T04:00:48Z"

PR #931, opened 2022-04-19
Make machine list command use flaps

shortdiv:
Updated this so list shows machines in a table format and --quiet/-q only shows machine ids

### DIFF
--- a/internal/cli/internal/command/machine/list.go
+++ b/internal/cli/internal/command/machine/list.go
@@ -77,15 +77,15 @@ func runMachineList(ctx context.Context) (err error) {
 	}
 
 	var listOfMachines []api.V1Machine
-	if err = json.Unmarshal(machines, listOfMachines); err != nil {
+	if err = json.Unmarshal(machines, &listOfMachines); err != nil {
 		return fmt.Errorf("list of machines could not be retrieved")
 	}
 
+	fmt.Fprintf(io.Out, "%d machines have been retrieved\n\n", len(listOfMachines))
 	for _, machine := range listOfMachines {
-		fmt.Fprintf(io.Out, "Success! A machine has been retrieved\n")
 		fmt.Fprintf(io.Out, " Machine ID: %s\n", machine.ID)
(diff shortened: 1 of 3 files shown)


internal/cli/internal/command/machine/list.go
```diff
@@ -55,54 +56,37 @@ func runMachineList(ctx context.Context) (err error) {
...
+		return fmt.Errorf("list of machines could not be retrieved")
+	}
 
+	fmt.Fprintf(io.Out, "%d machines have been retrieved\n\n", len(listOfMachines))
+	for _, machine := range listOfMachines {
+		fmt.Fprintf(io.Out, " Machine ID: %s\n", machine.ID)
+		fmt.Fprintf(io.Out, " Instance ID: %s\n", machine.InstanceID)
+		fmt.Fprintf(io.Out, " State: %s\n\n", machine.State)
 	}
-	_ = render.Table(stream.Out, appName, rows, "ID", "Image", "Created", "State", "Region", "Name", "IP Address")
```
jipperinbham, line 105: I think it’d be good to keep the same output format, given some apps could have a lot of machines, keeping the output for a machine on one line helps with readability. The response from list machines _should_ still contain enough information to populate the same columns and if not we can fix.

internal/cli/internal/command/machine/list.go
```diff
@@ -55,54 +57,44 @@ func runMachineList(ctx context.Context) (err error) {
 	var (
 		appName = app.NameFromContext(ctx)
 		client  = client.FromContext(ctx).API()
-		stream  = iostreams.FromContext(ctx)
-		cfg     = config.FromContext(ctx)
+		io      = iostreams.FromContext(ctx)
 	)
 
-	state := flag.GetString(ctx, "state")
```
jipperinbham, line 62: Do we need to keep these flags if we’re not using them? For `state`, we could filter in the CLI? and get rid of `all`. Also looks like `quiet` outputs only the IDs and my guess is that’s useful for scripting.

jipperinbham (approved):
Looks great!

PR #936, opened 2022-04-21
Feat: machine update via run

jipperinbham:
The following changes support updating an existing machine via `machine run` by including the `--id` flag. It will first get the latest machine config and apply any changes.


shortdiv (approved):
LGTM!

PR #960, opened 2022-05-03
Run Wireguard connection over websockets by default

jsierles:
Websocket-based Wireguard has helped a lot of `flyctl` users, and has been in the wild for some time, so this PR makes websockets the default. We also can view the current wireguard websocket status and disable websockets.

```
$ flyctl wg websockets
Wireguard is configured to connect over websockets. Switch back to UDP-based connections with 'fly wg websockets disable'.
$ flyctl wg websockets disable
Wireguard is configured to connect over UDP. Switch to connecting over websockets with 'fly wg websockets enable'.
```


jsierles:
We're not going to default to this for now. Instead, we should detect if standard wireguard fails and fall back.

PR #961, opened 2022-05-03
use AppCompact graphql query when interacting with machines API

jipperinbham:
Currently, the graphql query to get an App contains a lot of unnecessary data for what is needed for the machine commands. The following changes it to use the AppCompact query.

### DIFF
--- a/internal/command/machine/kill.go
+++ b/internal/command/machine/kill.go
@@ -51,7 +51,7 @@ func runMachineKill(ctx context.Context) (err error) {
 		return fmt.Errorf("app was not found")
 	}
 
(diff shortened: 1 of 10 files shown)


jipperinbham:
@rugwirobaker It's possible the postgres commands may not be able to use `AppCompact` if it's utilizing the `Machines` field of the `App` struct. If so, we can adjust the flaps client accordingly so either `App` or `AppCompact` can be used.

rugwirobaker (approved):
I've tested this on`flyctl services postgres` and it doesn't seem to break anything.

PR #969, opened 2022-05-16
tidy machine run flags

jipperinbham:


internal/command/machine/run.go
```diff
@@ -226,10 +235,6 @@ func runMachineRun(ctx context.Context) error {
 		if memory := flag.GetInt(ctx, "memory"); memory != 0 {
 			machineConf.Guest.MemoryMB = memory
 		}
-
-		if cpuKind := flag.GetString(ctx, "cpu-kind"); cpuKind != "" && cpuKind != "shared" {
```
jeromegn, line 230: I think removing that (which is fine) requires `web` changes. Perhaps the behaviour should change to: `memory` and `cpus` just override the properties of the selected `size` if any.

PR #970, opened 2022-05-16
add hidden kernel-arg flag to machine run

jipperinbham:
The following hidden flag can be useful to adjust certain parameters to the init to provide more debugging info.


jeromegn (commented):
Since this is a repeating flag, maybe it should be called kernel-arg?

jeromegn:
Or even karg

jipperinbham:
`karg` feels a bit too cryptic? maybe as the short form option?

PR #985, opened 2022-05-26
Obtain flyd lease for postgres configuration changes

rugwirobaker:
- [x] Obtain lease for postgres restarts.
- [ ] Obtain lease on config/settings update.
- [ ] I'd prefer to do the payload unmarshalling inside the flaps package

### DIFF
--- a/internal/command/services/postgres/restart.go
+++ b/internal/command/services/postgres/restart.go
@@ -2,15 +2,18 @@ package postgres
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/spf13/cobra"
 
+	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/internal/app"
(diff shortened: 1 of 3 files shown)


internal/command/services/postgres/restart.go
```diff
@@ -66,17 +72,31 @@ func runRestart(ctx context.Context) error {
 	}
 
 	for _, machine := range machines {
```
davissp14, line 74: We may want to acquire leases on all machines before we begin the restart process.

davissp14 (changes requested):
Overall looks good!

internal/command/services/postgres/restart.go
rugwirobaker: Yeah, I thought about that too.

davissp14 (approved):
LGTM

PR #990, opened 2022-06-01
launch command for Laravel

fideloper:
This gets the `fly launch` command working for PHP builds, based on work done here: https://github.com/fly-apps/fly-hello-laravel

## Basics

- [x] Add in template files for Dockerized Laravel
    - Octane and "vanilla" flavors of Laravel are supported.
    - There are 2 instances of duplicate template files between the 2 flavors - `crontab` and `php.ini`.
- [x] Get `fly launch` command working for Laravel applications

## Fun Extras

- [x] **RoadRunner**: Use a separate set of templates for Octane-powered Laravel using RoadRunner driver
- [x] **Swoole**: Use a separate set of templates for Octane-powered Laravel using the Swoole driver

## Possible Extras:

> This PR has enough going on, these will get worked on in a separate PR after some thought.

- Setup a default database (sqlite/pgsql)
- Generate the `APP_KEY` secret automatically
    - This would require some refactoring so the `Secret.Generate` flag calls some secret-generation code that is specific to the framework/project
- Ability to decide if this should run a CRON (for Laravel's scheduler) and/or queue worker
    - This would require adding in some features to `flyctl launch` to prompt for additional options and change the template file used. Seems out of scope.


internal/sourcecode/scan.go
```diff
@@ -601,6 +602,42 @@ For detailed documentation, see https://fly.dev/docs/django/
 	return s, nil
 }
 
+// setup Laravel with a sqlite database
+func configureLaravel(sourceDir string) (*SourceInfo, error) {
+	// Laravel projects contain the `artisan` command, all php projects contain `composer.json` (in theory!)
+	if !checksPass(sourceDir, fileExists("artisan", "composer.json")) {
+		return nil, nil
+	}
+
```
catflydotio, line 611: If I understand correctly, you want to find *both* of these files to be pretty confident we're detecting Laravel. 

Under its current behaviour, `fileExists()` will return `true` if it finds *any*. (https://github.com/superfly/flyctl/issues/1001)

internal/sourcecode/scan.go
```diff
@@ -601,6 +602,42 @@ For detailed documentation, see https://fly.dev/docs/django/
...
+	var t []SourceFile
+	if checksPass(sourceDir, dirContains("composer.json", "laravel/octane")) {
+		t = templates("templates/laravel/octane")
+	} else {
+		t = templates("templates/laravel/standard")
+	}
+
```
catflydotio, line 618: I think `dirContains()` also returns `true` if the directory contains either of these files.

internal/sourcecode/scan.go
fideloper: I remembered that, yep! Artisan is the most important by far, but I figured `composer.json` was a good approximation for something PHP-ish if not. Honestly this could just be `artisan` if we wanted.

internal/sourcecode/scan.go
fideloper: The `dirContains()` function (conveniently!) seems to be finding content within a file in the current directory (although perhaps there's an argument here for renaming the function).

image

So, for example, the `configureLucky` detector runs:

```go
checksPass(sourceDir, dirContains("shard.yml", "lucky"))
```

Which finds file `shard.yml` in the `sourceDir` and searches it for string `lucky`.

internal/sourcecode/scan.go
jsierles: Let's just use `artisan` then for now.

internal/sourcecode/scan.go
jsierles: Renaming this would be good, I agree.

internal/sourcecode/templates/laravel/octane/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+
+# add useful utilities
+RUN apk add curl \
+    zip \
+    unzip \
+    ssmtp \
+    tzdata
+
+# php, with assorted extensions we likely need
+RUN apk add php8 \
```
jsierles, line 15: Wow that's a lot of stuff! Is this this 'blessed' way to install PHP dependencies?

internal/sourcecode/templates/laravel/octane/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+# working directory
+RUN mkdir /var/www/html
+WORKDIR /var/www/html
+
+# copy app code across, skipping files based on .dockerignore
+COPY . /var/www/html
+# ... install Laravel dependencies
+RUN composer update && composer install --optimize-autoloader --no-dev
+
+# Download roadrunner binary, prepare needed directory
```
jsierles, line 88: Why do you need to delete this binary? Maybe good to explain in the comment.

internal/sourcecode/templates/laravel/octane/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+RUN /usr/bin/crontab docker/crontab
+
+# move the docker-related conf files out of the app folder to where on the vm they need to be
+RUN mv docker/supervisor.conf /etc/supervisord.conf
+RUN mv docker/nginx.conf /etc/nginx/nginx.conf
+RUN mv docker/php.ini /etc/php8/conf.d/php.ini
+
+# Laravel's demo app does not need mix however real apps likely will (optimize js/css) so install and then run that:
+RUN npm install
+RUN npx mix --production
```
jsierles, line 108: This line should be combined with the following one, again to prevent unnecessary layers.

internal/sourcecode/templates/laravel/octane/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+# ... now don't need /node_modules so might as well delete that
+RUN rm -rf node_modules
+# ... and don't need node and npm so might as well delete them
+RUN apk del nodejs npm
+
+# Octane needs to write to the storage directory
+RUN chmod -R ug+w /var/www/html/storage
+
+# clear Laravel cache that may be left over
```
jsierles, line 117: Where would this cache be left over from?

internal/sourcecode/templates/laravel/octane/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+RUN composer dump-autoload
+RUN php artisan optimize:clear
+
+# make sure can execute php files (since octane runs as app, it needs permission e.g for /storage/framework/views for caching views)
+RUN chmod -R 755 /var/www/html
+
+# the same port nginx.conf is set to listen on and fly.toml references (standard is 8080)
+EXPOSE 8080
```
jsierles, line 125: This line is a no-op on Fly - any port listened on is accessible.

internal/sourcecode/templates/laravel/octane/docker/nginx.conf
```diff
@@ -0,0 +1,130 @@
+user nginx;
+
+#error_log  /var/log/nginx/error.log warn;
```
jsierles, line 3: I'm in favor of removing commented lines from config.

internal/sourcecode/templates/laravel/octane/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+    php8-sockets \
+    php8-pcntl \
+    php8-posix \
+    php8-pecl-swoole
+
+# node, for Laravel mix
+RUN apk add nodejs npm
+
+# supervisor, to support running multiple processes in a single app
+RUN apk add supervisor
```
jsierles, line 59: We should pull all the `apk add` into a single line, otherwise every package gets its own layer. Generally the fewer layers the better for Docker images.

internal/sourcecode/templates/laravel/octane/docker/run.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env sh
+
+# Start Octane with the roadrunner binary if we find
```
jsierles, line 3: If we deleted roadrunner in the Dockerfile, will it be found here?

internal/sourcecode/templates/laravel/octane/docker/supervisor.conf
```diff
@@ -0,0 +1,63 @@
...
+stderr_logfile_maxbytes=0
+
+[program:octane]
+user=app
+priority=5
+autostart=true
+autorestart=true
+stdout_events_enabled=true
+stderr_events_enabled=true
+;rr command=php -d variables_order=EGPCS /var/www/html/artisan octane:start --server=roadrunner --host=0.0.0.0 --rpc-port=6001 --port=8000
```
jsierles, line 33: What is the purpose of these commented-out lines?

internal/sourcecode/templates/laravel/standard/.dockerignore
```diff
@@ -0,0 +1,31 @@
+# excludes from the docker image/build
+
+# 1. Ignore everything (ensures don't upload .git, .env etc)
+**
```
jsierles, line 4: Interesting! Never seen this technique used before.

internal/sourcecode/templates/laravel/standard/docker/nginx.conf
```diff
@@ -0,0 +1,108 @@
+user nginx;
```
jsierles, line 1: Perhaps to avoid duplicating this config across laravel runtimes, we could have a separate path in `templates/laravel` for common files, which could also be nginx includes if these `nginx.conf` files are not identical.

jsierles (changes requested):
Overall looks good, just suggested some changes and left some questions.

internal/sourcecode/templates/laravel/octane/Dockerfile
fideloper: Short answer: Yes. However I'm gonna try out the official PHP  base images which has much of this added in already.

internal/sourcecode/templates/laravel/octane/Dockerfile
fideloper: Oh yeah, this whole thing is an exercise in unnecessary layers! I kept Greg's work in place as the comments made is clearer. That being said, let's refactor this thing as we discussed! (I'm working on that)

internal/sourcecode/templates/laravel/octane/docker/run.sh
fideloper: The deletion you mentioned prior was just in case, as I had cases where the `rr` binary on my macos installed made its way into the docker container. I believe that was when `.dockerignore` wasn't added before I updated `go:embed`.

The get-binary command called in the container adds the `rr` binary back in that's appropriate for the OS type (ARM vs AMD, etc).

(This'll get refactored a but this concept will likely remain).

internal/sourcecode/templates/laravel/standard/.dockerignore
fideloper: Same here! I have a starting point from Greg's work (Greg Sanderson is a guest in the `#fly-content` channel). This seems pretty reasonable for a Laravel app to me.

internal/sourcecode/templates/laravel/octane/docker/supervisor.conf
fideloper: The `ENTRYPOINT` script uncomments the correct line depending on if they use Octane's `roadrunner` or `swoole` drivers

For added context: Octane is a way to run PHP apps more like other apps - PHP is a long-running process that can accept web requests. 

Traditional PHP spins up a new process and reloads the entire framework for every web request.

fideloper:
An alternative `.dockerignore` file might look like this:

```
# 1. Ignore files we don't want in an image
.env
.git
node_modules
vendor

# 2. Ignore unnecessary files that may be inside those allowed directories
**/*~
**/*.log
**/.DS_Store
**/Thumbs.db

```

fideloper:
@jsierles This is ready for a review again.

1. There's a `laravel/common` directory to reduce duplication (The launcher code finds the common templates files, and then "merges" in standard or octane files)
2. Theres only one `Dockerfile`, which now `RUN`'s some `sh` conditionals, and uses a **multi-stage** build for Node assets
    - I pulled in `node:14`, which I like because it gives dev's the ability to easily tweak that to change the Node version if they need. Some node packages (e.g. `node-sass`) are especially sensitive to the Node version used.
3. Nginx uses an `include` to grab the config file specific to the run time (standard vs octane).

jsierles (approved):

PR #1005, opened 2022-06-10
v2 builders w/ bug fixes

jipperinbham:
Converts the CLI to using v2 machines for builders as well as bug fixes for build-args and proxying.


davissp14 (approved):
Looks good to me!

PR #1011, opened 2022-06-17
fix merging -e env variables into config

### LINKED ISSUE #560: Specifying env vars with 'fly deploy -e' wipes out env vars set in fly.toml

michaeldwan:
Wrong type checks caused existing env to always get overwritten. Anything from the "env" map gets converted from `map[string]interface{}` to `map[string]string` before deploy-time env variables are set.

Fixes #560


jsierles (approved):
Super!

PR #1013, opened 2022-06-20
Smaller GraphQL requests

jeromegn:
This removes almost all calls to `GetApp` which gets a massive `App` instance.


jsierles:
This looks good! Note that I started migrating a new command to use a [different, code-gen based type safe GraphQL client](https://github.com/Khan/genqlient). That caused the conflict you see here. If you don't mind, I'd like to leave the new command on the new client.

jsierles:
This looks good on TestFlight, and worked nice (and fast!) for me locally. 🌡️

Issue #1026, opened 2022-06-26
fly launch for phoenix fails


chrismccord:
It appears we have a couple issues, which may or may not be related, that leave `fly launch` on latest phoenix in a broken state.

First, a warning is printing telling the user they need a phoenix version >= 1.6.3, but they are actually running the latest release 1.6.10, so there is some string/regex comparison that is failing to compare the actual semver. 

Second, the generated fly.toml after getting the warning lacks a PORT export, which leaves any attempt at fly deploy is a broken state because the server will fail to bind to 8080 as configured in fly.toml. 

I'm not sure if we have two separate issues, or if the version comparison bug short circuits the PORT export when the cli thinks the user's phoenix isn't prepped for the golden path

ref #1020 #1025

chrismccord:
I can confirm this is a single bug from the semver comparison code path. Specifying phoenix 1.6.9 results in proper PORT export, so it should come down to fixing the check that was attempted in #1025

jsierles:
This is mostly fixed, though we introduced a new problem for users who have no Elixir setup at all. We can open a new issue for this as needed.

PR #1030, opened 2022-06-26
Make version check more general with exit code

chrismccord:
This does not fix the PORT issue outlined in #1026 but is a much better generalized check for the docker generator


chrismccord:
I’ll fix up formatting when I’m back at keyboard

jsierles (approved):

PR #1035, opened 2022-06-27
feat: Add role to machine config type declaration

shortdiv:
Add role to machine config, currently we aren't using it for anything but eventually we'll use it to keep track of machines that serve different purposes i.e. serving web requests vs running background jobs


rugwirobaker:
Niceee, can I playing with this for Postgres on machines thing?

dangra:
sorry the dumb question but Is this still relevant? if so I'd like to merge and move on.

dangra:
hey, I closed it because its purpose wasn't clear and adding it back should be pretty easy as part of a PR that uses this field.

PR #1039, opened 2022-06-28
command to allocate private ipv6s to app

senyosimpson:
Adds the command `flyctl ips allocate-v6 --private` to allocate 6pn ips that are routable via our proxy. Also, move the IP command to the new CLI command chain.


shortdiv (approved):
Looks good to me

senyosimpson:
Updated the output of the `fly ips list` command. It uses a table for the output, introduces a new field, `Version` for the IP version and repurposing `Type` to mean, type of address (public or private). Image below

Screenshot from 2022-07-05 09-50-01

senyosimpson:
@jsierles updated both `fly ips list` and `fly ips private`

jsierles:
Great! The last step here is to move it to the new command chain here https://github.com/superfly/flyctl/blob/master/internal/command/root/root.go, and to the new internal/command dir. That requires a small change to the flag parsing.

senyosimpson:
@jsierles i think it's ready for review! Do i need to manually test the commands my side or does CI test all of them? And, I left the changes I made to `cmd/ips.go` in but I removed adding the command to `cmd/root.go`. I don't know if I should just delete `cmd/ips.go` entirely?

jsierles:
CI does not test most commands, so you'll want to do a quick manual test. Just remove `cmd/ips`, as we don't need it anymore.

jsierles:
I tested it, works for me!

PR #1049, opened 2022-06-30
Unrestrict personal organizations

jphenow:


jsierles (approved):
LGTM, assuming we never plan to have another org type.

jphenow:
> assuming we never plan to have another org type

Open to leaving the type query stuff in - it just seemed silly to keep it because one place in the code was using it (until I removed that use-case as well)

jsierles:
No, please take it out! 🗡️

PR #1058, opened 2022-07-05
Secret generation uses closure, allowing for custom logic

fideloper:
This allows us to add some custom logic for secret generation for both `fly launch` and `fly machine launch`.

Updated Secret struct:

```go
type Secret struct {
	Key      string
	Help     string
	Value    string
	Generate func() (string, error) // No longer bool
}
```

Code change example:

Before:

```go
s := &SourceInfo{
	Secrets: []Secret{
		{
				Key:  "APP_KEY",
				Help: "Laravel needs a unique application key.",
				Generate: true,
			},
		},
}
```

After:

```go
s := &SourceInfo{
	Secrets: []Secret{
		{
				Key:  "APP_KEY",
				Help: "Laravel needs a unique application key.",
				Generate: func() (string, error) {
					// Method used in RandBytes never returns an error
					r, _ := helpers.RandBytes(32)
					return "base64:" + base64.StdEncoding.EncodeToString(r), nil
				},
			},
		},
}
```

### DIFF
--- a/cmd/launch.go
+++ b/cmd/launch.go
@@ -326,7 +326,6 @@ func runLaunch(cmdCtx *cmdctx.CmdContext) error {
 			// If a secret should be a random default, just generate it without displaying
 			// Otherwise, prompt to type it in
 			if secret.Generate != nil {
-				// // helpers.RandString(64)
 				if val, err = secret.Generate(); err != nil {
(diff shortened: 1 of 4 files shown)


jsierles (approved):
Nice work!

PR #1077, opened 2022-07-13
Tweak Phoenix deploy message

catflydotio:
Based on my understanding that this message is shown for every Phoenix launch, regardless of Phoenix version, I've changed the deploy message for the Phoenix launcher so it doesn't imply that Phoenix <1.6.3 is detected and that action is needed.


catflydotio:
I see that it's _not_ meant to be shown if the Docker generator is present. So now I don't know why I got the message with 1.6.11.

catflydotio:
Finally looked at it sensibly and it's just the project. :fades into bushes:

PR #1107, opened 2022-07-26
Add a machines subcommand for updating machines

jsierles:
This replaces the confusing `fly machine run --id` flag, and allows update semantics and allowed flags to differ from `run`.

Also, we introduce the `--metadata` flag to run/update, in preparation for implementing deployments with [process groups](https://fly.io/docs/reference/configuration/#per-process-commands).

### DIFF
--- a/internal/command/machine/status.go
+++ b/internal/command/machine/status.go
@@ -87,7 +87,6 @@ func runMachineStatus(ctx context.Context) (err error) {
 			machine.InstanceID,
 			machine.State,
 			machine.FullImageRef(),
-			strings.Join(machine.Config.Init.Cmd, " "),
 			machine.Name,
 			machine.PrivateIP,
 			machine.Region,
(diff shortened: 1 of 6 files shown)


internal/command/machine/status.go
```diff
@@ -87,14 +87,17 @@ func runMachineStatus(ctx context.Context) (err error) {
 			machine.InstanceID,
 			machine.State,
 			machine.FullImageRef(),
+			strings.Join(machine.Config.Init.Cmd, " "),
```
dangra, line 90: why the Command and not the Entrypoint? should it be the last column as it sizes is the most variable one?

dangra (approved):

internal/command/machine/status.go
jsierles: I dunno. I figured most people would not be modifying the entrypoint, and can view the whole config with `-d`.

PR #1110, opened 2022-07-26
Improvements to the machines app launcher

jsierles:
This PR fixes some bugs in, and adds features to, the machines app launcher. Features are mostly about supporting more `fly.toml` entries for launcher options like `deploy`, `env`, etc. The goal here is to make the launcher useful enough for the community to run it and test machine app deployments without having to start 'from scratch'.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -73,24 +73,32 @@ type SlimConfig struct {
 
 // Config wraps the properties of app configuration.
 type Config struct {
-	AppName     string                 `toml:"app,omitempty"`
-	Build       *Build                 `toml:"build,omitempty"`
-	HttpService *HttpService           `toml:"http_service,omitempty"`
-	VM          *VM                    `toml:"vm,omitempty"`
-	Definition  map[string]interface{} `toml:"definition,omitempty"`
-	Path        string                 `toml:"path,omitempty"`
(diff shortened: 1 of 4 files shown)


internal/app/app.go
```diff
@@ -73,24 +73,32 @@ type SlimConfig struct {
 
 // Config wraps the properties of app configuration.
 type Config struct {
-	AppName     string                 `toml:"app,omitempty"`
-	Build       *Build                 `toml:"build,omitempty"`
-	HttpService *HttpService           `toml:"http_service,omitempty"`
-	VM          *VM                    `toml:"vm,omitempty"`
```
dangra, line 79: sorry I am new here. Why is that VM is not needed anymore?

dangra (approved):
Not sure I understand all the changes but as `machine launch` is a WIP command I am fine with al of it.

My only worry is how setting Rails port to 8080 could affect non-machine apps.

internal/app/app.go
jsierles: Setting the port to 8080 should have always been explicitly set on Rails apps. It's just that 8080 is the default for Nomad apps.

`VM` was prematurely added by me as a way to specify VM size in machine config. We backed out of that since we probably need to work out process groups first, then see how VM sizes might apply to those.

PR #1112, opened 2022-07-27
Fix Invalid MachineConfig field

jipperinbham:
ImageRef cannot be part of MachineConfig because it cannot be set by the user. We recently updated flaps with the latest API and it's causing some customers to have issues creating machines.


jsierles (approved):
Just added some fixes, looks good.

PR #1114, opened 2022-07-28
Manage remote builders from flyctl

jsierles:
This PR pulls logic to start or wake remote builder machines into `flyctl`, with the goal of reducing dependencies on the backend, and to fix some bugs with/start wake logic that depends on outdated machine states in the database.

This should also simplify debugging and keep code line with flyd/flaps.

### DIFF
--- a/api/resource_apps.go
+++ b/api/resource_apps.go
@@ -307,11 +307,11 @@ func (client *Client) GetAppPostgres(ctx context.Context, appName string) (*AppP
 	return &data.AppPostgres, nil
 }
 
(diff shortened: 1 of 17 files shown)


jsierles:
@dangra I left this PR in limbo because of a strange bug causing the `client` to become uninitialized in context. But this will be useful to fix some situations where builders are not starting up correctly via web.

dangra:
I was about to ask!   Ok then, I will make room next week to test it and hunt for that bug 🐶

PR #1118, opened 2022-07-29
Add -c flag back to `fly ssh console`

### LINKED ISSUE #1117: flyctl ssh console no longer supports -c flag
**Describe the bug**
Briefly, describe what broke and provide the following details:

This worked before I upgraded from version `  0.0.353 -> 0.0.362 `

```
flyctl ssh console -c fly.staging.toml
```

Now the `-c` flag is not recognized:

```
flyctl ssh console -c fly.staging.toml
Error: unknown shorthand flag: 'c' in -c
Usage:
  flyctl ssh console [flags]

Flags:
  -a, --app string       Application name
  -C, --command string   command to run on SSH session
  -h, --help             help for console
  -o, --org string       The organization to operate on
  -r, --region string    Region to create WireGuard connection in
  -s, --select           select available instances

Global Flags:
      --verbose               verbose output

Error unknown shorthand flag: 'c' in -c
```

Looks like the regression may have possibly happened with this change: https://github.com/superfly/flyctl/pull/1095

tvd0x2a:
Fixes #1117

### DIFF
--- a/internal/command/ssh/console.go
+++ b/internal/command/ssh/console.go
@@ -37,6 +37,7 @@ func newConsole() *cobra.Command {
 	flag.Add(cmd,
 		flag.Org(),
 		flag.App(),
+		flag.AppConfig(),
 		flag.String{
 			Name:        "command",
 			Shorthand:   "C",


tvd0x2a:
I didn't check if there were other regressions... 😬

lessless:
When is this going to be released? Even in the `pre`version. 
We rely on separate config files to deploy to different environments, and I figured this is broken just when I need to deploy a fix. 

At the moment, I will try to install earlier version, but it came with a bit of stress 😬

lessless:
This is still not working in the latest available version (sorry, I don't know how to get the exact version of the `flyctl` - it's not in the default output)

lessless:
My bad - it does work, just doesn't show up in the help anymore

lessless:
I added new issue for that - #1195

PR #1124, opened 2022-08-02
add --force to pg attach

mrkurt:


internal/command/services/postgres/attach.go
```diff
@@ -54,6 +54,11 @@ func newAttach() (cmd *cobra.Command) {
 			Default:     "DATABASE_URL",
 			Description: "The environment variable name that will be added to the consuming app. ",
 		},
+		flag.Bool{
+			Name: "force",
+			Default: false,
+			Description: "Force attach (bypass confirmation)",
+		})
```
jmillxyz, line 61: Thanks for this PR @mrkurt! I think this will fix it?
```suggestion
		},
```

PR #1149, opened 2022-08-11
Launch machines on preferred region

dangra:
This fixes a bug that ignores the region passed as
:
```
fly machines launch --region CODE
```

### DIFF
--- a/api/resource_apps.go
+++ b/api/resource_apps.go
@@ -169,6 +169,9 @@ func (client *Client) GetAppCompact(ctx context.Context, appName string) (*AppCo
 					repository
 					version
 				}
+				autoscaling {
+					preferredRegion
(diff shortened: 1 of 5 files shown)


jsierles (commented):
I'm not sure we want to rely on anything related to autoscaling, which will be implemented differently on machines apps. For this case, we can use the passed region without also passing it to CreateApp.

dangra:
@jsierles if I understand correctly, you would like to use appConfig.primaryRegion and pass that to deploy for machines so it is used on first deploy. Is that right?

I used preferredRegion because it looked natural as it was already in apps definition and used by Nomad. But can change it to fit whatever is better.

dangra:
Closed in favor of #1153

PR #1153, opened 2022-08-11
Obey primaryRegion on machine creation

dangra:
Supersedes #1149

Respect region passed at:
```
fly machine launch --region CODE
```
and in case of machine's app:
```
fly deploy --region CODE
```

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -107,10 +107,9 @@ func DeployWithConfig(ctx context.Context, appConfig *app.Config) (err error) {
 
 	if appConfig.ForMachines() {
 		return createMachinesRelease(ctx, appConfig, img, flag.GetString(ctx, "strategy"))
-	} else {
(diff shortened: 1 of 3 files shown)


jsierles:
One question here is to figure out what it means to deploy with `--region`. Here it would mean 'if there are no VMs, deploy in this region'. If you already have VMs, it would mean nothing, I believe.

dangra:
I do agree on your last statement. I didn't want to serialize primaryRegion yet because lack some of the context but IMO it should be.

The idea of respecting `--region` for `fly deploy` is only for the case when you skip deployment on `fly launch` but still want to deploy to a specific region the first time.  IMHO, we should store the primary region on the toml file.

jsierles:
Yeah, another use case for `primary_region` will be for choosing where to run release commands by default, for things like migrations, or where to default your deploys so they match the Redis primary region.

dangra:
looks like a strong case to store it on the toml file

PR #1154, opened 2022-08-11
Remove dead code

dangra:


jsierles (commented):
Looks fine, though some of the code in here may be useful in the future, as it was used at some point.

dangra:
Yes, that is the other side but keeping the codebase fit makes it easier for new comers and usually is less weight to carry on.

PR #1155, opened 2022-08-11
Enable govet linter

dangra:
Fixes for:

```
$ golangci-lint run
internal/command/ping/ping.go:142:2: lostcancel: the cancel function is not used on all paths (possible context leak) (govet)
        ctx, cancel := context.WithCancel(ctx)
        ^
internal/command/ping/ping.go:176:3: lostcancel: this return statement may be reached without using the cancel var defined on line 142 (govet)
                return err
                ^
```

### DIFF
--- a/internal/command/ping/ping.go
+++ b/internal/command/ping/ping.go
@@ -140,6 +140,7 @@ func run(ctx context.Context) error {
 	mu.Unlock()
 
 	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	if name != "" && name != "gateway" && !strings.HasPrefix(name, "fdaa:") {
 		// look up names in the background because I was too
@@ -275,7 +276,6 @@ func run(ctx context.Context) error {
 	for i := 0; count == 0 || i <= count; i++ {
 		select {
 		case <-stp:
-			cancel()
 			return nil
 		case <-ticker.C:
 		}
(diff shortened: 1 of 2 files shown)


internal/command/ping/ping.go
```diff
@@ -275,7 +276,6 @@ func run(ctx context.Context) error {
 	for i := 0; count == 0 || i <= count; i++ {
 		select {
 		case <-stp:
-			cancel()
```
jsierles, line 278: What is this change for?

jsierles (approved):
Apart from the comment left, I'm generally OK with this as long as the linter is not overly pedantic about things like style/whitespace. I'm not too familiar with go linters so happy to try this out.

internal/command/ping/ping.go
dangra: Good point. Sorry I didn't include the errors reported by govet in the description.

This is it:
```
$ golangci-lint run
internal/command/ping/ping.go:142:2: lostcancel: the cancel function is not used on all paths (possible context leak) (govet)
        ctx, cancel := context.WithCancel(ctx)
        ^
internal/command/ping/ping.go:176:3: lostcancel: this return statement may be reached without using the cancel var defined on line 142 (govet)
                return err
                ^
```

dangra:
>  as long as the linter is not overly pedantic about things like style/whitespace

👍 in general linters have an automatic mode for fixing code, particularly those that enforce formatting like gofmt.

PR #1166, opened 2022-08-13
Have NextJS template the default NextJS port

grantgustafson:
This template fills in port 8080 when nextJS apps default to 3000: https://nextjs.org/docs/api-reference/cli#production

### DIFF
--- a/scanner/nextjs.go
+++ b/scanner/nextjs.go
@@ -6,12 +6,12 @@ func configureNextJs(sourceDir string) (*SourceInfo, error) {
 	}
 
 	env := map[string]string{
-		"PORT": "8080",
+		"PORT": "3000",
 	}
 
 	s := &SourceInfo{
 		Family:       "NextJS",
-		Port:         8080,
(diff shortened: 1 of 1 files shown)


jsierles:
Is this just to match the default port, or is the a problem with the original template causing deployment failures?

grantgustafson:
The original template causes deployment errors because the health checks fail. I made a fly community forums post about it here: https://community.fly.io/t/nextjs-demo-failing-on-deploy-from-health-checks/6315

dangra:
@grantgustafson I am inclined to merge because a working template is better than nothing of course. But somehow the fix is not the right one, NextJS should listen on the value of the PORT environment variable according to docs https://nextjs.org/docs/api-reference/cli#production

image

grantgustafson:
@dangra I ran a fresh test by creating a new default nextJS app and deployed with fly and the ENV var templating as `8080` does work as intended, so this PR is erroneous.

PR #1169, opened 2022-08-15
Fix typos in launch wizard

jonathanhefner:
This fixes a few typos in the launch wizard (and elsewhere):

* "setup" => "set up"
  * "setup" is the noun form, whereas "set up" is the verb form.
* Add "--" to split up a run-on sentence
  * "Save your credentials in a secure place" and "you won't be able to see them again!" are both complete sentences.


jsierles (approved):
Many of these are Fly-isms. But I agree with the changes.

PR #1170, opened 2022-08-15
Remove Puma `-C config/puma.rb` option

jonathanhefner:
Since Puma v2.8.0 (puma/puma@7d7a1fbebe4dde2edf045ada80730ce4f41c4cd5), Puma will automatically use `config/puma.rb` if it is present, so specifying `-C config/puma.rb` is not necessary.  Also, specifying `-C config/puma.rb` precludes Puma from automatically using `config/puma/<environment>.rb` if the user has put their config there instead.  (See the [Configuration File section][1] of the Puma README for more information.)

[1]: https://github.com/puma/puma/blob/v5.6.4/README.md#configuration-file

### DIFF
--- a/scanner/rails.go
+++ b/scanner/rails.go
@@ -32,7 +32,7 @@ func configureRails(sourceDir string) (*SourceInfo, error) {
 		},
 		ReleaseCmd: "bundle exec rails db:migrate",
 		Env: map[string]string{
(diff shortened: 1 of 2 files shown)


jsierles:
Seems reasonable. Thanks!

jsierles (approved):

PR #1176, opened 2022-08-16
Serialize primary region in fly.toml and set PRIMARY_REGION envvar on deploy

dangra:


internal/command/deploy/deploy.go
```diff
@@ -182,12 +182,14 @@ func determineAppConfig(ctx context.Context) (cfg *app.Config, err error) {
...
-
 		cfg.SetEnvVariables(parsedEnv)
 	}
 
 	if regionCode := flag.GetString(ctx, flag.RegionName); regionCode != "" {
-		cfg.SetPrimaryRegion(regionCode)
+		cfg.PrimaryRegion = regionCode
+	}
+	if cfg.PrimaryRegion != "" {
+		cfg.SetEnvVariable("PRIMARY_REGION", cfg.PrimaryRegion)
```
jsierles, line 192: Do we want to explicitly set this in config, or always set `PRIMARY_REGION` without exposing to the user? I could argue both ways, but think hiding it will make things less confusing if the primary region needs to change.

dangra: Not entirely sure but I vote for setting PRIMARY_REGION env var dynamically and do not duplicate its purpose in fly.toml.

jsierles: Yup, sounds good to me.

PR #1203, opened 2022-08-23
Refactor:machine start/stop accept more than one machine ids

rugwirobaker:


internal/command/machine/start.go
```diff
@@ -38,31 +38,32 @@ func newStart() *cobra.Command {
...
+		app, err := appFromMachineOrName(ctx, machineID, appName)
+		if err != nil {
+			return fmt.Errorf("could not make flaps client: %w", err)
+		}
 
-	flapsClient, err := flaps.New(ctx, app)
-	if err != nil {
-		return fmt.Errorf("could not make flaps client: %w", err)
-	}
+		flapsClient, err := flaps.New(ctx, app)
```
jsierles, line 52: Just a comment here. Having to create a new flaps client for every app shows to me that the flaps package needs a bit of refactoring, so you can set the app on the client instead of creating a new one each time. Not required here of course.

jsierles (approved):

internal/command/machine/start.go
rugwirobaker: Yeah I’ve been thinking about that too. It’s kinda of a blocker for my restart postgres work.

jsierles: In general we have a lot of overlap now across postgres and machine deployment. It feels like a new abstraction would be helpful, but not sure exactly what yet. One problem we have is that flaps is a per-org endpoint, so I believe we need a client per org right now.

PR #1222, opened 2022-08-29
Add support for Rails multi-environment configurations

rubys:
See: https://blog.saeloun.com/2019/10/10/rails-6-adds-support-for-multi-environment-credentials.html
And: https://community.fly.io/t/rails-master-key-environment-variable-not-getting-fetched/6657/5

This worked on my machine, but I'm new to flyctl and go so I am requesting a review.

### DIFF
--- a/scanner/rails.go
+++ b/scanner/rails.go
@@ -59,9 +59,14 @@ func configureRails(sourceDir string) (*SourceInfo, error) {
 		"NODE_VERSION":    nodeVersion,
 	}
 
-	// master.key comes with Rails apps from v6 onwards, but may not be present
-	// if the app does not use Rails encrypted credentials
-	masterKey, err := ioutil.ReadFile("config/master.key")
+	// master.key comes with Rails apps from v5.2 onwards, but may not be present
+	// if the app does not use Rails encrypted credentials.  Rails v6 added
+	// support for multi-environment credentials.  Use the Rails searching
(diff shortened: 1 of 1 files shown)


jsierles (approved):

scanner/rails.go
```diff
@@ -59,9 +59,14 @@ func configureRails(sourceDir string) (*SourceInfo, error) {
...
 
-	// master.key comes with Rails apps from v6 onwards, but may not be present
-	// if the app does not use Rails encrypted credentials
-	masterKey, err := ioutil.ReadFile("config/master.key")
+	// master.key comes with Rails apps from v5.2 onwards, but may not be present
+	// if the app does not use Rails encrypted credentials.  Rails v6 added
+	// support for multi-environment credentials.  Use the Rails searching
+	// sequence for production credentials to determine the RAILS_MASTER_KEY.
+	masterKey, err := ioutil.ReadFile("config/credentials/production.key")
+	if err != nil {
```
rugwirobaker, line 67: Should probably if the error related to the file not being there? Some kind of errors.Is(err, NotFound)?

rubys: I think I disagree; either that or I think you have an issue with the current code.

Before this change, if `master.key` can't be read for any reason, no secret will be added.

After this change, if either `production.key` or `master.key` can be read, a secret will be added; otherwise no secret will be added.

The code didn't previously care about why the read failure occurred.  If it is important that it does so, that can be a follow-on change.

Issue #1215, opened 2022-08-26
Fly launch should return app details


shortdiv:
Current view:
<img width="727" alt="Screen Shot 2022-08-26 at 12 08 43 PM" src="https://user-images.githubusercontent.com/3229484/186956870-ec4feab4-9314-4655-b67c-5cce4671db36.png">

We'd probably want app information like app name, url and how to start interacting with the created app in the output

tvdfly:
Fixed with https://github.com/superfly/flyctl/pull/1224

PR #1224, opened 2022-08-29
Highlight app details on launch

shortdiv:
Addresses #1215


cmd/deploy.go
```diff
@@ -324,6 +325,11 @@ func watchReleaseCommand(ctx context.Context, cc *cmdctx.CmdContext, apiClient *
 }
 
 func watchDeployment(ctx context.Context, cmdCtx *cmdctx.CmdContext, evaluationID string) error {
+	var (
+		io       = iostreams.FromContext(ctx)
+		colorize = io.ColorScheme()
+		out      = iostreams.FromContext(ctx).Out
```
jsierles, line 331: Nit: you can do `io.Out` here since you already have `io`.

cmd/launch.go
```diff
@@ -86,27 +87,28 @@ func newLaunchCommand(client *client.Client) *Command {
...
-	ctx := cmdCtx.Command.Context()
-
-	dir := cmdCtx.Config.GetString("path")
+	var (
+		importedConfig bool
+		ctx            = cmdCtx.Command.Context()
+		dir            = cmdCtx.Config.GetString("path")
+		orgSlug        = cmdCtx.Config.GetString("org")
+		appConfig      = flyctl.NewAppConfig()
+		io             = iostreams.FromContext(ctx)
```
jsierles, line 96: Does this actually work? We have a general issue that the app context is not initialized for some old commands.

cmd/launch.go
```diff
@@ -312,7 +314,14 @@ func runLaunch(cmdCtx *cmdctx.CmdContext) error {
 		}
 	}
 
-	fmt.Printf("Created app %s in organization %s\n", cmdCtx.AppName, org.Slug)
+	out := iostreams.FromContext(ctx).Out
```
jsierles, line 317: Same as above, can use `io` since it's already set above.

cmd/launch.go
```diff
@@ -312,7 +314,14 @@ func runLaunch(cmdCtx *cmdctx.CmdContext) error {
...
+
+	adminLink := fmt.Sprintf("https://fly.io/apps/%s", appName)
+	appLink := fmt.Sprintf("%s.fly.dev", appName)
+	links := fmt.Sprintf("Admin URL: %s\nURL: %s \nApp ID: %s \n", adminLink, appLink, cmdCtx.AppName)
+
+	fmt.Fprintln(out, colorize.Yellow(aurora.Underline("App Created\n").String()))
```
jsierles, line 323: One reason we have `colorize` is so we can control whether ascii codes are used in the output or not. So, we might want `colorize.Underline` here to work instead of importing aurora directly.

jsierles (commented):
Looks good, just left some comments.

cmd/launch.go
shortdiv: I think so? It's used in machines commands as well

PR #1226, opened 2022-08-29
Add schedule flag to machine run

shortdiv:
`fly machine run --schedule hourly`

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -198,11 +198,6 @@ func runMachineRun(ctx context.Context) error {
 		},
 	}
 
-	schedule := flag.GetString(ctx, "schedule")
-	if schedule != "" {
-		machineConf.Schedule = schedule
-	}
-
 	input := api.LaunchMachineInput{
 		AppID:  app.Name,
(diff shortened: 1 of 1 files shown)


internal/command/machine/run.go
```diff
@@ -194,6 +198,11 @@ func runMachineRun(ctx context.Context) error {
 		},
 	}
 
+	schedule := flag.GetString(ctx, "schedule")
```
jipperinbham, line 201: set it directly and drop the if 

```suggestion
	machineConf.Schedule = flag.GetString(ctx, "schedule")
```

jsierles (commented):
I'd recommend instead to put this in [determineMachineConfig](https://github.com/superfly/flyctl/blob/master/internal/command/machine/run.go#L483-L484) so it can be used by `fly machine update`. In that case, you would want to keep the conditional mentioned in @jipperinbham's comment so the current config is not overwritten.

You'd also want to add the flag to the shared flag set used for both `update` and `run`. https://github.com/superfly/flyctl/blob/master/internal/command/machine/run.go#L34-L35

jipperinbham (approved):
:shipit:

PR #1227, opened 2022-08-29
Fly.rake

rubys:
This change effectively adds Ruby hooks in the form of rails tasks for key parts of the build process.

The motivation for this is to improve the initial experience people are seeing when attempting to migrate from places like Heroku.  What they often find is that their build process needs access to secrets, and the docs and advice they get on places like community.fly.io leads them into uncomfortable places like modifying a `Dockerfile` or a `toml` file that they don't understand.

This also greatly simplifies thinks like starting sidekiq.

To be clear, we want and need to retain the flexibility that modifying the `Dockerfile` and `toml` file provide for advanced use cases.  The goal if this change merely to make the initial experience less overwhelming for many and ideally to cover the 80% use case for build configuration.

### DIFF
--- a/scanner/rails.go
+++ b/scanner/rails.go
@@ -120,7 +120,6 @@ func configureRails(sourceDir string) (*SourceInfo, error) {
 	rake := strings.TrimSpace(`
 # commands used to deploy a Rails application
 namespace :fly do
-
   # BUILD step:
   #  - changes to the fs make here DO get deployed
   #  - NO access to secrets, volumes, databases
@@ -129,15 +128,15 @@ namespace :fly do
 
   # RELEASE step:
   #  - changes to the fs make here are DISCARDED
-  #  - access secrets, databases
-  #  - Failures here prevent deployment
+  #  - full access to secrets, databases
+  #  - failures here prevent deployment
   task :release => 'db:migrate'
 
   # SERVER step:
   #  - changes to the fs make here are deployed
-  #  - access secrets, databases
-  #  - Failures here result in VM being stated, shutdown, and rollback
-  #    to last successful VM.
+  #  - full access to secrets, databases
+  #  - failures here result in VM being stated, shutdown, and rolled back
+  #    to last successful deploy (if any).
   task :server do
     sh 'bin/rails server'
   end
--- a/scanner/rails.go
+++ b/scanner/rails.go
@@ -1,10 +1,12 @@
 package scanner
 
 import (
+	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
+	"os/exec"
 	"regexp"
+	"strings"
 )
 
 func configureRails(sourceDir string) (*SourceInfo, error) {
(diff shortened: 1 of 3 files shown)


jsierles:
How does this solve the problem of build-time secrets? A downside I see to prematurely overriding all the standard Rails tasks is that we shift the burden of support onto us. There isn't a ton of value there if we're just shimming existing commands.

rubys:
> How does this solve the problem of build-time secrets?

An all too common first time experience for Rails developers coming from other hosting providers is that they find out that their `assets:precompile` step needs to be run on the server instead of the build machine.  They hop on community.fly.io and are told to modify both their `Dockerfile` and `fly.toml` files.  Two files that they are not familiar with.  This user experience is sub-optimal.

With this change, they can instead be told go to into their [`lib/tasks/fly.rake` file](https://github.com/superfly/flyctl/blob/0e21977ff20b0cc2461ac2e7d279dc256ed86d22/scanner/rails.go#L121-L144) and move `=> 'assets:precompile'` from the `:build` step to the `:server` step.

Similarly, if they want to launch new services (sidekiq, anyone?), they can do so directly in the comfort of a language that they already are familiar with.

> A downside I see to prematurely overriding all the standard Rails tasks is that we shift the burden of support onto us. 

I'm confused.  This doesn't override all the standard Rails tasks.  In fact, it doesn't override any of them.  This adds three new tasks.

> There isn't a ton of value there if we're just shimming existing commands.

Again, I'm confused.  We aren't shimming existing commands (an earlier push included some such shims, but that was removed).

Just so we are on the same page, a recap of what this pull request does.

---

Before, in `Dockerfile`:

`RUN bundle exec rails assets:precompile`

After, in `Dockerfile`:

`RUN bin/rails fly:build`

After, in `lib/tasks/fly.rake`:

```ruby
task :build => 'assets:precompile'
```

---

Before, in `fly.toml`:

`release_command = "bundle exec rails db:migrate"`

After, in `fly.toml`:

`release_command = "bin/rails fly:release"`

After, in `lib/tasks/fly.rake`:

```ruby
task :release => 'db:migrate'
```

---

Before, in `fly.toml`:

`SERVER_COMMAND = "bundle exec puma"`

After, in `fly.toml`:

`SERVER_COMMAND = "bin/rails fly:server"`

After, in `lib/tasks/fly.rake`:

```ruby
task :server do
  sh 'bin/rails server'
end
```

Note: `bin/rails server` will run whatever web server you have in your Gemfile, and Rails generates a Gemfile including puma by default.

---

This change should not change the default behavior of the `Dockerfile` and `toml` file generated by `fly launch` (other than supporting whatever webserver is chosen in the `Gemfile`).  It should not increase our support costs.  All it does is provide an additional file in a language Rails developers already understand complete with comments that give guidance as to the implications of what happens in each of the three steps of the deploy process so that they can make informed decisions.

jsierles:
OK, I see your point here. Maybe worth posting on the forum for more feedbakc.

I see the sweet spot here being tasks that go into a gem that we can install at launch time. We can ask people if they want tasks overridden or not, while explaining what's going on in real time. This way, people who have not or do not want to use `launch` can still benefit, and we don't have to answer questions like 'why did your fly-specific task break my build?' (what I was referring to when mentioning "support")

So running `precompile` at server boot time means that every VM will run this potentially step which cannot benefit from cache. Maybe not a problem for smaller apps, but also not the way things work on Heroku.

rubys:
Posted to community: https://community.fly.io/t/rfc-have-fly-launch-produce-a-lib-tasks-fly-rake/6744

Again, this post doesn't override tasks.  I also see value in a gem that does override tasks, but this pull request doesn't do that.

Nothing this pull request does requires `fly launch`.  Users that avoid `fly launch` can chose to structure their Dockerfile and toml file this way, or not.

Any way that a developer breaks their own build using this mechanism is also equally as possible if they were to modify the Dockerfile and/or toml files.  In fact, I would argue that they are more likely to break the build in ways that they don't understand if they modify the Dockerfile and/or toml files than if they modify the fly.rake file.

Yes, `precompile` should by default run at build time.  This pull request doesn't change that default.

Yes, fly is different than Heroku.  This pull request doesn't change that.

jsierles:
Sorry if I wasn't being clear. I like the things this PR is doing, but think it would be better to move to a gem if we're going down this path, so if you wanted to add it to an existing app, you wouldn't need to use `fly launch` to do it. I will leave to your discretion on whether to merge, as a change like that could be done later.

scanner/rails.go
```diff
@@ -78,6 +117,38 @@ func configureRails(sourceDir string) (*SourceInfo, error) {
...
+  #    to last successful VM.
+  task :server do
+    sh 'bin/rails server'
+  end
+end
+`)
+
+	_, err = os.Stat("lib/tasks/fly.rake")
+	if errors.Is(err, os.ErrNotExist) {
+		os.WriteFile("lib/tasks/fly.rake", []byte(rake), 0o600)
```
jsierles, line 149: This could be done by adding this file in the same location as the Dockerfile, so it's installed by `template`, rather than writing out this way. Is there a reason you didn't do it that way?

rubys: Good catch!  Fixed.

rubys:
I sense that I'm still not explaining myself clearly.  The intent is to create a file that the Rails developer can tailor.  

If we later create a gem that adds tasks and generators, that can be layered on top.  That gem can use Thor to update the configuration, just like a bazillion Rails extensions already do.

jsierles:
Yep, makes sense. I think I was thinking of a generator that gets run from a gem, but this is simple and easy to deal with for now.

PR #1229, opened 2022-08-30
Re-organizing the machines list output and introducing volume

davissp14:
Reorganizes the `machine list` output and adds a `volume` column.

Screen Shot 2022-08-30 at 11 44 57 AM


rugwirobaker (approved):
LGTM

PR #1232, opened 2022-08-31
Bump go version in go.mod

jsierles:


jsierles:
@dangra For when you return, I had to disable two misbehaving linters with this upgrade. Maybe you have an idea on how to fix them.

PR #1236, opened 2022-08-31
Postgres: add hard restart

rugwirobaker:
`flyctl pg restart --hard` forces VMs restart instead of just the postgres process.

### DIFF
--- a/internal/command/postgres/restart.go
+++ b/internal/command/postgres/restart.go
@@ -99,40 +99,41 @@ func runRestart(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("list of machines could not be retrieved: %w", err)
 		}
+		ctx = flaps.NewContext(ctx, flapsClient)
 
 		members, err := flapsClient.List(ctx, "started")
 		if err != nil {
 			return fmt.Errorf("machines could not be retrieved %w", err)
 		}
-		leader, err := fetchPGLeader(ctx, app, members)
+		leader, replicas, err := nodeRoles(ctx, members)
 		if err != nil {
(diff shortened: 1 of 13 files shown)


internal/command/postgres/restart.go
```diff
@@ -52,136 +57,282 @@ func runRestart(ctx context.Context) error {
...
+			return machinesHardRestart(ctx, machines)
+		}
+		return machinesSoftRestart(ctx, machines)
 	}
 
 	return nil
 }
 
-func restartMachinesCluster(ctx context.Context, app *api.AppCompact) error {
+func machinesSoftRestart(ctx context.Context, machines []*api.Machine) error {
```
davissp14, line 111: We should test this, but I think performing a failover here would actually result in more downtime.  Process level restarts are super fast, so we probably don't need to worry about performing a failover here.

internal/command/postgres/restart.go
```diff
@@ -52,136 +57,282 @@ func runRestart(ctx context.Context) error {
...
+	if err := pgclient.RestartNodePG(ctx); err != nil {
+		return fmt.Errorf("failed to restart postgres on node: %w", err)
+	}
+	fmt.Fprintf(io.Out, "Postgres cluster has been successfully restarted!\n")
+
+	return
 }
 
-func restartNomadCluster(ctx context.Context, app *api.AppCompact) (err error) {
+func machinesHardRestart(ctx context.Context, machines []*api.Machine) (err error) {
```
davissp14, line 238: We should work to compare this process with the soft restart.  If they are very similar, we could work to remove the process level restarts.

davissp14 (approved):
Few notes, but overall looks good!

PR #1244, opened 2022-09-01
Fix typo in `fly scale vm --help`

thejager:


lubien (approved):

docstrings/gen.go
```diff
@@ -685,7 +685,7 @@ e.g. flyctl scale vm shared-cpu-1x --memory=2048
 
 For dedicated vms, this should be a multiple of 1024MB.
 
-For shared vms, this can be 256MB or a a multiple of 1024MB.
```
rugwirobaker, line 688: This file is generated from https://github.com/superfly/flyctl/blob/master/helpgen/flyctlhelp.toml. These changes will get deleted on the next docgen. You should find the scale key in the toml file and make the update there.

thejager: Oh ok! Will fix that

thejager:
I still kept the change in the original file, is that okay? I tried running the generator but not sure how. I noticed the `Makefile` but `make build` gives an error:

image

PR #1245, opened 2022-09-01
Always use wireguard-over-websockets in CI environments

jsierles:
This may increase chances of success in UDP-unfriendly CI environments.


tvdfly:
LGTM 🎉

PR #1255, opened 2022-09-06
Machine specific vm sizes

davissp14:
Notable changes:

* Fixes an issue where PG Machine provisions were prompting Nomad specific vm-sizes.
* PG Machine provisions now have their own list of pre-configured setups. 
* Machine specific vm-sizes have been hard-coded, but these should eventually come from flaps.
* Started to tease apart the CPU/Memory coupling.  Eventually i'd like memory to be specified independently of CPU when specifying a custom configuration. 

```
? Select configuration:  [Use arrows to move, type to filter]
  Development - Single node, 1x shared CPU, 256MB RAM, 1GB disk
  Production - Highly available, 2x shared CPUs, 4GB RAM, 40GB disk
  Production - Highly available, 4x shared CPUs, 8GB RAM, 80GB disk
> Specify custom configuration
```

### DIFF
--- a/flypg/launcher.go
+++ b/flypg/launcher.go
@@ -37,7 +37,7 @@ type CreateClusterInput struct {
 	Password           string
 	Region             string
 	VolumeSize         *int
-	VMSize             *string
+	VMSize             *api.VMSize
 	SnapshotID         *string
 }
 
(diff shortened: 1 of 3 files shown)


flypg/launcher.go
```diff
@@ -228,7 +228,11 @@ func (l *Launcher) getPostgresConfig(config *CreateClusterInput) *api.MachineCon
 		"PRIMARY_REGION": config.Region,
 	}
 
-	machineConfig.VMSize = *config.VMSize
```
davissp14, line 231: The next iteration will allow the user to specify memory separately from CPU.

PR #1259, opened 2022-09-08
Update NextJS port to 3000

spence:
Users following the blog article on [Deploy a NextJS Application](https://fly.io/docs/getting-started/nextjs/) will fail to deploy due to a port mismatch during `fly launch`.

There's a discrepancy between the generated `Dockerfile` and `fly.toml` files.
* The port generated in `fly.toml` is 8080 ([results in `services.internal_port = 8080`](https://github.com/superfly/flyctl/blob/b6d7c3cf51614c1a8b881ec752f8e15387fbd7b1/scanner/nextjs.go#L14)).
* The port generated in `Dockerfile` is 3000 ([`ENV PORT 3000`](https://github.com/superfly/flyctl/blob/30917786f10e8e64d03480460e65dbd274a350aa/scanner/templates/nextjs/Dockerfile#L38))
* The default port for [NextJS apps is 3000](https://nextjs.org/docs/deployment#docker-image). 

This updates the generated port (8080) to match the default (3000).

### DIFF
--- a/scanner/nextjs.go
+++ b/scanner/nextjs.go
@@ -6,12 +6,12 @@ func configureNextJs(sourceDir string) (*SourceInfo, error) {
 	}
 
 	env := map[string]string{
-		"PORT": "8080",
+		"PORT": "3000",
 	}
 
 	s := &SourceInfo{
 		Family:       "NextJS",
-		Port:         8080,
+		Port:         3000,
 		SkipDatabase: true,
 	}
 


cadbox1:
I ran into the same issue! 

looking at the other scanners, it seems they all listen on port 8080 but they also pass in an environment variable to make that happen.

the nextjs appears to do the same thing but it doesn't seem to work.

rather than change internal port to match nexts, I've tried to align the nextjs scanner with the other ones by setting the environment variable in the same way so they're consistent.

https://github.com/superfly/flyctl/pull/1269

rugwirobaker (approved):
LGTM

lubien:
Thanks for contributing, it'll be fixed from here: https://github.com/superfly/flyctl/pull/1269

PR #1269, opened 2022-09-13
Fix NextJS Port

cadbox1:
running flyctl on a nextjs package fails right now.

this seems to be happening because the port specified in the fly.toml, 8080, doesn't match the port nextjs is listening on, 3000.

diving deeper, it looks like the environment variables required for nextjs listening on port 8080 are missing. they seem to be set differently to other scanners, like this phoneix one: https://github.com/superfly/flyctl/blob/master/scanner/phoenix.go#L30

I've made the following change to the nextjs scanner to align it with the other scanners.

I've also made the change to nuxtjs and remix because they follow the same convention as nextjs and probably don't work either

all scanners follow the same convention for ports and environment variables with this change

### DIFF
--- a/scanner/nextjs.go
+++ b/scanner/nextjs.go
@@ -5,13 +5,15 @@ func configureNextJs(sourceDir string) (*SourceInfo, error) {
 		return nil, nil
 	}
 
+	env := map[string]string{
+		"PORT": "8080",
+	}
+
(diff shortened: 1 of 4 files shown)


lubien:
I've simplified the code a little bit

jsierles (approved):

PR #1271, opened 2022-09-13
fix: Add error msg for better clarity on why machine could not be retrieved

shortdiv:
## Machine Status 
Before:
before

After:
after

## Machine List
Before:
before

After:
after


jipperinbham:
🤔 I'm not sure how I feel about it being an error to list machines for an app that doesn't have any. I do like the change for attempting to get a specific machine!

internal/command/machine/list.go
```diff
@@ -71,14 +71,19 @@ func runMachineList(ctx context.Context) (err error) {
 		return fmt.Errorf("machines could not be retrieved")
 	}
 
+	if len(machines) == 0 {
+		fmt.Printf("No machines are available on this app %s", appName)
```
jipperinbham, line 75: Is there a reason to not use `fmt.Fprintf(io.Out…` here like done below?

jipperinbham (approved):

PR #1273, opened 2022-09-13
added alias volume to `fly volumes`

eliferrous:
this was a request from the forums:
- https://community.fly.io/t/fly-volume-alias-does-not-exist/7122
this change reflects the current docs

since I make this mistake all the time too, I thought it might be nice to add it.


catflydotio:
Ah, interesting approach. I was going to fix the docs! 😆

thejager:
Although I like the alias, I think it should be considered that there's many such commands that _don't_ have this alias.

Why is `fly volume` a thing, but `fly app` is not? Isn't it making things more confusing?

PR #1276, opened 2022-09-14
Updates to Laravel launcher

fideloper:
## Changes

1️⃣

Using `.fly` instead of `docker` directory for files that `flyctl` adds to users codebase (renaming the files make Git's diff think I added files for some reason, but really I just moved them).

So user's projects will have a `.fly` directory instead of `docker`. The `docker` directory might be used by other projects, such as Laravel Sail. Also, we'd like to standardize a `.fly` directory across supported frameworks.

---

2️⃣

Nginx sets `X-Forwarded-For` to the value of `Fly-Client-Ip` to help with limitations of Symfony's Trusted Proxy concept (see https://debug.fly.dev/ to see headers a Fly app sees)

---

3️⃣

User's can add custom scripts to `.fly/scripts`. Any `*.sh` script in there will be run with bash when the container starts up for *any* command (default ENTRYPOINT or ENTRYPOINT + CMD combo).

Our default is to run the following artisan commands via `.fly/scripts/caches.sh` (but it can be customized of course):

```bash
/usr/bin/php /var/www/html/artisan config:cache --no-ansi -q
/usr/bin/php /var/www/html/artisan route:cache --no-ansi -q
/usr/bin/php /var/www/html/artisan view:cache --no-ansi -q
```


scanner/templates/laravel/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# syntax = docker/dockerfile:experimental
+# syntax = .fly/dockerfile:experimental
```
thejager, line 1: This should not be changed to `.fly` right? Seems like a find-and-replace error

scanner/templates/laravel/.fly/entrypoint.sh
```diff
@@ -12,5 +12,12 @@ else
     /usr/bin/php /var/www/html/artisan view:cache --no-ansi -q
     chown -R webuser:webgroup /var/www/html
 
+    ## Run user scripts, if they exist
+    for f in /var/www/html/.fly/scripts/*.sh; do
```
thejager, line 16: Nice, should cache preparation go to `.fly/scripts`?

scanner/templates/laravel/.fly/nginx-default
```diff
@@ -24,6 +24,7 @@ server {
 	# Pass "*.php" files to PHP-FPM
 	location ~ \.php$ {
 			include snippets/fastcgi-php.conf;
+			fastcgi_param  HTTP_X_FORWARDED_FOR $http_fly_client_ip;
```
thejager, line 27: I know I'm nitpicking, but there's 2 spaces after `fastcgi_param` 👀

scanner/templates/laravel/Dockerfile
fideloper: Yep, thanks!

scanner/templates/laravel/.fly/entrypoint.sh
fideloper: Yep, altho I may tweak this some more so be able to hook into key points:

1. When the Docker image is being built
2. When a container spins up (called from the `ENTRYPOINT`)
3. **Can you think of anything else?**

The main things to things to think about in terms of how Fly works and what Laravel needs for deployments (migrations, setting caches, etc):

1. The `docker build` doesn't have access to env vars / build secrets without extra work
2. `release_command` commands run in a separate VM just before deployments ([reference](https://fly.io/docs/reference/configuration/#the-deploy-section)) 
    - Doesn't have access to mounted volumes
    - Any file-system changes are lost (e.g. laravel view cache files). This means the `artisan view:cache` and other commands should run somewhere else, generally
    - Has access to the network, so migrations against databases running separately should be fine (but a local sqlite database change is not fine, see point 2)

scanner/templates/laravel/.fly/nginx-default
fideloper: thanks!

thejager: Could the spacing in `nginx-default`also be improved in this PR? It looks a bit weird. When I run `fly launch` the first thing I do is fix the weird spacing in `nginx-default` :)) My preference would be:

- Spaces instead of tabs
- 4 spaces for indentation

image

fideloper: Agree!

scanner/templates/laravel/.fly/nginx-default
```diff
@@ -0,0 +1,33 @@
...
+
+    # Have NGINX try searching for PHP files as well
+    location / {
+        try_files $uri $uri/ /index.php?$query_string;
+    }
+
+    # Pass "*.php" files to PHP-FPM
+    location ~ \.php$ {
+        include snippets/fastcgi-php.conf;
+        fastcgi_param HTTP_X_FORWARDED_FOR $http_fly_client_ip;
```
thejager, line 27: I would suggest adding this line as well:

```
fastcgi_param HTTPS on;
```

Without this line, `$request->isSecure()` will return false. This affects how URLs are generated in Laravel/Symfony

fideloper: That's set already in `/etc/nginx/fastcgi.conf` which is itself included via `include snippets/fastcgi-php.conf`

thejager: This line is specified as:

`fastcgi_param  HTTPS              $https if_not_empty;`

I am not too proficient in NGINX configurations but this passes the `$https` value as `HTTPS` header. From the docs:

```
$https
    “on” if connection operates in SSL mode, or an empty string otherwise
```

The connection to the VM does not operate in SSL mode so this will be an empty string I suppose?

I doubt this is working out of the box with `fastcgi-php.conf`. We were getting `$request->isSecure() === false` until we added `fastcgi_param HTTPS on;`

thejager: Hm, I just created a new Laravel project, ran `fly launch` and `dd(request()->isSecure())` and it shows `true`. Nvm then, I guess there's something different between our configuration and the current config given by `fly launch`...

thejager: Oddly enough this does not work in Symfony. I use the Laravel configuration for Symfony (with some adjustments), and `isSecure()` returns false if I don't add that `fastcgi_param HTTPS on;` line.

fideloper: I suspect Laravel is reading the `X-Forwarded-Proto` header to determine the result of `isSecure()` (since the TrustedProxy middleware is being used/enabled), and in that case, it likely ignores `HTTPS`.

That's a good point tho, Fly's proxy is sending `http://` requests to the VM after terminating the TLS connection 🤔

thejager:
@fideloper just putting a comment here because I think you won't get notified about comments underneath resolved comments.

> Just looked it up to be sure - That's set already in /etc/nginx/fastcgi.conf which is itself included via include snippets/fastcgi-php.conf
I noticed `isSecure()` returns false in Symfony apps but _not_ in Laravel apps. I was curious as to why that is so I dove into it... 



This does _not_ make `$request->isSecure()` return `true`. When you dump `$_SERVER` it will not show the `HTTPS` header. It's not present because the connection to the VM is not a connection operating in SSL mode.

What _does_ make this work is the `TrustProxies` middleware. Because you configured the `TrustProxies` middleware, `Request::isFromTrustedProxy()` returns `true` because the proxy is trusted AND the `HTTP_X_FORWARDED_PROTO` is equivalent to `https`.

thejager:
@fideloper just putting a comment here because I think you won't get notified about comments underneath resolved comments.

I noticed `isSecure()` returns false in Symfony apps but _not_ in Laravel apps. I was curious as to why that is so I dove into it... You mentioned:

> Just looked it up to be sure - That's set already in /etc/nginx/fastcgi.conf which is itself included via include snippets/fastcgi-php.conf

I don't think this makes `$request->isSecure()` return `true`. When you dump `$_SERVER` it will not show the `HTTPS` header. It's not present because the connection to the VM is not a connection operating in SSL mode.

What _does_ make this work is the `TrustProxies` middleware. Because you configured the `TrustProxies` middleware, `Request::isFromTrustedProxy()` returns `true` which makes `Request::isSecure()` look at the `HTTP_X_FORWARDED_PROTO` which is configured to be `https`.

This also explains why this wasn't working in my Symfony app which has the exact same nginx configuration (I diff'd the `nginx -T` outputs 😬). My Symfony app does not have trusted proxies configured. https://symfony.com/doc/current/deployment/proxies.html

jsierles (approved):

shpetimislami:
Having launched a laravel app before this was merged when instead, I could not deploy yesterday and the container was exiting with error code 100. I fixed the issue by replacing the changes of the Dockerfile and docker folder to .fly. 
Is there no way to update the docker files generated from fly launch when there are new updates these files?

PR #1282, opened 2022-09-16
fix: Add details on machine update

shortdiv:
Screen Shot 2022-09-16 at 1 50 03 PM

### DIFF
--- a/internal/command/machine/update.go
+++ b/internal/command/machine/update.go
@@ -99,15 +99,25 @@ func runUpdate(ctx context.Context) (err error) {
 	fmt.Fprintln(out, colorize.Yellow(fmt.Sprintf("Machine %s has been updated\n", machine.ID)))
 	fmt.Fprintf(out, "Instance ID has been updated:\n")
 	fmt.Fprintf(out, "%s -> %s\n\n", prevInstanceID, machine.InstanceID)
-	fmt.Fprintf(out, "Image: %s\n", machine.Config.Image)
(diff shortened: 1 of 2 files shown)


jipperinbham (approved):

internal/command/machine/update.go
```diff
@@ -92,6 +95,15 @@ func runUpdate(ctx context.Context) (err error) {
 		waitForAction = "stop"
 	}
 
+	out := io.Out
+	fmt.Fprintln(out, colorize.Yellow(fmt.Sprintf("Machine %s has been updated\n", machine.ID)))
+	fmt.Fprintf(out, "Instance ID has been updated:\n")
+	fmt.Fprintf(out, "%s -> %s\n\n", prevInstanceID, machine.InstanceID)
+	fmt.Fprintf(out, "Image: %s\n", machine.Config.Image)
+	fmt.Fprintf(out, "State: %s\n\n", machine.State)
```
jipperinbham, line 103: Oh! I think we need to reload the machine state here?

shortdiv: oh, you're ttly right, we do!

shortdiv: I assume that since we call wait for start/stop and we know what state it's in, we can just assume that state without having to make another call to flaps?

jipperinbham: Yea, I honestly think that's fine to do.

jipperinbham (approved):

PR #1289, opened 2022-09-20
fix: prevent schedule override

shortdiv:
Schedule was being overridden when an update happened via flyctl. This made machine states inconsistent since a machine would end up stopped if it had been scheduled previously. 
schedule

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -527,7 +527,9 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


jipperinbham (approved):
LGTM! 

Should we also have a guard in flyd to prevent a machine with a schedule from being updated without a schedule?

shortdiv:
> Should we also have a guard in flyd to prevent a machine with a schedule from being updated without a schedule?

Yeah probably, I think there's a high likelihood that someone may send in a config without a schedule for a scheduled machine and accidentally override. We may want to let it happen with a warning or prevent it and send an error saying that they're overriding a scheduled machine

PR #1295, opened 2022-09-23
fix: Catch error if app is running on machines

shortdiv:
- [x]  fly vm status →fly machine status . 
- [x] fly scale vm→ fly machine update 
- [x]  fly scale memory→ fly machine update 
- [x] fly scale show→ fly machine status 
- [x] fly scale count → fly machine clone 

Screen Shot 2022-09-22 at 5 03 54 PM


davissp14 (approved):
LGTM!

dangra (approved):

dangra:
Nice addition! I'd extract much of it into a single function that gets the client from context, queries app and check for the platform to be Nomad or return the "unsupported" error.

PR #1299, opened 2022-09-23
add name, no-replicas, enable/disable-eviction, and plan flag to redi…

rubys:
…s create

### DIFF
--- a/internal/command/redis/create.go
+++ b/internal/command/redis/create.go
@@ -30,6 +30,27 @@ func newCreate() (cmd *cobra.Command) {
 	flag.Add(cmd,
 		flag.Org(),
 		flag.Region(),
+		flag.String{
+			Name:        "name",
+			Shorthand:   "n",
+			Description: "The name of your Redis cluster",
+		},
+		flag.Bool{
+			Name:        "no-replicas",
+			Description: "No replica regions",
+		},
+		flag.Bool{
+			Name:        "enable-eviction",
+			Description: "Evict objects when memory is full",
+		},
+		flag.Bool{
+			Name:        "disable-eviction",
+			Description: "Disallow writes when the max data size limit has been reached",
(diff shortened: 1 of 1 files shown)


internal/command/redis/create.go
```diff
@@ -30,6 +30,27 @@ func newCreate() (cmd *cobra.Command) {
...
+		},
+		flag.Bool{
+			Name:        "no-replicas",
+			Description: "No replica regions",
+		},
+		flag.Bool{
+			Name:        "enable-eviction",
+			Description: "Evict objects when memory is full",
+		},
+		flag.Bool{
```
jsierles, line 46: What's the purpose of this flag if 'disabled' is the default?

internal/command/redis/create.go
```diff
@@ -30,6 +30,27 @@ func newCreate() (cmd *cobra.Command) {
 	flag.Add(cmd,
 		flag.Org(),
 		flag.Region(),
+		flag.String{
+			Name:        "name",
+			Shorthand:   "n",
+			Description: "The name of your Redis cluster",
```
jsierles, line 36: Probably should name this 'database' instead of 'cluster', as 'database' is used everywhere else in this command.

jsierles (approved):

internal/command/redis/create.go
rubys: Default actually is prompt.  If you specify either disabled or enabled you won't get prompted.

internal/command/redis/create.go
rubys: ok, will fix

rubys: Fixed: https://github.com/superfly/flyctl/commit/2fa4ff1e0d565b11e36f28b4a14edc8139a86130

PR #1303, opened 2022-09-26
Watch machine heatlh checks during update/deployments

rugwirobaker:


internal/watch/watch.go
```diff
@@ -288,3 +289,81 @@ func renderLogs(ctx context.Context, alloc *api.AllocationStatus) {
...
+
+	tb := render.NewTextBlock(ctx, "Monitoring health checks")
+
+	tb.Println()
+
+	var pass, warn, fail int
+
+	var errCount int
+
+	ctx, cancel := context.WithTimeout(ctx, 120*time.Second)
```
davissp14, line 307: We should avoid using a global timeout here.  What's reasonable will be based on the number of machines we are monitoring.

internal/command/postgres/update.go
```diff
@@ -192,6 +194,14 @@ func runUpdate(ctx context.Context) error {
 		}
 	}
 
+	if !flag.GetBool(ctx, "detach") {
```
davissp14, line 197: Instead of running health checks at the end of the update process, we should look to monitor the health of the individual machine after each update.  Ideally the cluster should be in a healthy state before any new operation is executed. 

E.G.
Update Machine 1  ( Wait for passing health checks )
Update Machine 2 ( Wait for passing health checks )
Step down ...          ( Wait for passing health checks )

davissp14 (changes requested):
Awesome start!  Just a few comments.

internal/watch/watch.go
```diff
@@ -288,3 +289,95 @@ func renderLogs(ctx context.Context, alloc *api.AllocationStatus) {
...
+	defer cancel()
+
+	fmt.Fprintln(io.Out)
+	fmt.Fprintln(io.Out, colorize.Green("==> "+"Monitoring health checks"))
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(100 * time.Millisecond):
```
davissp14, line 311: My understanding is that this can lead to a memory leak: https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/

Also, we should extend this interval a bit. 100ms is pretty aggressive for a loop that hits an external API.  1-2 seconds should be ok.

internal/watch/watch.go
```diff
@@ -288,3 +289,95 @@ func renderLogs(ctx context.Context, alloc *api.AllocationStatus) {
 		)
 	}
 }
+
+func MachinesChecks(ctx context.Context) error {
```
davissp14, line 293: We may want to adjust this logic a bit.  

So instead of having a single function that monitors health checks for all machines, we should consider having a single function that takes a machine and monitors its associated health checks.   There may be times where we need to be explicit about which Machines we want to monitor, for example, we only care about PG machines within the context of a PG app.  We can ignore any non-PG Machines that may have been provisioned.  

I think this will make it a bit more flexible as we can leverage the same function for restarts/updates and provisions.  Restarts/updates will need to monitor machine independently one-by-one, while post-provision we could look to monitor Machines concurrently.

davissp14 (changes requested):
Few notes.

davissp14 (approved):

PR #1304, opened 2022-09-26
Clarify --org takes lowercase

dekimir:
If I just copy the capitalized org name from the dashboard, `launch` errors out with "organization CapOrgName not found".


dangra:
Hi @dekimir, thanks for pointing this out.  The problem comes from the dashboard showing the *org name* although flyctl takes the *organization slug* which is lowercased and dashed. 

The slug is reflected in the url of the dashboard. That said, I think we can do better on dashboard to make it clear what id to use to reference an organization.  

image
 

Your PR is fine, it improves the situation for flyctl users, but what do you think about referencing it as Organization slug in the flag description?

dekimir:
Thanks for clarifying, @dangra.  It's definitely better to refer to "slug" in the flag description.  Let me think how best to do that, so it's obvious to the reader what "slug" refers to.

dangra:
> The best I could come up with is `Organization that will own the app (use org slug here; see "fly orgs show")`.

that's a good idea. I'd use `fly orgs list` tho

dangra (approved):

dangra:
grrr... trying to merge but Github is making up 
image

dangra:
@dekimir I've asked a college to try merge this PR and the error is the same. GH is not responding to support requests on this one so I opted to create https://github.com/superfly/flyctl/pull/1329 keeping you as commit author. Closing this one.

dangra:
#1329 is merged. Kudos to @dekimir. thanks!

PR #1308, opened 2022-09-26
Faster PG status by using role from check output

dangra:


internal/command/status/machines.go
```diff
@@ -81,18 +43,25 @@ func renderMachineStatus(ctx context.Context, app *api.AppCompact) (err error) {
...
-
+func renderPGStatus(ctx context.Context, app *api.AppCompact, machines []*api.Machine) (err error) {
+	io := iostreams.FromContext(ctx)
 	rows := [][]string{}
 
-	for role, machine := range machines {
+	for _, machine := range machines {
+		role := "unknown"
+		for _, check := range machine.Checks {
+			if check.Name == "role" {
```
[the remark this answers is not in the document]
dangra: not sure what is the final intention, but no, it won't .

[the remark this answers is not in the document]
dangra: I don't think it is a bad thing tho. We want to be transparent about the lack of writeable leader, don't we?

internal/command/status/machines.go
```diff
@@ -81,18 +43,25 @@ func renderMachineStatus(ctx context.Context, app *api.AppCompact) (err error) {
...
+				if check.Status == "passing" {
+					role = check.Output
+				} else {
+					role = "failing"
```
rugwirobaker, line 60: We usually print `error` when the role is not retrievable.

dangra: No hard feelings on this one. I'll change to "error" if that fits better :)

rugwirobaker:
@davissp14 for some weird reason it's showing that both machines have `replica` as role.

davissp14:
@rugwirobaker Interesting. Do you see the same when you hit the role health check endpoint manually?

rugwirobaker:
@davissp14 nope, in fact on the main branch where we are still doing it manually it works fine.

davissp14 (approved):

internal/command/status/machines.go
dangra: addressed by https://github.com/superfly/flyctl/pull/1308/commits/5c14a6940ef17b8a4a888bdd068bedbd77582abb

Issue #1310, opened 2022-09-26
`fly deploy` resets vm size to shared-cpu-1x 256MB on machine-based apps


lewisl9029:
**Describe the bug**

Was asked to report here from the forum thread: https://community.fly.io/t/preview-deploying-applications-on-machines-with-flyctl/6171/8?u=lewis

> I had my app set to use 2048MB memory and 2 shared cpus through fly machine update. But running fly deploy seems to have reset it to 1 shared cpu and 256MB memory.

This is reproducible every time I deploy. For now I've been working around it by updating my CI script to set the size immediately after each deploy. Works for now since there's barely any traffic, but obviously not sustainable.

Doesn't seem to be OS specific since it happens when I do `fly deploy` in CI (linux) as well as locally (macOS).

* Operating system: macOS Monterey 12.6
* `fly version`: fly v0.0.396 darwin/arm64 Commit: edf8f512 BuildDate: 2022-09-23T19:34:38Z

** Paste your `fly.toml`

```toml
app = "reflame-deployer"

[build]

[http_service]
  internal_port = 58001
  force_https = true

  [http_service.concurrency]
    hard_limit = 10
    soft_limit = 2
    type = "requests"

[deploy]
```

ignoramous:
> This is reproducible every time I deploy. For now I've been working around it by updating my CI script to set the size immediately after each deploy. Works for now since there's barely any traffic, but obviously not sustainable.

There's a bunch of things `flyctl m run` does differently to `flyctl deploy` which causes this. I am not sure why, because to me, `flyctl deploy` should be equivalent to calling `flyctl m run` a bunch of times.

remorses:
I had this problem today, not sure if it's a new issue

PR #1311, opened 2022-09-26
machines: preserve existing config on deploy

ignoramous:
Ref: https://github.com/superfly/flyctl/issues/1310


shortdiv (approved):

fideloper:
Tested this PR locally and it seems to work great! 

The web UI never updated as a result of me running `fly m update <machineid> --cpus 2 --memory 2048`, but SSHing and and running `htop` / `free -h` showed me that it did indeed scale, and stayed scaled after a deployment.

I'm curious if you've used other commands to scale a machine?

ignoramous:
> I'm curious if you've used other commands to scale a machine?

I've only ever used the `fly m update` command. The thing with `fly deploy` was, all machines vms end up with the *same* vm-config.

PR #1314, opened 2022-09-28
Add fly pg failover command

rugwirobaker:
- Add fly pg failover command

### DIFF
--- a/internal/command/postgres/failover.go
+++ b/internal/command/postgres/failover.go
@@ -50,12 +50,12 @@ func runFailover(ctx context.Context) (err error) {
 		return fmt.Errorf("get app: %w", err)
 	}
 
-	if app.PlatformVersion != "machines" {
-		return fmt.Errorf("failerover is only supported for machines apps")
-	}
-
 	if !app.IsPostgresApp() {
 		return fmt.Errorf("app %s is not a Postgres app", app.Name)
+	}
+
+	if app.PlatformVersion != "machines" {
+		return fmt.Errorf("failerover is only supported for machines apps")
 	}
 
 	agentclient, err := agent.Establish(ctx, client)
--- a/internal/command/postgres/failover.go
+++ b/internal/command/postgres/failover.go
@@ -49,6 +49,10 @@ func runFailover(ctx context.Context) (err error) {
 		return fmt.Errorf("get app: %w", err)
 	}
 
+	if app.PlatformVersion != "machines" {
+		return fmt.Errorf("failerover is only supported for machines apps")
+	}
+
 	if !app.IsPostgresApp() {
 		return fmt.Errorf("app %s is not a Postgres app", app.Name)
 	}
--- a/internal/command/postgres/failover.go
+++ b/internal/command/postgres/failover.go
@@ -1,0 +1,96 @@
+package postgres
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


internal/command/postgres/failover.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("list of machines could not be retrieved: %w", err)
+	}
+	ctx = flaps.NewContext(ctx, flapsClient)
+
+	machines, err := flapsClient.List(ctx, "started")
+	if err != nil {
+		return fmt.Errorf("machines could not be retrieved %w", err)
+	}
+
+	leader, err := fetchPGLeader(ctx, machines)
```
davissp14, line 78: This is fine as is, but anything that results in a `stolonctl` call can be  issued from any member in the cluster.

internal/command/postgres/failover.go
```diff
@@ -0,0 +1,101 @@
...
+
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("get app: %w", err)
+	}
+
+	if !app.IsPostgresApp() {
+		return fmt.Errorf("app %s is not a Postgres app", app.Name)
+	}
+
```
davissp14, line 55: We may want to toss an error if the App is on Nomad.

internal/command/postgres/failover.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return fmt.Errorf("list of machines could not be retrieved: %w", err)
+	}
+	ctx = flaps.NewContext(ctx, flapsClient)
+
+	machines, err := flapsClient.List(ctx, "started")
+	if err != nil {
+		return fmt.Errorf("machines could not be retrieved %w", err)
+	}
+
```
davissp14, line 77: We should look to acquire a cluster wide lease before issuing the failover.

internal/command/postgres/failover.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("failover is not available for standalone postgres")
+	}
+
+	pgclient := flypg.New(appName, dialer)
+
+	fmt.Fprintf(io.Out, "Performing a failover\n")
+	if err := pgclient.Failover(ctx); err != nil {
+		return fmt.Errorf("failed to trigger failover %w", err)
+	}
+
```
davissp14, line 98: What are your thoughts on waiting on health checks here?

davissp14 (changes requested):
Great start!  Just a few comments.

internal/command/postgres/failover.go
rugwirobaker: I'm only fetching the leader to check the image version It's running. That's going to change in https://github.com/superfly/flyctl/pull/1315

davissp14 (approved):
LGTM!

internal/command/postgres/failover.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return fmt.Errorf("machines could not be retrieved %w", err)
+	}
+
+	if err := hasRequiredVersionOnMachines(machines, MinPostgresHaVersion, MinPostgresHaVersion); err != nil {
+		return err
+	}
+
+	// acquire cluster wide lock
+	for _, machine := range machines {
```
davissp14, line 88: We should perform this after the single node check below.

PR #1318, opened 2022-09-28
Use flaps.Restart to restart postgres on machines

rugwirobaker:
Waiting for health checks in follow-up PR: https://github.com/superfly/flyctl/pull/1303


internal/command/postgres/restart.go
```diff
@@ -269,82 +188,30 @@ func machinesHardRestart(ctx context.Context, machines []*api.Machine) (err erro
...
+			if err = flapsClient.Restart(ctx, in); err != nil {
 				return fmt.Errorf("failed to restart vm %s: %w", replica.ID, err)
 			}
 			// wait for health checks to pass
 		}
 	}
 
-	// Don't perform failover if the cluster is only running a
-	// single node.
-	if len(allocs) > 1 {
```
davissp14, line 333: I think we should still look to perform a failover here.

davissp14 (changes requested):
Overall looks great!  Just one small comment.

PR #1322, opened 2022-09-29
Reworking pg upgrade process and adding confirmation prompt

davissp14:
**Notable changes:**

* Adds a confirmation prompt before performing update.
* Ensures any non-pg machines running under the same app are ignored.
* Added check to ensure PG machines do not span major versions. 

Screen Shot 2022-09-29 at 4 17 15 PM

### DIFF
--- a/internal/command/postgres/update.go
+++ b/internal/command/postgres/update.go
@@ -96,7 +96,7 @@ func runUpdate(ctx context.Context) error {
 		image := fmt.Sprintf("%s:%s", machine.ImageRef.Repository, machine.ImageRef.Tag)
 		latestImage, err := client.GetLatestImageDetails(ctx, image)
 		if err != nil {
-			return fmt.Errorf("unalbe to fetch latest image details for %s: %w", image, err)
+			return fmt.Errorf("unable to fetch latest image details for %s: %w", image, err)
 		}
 
 		if latest == nil {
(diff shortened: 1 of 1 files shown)


dangra (approved):
Looks great! Are you going to add the health check test before merging or is that for a follow up PR?

internal/command/postgres/update.go
```diff
@@ -150,43 +191,34 @@ func runUpdate(ctx context.Context) error {
...
 			pgclient := flypg.New(app.Name, dialer)
 
-			fmt.Fprintf(io.Out, "Performing a failover\n")
+			fmt.Fprintf(io.Out, "Performing failover\n")
 			if err := pgclient.Failover(ctx); err != nil {
 				return fmt.Errorf("failed to trigger failover %w", err)
 			}
 		}
 
 		fmt.Fprintf(io.Out, "Updating leader\n")
```
dangra, line 218: nitpick: the message is miss leading (pun-intended) in case of a failover  🔍

davissp14:
> Looks great! Are you going to add the health check test before merging or is that for a follow up PR?

That will have to be a follow-up PR as some of it will rely on @rugwirobaker 's on-going work.

internal/command/postgres/update.go
dangra: heh, the weird part is that the failover isn't happening for single node postgres. Not a big deal

PR #1333, opened 2022-10-05
WIP: SFTP

tqbf:
I'm not even trying to get the UX right here, I just want to get the basics working so we can play with it a bit.


tqbf:
Some updates:

SFTP commands are now accessible from `flyctl sftp`.

The commands now:

* `flyctl sftp shell` - an interactive SFTP shell with `get`, `put`, `cd`, and `ls`.

* `flyctl sftp get <file>` - fetch a remote file non-interactively.

* `flyctl sftp find [root]` - dump a list of files starting from `root`; this will probably go away, since I just wrote it to POC SFTP and it doesn't do much.

matteing:
This is awesome work! Thanks so much for shipping this 👍

PR #1342, opened 2022-10-07
machines image updates

rugwirobaker:
- A little bit of refactoring first
- fly image update supports machines now
- Simply postgres image updates

### DIFF
--- a/flypg/launcher.go
+++ b/flypg/launcher.go
@@ -282,6 +282,11 @@ func (l *Launcher) getPostgresConfig(config *CreateClusterInput) *api.MachineCon
 		},
 	}
 
+	// medatata
+	machineConfig.Metadata = map[string]string{
+		"image_version_tracking_enabled": "true",
+	}
+
 	// Restart policy
 	machineConfig.Restart.Policy = api.MachineRestartPolicyAlways
 
--- a/internal/command/image/update.go
+++ b/internal/command/image/update.go
@@ -5,20 +5,28 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
+	"github.com/superfly/flyctl/agent"
+	"github.com/superfly/flyctl/flaps"
+	"github.com/superfly/flyctl/flypg"
 	"github.com/superfly/flyctl/iostreams"
 
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/client"
 	"github.com/superfly/flyctl/internal/app"
 	"github.com/superfly/flyctl/internal/command"
+	machines "github.com/superfly/flyctl/internal/command/machine"
 	"github.com/superfly/flyctl/internal/flag"
 	"github.com/superfly/flyctl/internal/prompt"
 	"github.com/superfly/flyctl/internal/render"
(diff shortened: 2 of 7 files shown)


internal/command/image/update.go
```diff
@@ -142,6 +174,230 @@ func runUpdate(ctx context.Context) error {
...
+
+	// Tracks latest eligible version
+	var latest *api.ImageVersion
+
+	// Filter out eligible that are not eligible for updates
+	var eligible []*api.Machine
+
+	for _, machine := range candidates {
+		// Skip machines that are not running our internal images.
+		if machine.ImageRef.Repository != validRepository && machine.ImageVersionTrackingEnabled() {
```
davissp14, line 222: We should ditch the `machine.ImageRef.Repository != validRepository` condition.

internal/command/image/update.go
```diff
@@ -49,12 +57,36 @@ The update will perform a rolling restart against each VM, which may result in a
...
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("get app: %w", err)
+	}
+
+	switch app.PlatformVersion {
+	case "nomad":
+		return updateImageForNomad(ctx)
+	case "machines":
+		return updateImageForMachines(ctx)
```
davissp14, line 84: What do you think about just calling the postgres specific update logic here in the event the app has role `postgres_cluster` and work to keep the image update logic non-specific.

davissp14 (changes requested):

internal/command/image/update.go
davissp14: Right, we should do our best to keep PG specific orchestration out of the general update logic.

internal/command/image/update.go
```diff
@@ -49,12 +54,36 @@ The update will perform a rolling restart against each VM, which may result in a
...
+	var (
+		appName = app.NameFromContext(ctx)
+		client  = client.FromContext(ctx).API()
+	)
+
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("get app: %w", err)
+	}
+
```
davissp14, line 76: Looks like we could save a graphql query if we just pass app into updateImageForNomad and updateImageForMachines

davissp14 (changes requested):

flypg/launcher.go
```diff
@@ -282,6 +282,11 @@ func (l *Launcher) getPostgresConfig(config *CreateClusterInput) *api.MachineCon
 		},
 	}
 
+	// medatata
```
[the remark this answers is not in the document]
rugwirobaker: How do you propose we do it?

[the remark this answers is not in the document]
davissp14: Yeah, probably fine for now.  The two cases where it could lead to issues are:

1. If someone moves to a fork or transitions their image to something other than `flyio/postgres`.
This is fine, as the call to `client.GetLatestImageDetails(ctx, image)` will fail. 

2. If someone transitions away from a fork to our managed image.
This case is less ideal, as even though they are running our managed image the metadata will prevent the operation from going through.

davissp14 (approved):

PR #1353, opened 2022-10-09
handle Windows colors

mattn:
On Windows, `flyctl pg connect` or `flyctl ssh connect` break screen.
image


mattn:
@jsierles Could you please take a look?

PR #1356, opened 2022-10-10
Nudge users to update machine images if we are tracking the version

rugwirobaker:


internal/command/status/machines.go
```diff
@@ -31,6 +37,47 @@ func renderMachineStatus(ctx context.Context, app *api.AppCompact) error {
 		return renderPGStatus(ctx, app, machines)
 	}
 
+	// Tracks latest eligible version
+	var latest *api.ImageVersion
+
```
davissp14, line 42: Doesn't look like this is working.
`fly image show`
Screen Shot 2022-10-10 at 12 24 11 PM

`fly status`
Screen Shot 2022-10-10 at 12 25 23 PM

davissp14 (approved):
Looks good!

PR #1371, opened 2022-10-12
Default PG provisions to Machines

davissp14:


lubien (approved):

dangra (approved):

dangra:
@davissp14 don't you think it is important to keep the `--machines` flag, add a warning in case it is used, and add the `--nomad` flag just in case we break some automated process.

PR #1372, opened 2022-10-12
Simplify the watch.MachineChecks function output

rugwirobaker:


internal/command/postgres/postgres.go
```diff
@@ -173,3 +161,18 @@ func nomadNodeRoles(ctx context.Context, allocs []*api.AllocationStatus) (leader
...
+func machineRole(machine *api.Machine) (role string) {
+	role = "unknown"
+
+	for _, check := range machine.Checks {
+		if check.Name == "role" {
+			if check.Status == "passing" {
+				role = check.Output
+			} else {
+				role = "error"
+			}
```
dangra, line 174: it should return asap after it founds the check that contains the role.

davissp14 (approved):

PR #1405, opened 2022-10-24
change node from buildpacks to docerfiles

rubys:


scanner/templates/node/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN volta install node@${NODE_VERSION} yarn@${YARN_VERSION}
+{{ else -}}
+RUN volta install node@${NODE_VERSION}
+{{ end }}
+#######################################################################
+
+RUN mkdir /app
+WORKDIR /app
+
+ENV NODE_ENV production
```
jphenow, line 21: Would we also want node-env set to production in the runner definition below?

scanner/node.go
```diff
@@ -1,18 +1,90 @@
...
+		Family: "NodeJS",
+		Port:   8080,
 		Env: map[string]string{
 			"PORT": "8080",
 		},
 	}
 
+	scripts, ok := result["scripts"].(map[string]interface{})
+
+	if !ok || scripts["start"] == nil {
```
jphenow, line 36: Probably doesn't need to be this version but we could prompt for the script or output info about why we did buildpacks instead?

scanner/node.go
```diff
@@ -1,18 +1,90 @@
...
+	if err != nil {
+		return nil, nil
+	}
+
 	s := &SourceInfo{
-		Builder: "heroku/buildpacks:20",
-		Family:  "NodeJS",
-		Port:    8080,
+		Family: "NodeJS",
+		Port:   8080,
```
jphenow, line 28: Might be good to include information about using `PORT` and `8080` in the final output in case folks didn't follow standards here.

scanner/templates/node/Dockerfile
rubys: good catch!

scanner/node.go
rubys: I'm hesitant to rewrite the package.json.  What's more likely to happen is people are going to identify other scripts that should be used.  For example, I've seen package.json files with client and server scripts - we would be interested in the latter.

scanner/node.go
rubys: Good idea.

scanner/node.go
jphenow: Good point, weird to have this update code in their realm. Perhaps a note that we've configured the build to run `run build` is probably enough - maybe enough to point them in a direction if they choose to update it themselves.

PR #1406, opened 2022-10-24
Use supevisord over s6 init

fideloper:
This avoids issues of S6 v3, which requires that it is run as pid 1. We cannot do on Fly - Fly's VMs have their own `init` program that runs as pid 1.

## TODO:

- [x] Use supervisor over s6
- [ ] Devise a system where serversideup-defined (the base container) env vars can work with this setup
    - Or take away those (largely unused) config options
    - S6 does magic to replace config files with content from env var values


fideloper:
This PR will be obsolesced by the use of https://hub.docker.com/r/fideloper/fly-laravel base containers

PR #1410, opened 2022-10-25
Fallback to legacy pg role endpoint when modern one 404s

tvdfly:


tvdfly:
tested that the legacy endpoint parsing code works by building+running `fly pg connect` with this patch:

```patch
diff --git a/flypg/pg.go b/flypg/pg.go
index 03e62813..1f232704 100644
--- a/flypg/pg.go
+++ b/flypg/pg.go
@@ -1,8 +1,10 @@
 package flypg
 
 import (
+       "bytes"
        "context"
        "fmt"
+       "io/ioutil"
        "net/http"
 
@@ -128,6 +130,10 @@ func (c *Client) NodeRole(ctx context.Context) (string, error) {
        out := new(NodeRoleResponse)
 
        err := c.Do(ctx, http.MethodGet, endpoint, nil, out)
+       // FIXME: remove this line
+       err = newError(http.StatusNotFound, &http.Response{
+               Body: ioutil.NopCloser(bytes.NewBufferString("FIXME: this is a test that should be removed!!!")),
+       })
        if err != nil && ErrorStatus(err) == http.StatusNotFound {
                terminal.Debugf("404 response from %s endpoint. Calling legacy endpoint.\n", endpoint)
                return c.legacyNodeRole(ctx)

```

PR #1414, opened 2022-10-25
Fix/add update text

shortdiv:
With a machine update, you don't get notified of the changes to config unless you run a machine status right after, this PR highlights the config bits that get changed when an update is made

updateconfig

Edit: Moved to using a different library, now the diff looks like this:
Screen Shot 2022-10-28 at 2 21 15 AM

I think color coding it would be cool but for now it shows the diff!

Edit OMGeeeee
Screen Shot 2022-10-28 at 3 24 40 AM

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -211,8 +211,8 @@ type MachineConfig struct {
 
 type MachineDiff struct {
 	Key     string
-	Initial interface{}
-	New     interface{}
+	Initial string
+	New     string
 }
 
 type MachineLease struct {
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -209,6 +209,12 @@ type MachineConfig struct {
 	Checks    map[string]MachineCheck `json:"checks,omitempty"`
 }
 
+type MachineDiff struct {
+	Key     string
+	Initial interface{}
+	New     interface{}
+}
+
 type MachineLease struct {
 	Status string `json:"status"`
 	Data   struct {
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -209,12 +209,6 @@ type MachineConfig struct {
 	Checks    map[string]MachineCheck `json:"checks,omitempty"`
 }
 
-type MachineDiff struct {
-	Key     string
-	Initial string
-	New     string
-}
-
 type MachineLease struct {
 	Status string `json:"status"`
 	Data   struct {
--- a/go.mod
+++ b/go.mod
@@ -70,6 +70,7 @@ require (
 	github.com/go-playground/locales v0.14.0 // indirect
 	github.com/go-playground/universal-translator v0.18.0 // indirect
 	github.com/google/btree v1.0.1 // indirect
+	github.com/google/go-cmp v0.5.9 // indirect
 	github.com/kr/fs v0.1.0 // indirect
 	github.com/leodido/go-urn v1.2.1 // indirect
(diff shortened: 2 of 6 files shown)


api/machine_types.go
```diff
@@ -209,6 +209,12 @@ type MachineConfig struct {
 	Checks    map[string]MachineCheck `json:"checks,omitempty"`
 }
 
+type MachineDiff struct {
```
jipperinbham, line 212: I don’t think this needs to be exported (same goes for the fields).

internal/command/machine/run.go
```diff
@@ -514,24 +517,68 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 			return
 		}
 		machineConf.Guest = guest
+
```
jipperinbham, line 520: This seems like a lot to keep track of and potentially easy to miss when adding new fields. I feel like I’ve seen some library that can show a JSON diff in a meaningful way and wonder if that could be easier to maintain while still keeping the valuable diff.

shortdiv:
I did an exploration using a diffing library https://play.golang.com/p/Ly5YQX671QK and honestly it's not great :/ Because config is so deeply nested, it's a struggle to surface this information without a lot of extra work. I think maybe the best thing to do is to run status after an update? It won't show the changes but it'll show the new config at least?

internal/command/machine/run.go
rugwirobaker: Perhaps https://github.com/google/go-cmp?

shortdiv: Oh! we were looking into https://github.com/nsf/jsondiff but the library you mentioned looks cool. I like that the diff shows (+) and (-) so we don't have to format the diff ourselves 

https://play.golang.com/p/SQfAG9nPsoN

go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/go-playground/locales v0.14.0 // indirect
 	github.com/go-playground/universal-translator v0.18.0 // indirect
 	github.com/google/btree v1.0.1 // indirect
+	github.com/google/go-cmp v0.5.9 // indirect
```
jipperinbham, line 73: need to `go mod tidy` since this isn't an indirect dep

internal/command/machine/update.go
```diff
@@ -106,6 +106,8 @@ func runUpdate(ctx context.Context) (err error) {
 	fmt.Fprintln(out, colorize.Yellow(fmt.Sprintf("Machine %s has been updated\n", machine.ID)))
 	fmt.Fprintf(out, "Instance ID has been updated:\n")
 	fmt.Fprintf(out, "%s -> %s\n\n", prevInstanceID, machine.InstanceID)
+	fmt.Fprintln(out, "The following config has been updated")
```
jipperinbham, line 109: I wonder if we shouldn't show the diff *before* sending the update? It would make things interactive but seems like that's an ok change to make?

shortdiv: oh yeah I'm good with doing that, esp so users don't accidentally change their config without double checking

shortdiv: Wonder if that can be addressed in another pr since it'll be adding an interactive component to updates

jipperinbham: Yea, a followup PR is good. Maybe we show them what's changing still before sending the update so the followup change is adding a prompt and not also change the order of output?

PR #1427, opened 2022-10-28
Add interactive prompt for machine update

shortdiv:
This builds off of #1414 to add an interactive element to machine update so users can check the changes to their config before confirming an update

Screen Shot 2022-10-28 at 5 12 56 PM

Screen Shot 2022-10-28 at 5 12 27 PM

### DIFF
--- a/internal/command/machine/update.go
+++ b/internal/command/machine/update.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/AlecAivazis/survey/v2"
 	"github.com/spf13/cobra"
 
 	"github.com/superfly/flyctl/api"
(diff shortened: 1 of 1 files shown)


internal/command/machine/update.go
```diff
@@ -86,6 +88,33 @@ func runUpdate(ctx context.Context) (err error) {
...
+		} else {
+			str += val
+		}
+	}
+	fmt.Fprintln(out, "The following config is being updated")
+	fmt.Fprintln(out, str)
+
+	// interactive update //
+	name := false
+	prompt := &survey.Confirm{
```
rugwirobaker, line 109: The prompt package has you covered: https://github.com/superfly/flyctl/blob/b9e83116ebbb045b94fa0bb25930f90ab4380797/internal/prompt/prompt.go#L125-L142

PR #1431, opened 2022-10-29
Fix launchers by ensuring the app name is always determined correctly

jsierles:
App names can come from config, a flag, or from a newly created app during launch. Previously, deployments were relying only on the app value added to context by the deploy command. Since `launch` skips the context step, we moved app name logic into `determineAppConfig`. But, we neglected to ensure that the app name placed in the config is what's used everywhere during deployments. So, deployments initiated during the launch phase would fail.

### DIFF
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -497,7 +497,7 @@ func run(ctx context.Context) (err error) {
 			deployNow = true
 		}
 	}
-	fmt.Printf("%#v", appConfig)
+
 	if deployNow {
 		return deploy.DeployWithConfig(ctx, appConfig)
 	}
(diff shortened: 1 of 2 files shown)


rugwirobaker (approved):

internal/command/launch/launch.go
```diff
@@ -497,7 +497,7 @@ func run(ctx context.Context) (err error) {
 			deployNow = true
 		}
 	}
-
+	fmt.Printf("%#v", appConfig)
```
jeromegn, line 500: Probably want to remove this.

jsierles: 🤔

PR #1436, opened 2022-11-02
Check the container environment matches Gemfile.lock

bradgessler:
Putting together a Rake task that checks if the `Gemfile.lock` matches the configuration of the image.

### DIFF
--- a/scanner/templates/rails/standard/Dockerfile
+++ b/scanner/templates/rails/standard/Dockerfile
@@ -76,7 +76,8 @@ RUN gem update --system --no-document && \
     gem install -N bundler -v ${BUNDLER_VERSION}
 
 COPY Gemfile* ./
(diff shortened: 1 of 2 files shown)


bradgessler:
Not a fan of how late in the game I verify the environment. I should do it right after I copy in `Gemfile` and `Gemfile.lock` in a separate script, but then I'd have to copy another file into the image.

I'm also not a fan of `fly:verify` being in a user-facing Rakefile. We should be keeping that free of Fly systems checks.

The idea is I could copy this file into the image, run the check, and get it out of there, without this file ever being seen by the user. If it gives them trouble or the don't want the check, they could remove it from the `RUN` command in Docker.

dangra:
Hi @bradgessler , I am closing as the draft PR is dusty and the Rails launcher has changed dramatically since the introduction of dockerfile-rails. thanks!

Issue #1449, opened 2022-11-11
Rails: fly deploy error, ‘ruby\r’: No such file or directory


rahul3v:
**Describe the bug**
on fly deploy rails app, throwing error  ‘ruby\r’: No such file or directory`

* Operating system : windows 11
* `fly version`  :  v0.0.429 
* `Ruby version`  :  v3.1.2 

**Command output:**
```text
......
 => CACHED [stage-3 7/9] RUN ls -l                                                                                  0.0s 
 => CACHED [stage-3 8/9] RUN pwd                                                                                    0.0s 
 => ERROR [stage-3 9/9] RUN bin/rails fly:build                                                                     0.3s 
------
 > [stage-3 9/9] RUN bin/rails fly:build:
#22 0.276 /usr/bin/env: ‘ruby\r’: No such file or directory
------
Error failed to fetch an image or build from source: error building: executor failed running [/bin/bash -o pipefail -c ${BUILD_COMMAND}]: exit code: 127
```
![error-ruby-r-not-found](https://user-images.githubusercontent.com/24385409/201338938-bdcb9fe1-dd07-480e-9092-a7803103a51b.jpg)

PR #1450, opened 2022-11-11
Fix :  ‘ruby\r’: No such file or directory #1449

rahul3v:
Fixing Error for ruby and rails app #1449

Works fine after making the specified changes

**Kindly verify from your side**
initially tried this but my app didn't load after that, after deployment
```
sed -i 's/ruby.exe/ruby/' /app/bin/* && \
sed -i 's/ruby\r/ruby/' /app/bin/* && \
```
but in this case everything works just fine 
```
sed -i 's/ruby\r/ruby/' /app/bin/* && \
```

**For existing ruby applications**, having `Dockerfile` just replace in that file
```
sed -i 's/ruby.exe/ruby/' /app/bin/* && \
```
into 
```
sed -i 's/ruby\r/ruby/' /app/bin/* && \
```


rubys (changes requested):
Hmm.  When I run on windows, I definitely get `.exe` in the path so I don't want to lose that behavior.  sed's support for regexes is subpar, and I previously merged a change in that adds `\r*` when what I really wanted was `\r?`. What I guess we really want is the equivalent of s/ruby(\.exe)?\n?/ruby/.

rahul3v:
Well I have tried that `s/ruby.exe\r*/ruby/`,  and `s/ruby(.exe)?\n?/ruby/` now and many more possible combination
but the only thing that works for me is  `s/ruby\r/ruby/`

Just one more thing (I might be wrong) : 
I don't think `.exe` is dependable as the ruby bundle we use 
```
FROM quay.io/evl.ms/fullstaq-ruby:${RUBY_VERSION}-${VARIANT} as base
```
which is I believe of linux and after copying the file which is created **under windows installed ruby** 
we copy that file and put that into the docker we created and then run the `s/ruby\r/ruby/` to replace all that 

I hope that make sense to you

rubys:
Just curious, how did you install ruby?  I used RubyInstaller.

In any case, it looks handling this case without regressing what the current code supports will either require:

- Multiple sed lines, one with .exe and one without
- Actually reading either one or all of the files in the bin directory, and emitting just the sed lines that are needed.  For non windows users, that will be zero sed lines.  For most windows users one will do, but which one will depend on what your installed version of Ruby actually does.

rahul3v:
Well I also used RubyInstaller,  version `v3.1.2`

Have you tried deploying your project on fly.io, with your changes

As you mentioned 
 - Multiple sed lines, one with .exe and one without
   - >Actually I tried this one but after deployment I was not able to access the project via url

 - depend on what your installed version of Ruby actually does.
    - > Have you tried with older versions to deploy it on fly.io with yours and mine changes 

If there is any possibility we can connect, let me know about that :)

rahul3v:
Have **verified again with Multiple sed lines**, one with .exe and one without, **everything works normal now**, after deployment, Just confirm it once under older version from your side and then you can merge it :)

rubys (approved):

rubys:
I completely uninstalled ruby installer and reinstalled it.  I still see `ruby.exe` in the `bin/rails` file.

Your latest patch doesn't have the `\r*` on the `ruby.exe` line, so I added it.  With that in place, I'm now going to merge it.

In the future, I'd like to see something like the following be put in place, which will produce as many sed lines as necessary (typically zero or one, but who knows?):

```ruby
shebangs = Dir['bin/*'].map {|file| IO.read(file).lines.first}.join
rubies = shebangs.scan(%r{#!/usr/bin/env (ruby.*)}).flatten.uniq - ['ruby']

puts rubies.map {|ruby| "sed -i 's/#{ruby.inspect[1...-1]}/ruby/' /app/bin/* && \\"}
```

This is relatively straightforward in Ruby, but will require a bit more code to implement in go.

rubys:
Slightly improved version in a gem I'm working on: https://github.com/superfly/fly-rails/commit/011c83292f1512c3aa219ddec0edc5c899b357db

rahul3v:
> I completely uninstalled ruby installer and reinstalled it. I still see `ruby.exe` in the `bin/rails` file.

Yup, you can see that in the ruby installed file directory,

But I believe we are not transferring any window installed ruby files into our docker(getting ruby linux based files from mentioned link under `Dockerfile`), we are only transferring the project files into our docker which is generated via windows installed ruby in this case

PR #1453, opened 2022-11-11
WIP - `fly restart` rework 

davissp14:
**Notable changes:**

*  It's now compatible with Machines
*  Postgres apps are now supported
* `fly pg restart` will be deprecated
* Machine restarts will now require leases

Kept the restart logic within the restart package due to cyclic dependency issues.    

cc:// @jsierles


rugwirobaker (approved):
LGTM!

PR #1457, opened 2022-11-14
Fly restart rework

davissp14:


internal/machine/query.go
```diff
@@ -0,0 +1,27 @@
+package machine
+
+import (
+	"context"
+
+	"github.com/samber/lo"
+	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/flaps"
+)
+
```
davissp14, line 10: The idea would be to keep the flaps package endpoint specific and any modifications or enhancements that outside of the flaps scope could be handled here.

internal/command/apps/restart.go
```diff
@@ -6,41 +6,86 @@ import (
...
 	"github.com/superfly/flyctl/internal/flag"
+	"github.com/superfly/flyctl/internal/machine"
+	"github.com/superfly/flyctl/iostreams"
 )
 
 func newRestart() *cobra.Command {
 	const (
-		long = `The APPS RESTART command will restart all running vms.
-`
+		long  = `The APPS RESTART command will perform a rolling restart against all running VM's`
```
jsierles, line 19: Should just be "VMs" here

internal/command/apps/apps.go
```diff
@@ -26,13 +33,38 @@ The LIST command will list all currently registered applications.
...
-		newRestart(),
 		NewOpen(),
 		NewReleases(),
 	)
 
 	return apps
 }
+
+// BuildContext is a helper that builds out commonly required context requirements
+func BuildContext(ctx context.Context, app *api.AppCompact) (context.Context, error) {
```
jsierles, line 48: Is this mainly to get the flaps client into context?

internal/machine/lease.go
```diff
@@ -0,0 +1,52 @@
+package machine
```
jsierles, line 1: Is it a problem that we have the machines package split across `internal/command/machine` and `internal/machine`?

jsierles (approved):
Nice to see the machines stuff being split out! Just left some comments.

internal/machine/lease.go
davissp14: We can rename the package if it's confusing, but it's not really an issue.  If both packages need to be used at the same time we will just need to define an alias during import.

internal/command/apps/apps.go
davissp14: Both the flaps client and dialer.

shortdiv (approved):
Someday we can get rid of that platform version if conditional

PR #1462, opened 2022-11-16
Update base Rails image to point to Debian 11 (from Debian 10)

bradgessler:
Based on feedback from https://community.fly.io/t/help-migrating-postgres-from-heroku-rails-app/7863/16

### DIFF
--- a/scanner/templates/rails/standard/Dockerfile
+++ b/scanner/templates/rails/standard/Dockerfile
@@ -18,7 +18,7 @@
 # performance.
 
 ARG RUBY_VERSION={{ .rubyVersion }}
-ARG VARIANT=jemalloc-slim
+ARG VARIANT=jemalloc-bullseye-slim
 FROM quay.io/evl.ms/fullstaq-ruby:${RUBY_VERSION}-${VARIANT} as base
 
 LABEL fly_launch_runtime="rails"


rubys:
Interesting.  From https://github.com/evilmartians/fullstaq-ruby-docker/blob/master/README.md :

> Latest patch versions for Ruby 3.1 on Debian 11 (bullseye) are also aliased with shortened tags including major and minor versions only: 3.1.2-jemalloc-bullseye → 3.1-jemalloc
> . . .
> For Ruby 3.0 and older, short aliases for latest patch versions are made against Debian 10 (buster): 3.0.4-jemalloc-buster → 3.0-jemalloc

I definitely think that Debian 11 is a safe bet, but also I understand the logic of Ruby 2.7.6 being originally and forever based on Debian 10.

The only downside I see with explicitly selecting Debian 11 is that this will need to be revisited once Debian 12 comes out.

bradgessler:
Good point. Maybe the right approach is to bug that project for a new tag that points to the most recent version of Debian. I know there's the implicit `:latest` tag on a project, but we'd want something like `:edge`

rubys:
Merging as this is a significant step forward, and can be refined later.  One closing thought: the latest stable Debian version should be picked at launch time and not changed for existing deployments without developer interaction.  That's essentially what this patch does, but requires the community to watch for new Debian releases.

Issue #1459, opened 2022-11-15
`fly launch` on Rails app includes `pg` gem if I select `N` for pg


bradgessler:
When I run `fly launch` in a Rails app and select `N` for install Postgres, the `pg` gem is still added to my project.

**Describe the bug**

```sh
actioncable_tester [main] → fly launch
Update available 0.0.429 -> 0.0.430.
Run "fly version update" to upgrade.
Creating app in /Users/bradgessler/Projects/fly-apps/actioncable_tester
Scanning source code
Detected a Rails app
? Choose an app name (leave blank to generate one): actioncable
? Select Organization: fly-apps (fly-apps)
? Choose a region for deployment: San Jose, California (US) (sjc)
Created app actioncable in organization fly-apps
Set secrets on actioncable: RAILS_MASTER_KEY
Wrote config file fly.toml
? Would you like to set up a Postgresql database now? No
? Would you like to set up an Upstash Redis database now? Yes
? Select an Upstash Redis plan Free: 100 MB Max Data Size
input:3: createAddOn You may only provision 1 Free plan per organization

Adding the 'pg' gem for Postgres database support
```

Then the following appears in my `Gemfile`:

```
gem "pg", "~> 1.4"
```

Your Rails app is prepared for deployment.
```git Dif

* Operating system
* `fly version`

Update available 0.0.429 -> 0.0.431.
Run "fly version update" to upgrade.
fly v0.0.429 darwin/arm64 Commit: 6db701e0 BuildDate: 2022-11-04T13:45:42Z

** Paste your `fly.toml`

```toml
# fly.toml file generated for actioncable on 2022-11-15T09:35:51-08:00

app = "actioncable"
kill_signal = "SIGINT"
kill_timeout = 5
processes = []

[build]
  [build.args]
    BUILD_COMMAND = "bin/rails fly:build"
    SERVER_COMMAND = "bin/rails fly:server"

[deploy]
  release_command = "bin/rails fly:release"

[env]
  PORT = "8080"

[experimental]
  allowed_public_ports = []
  auto_rollback = true

[[services]]
  http_checks = []
  internal_port = 8080
  processes = ["app"]
  protocol = "tcp"
  script_checks = []
  [services.concurrency]
    hard_limit = 25
    soft_limit = 20
    type = "connections"

  [[services.ports]]
    force_https = true
    handlers = ["http"]
    port = 80

  [[services.ports]]
    handlers = ["tls", "http"]
    port = 443

[...]

PR #1463, opened 2022-11-16
Don't run PostgresInitCommands if the user did not install Postgres

bradgessler:
Fixes https://github.com/superfly/flyctl/issues/1459


bradgessler:
I tested this locally by launching a Rails app and verified that it didn't run the pg PostgresInitCommands when I selected `N` to the question of launching a PG database.

jsierles (approved):

PR #1464, opened 2022-11-17
 Fly machine update refactor

davissp14:
**Notable changes:**

* Abstracted out the Machine update process so it can be re-used across commands.
* Abstracted out the Machine config diff prompt, so it can be re-used across all commands.
* Machine update and image update have been re-worked to use the new primitives.
* Machine update will now properly acquire a lease and wait for health checks when present.
* Since we currently do not enforce configuration consistency across Machines, any process that updates a group of non-Postgres Machines will require the end-user to confirm the configuration changes for each Machine. 
* Users can now pass in a custom image to the `fly image update` via the `--image` flag.
* Dedicated VM sizes have been removed, since they are not supported.
* `GetLease` has been renamed to `AcquireLease` since there's now an endpoint for querying lease information. 

Example of an individual Machine update:

Screen Shot 2022-11-18 at 4 22 11 PM


api/machine_types.go
```diff
@@ -143,15 +143,16 @@ const (
 	MEMORY_MB_PER_CPU        = 2048
 )
 
+// TODO - Determine if we want allocate max memory allocation, or minimum per # cpus.
 var MachinePresets map[string]*MachineGuest = map[string]*MachineGuest{
```
davissp14, line 147: Dedicated is not supported for Machines.

internal/command/machine/run.go
```diff
@@ -536,22 +500,26 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
...
+		return machineConf, err
 	}
 
 	if flag.GetString(ctx, "schedule") != "" {
 		machineConf.Schedule = flag.GetString(ctx, "schedule")
 	}
 
-	machineConf.Metadata, err = parseKVFlag(ctx, "metadata", machineConf.Metadata)
-
+	// Metadata
```
davissp14, line 510: This will make metadata append only when issued through `fly machine update`.

internal/machine/lease.go
```diff
@@ -8,7 +8,8 @@ import (
 	"github.com/superfly/flyctl/flaps"
 )
 
-// AcquireLeases works to acquire a leases for each active machine.
+// AcquireLeases works to acquire a lease for each active machine and returns
+// a reference with the lease nonce attached.
 // WARNING: Make sure you defer the lease release process.
```
davissp14, line 13: Not crazy about this function name.

internal/machine/lease.go
```diff
@@ -8,14 +8,21 @@ import (
...
 		return nil, err
 	}
 
+	return AcquireLeases(ctx, machines)
+
+}
+
+// AcquireLeases works to acquire/attach a lease for each machine specified.
+// WARNING: Make sure you defer the lease release process.
+func AcquireLeases(ctx context.Context, machines []*api.Machine) ([]*api.Machine, error) {
```
rugwirobaker, line 25: Isn't there a risk of callers forgetting to call release on these returned machines? I return a release function that callers can defer. It's name would make it had to forget.

Issue #1484, opened 2022-11-23
GetAppLogs panics when using api module


lucagez:
**Describe the bug**
Invoking `GetAppLogs` from `api` module panics.

The module is installed by
```
go get github.com/superfly/flyctl/api
```

This happens as the `GetAppLogs` function [tries to retrieve a logger from context](https://github.com/superfly/flyctl/blob/master/api/resource_logs.go#L24).
It is not possible to create a context that carries a logger from the caller as the `logger` module is internal.

Anyway, the `logger` [appears to be created](https://github.com/superfly/flyctl/blob/master/api/resource_logs.go#L26) just for creating an instance of `httpClient` which is actually already initialized in the client
https://github.com/superfly/flyctl/blob/df8c42a6f3c645374c3b41b4cc274f65ea6ded16/api/client.go#L32

So, unless this is done for other reasons, the solution might be using the http client already initialized.

PR #1485, opened 2022-11-23
Use existing http client in GetAppLogs

lucagez:
Fixes: https://github.com/superfly/flyctl/issues/1484

Building locally and running:
```
flyctl logs --app <app-name>
```
and
```
./bin/flyctl logs --app <app-name>
```
Produce the same output


jsierles:
Good catch!

PR #1487, opened 2022-11-23
Use -jemalloc-slim as ruby stack variant that always point to existen…

dangra:
…t image

As seen in
https://community.fly.io/t/error-jemalloc-bullseye-slim-not-found/8925/8

### DIFF
--- a/scanner/templates/rails/standard/Dockerfile
+++ b/scanner/templates/rails/standard/Dockerfile
@@ -18,7 +18,7 @@
 # performance.
 
 ARG RUBY_VERSION={{ .rubyVersion }}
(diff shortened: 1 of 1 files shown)


rubys:
This essentially reverts https://github.com/superfly/flyctl/pull/1462 , which was originally created to address a problem: https://community.fly.io/t/help-migrating-postgres-from-heroku-rails-app/7863/16

It is the right short term step (getting an older version of Debian is a better problem to have than identifying an image that doesn't exist).  I'll integrate a better solution into the fly.io-ruby gem.

rubys (approved):

PR #1493, opened 2022-11-30
Validate configuration before building.

### LINKED ISSUE #932: flyctl should validate fly.toml before initiating a build
The validation method we are using to catch mistakes in `fly.toml` before initializing a  build on commands like `fly deploy` might be different from what the API is using. We sometimes get a validation error just as a new release is about to be deployed, which can be frustrating after waiting for a very large build.

ndarilek:
Fixes #932

Some of the existing validation checks may now be redundant. I opted to not remove otherwise working code without knowing for sure. If the remaining checks are already covered by the API endpoint, please let me know and I'll remove them.

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/dustin/go-humanize"
+	"github.com/logrusorgru/aurora"
 	"github.com/spf13/cobra"
 
 	"github.com/superfly/flyctl/iostreams"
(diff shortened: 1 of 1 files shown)


tvdfly:
Thanks for implementing this! I made some minor changes and will merge with #1500. Example output when there is a validation error is:

```
% fly deploy
==> Verifying app config


Configuration errors in /home/tvd/src/superfly/flyctl/example/fly.toml:

    ✘ services: [0] Concurrency Type is not included in the list
    ✘ services: [0] Checks [1] HTTP method is not included in the list

Error App configuration is not valid
```

PR #1494, opened 2022-11-30
Rename `redis delete` and require confirmation before destroying instances

ndarilek:
- Rename `redis delete` to `redis destroy` for consistency.
- Prompt for confirmation when destroying Redis instances.


ndarilek:
Ack, meant to finish this one off in the browser.

Anyhow, I wasn't able to test whether destroying a Redis instance did in fact require confirmation because I can't create one. Can't seem to get that working under Windows, since every command I enter seems to be flagged as running in a non-interactive session and something isn't passing through. But the code is a direct copy-paste from the `apps destroy` command, so I'm hopeful. :) Further testing welcome.

ndarilek:
I added a `delete` alias so existing scripts should continue working. `help` correctly lists `destroy` and shows the `delete` alias.

jphenow (approved):

PR #1504, opened 2022-12-05
chore: Add URL to the app's monitoring page during deployment

pavelshtanko:
# Problem
During deployment my routine is quite often like this: 
 - `fly deploy`
 - wait for build and deploy
 - check logs on the website in order to comfortably understand what is going on with the deployment

Therefore, I need to keep the right tab open or open it manually one more time. 

# Solution 
In order to speed up this routine I would love to have the URL to the monitoring tab of the app right in front of my eyes. 

WDYT?


dangra:
thanks!

PR #1505, opened 2022-12-05
remove reference to non-existent connection string

thomas-mcdonald:
I'm not sure this separate block with connection info makes sense since it now only clarifies what can connect to the database, but I have removed the reference to the connection string that is no longer printed below


rugwirobaker (approved):

dangra:
thanks!

Issue #1330, opened 2022-10-03
`fly.toml` is not present in image produced by `flyctl.deploy


vbrandl:
I have an app that relies on the state of the git repository at build time to generate version information in the footer. If there are any uncommitted changes in the repository at build time, a `-dirty` flag is appended to the git tag or commit ref.

When building the docker image directly using `docker build`, everything works as expected, the repository is clean and the proper version info is generated. But when building the image using `flyctl deploy --build-only --local-only`, the `fly.toml` is not present, which causes all deployments to include the `-dirty` flag. I'm not sure if `fly.toml` is actively deleted or if it is ignored because flyctl extends `.dockerignore` or something like that.

Is there a way to include and not delete `fly.toml` when building and deploying an image?

* NixOS 22.05
* `flyctl v0.0.328 linux/amd64 Commit: v0.0.328 BuildDate: 1970-01-01T00:00:00Z`
* The repoistory in question is https://github.com/vbrandl/petnames-generator but I can create a minimal testcase if that helps.

```toml
app = "petnames"

kill_signal = "SIGINT"
kill_timeout = 5

[experimental]
cmd = "./petnames-generator"

[env]
PORT = "8080"

[metrics]
port = 3000
path = "/metrics"

[[services]]
internal_port = 8080
protocol = "tcp"

[services.concurrency]
hard_limit = 25
soft_limit = 20

[[services.ports]]
handlers = ["http"]
port = "80"

[[services.ports]]
handlers = ["tls", "http"]
port = "443"

[[services.tcp_checks]]
grace_period = "1s"
interval = "15s"
port = "8080"
restart_limit = 6
timeout = "2s"
```

bradgessler:
I'm running into the same issue. I'd like to copy the `fly.toml` file so I can pull some config from it at runtime in this project: https://github.com/fly-apps/railsite/blob/main/config/application.rb#L17-L20

Workaround: I can copy the `fly.toml` file to`mcfly.toml` on my host machine, then in my `Dockerfile` add:

```Dockerfile
COPY mcfly.toml ./fly.toml
```

And I get my `fly.toml` file in the image.

redjonzaci:
Hey @vbrandl and @bradgessler!
I see @lillianberryfly has merged a PR related to this issue.
Does that solve your issues or you need more help?

tomr-a5:
[...]

PR #1508, opened 2022-12-07
More sensible behaviour with fly.toml in dockerignore

lillianberryfly:
- add `fly.toml` to dockerignore when creating a new app with `fly launch`
- only add `fly.toml` to dockerignore when building if no dockerignore file was found

The current behaviour (silently add `fly.toml` to dockerignore when building) is confusing -> #1330


lillianberryfly:
not sure why deepsource is failing with "new issues introduced" as I didn't touch the code it mentions

rubys:
These changes look good, but I'm concerned that there is no fly.toml in the following files:

https://github.com/superfly/flyctl/tree/master/scanner/templates

```
% find . -name .dockerignore | xargs grep fly.toml | wc -l
       0
```

jphenow (approved):

PR #1509, opened 2022-12-07
give scanner functions access to Context; Phoenix scanner changes for PR deploy

brainlid:
## Changes

Creates a new ScannerConfig that the `launch` package uses. When the CLI flags `--copy-config` and `--now` and used together, the `ScannerConfig.Mode` becomes `"clone"`. Without that combination the default value is `"launch"`.

The `ScannerConfig` is passed into each `scanner.Scan` function for framework and language detection. 

## Goal

These CLI flags are set when running [`fly-pr-review-apps`](https://github.com/superfly/fly-pr-review-apps).

In this context, the `launch` command is still used, but it is not the first time the app has been configured. This PR use-case applies to when an app has already been setup and deployed. Now, we _only_ want to copy the config and change the name. However, the app's name is also set in ENV values for Phoenix, which is how we tell the server what the host name should be for when URL are generated.

This change limits all `fly launch` changes to only impact the `fly.toml` file. No Dockerfile configuation, no dev tooling `mix` commands get executed, etc.

## Extra Behavior

This takes the extra step of setting an additional ENV:

```toml
[env]
  FLY_LAUNCH_MODE = "clone"
```

When deploying a PR branch, it is important to be able to detect that it is a PR deploy. When that is detected, then the deployment may choose to seed the database with test data for instance.

## Frameworks Impacted

Only the Phoenix framework takes advantage of this new Mode. What is appropriate varies by framework and Phoenix is the first experiment.


scanner/deno.go
```diff
@@ -1,6 +1,8 @@
 package scanner
 
-func configureDeno(sourceDir string) (*SourceInfo, error) {
+import "context"
+
+func configureDeno(sourceDir string, ctx context.Context) (*SourceInfo, error) {
```
rugwirobaker, line 5: The general convention is to pass the context object before everything else.

brainlid:
@rugwirobaker - moved the context to be first arg on the Scan function to better match project conventions.

jsierles (commented):
This looks good for what your goal was, but I have a few suggestions for improvements. All this is optional, but

First, there's now duplicated code in the Phoenix scanner. You could try the common 'guard' pattern here: create the `srcInfo` struct, add the secret key to it, then place your new conditional block. That block would add the extra ENV entry and then return. Otherwise, the function continues adding the rest of the `srcInfo` entries.

Next, while it's useful to pass context everywhere, I would consider not passing flags all the way down and being more explicit about your intent. Flags are specific to the higher-level command, and could change down the road. We had some problems in the past with changes to flags being hard to trace through the code base.

This would also allow other scanners to do something similar without having to duplicate conditionals around specific flags.

So an option could be to create an extendable configuration struct for scanners that gets passed instead of context, like:

```
type LaunchConfig struct {
  Mode string
}
```

Then something like:

```
config := &LaunchConfig{}

if flag.GetBool(ctx, "copy-config") && flag.GetBool(ctx, "now") {
  config.Mode = "clone"
}
scanner.Scan(dir, config)
```

brainlid:
Thanks @jsierles for the feedback. I'll try to apply those changes. I appreciate it!

brainlid:
@jsierles I'm ready for another look! I applied your suggested changes and updated the PR description.

jsierles (approved):

PR #1512, opened 2022-12-12
Clone machine and its volumes from last snapshot

dangra:
Before, there was a special case to clone machine and its volumes only for _postgres_cluster_ role. This PR generalizes it to create volumes for any cloned machine and optionally to restore volume's content to latest snapshot from the original volumes.

The biggest change there is that we are creating new volumes no matter what app role the machine belongs to as long as the volume is defined in the source machine.

Optionally, if `--from-snapshot=last` is passed, it restore volumes from most recent snapshot.

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -146,7 +146,6 @@ func runMachineClone(ctx context.Context) (err error) {
 			SnapshotID:        snapshotID,
 			RequireUniqueZone: false,
 		}
-		fmt.Printf("%#v\n", volInput)
 		vol, err := client.CreateVolume(ctx, volInput)
 		if err != nil {
 			return err
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/samber/lo"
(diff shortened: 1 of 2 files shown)


internal/command/machine/clone.go
```diff
@@ -44,6 +45,10 @@ func newClone() *cobra.Command {
 			Name:        "name",
 			Description: "Optional name for the new machine",
 		},
+		flag.Bool{
```
davissp14, line 48: What should the user do if they want to restore from an older snapshot?

dangra: Good point :)

We can add another flag `--from-snapshot=ID` and even use `last` as a reference for most recent snapshot. wdyt?

davissp14: Wow, sorry I missed this.  That sounds good to me!

dangra: Addressed in [this commit](https://github.com/superfly/flyctl/pull/1512/commits/94b98610c024a6a005d7a03b33a848f1ff2bce30#diff-4a3436ee7e8af3a8b58c6dc54d42485e71f9367bc5706d7a6adaaf143e9bcf93R48-R50)

davissp14 (approved):
Looks good!

PR #1514, opened 2022-12-13
Add fly m leases command

rugwirobaker:


internal/command/machine/leases.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func newLeaseView() *cobra.Command {
+	const (
+		short = "View machine leases"
+		long  = short + "\n"
+		usage = "view"
+	)
+
+	cmd := command.New(usage, short, long, runLeaseView,
```
davissp14, line 50: How can we make this a better experience for users who want to view/clear all leases at the same time?

internal/command/machine/leases.go
```diff
@@ -0,0 +1,201 @@
...
+	ctx, err = apps.BuildContext(ctx, app)
+	if err != nil {
+		return
+	}
+
+	flapsClient := flaps.FromContext(ctx)
+
+	for _, machineID := range args {
+		lease, err := flapsClient.FindLease(ctx, machineID)
+		if err != nil {
```
davissp14, line 189: We shouldn't fail in the event err == "lease not found".

davissp14 (changes requested):

internal/command/machine/leases.go
davissp14: I'm honestly not sure what the best UX is for this.  I think this is a good start though, we can swing back to this later if we need to.

internal/command/machine/leases.go
rugwirobaker: I thought it should have the same UX as `fly machine list`.

davissp14: Seems like we'd want the same logic you have above:

```
	if strings.Contains(err.Error(), " lease not found") {
		continue
	}
	return err
```

Since there may be cases where some Machines have leases and some don't.

rugwirobaker: My bad, I misread this.

PR #1520, opened 2022-12-15
App move fix

davissp14:
Adds support for Machines.

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -204,7 +204,12 @@ type MachineConfig struct {
 	Guest     *MachineGuest           `json:"guest,omitempty"`
 	Metrics   *MachineMetrics         `json:"metrics"`
 	Schedule  string                  `json:"schedule,omitempty"`
+	Network   MachineNetwork          `json:"network,omitempty"`
 	Checks    map[string]MachineCheck `json:"checks,omitempty"`
+}
+
+type MachineNetwork struct {
+	ID int `json:"id"`
 }
 
 type MachineLease struct {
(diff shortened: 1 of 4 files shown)


internal/machine/lease.go
```diff
@@ -34,7 +35,9 @@ func AcquireLeases(ctx context.Context, machines []*api.Machine) ([]*api.Machine
 	releaseFunc := func(ctx context.Context, machines []*api.Machine) {
 		for _, m := range machines {
 			if err := flapsClient.ReleaseLease(ctx, m.ID, m.LeaseNonce); err != nil {
-				fmt.Fprintf(io.Out, "failed to release lease for machine %s: %s", m.ID, err.Error())
+				if strings.Contains(err.Error(), "lease not found") {
+					fmt.Fprintf(io.Out, "failed to release lease for machine %s: %s", m.ID, err.Error())
```
jipperinbham, line 39: Somewhat tangential to his overall change but I think this confuses customers because we've not really educated them on what a lease is. Do we actually create if we failed to release the lease? Does it give the customer a signal that something is wrong or they need to do take action?

davissp14: Let me remove this and push it into a separate PR. 

This will just hide the lease error in the event of an "lease is not found" error, this is because the TTL expires before the operation finishes.  This is a separate issue that we are working through.  

> Do we actually create if we failed to release the lease? Does it give the customer a signal that something is wrong or they need to do take action?

The release happens at the very end, but eventually the action we will prompt the user with is to manually release the leases using the new cli command @rugwirobaker  introduced.

jipperinbham: 👍

PR #1525, opened 2022-12-17
Improve scan for django projects

walison17:
Currently projects that use tools like `pip-tools` that generate a `requirements.txt` with all declared packages in lower case are not reconized as a Django project.


mrkurt:
Nice.

PR #1526, opened 2022-12-19
Fix `fly volume create` docstring to mention actual default size of 3.

Jesse-Bakker:
The docstring previously mentioned the default being 10 instead of 3,
which is in conflict with the flag's doc and the actual default.


dangra:
thanks!

PR #1533, opened 2022-12-22
Add:fly pg attachments command

rugwirobaker:
* Consolidates the likes of attach and detach.
* Adds a new list attachments command.
* There old commands `fly pg attach` and `fly pg detach` are still there for now.

### DIFF
--- a/internal/command/postgres/attachments.go
+++ b/internal/command/postgres/attachments.go
@@ -1,0 +1,102 @@
+package postgres
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/internal/app"
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/config"
+	"github.com/superfly/flyctl/internal/flag"
+	"github.com/superfly/flyctl/internal/render"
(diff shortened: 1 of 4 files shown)


internal/command/postgres/attachments.go
```diff
@@ -0,0 +1,102 @@
...
+		short = "manage database attachments to apps"
+		long  = short + "\n"
+		usage = "attachments [command]"
+	)
+
+	cmd := command.New(usage, short, long, nil)
+
+	cmd.AddCommand(
+		newAttachmentsList(),
+		newAttach(),
```
davissp14, line 28: Do you think we should rename the attach/detach sub commands with this? 

```
fly pg attachments attach` doesn't feel quite right.

davissp14 (changes requested):
I added a quick note, but we may want to consider renaming the subcommands. 

This change will also require updates to our documentation.

internal/command/postgres/attachments.go
rugwirobaker: @davissp14 I think people are too used to it and it makes sense. I'd rather find a new name for the new attachments command.

guillaumervls:
@rugwirobaker what's in the way of us merging / shipping / closing this? How can I help?

dangra:
Closing, we can reopen if we ever revisit the idea.

PR #1537, opened 2022-12-26
Detect process name from app config when scaling

sosedoff:
When running scaling command (`flyctl scale count 5`) i would get a cryptic error message like `Error Invalid VM group:`. While `flyctl scale web=5` works, it would be nice to automatically detect then process name from the config itself (if available), we would still default to `app`.

My config:

```
app = "..."
kill_signal = "SIGINT"
kill_timeout = 5

[processes]
web = "bundle exec ruby app.rb"

[[services]]
  http_checks = []
  internal_port = 8080
  processes = ["web"]
  protocol = "tcp"
  script_checks = []

# ...
```


dangra:
thanks!

PR #1551, opened 2023-01-06
WIP: fly launch and fly deploy for machines for apps v2

tvdfly:
We're pushing to migrate the Fly Apps platform to machines instead of nomad to improve developer experience. We're calling this Apps v2. The first big step will be making new apps use the machines platform instead of nomad. There's a bunch of small steps to get there, and this is one of them...

Update `fly deploy` when we're working on an app with machines to:

* Prompt developer to migrate the machines into the "Fly Apps platform". The machines being deployed will become the instances of the apps. Specifically, we put some metadata on those machines indicating they are part of the Fly Apps platform. Subsequent deploys will only operate on machines with that metadata.

* Support rolling deployments, release_command, and many of the other standard fly.toml configuration.

Update `fly launch` to use machines instead of nomad.

Things that are known not to work, at least for now, with Apps v2 `fly deploy` are:
* [x] ~~`[processes]` are not supported and will be ignored~~
* [x] ~~images get built twice~~
* [ ] `[statics]` are not supported with machines


internal/command/deploy/machines.go
```diff
@@ -312,3 +341,117 @@ func releaseLease(ctx context.Context, machine *api.Machine) error {
...
+
+func (md *machineDeployment) flyRelease() string {
+	return fmt.Sprintf("%s_%d", md.releaseId, md.releaseVersion)
+}
+
+func (md *machineDeployment) baseMachineConfig() *api.MachineConfig {
+	machineConfig := &api.MachineConfig{}
+	machineConfig.Metadata = map[string]string{
+		api.MachineConfigMetadataKeyFlyPlatformVersion: api.MachineFlyPlatformVersion2,
+		api.MachineConfigMetadataKeyFlyRelease:         md.flyRelease(),
```
tvdfly, line 394: consider splitting this out into two keys:

fly_release_id = md.releaseId
fly_release_version = md.releaseVersion

internal/command/deploy/machines.go
```diff
@@ -107,6 +116,62 @@ func (lm *leasableMachine) Update(ctx context.Context, input api.LaunchMachineIn
...
+func (lm *leasableMachine) logClearLinesAbove(count int) {
+	if lm.io.IsInteractive() {
+		builder := aec.EmptyBuilder
+		str := builder.Up(uint(count)).EraseLine(aec.EraseModes.All).ANSI
+		fmt.Fprint(lm.io.ErrOut, str.String())
+	}
+}
+
+func (lm *leasableMachine) logStatus(desired, current string) {
+	cur := aurora.Green(current)
```
jsierles, line 136: Just a note here that people sometimes turn colors off, so we're using `colorize` from the `iostreams` package, which is available on the global context. See [[So about storj, ‘amazingly well’ only for registry, or for storage in general? ](https://github.com/superfly/flyctl/blob/8a55d4067be1d5c7c78890ea7a28ff3016d8a848/internal/machine/restart.go#L32-L35)]

jsierles (approved):

PR #1554, opened 2023-01-06
Reattach existing volume to a new machine

rugwirobaker:
`fly clone --volume-id`

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -49,6 +49,10 @@ func newClone() *cobra.Command {
 			Name:        "from-snapshot",
 			Description: "Clone attached volumes and restore from snapshot, use 'last' for most recent snapshot. The default is an empty volume",
 		},
+		flag.String{
+			Name:        "volume-id",
+			Description: "Existing volume to attach to the new machine",
+		},
 	)
 
 	return cmd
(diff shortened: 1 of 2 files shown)


internal/command/machine/clone.go
```diff
@@ -106,39 +110,68 @@ func runMachineClone(ctx context.Context) (err error) {
...
+		} else {
+			volName := mnt.Name
+			if mnt.Name == "" {
+				// Fallback for mounts that doesn't have pool name set
+				vol, err := client.GetVolume(ctx, mnt.Volume)
+				if err != nil {
+					return err
+				}
+				volName = vol.Name
+			}
```
dangra, line 137: This section was removed by https://github.com/superfly/flyctl/pull/1556/files

internal/command/machine/clone.go
```diff
@@ -49,6 +49,10 @@ func newClone() *cobra.Command {
 			Name:        "from-snapshot",
 			Description: "Clone attached volumes and restore from snapshot, use 'last' for most recent snapshot. The default is an empty volume",
 		},
+		flag.String{
+			Name:        "volume-id",
```
dangra, line 53: I would call it `--attach-volume`

dangra (approved):

PR #1560, opened 2023-01-10
parse port range in machine run command

lillianberryfly:


internal/command/machine/run.go
```diff
@@ -426,35 +426,83 @@ func determineServices(ctx context.Context) ([]api.MachineService, error) {
...
+		var external_port int
+		external_port, err = strconv.Atoi(split[0])
+		if err != nil {
+			err = errors.Wrap(err, "invalid port")
+			return
+		}
+
+		p := int32(external_port)
+		port = &p
+	} else {
```
DAlperin, line 462: It looks like we might want to sanity check that `split` doesn't have more than two items? For instance `80:8080:3000` is ambiguous and doesn't mean anything.

internal/command/machine/run.go
```diff
@@ -426,35 +426,83 @@ func determineServices(ctx context.Context) ([]api.MachineService, error) {
...
+			var external_port int
+			external_port, err = strconv.Atoi(external_split[0])
+			if err != nil {
+				err = errors.Wrap(err, "invalid external port")
+				return
+			}
+
+			p := int32(external_port)
+			port = &p
+		} else {
```
DAlperin, line 480: same thing here

DAlperin (approved):
lgtm!

PR #1569, opened 2023-01-11
Postgres flex config

DAlperin:


internal/command/postgres/config_update.go
```diff
@@ -251,7 +251,13 @@ func updateStolonConfig(ctx context.Context, app *api.AppCompact, leaderIP strin
 	if !force {
 		// Query PG settings
 		pgclient := flypg.NewFromInstance(leaderIP, dialer)
-		settings, err := pgclient.ViewSettings(ctx, keys)
+
+		_, flex := os.LookupEnv("FORCE_FLEX")
```
davissp14, line 255: What does this do?

DAlperin: When running in development `app.ImageDetails.Repository == "flyio/postgres-flex"` will not be truthy, so you can bypass that check with the `FORCE_FLEX` env variable.

tvdfly (approved):
approving for prerelease

PR #1583, opened 2023-01-18
Pg config fixes

davissp14:
Notable changes:

* FORCE_FLEX and associated env vars are no longer required.
* Disabled the pg failover process for PG's running the postgres-flex image type.
* Configuration changes will now sync with each node in the cluster. 
* Refactored the pg config logic a bit.

### DIFF
--- a/internal/command/postgres/config_show.go
+++ b/internal/command/postgres/config_show.go
@@ -198,7 +198,7 @@ func showSettings(ctx context.Context, app *api.AppCompact, manager string, lead
 			value = fmt.Sprintf("%s -> %s", value, p)
 		}
 		rows = append(rows, []string{
-			strings.Replace(setting.Name, "_", "-", -1),
+			strings.ReplaceAll(setting.Name, "_", "-"),
 			value,
 			setting.Unit,
 			desc,
--- a/internal/command/postgres/config_show.go
+++ b/internal/command/postgres/config_show.go
@@ -103,7 +103,12 @@ func runMachineConfigShow(ctx context.Context, app *api.AppCompact) (err error) 
 		return err
 	}
 
-	return showSettings(ctx, app, leader.PrivateIP)
+	manager := flypg.StolonManager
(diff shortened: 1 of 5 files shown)


internal/command/postgres/config_show.go
```diff
@@ -103,7 +103,12 @@ func runMachineConfigShow(ctx context.Context, app *api.AppCompact) (err error)
 		return err
 	}
 
-	return showSettings(ctx, app, leader.PrivateIP)
+	manager := flypg.StolonManager
+	if leader.ImageRef.Repository == "flyio/postgres-flex" {
```
DAlperin, line 107: still wish there was a flag or something so when I am developing and my image comes from my personal repo it could still work. But also I'm fine just commenting stuff out for now

davissp14: Yeah, we can do that.  What about something like `PG_MANAGER`?

DAlperin: Yeah I can open a different PR for it if you want

internal/command/postgres/config_update.go
```diff
@@ -164,70 +179,69 @@ func runMachineConfigUpdate(ctx context.Context, app *api.AppCompact) error {
...
-				return err
-			}
-		}
-
-		if err := nomadRestart(ctx, app); err != nil {
-			return err
+		// Pull configuration settings down from Consul for each node and reload the config.
+		err := client.SyncSettings(ctx)
+		if err != nil {
+			return false, fmt.Errorf("failed to sync configuration on %s: %s", machine.ID, err)
```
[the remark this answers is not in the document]
davissp14: We may want to introduce a `fly pg config reload/sync` command so this can be re-run.  It would be a bummer to have the user do a full restart to apply run-time settings.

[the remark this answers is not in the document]
davissp14: I created a ticket for this and will swing back to it.

PR #1585, opened 2023-01-18
Use SECRET_KEY_BASE_DUMMY for asset precompilation

ghiculescu:
ref: https://github.com/rails/rails/pull/46760


ghiculescu:
Cool. Thanks!

PR #1588, opened 2023-01-18
Rails Dockerfile template: include `libvips` for Active Storage

ghiculescu:
Rails [includes this](https://github.com/rails/rails/pull/46991/files#diff-a83245368df35c610a37260d0a51041de89211857d47cfd0b2d1301a11709164R527) if Active Storage is enabled. I don't know if there's a way fly can know that, but it seems better to assume it is and install the package since in most cases it will be (and advanced users can remove it from the Dockerfile).

### DIFF
--- a/scanner/templates/rails/standard/Dockerfile
+++ b/scanner/templates/rails/standard/Dockerfile
@@ -102,7 +102,7 @@ RUN npm install
 
 FROM base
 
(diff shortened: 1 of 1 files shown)


rubys:
The plan is to convert over to https://github.com/rubys/dockerfile-rails which will make the determination based on what gems you are actually using.  But meanwhile, this change is worthwhile - it adds a small library which is often needed.

PR #1590, opened 2023-01-18
Request app name for `machine status`.

### LINKED ISSUE #1264: Prompt the user to input flag for machine status
fly machine status without an app name should prompt a user to input an app name instead of erroring out completely

Screen Shot 2022-09-09 at 5 00 18 PM

ndarilek:
Note that this technically fixes the issue, but if no app name is entered then it works anyway. I'm leaving this flag in as requested.

Fixes #1264


ndarilek:
Not entirely convinced this is the correct path forward, so closing for now.

PR #1597, opened 2023-01-19
Add a `--workdir`/`-w` option to set the working directory of `fly ssh console`.

### LINKED ISSUE #1220: Allow fly ssh console to have multiple or chained commands.
Every time I enter the ssh console, I have to cd into a directly and then open a rails console. It would be amazing if this was one line.

Works fine with one command
```
fly ssh console -C "ls"
Connecting to top1.nearest.of.app-name.internal... complete
app  boot  etc   lib    media  opt   root  sbin  sys  usr
bin  dev   home  lib64  mnt    proc  run   srv   tmp  var
```

But with two:

```
fly ssh console -C "ls && cd app"
Connecting to top1.nearest.of.app-name.internal... complete
ls: cannot access '&&': No such file or directory
ls: cannot access 'cd': No such file or directory
app:
Dockerfile    other_files
```

ndarilek:
Unfortunately this prints the initial `cd` command and output to stdout/stderr when launching the session. Go's SSH client (or more probably SSH itself) has no way to start a command in a given working directory. Better solutions welcome.

Partially fixes #1220

### DIFF
--- a/internal/command/ssh/console.go
+++ b/internal/command/ssh/console.go
@@ -58,6 +58,12 @@ func stdArgsSSH(cmd *cobra.Command) {
 			Shorthand:   "A",
 			Description: "Address of VM to connect to",
 		},
+		flag.String{
+			Name:        "workdir",
+			Shorthand:   "w",
+			Default:     "",
+			Description: "working directory inside the SSH session",
+		},
 	)
 }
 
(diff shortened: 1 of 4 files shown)


ndarilek:
Doesn't work for non-shell command invocations, and because we can't make many assumptions about the environment, automatically invoking `sh -c` would probably introduce all sorts of complexity. I also didn't anticipate limitations in SSH that seem to make this more difficult.

Going to have to think about this one. It might be a dead end, and the recommendation to users might just be packaging utility scripts for more complex tasks.

ndarilek:
I've unilaterally decided to close this one by virtue of it being too hacky. Didn't realize before I started on it that the SSH protocol didn't expose this functionality for a good reason, so neither should a thing calling itself SSH.

PR #1602, opened 2023-01-20
Apps v2 service checks

dangra:
Respect checks defined for services so fly-proxy can relate them and use their status to decide when to route traffic to instances serving that service or not.


internal/app/app.go
```diff
@@ -637,45 +655,31 @@ func (c *Config) SetVolumes(volumes []scanner.Volume) {
...
-	defaultProcessName := ""
-	firstProcessNameOrDefault := ""
-	if processCount == 1 {
-		for procName := range c.Processes {
-			firstProcessNameOrDefault = procName
-			break
-		}
+	processCount := len(c.Processes)
+	if processCount == 0 {
+		c.Processes = map[string]string{"": ""}
```
tvdfly, line 667: the original was causing a panic when c.Processes was nil, and this version fixes that. Setting c.Processes here does break assumptions elsewhere that empty processes means no processes were set.  i'll try not overwriting c.Processes and see if that fixes stuff...

Might be good to do some more testing before merging this.

internal/app/app.go
```diff
@@ -133,26 +134,33 @@ func (hs *HttpService) ToMachineService() *api.MachineService {
 		}},
 		Concurrency: concurrency,
 	}
-
 }
 
-func (s *Service) ToMachineService() *api.MachineService {
+func (s *Service) toMachineService() *api.MachineService {
+	checks := make([]api.Check, len(s.TCPChecks)+len(s.HTTPChecks))
```
tvdfly, line 140: I'm pretty sure this makes a slice with `len(s.TCPChecks)+len(s.HTTPChecks)` items all set to nil. the appends add to that list without overwriting any of the nils. 

```suggestion
	checks := make([]api.Check, 0)
```

tvdfly: I'm playing around with this, and will push an update in a sec.

tvdfly (commented):
it looks like the per service check is getting passed along to the machines, and flaps doesn't yet support those checks (which I'm thinking is expected and the flaps change will come next).

setting the top level checks in fly.toml still works. it looks like we're not letting it use the newly supported attributes (like grace period). Is it safe to add those?

tvdfly:
btw, thank you for cleaning up this code and making it simpler! It's much easier to go through now.

internal/app/app.go
```diff
@@ -178,7 +186,7 @@ type Check struct {
 	Headers       map[string]string `json:"headers,omitempty" toml:"headers,omitempty"`
 }
 
-func (c *Check) toMachineCheck(launching bool) (*api.MachineCheck, error) {
+func (c *ToplevelCheck) toMachineCheck(launching bool) (*api.MachineCheck, error) {
```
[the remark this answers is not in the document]
dangra: Yup, I was waiting on the server side changes to be ready before pushing it. Will do soon.

internal/app/app.go
dangra: thanks for this!

internal/app/app.go
```diff
@@ -137,7 +137,7 @@ func (hs *HTTPService) toMachineService() *api.MachineService {
 }
 
 func (s *Service) toMachineService() *api.MachineService {
-	checks := make([]api.Check, len(s.TCPChecks)+len(s.HTTPChecks))
+	checks := make([]api.Check, 0)
```
dangra, line 140: Oh, my mistake, the intention was to create an empty slice with the expected capacity

```
checks := make([]api.Check, 0, len(s.TCPChecks)+len(s.HTTPChecks))
```

tvdfly: Oh that makes sense. Thanks!

internal/app/app.go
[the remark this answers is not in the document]
tvdfly: Awesome, thanks!

PR #1609, opened 2023-01-22
avoids infinite loop when using directory as rootname in ssh issue

drio:
Hi team. 

I run into an issue when running: `flyctl ssh issue`. If you provide a directory as a path you enter in an infinite loop. By setting `rootname` back to the empty string we break the infinite loop and let the user enter another file. 

Probably you have a better approach, let me know and I will update the PR.


drio:
Can someone rerun the workflow? I believe there was an issue with github see below:

screen-capture-drio 2023-01-25 at 07 58 51

dangra:
this makes sense. thanks for fixing it.

PR #1618, opened 2023-01-24
If a machine won't start, don't hang forever when attempting an update.

ndarilek:
To reproduce:

1. Check out https://github.com/fly-apps/hello-rust.
2. Launch with `fly m run . -p 443:8080/tcp:tls:http -p 80:8080/tcp:http -a mytestapp`.
3. Note that, I suspect since we haven't specified a command, the machine fails to start.
4. Run something like `fly m update ... -e FOO=BAR` to force an update that doesn't actually fix the problem.

For me, this command hangs indefinitely.

New to Go and I don't know if there's a better way of matching against bespoke errors like this. Maybe matching against HTTP 409 is better, but I don't know enough about the backend API to know if this is returned anywhere else. It also appears that flaps\flaps.go returns a string here anyway. Please let me know if there should be a better/different fix for this.

Thanks.

### DIFF
--- a/internal/machine/wait.go
+++ b/internal/machine/wait.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/jpillora/backoff"
@@ -40,6 +41,8 @@ func WaitForStartOrStop(ctx context.Context, machine *api.Machine, action string
 			return err
 		case errors.Is(err, context.DeadlineExceeded):
 			return fmt.Errorf("timeout reached waiting for machine to %s %w", waitOnAction, err)
+		case strings.Contains(err.Error(), "currently stopped"):
+			return err
 		case err != nil:
 			time.Sleep(b.Duration())
 			continue


catflydotio:
I reproduced this. I wonder if the solution lies not in erroring out when flyd sends a 409 (which you could imagine wanting to treat as informational, not fatal), but in fixing the reason we're neither cancelling after some timeout, nor getting a DeadlineExceeded error.

update.go calls this with a 5-minute timeout. I tried your reproduction instructions and set a 5-minute timer on my computer. Just when my timer went off, the api stopped responding. flyctl kept sending GET requests every two seconds forever (40 minutes) though.

I don't read Go very well. I'd bet that tomorrow @jsierles or @davissp14 could tell us pretty quick what's going on.

tvdfly:
The flaps client should probably be looking at the status code of the response in addition to the response body, and communicating timeouts separate from other errors (since timeouts in this case probably mean try again). I think right now it's doing the same general error checking for the wait endpoint as all the other endpoints.

Note that there are at least two proxies in the mix here that can timeout the request: fly-proxy and envoy (if using api.machines.dev). Each proxy will give a different timeout response: I think 504 from envoy, and 408 from fly-proxy.

the [code in here](https://github.com/superfly/test-registry/blob/main/scripts/runtest/main.go#L271-L319) does some of the parsing of status codes and bodies. I wouldn't recommend copying it, but it may provide some inspiration for updating the flaps client.

```go
func waitForMachine(ctx context.Context, desiredState, reqUrl, machId, machInstanceId, machName, region, authToken string, httpClient *http.Client) bool {
	startTime := time.Now()
	defer func() {
		elapsed := time.Since(startTime)
		fmt.Printf("[%s] machine %s (%s) one wait request for %s state took %.2f seconds\n", region, machName, machId, desiredState, elapsed.Seconds())
	}()
	waitUrl := fmt.Sprintf("%s/%s/wait?instance_id=%s&state=%s&timeout=%d", reqUrl, machId, machInstanceId, desiredState, int(waitTimeout.Seconds()))
	waitReq, err := http.NewRequestWithContext(ctx, http.MethodGet, waitUrl, nil)
	if err != nil {
		fmt.Printf("[%s] error creating wait request: %v\n", region, err)
		return false
	}
	waitReq.Header.Set("Authorization", fmt.Sprintf("Bearer %s", authToken))
	waitResp, err := httpClient.Do(waitReq)
	if err != nil {
		fmt.Printf("[%s] wait response error: %v\n", region, err)
		return false
	}
	if waitResp.StatusCode == http.StatusRequestTimeout {
		return false
	} else if waitResp.StatusCode != http.StatusOK {
		errBody, _ := ioutil.ReadAll(waitResp.Body)
		fmt.Printf("[%s] non-200 wait response code: %d body: %s\n", region, waitResp.StatusCode, string(errBody))
		return false
	}
	waitRespBody, err := ioutil.ReadAll(waitResp.Body)
	if err != nil {
		fmt.Printf("[%s] error reading wait response body: %v\n", region, err)
	}
	waitJson := map[string]interface{}{}
	err = json.Unmarshal(waitRespBody, &waitJson)
	if err != nil {
		fmt.Printf("[%s] error parsing wait json: %v\n", region, err)
		return false
	}
	waitRespStr := strings.TrimSpace(string(waitRespBody))
	okVal, present := waitJson["ok"]
	if okVal, castOk := okVal.(bool); present && castOk && okVal {
		fmt.Printf("[%s] machine %s (%s) has enter %s state\n", region, machName, machId, desiredState)
		return true
	}
	errorVal, present := waitJson["error"]
	if present {
		fmt.Printf("[%s] error in wait response, will retry: %s\n", region, errorVal)
	} else {
		fmt.Printf("[%s] machine %s (%s) unknown wait response: %s\n", region, machName, machId, waitRespStr)
	}
	return false
}
```

ndarilek:
Revisiting this code after a few weeks away and I'm a bit confused. How do timeouts enter into the picture here?

My assumption on what was happening is that I configured the machine incorrectly so it didn't actually start. Then the second update failed because the machine I was updating wasn't started (or maybe couldn't start), and maybe some API call expected it to be started before going through? It seemed like I was getting responses back--409, conflict, which to me doesn't suggest a timeout. I also get them back pretty regularly...

...or at least, I *remember* getting them back regularly. Trying to run the steps above prompts if I want to create mytestapp then dumps me back to the shell. No build step, nothing. Did something change and invalidate these reproduction steps in the last few weeks?

ndarilek:
Ah, here we go:

```
DEBUG --> POST https://api.fly.io/graphql                                                                               
                                                                                                                        
{                                                                                                                       
  "query": "mutation($input: CreateAppInput!) { createApp(input: $input) { app { id name organization { slug } config { definition } regions { name code } } } }",                                                                              
  "variables": {                                                                                                        
    "input": {                                                                                                          
      "organizationId": "m8Y4w6lakLYJ3IxBxqpZLXjpZYhxOy",                                                               
      "name": "mytestapp"                                                                                               
    }                                                                                                                   
  }                                                                                                                     
}                                                                                                                       
                                                                                                                        
DEBUG {}                                                                                                                
DEBUG <-- 200 https://api.fly.io/graphql (226.25ms)                                                                     
                                                                                                                        
{                                                                                                                       
  "data": {                                                                                                             
    "createApp": null                                                                                                   
  },                                                                                                                    
  "errors": [                                                                                                           
    {                                                                                                                   
      "message": "Validation failed: Name has already been taken",                                                      
      "locations": [                                                                                                    
        {                                                                                                               
          "line": 1,                                                                                                    
          "column": 37                                                                                                  
        }                                                                                                               
      ],                                                                                                                
      "path": [                                                                                                         
        "createApp"                                                                                                     
      ],                                                                                                                
      "extensions": {                                                                                                   
        "code": "UNPROCESSABLE"                                                                                         
      }                                                                                                                 
    }                                                                                                                   
  ]                                                                                                                     
}                                                                                                                       
                                                                                                                        
Nolan@Desktop MSYS ~/src/flyctl/hello-rust (main)                                                                       
$
```

Looks like that error is never printed to the console.

ndarilek:
Seems to have been resolved in some of the recent fixes for machines-related hangs. Closing.

PR #1620, opened 2023-01-25
Switch to using dockerfile-rails generator

rubys:


rubys:
See https://community.fly.io/t/cut-over-to-rails-dockerfile-generator-on-sunday-29-jan-2023/10350 for announcement and plans.

internal/command/launch/launch.go
```diff
@@ -418,6 +368,44 @@ func run(ctx context.Context) (err error) {
 		}
 	}
 
+	options := make(map[string]bool)
+
+	if !flag.GetBool(ctx, "no-deploy") && !flag.GetBool(ctx, "now") && !srcInfo.SkipDatabase {
```
tom93, line 373: Here `srcInfo` may be nil, leading to a nil pointer dereference.
(Before this commit the code was protected by an `if srcInfo == nil { return nil }`.)

Steps to reproduce:

```
$ fly launch
Creating app in /home/user/app2
Scanning source code
? Choose an app name (leave blank to generate one): 
automatically selected personal organization: <redacted>
? Choose a region for deployment: Chicago, Illinois (US) (ord)
Created app fragrant-dust-4170 in organization personal
Admin URL: https://fly.io/apps/fragrant-dust-4170
Hostname: fragrant-dust-4170.fly.dev
```

Incidentally, it would be very helpful if [sentry.printError](https://github.com/superfly/flyctl/blob/5a5b5b1fe5793907442b7c57d2f51cdd22baca51/internal/sentry/sentry.go#L106) included the stack trace when LOG_LEVEL was set to an appropriate value (currently it only includes the stack trace in development).

PR #1628, opened 2023-01-26
`fly machine update` accepts a `--command`/`-C` flag to update `CMD`.

ndarilek:


internal/command/machine/run.go
```diff
@@ -629,6 +629,11 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 		machineConf.Schedule = flag.GetString(ctx, "schedule")
 	}
 
+	if command := flag.GetString(ctx, "command"); command != "" {
+		command := []string{command}
```
dangra, line 633: I think in this case we want to use a shell parser like `shlex.Parse()`, wdyt?

ndarilek: Right, good catch. I'll circle back to this on Monday and will look into this--new to Go and not familiar with that package but should be simple enough. Thanks!

ndarilek:
Should be fixed. Thanks for the feedback!

internal/command/machine/run.go
```diff
@@ -659,8 +667,13 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 		machineConf.Init.Entrypoint = splitted
 	}
 
-	if cmd := flag.Args(ctx)[1:]; len(cmd) > 0 {
-		machineConf.Init.Cmd = cmd
+	if cmd := flag.Args(ctx)[1:]; len(cmd) > 0 && len(machineConf.Init.Cmd) == 0 {
+		cmd := strings.Join(cmd, " ")
+		split, err := shlex.Split(cmd)
```
dangra, line 672: no need to join/split the args here, they were already splitted correctly by the shell when call flyctl.

dangra (changes requested):
looking good except for the join/split of already correctly parsed args.

internal/command/machine/run.go
```diff
@@ -659,7 +667,7 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 		machineConf.Init.Entrypoint = splitted
 	}
 
-	if cmd := flag.Args(ctx)[1:]; len(cmd) > 0 {
+	if cmd := flag.Args(ctx)[1:]; len(cmd) > 0 && len(machineConf.Init.Cmd) == 0 {
```
dangra, line 670: hmm, the check on `Init.Cmd` length is preventing the update if already set, is that the expected behavior?

ndarilek: Yeah. If I'm reading the code correctly, `determineMachineConfig` is used from both `fly m run` and `fly m update`. `run` accepts its command as a positional argument while `update` accepts it via flag. I was trying to ensure that the command wouldn't be trashed, and since these methods of injecting a command are mutually exclusive, the cleanest way to not trash the configured command seemed to be checking for an existing configuration. Otherwise `fly m update /bin/ls` might work when it shouldn't.

Unless Cobra is smart enough to not pass the positional arg through based on the docstring? Haven't figured out if that is a thing yet.

Thanks again.

ndarilek: I added a comment clarifying that check. Arguably it might need to be rearchitected to accept the command as a separate argument to `determineMachineConfig`--I'm perpetually faced with choosing between refactoring or preserving, so please do let me know if I land on the wrong side of that choice. :)

dangra: Considering all the other flags in use in this function I think we "preserve" and let your comment live like a _HERE BE DRAGONS_ warning.

dangra (approved):

PR #1660, opened 2023-02-03
Remove deprecated list command

jsierles:
The filtering behavior from this command should be reimplemented in `fly apps list`.


tvdfly:
Will this command still work and just not show up in the help output? Or will this break for anyone using `fly list` now?

jsierles:
This removes the command completely. We had it slated for deprecation since its behavior is duplicated and we're aiming for removing or porting commands in the old flyctl format.

PR #1664, opened 2023-02-03
Use flycast for pg attach if it exists

DAlperin:


internal/command/postgres/attach.go
```diff
@@ -101,9 +101,23 @@ func runAttach(ctx context.Context) error {
...
 
+	pgAppFull, err := client.GetApp(ctx, pgAppName)
+	if err != nil {
+		return fmt.Errorf("failed retrieving postgres app %s: %w", pgAppName, err)
+	}
+
+	var flycast *string
+
+	for _, ip := range pgAppFull.IPAddresses.Nodes {
+		fmt.Println(ip)
```
davissp14, line 112: We can ditch the println here.

internal/command/postgres/attach.go
```diff
@@ -101,9 +101,23 @@ func runAttach(ctx context.Context) error {
...
+		if ip.Type == "private_v6" {
+			flycast = &ip.Address
```
davissp14, line 114: Can there only be one?

davissp14 (approved):
Few questions, but if nothing critical. LGMT

internal/command/postgres/attach.go
DAlperin: oh oops. from debugging the web issue

PR #1675, opened 2023-02-06
Adds command to install flycast on existing machine pgs

DAlperin:


internal/command/postgres/add_flycast.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	switch app.PlatformVersion {
+	case "machines":
+		err := doAddFlycast(ctx)
+		if err != nil {
+			return err
+		}
+
+		fmt.Println("Flycast added!")
```
davissp14, line 68: We should be using Fprintln here:
Reference: https://github.com/superfly/flyctl/blob/master/internal/command/postgres/attach.go#L281

internal/command/postgres/add_flycast.go
```diff
@@ -0,0 +1,148 @@
...
+			return nil
+		}
+
+		conf := machine.Config
+		conf.Services =
+			[]api.MachineService{
+				{
+					Protocol:     "tcp",
+					InternalPort: 5432,
+					Ports: []api.MachinePort{
```
davissp14, line 112: Shouldn't this just reference the port variable defined above?

internal/command/postgres/add_flycast.go
```diff
@@ -0,0 +1,148 @@
...
+								"pg_tls",
+							},
+							ForceHttps: false,
+						},
+					},
+					Concurrency: nil,
+				},
+				{
+					Protocol:     "tcp",
+					InternalPort: 5433,
```
davissp14, line 125: Should just use defined port variable.

internal/command/postgres/add_flycast.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	ctx, err = apps.BuildContext(ctx, app)
+	if err != nil {
+		return err
+	}
+
+	switch app.PlatformVersion {
+	case "machines":
+		err := doAddFlycast(ctx)
```
davissp14, line 63: These should usually be: `if err := ... ; err != nil {}`

davissp14 (approved):
Just a few nits, but overall looks good!

Issue #1632, opened 2023-01-27
`fly info` doesn't list shared IPv4


catflydotio:
From #1631:

```
$ fly ips list
VERSION IP                      TYPE            REGION  CREATED AT           
v6      2a09:8280:1::d285       public          global  2023-01-25T21:35:29Z
v4      66.241.125.211          public (shared)                             
```

```
$ fly info
App
  Name     = appname       
  Owner    = personal       
  Platform = machines       
  Hostname = appname.fly.dev  

Services
PROTOCOL        PORTS                                   FORCE HTTPS 
TCP             1374404984352 => 4999 [HTTP]            True       
                1374404984368 => 4999 [TLS,HTTP]        False      

IP Addresses
TYPE    ADDRESS                 REGION  CREATED AT           
v6      2a09:8280:1::d285       global  2023-01-25T21:35:29Z
```

catflydotio:
`fly info` is gone so we can close this :)

PR #1677, opened 2023-02-07
Deprecated the info command and added the services command

### LINKED ISSUE #1631: Drop `fly info`?
Waay back in https://github.com/superfly/flyctl/issues/78 I see it was decided that `fly info` wasn't needed, and I don't see anything since.

```
$ fly status
App
  Name     = appname          
  Owner    = personal       
  Hostname = appname.fly.dev  
  Platform = machines       

ID              STATE   REGION  HEALTH CHECKS   IMAGE                                           CREATED                   UPDATED              
e286065f969386  started iad                     appname:deployment-01GQR4GN7EA8G12VE1DBRQG4KQ     2023-01-25T21:35:32Z      2023-01-26T22:52:36Z
```

```
$ fly ips list
VERSION IP                      TYPE            REGION  CREATED AT           
v6      2a09:8280:1::d285       public          global  2023-01-25T21:35:29Z
v4      66.241.125.211          public (shared)                             
```

```
$ fly info
App
  Name     = appname       
  Owner    = personal       
  Platform = machines       
  Hostname = appname.fly.dev  

Services
PROTOCOL        PORTS                                   FORCE HTTPS 
TCP             1374404984352 => 4999 [HTTP]            True       
                1374404984368 => 4999 [TLS,HTTP]        False      

IP Addresses
TYPE    ADDRESS                 REGION  CREATED AT           
v6      2a09:8280:1::d285       global  2023-01-25T21:35:29Z
```

I like `fly ips list` because I can remember it. If we wanted to list services too, we could have `fly services list` -- I'd remember that.

billyb2:
Should fix #1631 and #1632


billyb2:
Just realized that I need to make sure commands that the ability to view services from `flyctl` will be removed if this gets merged, changing this to a draft PR.

ndarilek (approved):
I'm new at this, so take with a grain of salt, but LGTM. It *might* be wise to separate out the removal of the info command from the introduction of the new commands, but I suspect `info` likely isn't used directly in automations anyway. Seems safe to merge.

billyb2:
> I'm new at this, so take with a grain of salt, but LGTM. It _might_ be wise to separate out the removal of the info command from the introduction of the new commands, but I suspect `info` likely isn't used directly in automations anyway. Seems safe to merge.

Hmm, it could now that you bring it up. I thought this would be a safe PR since two pretty old issues mentioned removing it, just because the command is superfluous. If you want, I could request a second review to double check this would be fine.

ndarilek:
[This thread](https://flyio.discourse.team/t/flyctl-command-inconsistencies/1342) has a bit of attention, it might be worth posting there for increased visibility.

Spitballing: it might be worth bringing back `fly info` as a sort of "call all these other `info` commands" super-command that just calls the various run functions? If the flags and general command structure all agree, it should be potentially less non-breaking.

internal/command/info/info.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// TODO: Deprecate and remove
```
catflydotio, line 17: This line has multiplied throughout the repo, but I don't think we need it, since you just did deprecate it and we may not want to remove it.

catflydotio (approved):
lgtm

internal/command/info/info.go
billyb2: Removed it

PR #1681, opened 2023-02-07
Machine API now launches when an existing dev agent is running

billyb2:
This is catalogued a lot more here:
https://flyio.discourse.team/t/launching-proxy-fails-on-dev-version/1802/2

Basically, this fixes a weird version number bug when a daemon fails to launch and tries again. Pretty obscure, but it was preventing me from testing the machines api.

### DIFF
--- a/agent/start.go
+++ b/agent/start.go
@@ -41,7 +41,7 @@ func StartDaemon(ctx context.Context) (*Client, error) {
 
 	env := os.Environ()
 
-	env = append(env, fmt.Sprintf("DEV_VERSION_NUM=%d", versionNum))
+	env = append(env, fmt.Sprintf("FLY_DEV_VERSION_NUM=%d", versionNum))
 	env = append(env, "FLY_NO_UPDATE_CHECK=1")
 
 	cmd.Env = env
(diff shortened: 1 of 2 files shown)


agent/start.go
```diff
@@ -35,7 +36,15 @@ func StartDaemon(ctx context.Context) (*Client, error) {
 	}
 
 	cmd := exec.Command(os.Args[0], "agent", "run", logFile)
-	cmd.Env = append(os.Environ(), "FLY_NO_UPDATE_CHECK=1")
+
+	versionNum := buildinfo.Version().Pre[0].VersionNum
+
+	env := os.Environ()
+
+	env = append(env, fmt.Sprintf("DEV_VERSION_NUM=%d", versionNum))
```
tvdfly, line 44: how about prefixing this env var name with `FLY_` ?

```suggestion
	env = append(env, fmt.Sprintf("FLY_DEV_VERSION_NUM=%d", versionNum))
```

billyb2: You're right, that's a good idea

lubien (approved):
Not sure I know exatcly whats happening but ✅

billyb2:
> Not sure I know exatcly whats happening but white_check_mark

Do any of us

PR #1684, opened 2023-02-08
Add Poetry detection to the Python scanner

antdking:
Support detecting a Poetry environment.
Saves users needing to run `poetry export` if they're just looking to spin up a python project quickly

Using `poetry.lock` as an indicator, Paketo doesn't support other dependency managers. This prevents using the builder for cases it won't work with.

note: Paketo detects based on `pyproject.toml` having `tool.poetry.dependencies.python` set

### DIFF
--- a/scanner/python.go
+++ b/scanner/python.go
@@ -1,7 +1,7 @@
 package scanner
 
 func configurePython(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
-	if !checksPass(sourceDir, fileExists("requirements.txt", "environment.yml")) {
+	if !checksPass(sourceDir, fileExists("requirements.txt", "environment.yml", "poetry.lock")) {
 		return nil, nil
 	}
 
--- a/scanner/python.go
+++ b/scanner/python.go
@@ -1,7 +1,8 @@
 package scanner
 
 func configurePython(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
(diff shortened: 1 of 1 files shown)


antdking:
Let me know if there's anything I missed, such as a Community request post, an Issue, or Tests. I can't find any contributing guidelines, so just went ahead with the PR (it's a 1-liner)

dangra:
Hi @antdking, thanks for your contribution. I was checking the state of the art on python side and seems checking for `pyproject.toml` is more general way of detecting a python project and should be fine with Poetry too ([1](https://python-poetry.org/docs/basic-usage/), [2](https://python-poetry.org/docs/pyproject/))

antdking:
While it's fine for detecting poetry, Paketo doesn't seem to support the other tools that use `pyproject.toml` (flit, hatch, pep517).

dangra:
that sounds reasonable. While we are on it, wdyt about adding "Pipfile" to the list?

dangra:
Also, I think we should add a comment stating why we can't just check on pyproject.toml presence

antdking:
> that sounds reasonable. While we are on it, wdyt about adding "Pipfile" to the list?

Good thinking; I've added it, however it's untested. Paketo should support it.

> Also, I think we should add a comment stating why we can't just check on pyproject.toml presence

Added

dangra:
Thanks @antdking

PR #1689, opened 2023-02-08
Fix shell selection logic for `fly version update` under Windows.

ndarilek:
Previously, this checked the $SHELL environment variable and used that
if set. Unfortunately, the Windows update logic is powershell-specific,
while all other platforms use bash.

If someone is running git-bash as their shell under Windows, the script
picks up bash from $SHELL, then happily sends it powershell commands
that fail, after renaming the existing executable and DLL first, thus
trashing the existing installation.


tvdfly (approved):
thank you!

PR #1693, opened 2023-02-08
Machine API now launches when an existing dev agent is running

billyb2:
I totally spaced that the Pre part of a version can just not exist. This should implement my old PR but with proper checks in place.

### DIFF
--- a/agent/start.go
+++ b/agent/start.go
@@ -42,7 +42,7 @@ func StartDaemon(ctx context.Context) (*Client, error) {
 
 	versionPre := buildinfo.Version().Pre
 
-	if versionPre != nil {
+	if len(versionPre) > 0 {
 		versionNum := versionPre[0].VersionNum
 		env = append(env, fmt.Sprintf("FLY_DEV_VERSION_NUM=%d", versionNum))
 	}
--- a/agent/start.go
+++ b/agent/start.go
@@ -13,6 +13,7 @@ import (
 	"github.com/azazeal/pause"
 
 	"github.com/superfly/flyctl/flyctl"
+	"github.com/superfly/flyctl/internal/buildinfo"
 	"github.com/superfly/flyctl/internal/filemu"
 	"github.com/superfly/flyctl/internal/logger"
 	"github.com/superfly/flyctl/internal/sentry"
@@ -35,7 +36,18 @@ func StartDaemon(ctx context.Context) (*Client, error) {
 	}
 
 	cmd := exec.Command(os.Args[0], "agent", "run", logFile)
-	cmd.Env = append(os.Environ(), "FLY_NO_UPDATE_CHECK=1")
+
+	env := os.Environ()
+	env = append(env, "FLY_NO_UPDATE_CHECK=1")
+
+	versionPre := buildinfo.Version().Pre
+
+	if versionPre != nil {
+		versionNum := versionPre[0].VersionNum
+		env = append(env, fmt.Sprintf("DEV_VERSION_NUM=%d", versionNum))
+	}
+
+	cmd.Env = env
 	setSysProcAttributes(cmd)
 
 	if err := cmd.Start(); err != nil {
--- a/agent/start.go
+++ b/agent/start.go
@@ -44,7 +44,7 @@ func StartDaemon(ctx context.Context) (*Client, error) {
 
 	if versionPre != nil {
 		versionNum := versionPre[0].VersionNum
-		env = append(env, fmt.Sprintf("DEV_VERSION_NUM=%d", versionNum))
+		env = append(env, fmt.Sprintf("FLY_DEV_VERSION_NUM=%d", versionNum))
 	}
 
 	cmd.Env = env
--- a/internal/buildinfo/version.go
+++ b/internal/buildinfo/version.go
@@ -37,7 +37,7 @@ func loadMeta() {
 	}
 
 	if IsDev() {
-		envVersionNum := os.Getenv("DEV_VERSION_NUM")
+		envVersionNum := os.Getenv("FLY_DEV_VERSION_NUM")
 		versionNum := uint64(parsedBuildDate.Unix())
 
 		if envVersionNum != "" {
--- a/internal/buildinfo/version.go
+++ b/internal/buildinfo/version.go
@@ -2,6 +2,8 @@ package buildinfo
 
 import (
 	"errors"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/blang/semver"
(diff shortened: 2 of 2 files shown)


agent/start.go
```diff
@@ -35,7 +36,18 @@ func StartDaemon(ctx context.Context) (*Client, error) {
...
 
 	cmd := exec.Command(os.Args[0], "agent", "run", logFile)
-	cmd.Env = append(os.Environ(), "FLY_NO_UPDATE_CHECK=1")
+
+	env := os.Environ()
+	env = append(env, "FLY_NO_UPDATE_CHECK=1")
+
+	versionPre := buildinfo.Version().Pre
+
+	if versionPre != nil {
```
tvdfly, line 45: probably need to check len instead of nil, because an empty list can be non-nil in go.

```go
l := make([]string, 0)
fmt.Println(l == nil)
// false
```

but a nil list will return len 0:

```go
var n []string
fmt.Println(n == nil)
// true
fmt.Println(len(n) == 0)
// true
```

tvdfly (commented):
how are you testing this?

agent/start.go
tvdfly:
```suggestion
	if len(versionPre) > 0 {
```

billyb2: Thank you, I wasnt aware of that!

billyb2:
> how are you testing this?

Would you want me to add a unit test?

jeromegn:
Tests are always welcome. Might not be the easiest parts to test though.

tvdfly:
> > how are you testing this?
> 
> Would you want me to add a unit test?

nope...

I am interested in the scenarios you're testing with the local build to know that this is working. some scenarios I would test with (as an example):

* find the set of commands that repros the issue without this change, and run them to see that it now works
* since we had a bug last time this was released, build locally with different values for version (at least one that is identical in format to what will get released) and run scenarios where the agent is stopped and started, older/same version/newer and ensure it doesn't have any errors

billyb2:
@tvdfly by killing the previous `flyctl` and adding a breakpoint to the StartDaemon function, I can verify that the function's being run, and in the way I want to. 

Upon building a debug build (verified by setting a break point at IsDev and checking the return value), when we hit StartDaemon, it's possible to quickly check the value. Upon printing the version value, we get
```
github.com/blang/semver.Version {
		Major: 0,
		Minor: 0,
		Patch: 0,
		Pre: []github.com/blang/semver.PRVersion len: 1, cap: 1, [
			(*"github.com/blang/semver.PRVersion")(0xc0002ce5c0),
		],
		Build: []string len: 1, cap: 1, ["dev"],}
```

Printing the .Pre value specifically gives us
```
[]github.com/blang/semver.PRVersion len: 1, cap: 1, [
	{VersionStr: "", VersionNum: 1676042253, IsNum: true},
]
```

The program then continues running, and we get confirmation that the proxy was successfully started:
```
Proxying local port 4280 to remote [_api.internal]:4280
```

If we kill the currently running program and run the same commands as before, we instead get a warning, then a confirmation:
```
WARN The running flyctl agent (v0.0.0-1676042596+dev) is older than the current flyctl (v0.0.0-1676042617+dev).
WARN The out-of-date agent will be shut down along with existing wireguard connections. The new agent will start automatically as needed.

Proxying local port 4280 to remote [_api.internal]:4280
```

This is good. to verify that the fix is any different then the previous commit, we have to checkout 43371b58dd8bfda3d782f909924102815d8d9368.

Stepping out of the debugger, we first need to kill the running flyctl agent. Next, we run the `machine api-proxy` command:
```
billy@laptop2 ~/W/flyctl ((v0.0.456))> go run . machine api-proxy
Proxying local port 4280 to remote [_api.internal]:4280
```

All good so far. Next we kill the running command and try it again, to see what happens:
```
WARN The running flyctl agent (v0.0.0-1676043076+dev) is older than the current flyctl (v0.0.0-1676043075+dev).
WARN The out-of-date agent will be shut down along with existing wireguard connections. The new agent will start automatically as needed.
Error _api.internal: tunnel unavailable


exit status 1
```

As expected, the program crashes. Next, in order to make sure the same crash as before doesn't happen again, we need to build a release build. First, of course, we need to return to We can do so by setting the `IsDev` function to always return false, and using `ldflags` to set the version number.

`go build -o bin/flyctl -ldflags="-X 'github.com/superfly/flyctl/internal/buildinfo.buildDate=2023-02-10T15:50:08Z' -X 'github.com/superfly/flyctl/internal/buildinfo.version=1.2.3'" `

Next, let's kill the currently running agent, to ensure that the daemon gets started. If we add a print statement to the beginning of StartDaemon (since I can't figure out the ldflags stuff with delve at this moment), we get

```
Version: 1.2.3

Proxying local port 4280 to remote [_api.internal]:4280
```

Good. Now let's rerun the command, since the agent will  still  be running in the background. This *shouldn't* print a warning, or the version number.

```
Proxying local port 4280 to remote [_api.internal]:4280
```

Yes! Seems to be working so far. Let's try running the same command again, but with a higher version number.

```
billy@laptop2 ~/W/flyctl (fix_machines_api_proxy)> go build -o bin/flyctl -ldflags="-X 'github.com/superfly/flyctl/internal/buildinfo.buildDate=2023-02-10T15:50:08Z' -X 'github.com/superfly/flyctl/internal/buildinfo.version=1.2.4'"
billy@laptop2 ~/W/flyctl (fix_machines_api_proxy)> ./bin/flyctl machine api-proxy
WARN The running flyctl agent (v1.2.3) is older than the current flyctl (v1.2.4).
WARN The out-of-date agent will be shut down along with existing wireguard connections. The new agent will start automatically as needed.
Version: 1.2.4

Proxying local port 4280 to remote [_api.internal]:4280
```

Working as expected, thank god. Finally, let's try adding a Pre to the version number (not entirely sure what it's called, but the -[versionPre] thing).

```
billy@laptop2 ~/W/flyctl (fix_machines_api_proxy)> go build -o bin/flyctl -ldflags="-X 'github.com/superfly/flyctl/internal/buildinfo.buildDate=2023-02-10T15:50:08Z' -X 'github.com/superfly/flyctl/internal/buildinfo.version=1.2.4-1'"
.⏎                                                                                            billy@laptop2 ~/W/flyctl (fix_machines_api_proxy)> ./bin/flyctl machine api-proxy
WARN The running flyctl agent (v1.2.4) is older than the current flyctl (v1.2.4-1).
WARN The out-of-date agent will be shut down along with existing wireguard connections. The new agent will start automatically as needed.
Version: 1.2.4-1

Proxying local port 4280 to remote [_api.internal]:4280
```

Rerunning the same command shouldn't print any warnings.

```
billy@laptop2 ~/W/flyctl (fix_machines_api_proxy)> ./bin/flyctl machine api-proxy
Proxying local port 4280 to remote [_api.internal]:4280
```

And it doesn't. From the testing done above, I'm fairly confident that this change won't introduce any crashes or regressions.

tvdfly:
Thanks for the details @billyb2 ! Sounds like this is well tested and ready to go. Nice work!

PR #1694, opened 2023-02-08
Add support for creating machines from a JSON file

### LINKED ISSUE #1683: Add support for creating machine from a json file
Something like `fly machine run --machine-config ./my-machine.json`, which would essentially be a wrapper around the machines API call and would use the json file as the request body.

alichay:
This would close #1683

This isn't the prettiest, but I tried to share most of the code between the two path so that little bugs didn't sneak in over time

### DIFF
--- a/internal/command/machine/flaps_api/post.go
+++ b/internal/command/machine/flaps_api/post.go
@@ -121,6 +121,9 @@ func runPost(ctx context.Context) error {
 		app, err = client.GetAppCompact(ctx, appName)
 		if err != nil && strings.Contains(err.Error(), "Could not find App") {
 			app, err = mach.CreateApp(ctx, fmt.Sprintf("App '%s' does not exist, would you like to create it?", appName), appName, client)
+			if err != nil {
+				return err
+			}
 			if app == nil {
 				return nil
 			}
--- a/internal/command/machine/flaps_api/post.go
+++ b/internal/command/machine/flaps_api/post.go
@@ -1,4 +1,4 @@
-package machine
+package flaps_api
 
 import (
 	"context"
@@ -24,15 +24,21 @@ import (
 	"github.com/superfly/flyctl/internal/watch"
 )
 
-func newRunFromJson() *cobra.Command {
+var sharedFlags = flag.Set{
+	flag.App(),
+	flag.AppConfig(),
+	flag.Detach(),
+}
+
+func newPost() *cobra.Command {
 	const (
 		short = "Run a machine from a JSON configuration file"
 		long  = short + "\n"
 
-		usage = "run_from_json <json_file> [command]"
+		usage = "post <json_file> [command]"
 	)
 
-	cmd := command.New(usage, short, long, runMachineRunFromJson,
+	cmd := command.New(usage, short, long, runPost,
 		command.RequireSession,
 		command.LoadAppNameIfPresent,
 	)
@@ -40,11 +46,6 @@ func newRunFromJson() *cobra.Command {
 	flag.Add(
 		cmd,
 		flag.Region(),
-		// deprecated in favor of `flyctl machine update`
-		flag.String{
-			Name:        "id",
-			Description: "Machine ID, if previously known",
-		},
 		flag.String{
 			Name:        "name",
 			Shorthand:   "n",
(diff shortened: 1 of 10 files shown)


internal/command/machine/run.go
```diff
@@ -160,21 +166,73 @@ func newRun() *cobra.Command {
 		sharedFlags,
 	)
 
-	cmd.Args = cobra.MinimumNArgs(1)
+	// If we pass a JSON machine configuration, we don't build a local image.
+	// That introduces this ugliness, which changes the expected arg count depending
+	// on flags passed.
```
tvdfly, line 171: this complexity is perhaps a good sign that this functionality should live in a different command. How about moving this feature to something like `fly m create` or `fly m create_in_api` or something like that?

tvdfly: a totally different idea:

```
fly machine api post
```

so it's super clear this is simply a passthrough to the machines api, and won't get confused with the other machine sub commands.

In the future, we could add other methods too. Like `fly machine api put`.

tvdfly: or maybe `fly machine flaps-api post`, to maybe be less likely confused with the api-proxy subcommand

alichay: How about `json-api`? flaps feels more like an implementation detail imo

alichay:
@tvdfly How are we feeling about this version of this?

internal/command/machine/flaps_api/post.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	input.AppID = app.Name
+	// Support a few command-line overrides, but this shouldn't be a clone of `machine run`
+	if name := flag.GetString(ctx, "name"); name != "" {
+		input.Name = name
+	}
+	if region := flag.GetString(ctx, "region"); region != "" {
+		input.Region = region
+	}
```
tvdfly, line 156: You might be aware of this already and have other reasons for implementing this way, but just in case not: this can be written without the if statement wrapper because the default value for a string in go is `""`:

```suggestion
	input.Region = flag.GetString(ctx, "region")
```

internal/command/machine/flaps_api/post.go
```diff
@@ -0,0 +1,189 @@
...
+	fmt.Fprintf(io.Out, " Instance ID: %s\n", instanceID)
+	fmt.Fprintf(io.Out, " State: %s\n", state)
+
+	// wait for machine to be started
+	if err := mach.WaitForStartOrStop(ctx, machine, "start", time.Minute*5); err != nil {
+		return err
+	}
+
+	if !flag.GetDetach(ctx) {
+		fmt.Fprintln(io.Out, colorize.Green("==> "+"Monitoring health checks"))
```
tvdfly, line 177:
```suggestion
		fmt.Fprintln(io.Out, colorize.Green("==> Monitoring health checks"))
```

tvdfly (commented):
I suspect in most cases if the app doesn't exist it will mean the caller forgot to specify it or is in the wrong directory, since this will be a more advanced flyctl subcommand.

What if we had this command fail if it cannot find an app from context? It could include in the error message something like
> specify app or fly.toml or use `fly apps create` to create a new app

so the caller has the instructions for resolving on their own.

internal/command/machine/flaps_api/post.go
alichay: I wanted to support region being included in the json, so I only set it when there's an override.

internal/command/machine/flaps_api/post.go
alichay: yeah idk why the original run was written like that lol

shortdiv:
This is awesome! Though I'm not a fan of the flaps-api command in flyctl. It looks like the machines api refers to it loosely but we should probably abstract that away since it adds extra complexity/overhead for a user who just wants to create a machine with a json file. Considering that other machine commands that go through flaps aren't exposed, we should offer a similar user experience. So I think a flag would be better, something like --machine-config perhaps?

alichay:
> Though I'm not a fan of the flaps-api command in flyctl

I also wasn't too keen on it at first. The version zero of this was something like `fly m run --from-json some_file.json`, but it turned into a hack where the number of arguments that `fly m run` expected depended on the flags passed to the command, and it was _fine_, but the code was pretty ugly. (because when you aren't passing in a json file, you're expected to provide a docker image as a positional argument)

T and I agreed that it was better refactored out into a separate command, and between `fly m run_from_json` (or something like that) and `fly m flaps_api post`, post won out because it was more extensible.

The idea being, I believe, to eventually flesh-out `flaps_api` into like a one-step-above-curl kind of hub for testing raw API calls for machines. it's been a while, though, since i was knee-deep in all of this

PR #1700, opened 2023-02-11
Fix inconsistency between optional/required arguments to commands, enforce argument requirements

### LINKED ISSUE #1663: Minor readability issue in command parameters optionality
**Describe the bug**
The documentation for commands is inconsistent with the convention of `<required>` and `[optional]` parameters.

For example,

```
> flyctl apps restart
Error: accepts 1 arg(s), received 0
Usage:
  flyctl apps restart [APPNAME] [flags]
```

This implies `[APPNAME]` is optional, but it is in fact required. Other parameters in other commands use `<required>` when appropriate. This should be `<APPNAME>` to follow convention..

* Operating system

macOS 13.2

* `fly version`

flyctl v0.0.451 darwin/arm64 Commit: 5a5b5b1f BuildDate: 2023-01-29T21:49:28Z

** Paste your `fly.toml`

not relevant

** Command output: **

see above

alichay:
This does two things, separated into the two commits:

1. Canonicalize `[FOO]` as optional, and `<FOO>` as required in usage info.
   This would close #1663, and is a cosmetic change.
2. Make better use of `cobra.ExactArgs` where it seems useful.

### DIFF
--- a/internal/command/apps/destroy.go
+++ b/internal/command/apps/destroy.go
@@ -20,7 +20,7 @@ func newDestroy() *cobra.Command {
 from the Fly platform.
 `
 		short = "Permanently destroys an app"
-		usage = "destroy [APPNAME]"
+		usage = "destroy <APPNAME>"
 	)
 
 	destroy := command.New(usage, short, long, RunDestroy,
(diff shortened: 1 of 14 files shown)


tvdfly (commented):
awesome, thank you!

I spot checked a couple of the exact arg changes and they looked reasonable based on surrounding code.

How did you test those exact args changes? Were you able to successfully run all those commands after making the changes?

alichay:
> Were you able to successfully run all those commands after making the changes

I checked everything but turboku

dangra (approved):
I didn't test it but looks good.

PR #1708, opened 2023-02-13
fly machine restart: inherit app from first machine, require all machines to share app

### LINKED ISSUE #1703: `fly machine restart` requires an app name where other `machine` commands don't
```
$ fly m restart 06e82990b37387
Error could not get app: Could not find App
```

It's confusing to hit that when `fly m start` works with just the machine id. 

Though I wonder what it should do if you give it ids of multiple machines that don't belong to the same app. `fly m start` claims to accept multiple ids as well, so whatever it is should be consistent.

alichay:
This brings behavior in line with other machine commands, like `m start`

(would fix #1703)

### DIFF
--- a/internal/command/machine/restart.go
+++ b/internal/command/machine/restart.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/api"
-	"github.com/superfly/flyctl/client"
 	"github.com/superfly/flyctl/flaps"
 	"github.com/superfly/flyctl/internal/app"
 	"github.com/superfly/flyctl/internal/command"
@@ -61,16 +60,29 @@ func newRestart() *cobra.Command {
 	return cmd
 }
 
-func runMachineRestart(ctx context.Context) error {
+func runMachineRestart(ctx context.Context) (err error) {
+
 	var (
-		args    = flag.Args(ctx)
+		args = flag.Args(ctx)
+	)
+
+	for _, machineID := range args {
+		if err = restart(ctx, machineID); err != nil {
+			return
+		}
+	}
+	return
+}
+
+func restart(ctx context.Context, machineID string) (err error) {
+
+	var (
 		signal  = flag.GetString(ctx, "signal")
 		timeout = flag.GetInt(ctx, "time")
 		appName = app.NameFromContext(ctx)
-		client  = client.FromContext(ctx).API()
 	)
 
-	app, err := client.GetAppCompact(ctx, appName)
+	app, err := appFromMachineOrName(ctx, machineID, appName)
 	if err != nil {
 		return fmt.Errorf("could not get app: %w", err)
 	}
@@ -100,28 +112,21 @@ func runMachineRestart(ctx context.Context) error {
 
 	flapsClient := flaps.FromContext(ctx)
 
-	var machines []*api.Machine
-	// Resolve machines
-	for _, machineID := range args {
-		machine, err := flapsClient.Get(ctx, machineID)
-		if err != nil {
-			return fmt.Errorf("could not get machine %s: %w", machineID, err)
-		}
-		machines = append(machines, machine)
+	// Resolve machine
+	machine, err := flapsClient.Get(ctx, machineID)
+	if err != nil {
+		return fmt.Errorf("could not get machine %s: %w", machineID, err)
 	}
 
-	// Acquire leases
-	machines, releaseLeaseFunc, err := mach.AcquireLeases(ctx, machines)
+	// Acquire lease
+	machines, releaseLeaseFunc, err := mach.AcquireLease(ctx, machine)
 	defer releaseLeaseFunc(ctx, machines)
 	if err != nil {
 		return err
 	}
 
-	// Restart each machine
-	for _, machine := range machines {
-		if err := mach.Restart(ctx, machine, input); err != nil {
-			return fmt.Errorf("failed to restart machine %s: %w", machine.ID, err)
-		}
+	if err := mach.Restart(ctx, machine, input); err != nil {
+		return fmt.Errorf("failed to restart machine %s: %w", machine.ID, err)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


internal/command/machine/restart.go
```diff
@@ -61,16 +60,29 @@ func newRestart() *cobra.Command {
...
+
+	for _, machineID := range args {
+		if err = restart(ctx, machineID); err != nil {
+			return
+		}
+	}
+	return
+}
+
+func restart(ctx context.Context, machineID string) (err error) {
```
tvdfly, line 77: it doesn't look like the return variable is being used (all the returns specify what they're returning rather than returning empty), so probably safer to remove it.

```suggestion
func restart(ctx context.Context, machineID string) error {
```

tvdfly (commented):
It looks like we're changing behavior from:

* acquire leases on all machines that are being restarted
* restart all machines
* release all leases

to:

* acquire lease on one machine
* restart one machine
* release one machine
* repeat for the rest of the machines

I think in the other places we acquire leases for multiple machines, we typically use the original form of acquiring leases on all machines. The benefit of doing that is if any of the leases are unable to be acquired, none of the operations on the machines succeed. This is helpful for resolving conflicts when multiple entities are trying to get leases on multiple machines at the same time, because both will fail without making changes (and then can retry until one acquires the full set of leases).

I'm a little nervous about changing the lease acquire behavior, since some folks might be relying on it (or at least surprised if it changed).

alichay:
you're totally right, I missed that behavior change. fixing that seems messy, though. we're left with three ways forward.

1. keep this as-is, and change leasing behavior... this is a bad idea.
2. store a struct containing `(app, ctx)` in an array next to the machines array, and make everything a little more messy. this brings the UX in line with `m start` and `m stop`, though, which would have nice symmetry and be very predictable.
3. enforce that all the machines are in the same app. this makes the implementation a lot simpler and cleaner, but makes the command function differently to its counterparts.

i'm leaning towards 2, but it makes things messy (for example, having to reimplement a local version `mach.AcquireLeases` so that it can use each machine's context)

internal/command/machine/restart.go
```diff
@@ -70,11 +70,24 @@ func runMachineRestart(ctx context.Context) error {
...
-	app, err := client.GetAppCompact(ctx, appName)
+	app, err := appFromMachineOrName(ctx, args[0], appName)
 	if err != nil {
 		return fmt.Errorf("could not get app: %w", err)
 	}
 
+	// Ensure that all machines are on the same app
+	for _, machineID := range args {
+
+		machine, err := client.GetMachine(ctx, machineID)
```
dangra, line 81: This is making a Web API call per machine where few lines below there is another loop that does the same but hitting Machines API endpoint:  https://github.com/superfly/flyctl/blob/1e084c967e42042a20b43f3c9d80ca900accd237/internal/command/machine/restart.go#L114-L124

It will return a 404 if the machine doesn't belong to the app tho, which should be enough, isn't it?

I would check [line 144](https://github.com/superfly/flyctl/blob/1e084c967e42042a20b43f3c9d80ca900accd237/internal/command/machine/restart.go#L114) too, that is pulling the flaps client (which needs an app to operate) that somehow could be different (or always the same) than what we get in https://github.com/superfly/flyctl/blob/1e084c967e42042a20b43f3c9d80ca900accd237/internal/command/machine/restart.go#L73

alichay: I double-hit the API because the second one only says something like "machine not found," which could confusingly imply that either the machine ID was wrong *or* that the machine wasn't in the app. When the app is specified implicitly by the first machine, this gets confusing.

That said, I guess it would be more reasonable to just catch the error in the second loop, and if it's a machine not found error, turn the error into something like "machine {id} not found. perhaps this machine is not in the app {app}"

alichay: > I would check [line 144](https://github.com/superfly/flyctl/blob/1e084c967e42042a20b43f3c9d80ca900accd237/internal/command/machine/restart.go#L114) too, that is pulling the flaps client (which needs an app to operate) that somehow could be different (or always the same) than what we get in

I'm sorry, I'm not sure what you're getting at here. I rechecked the flow and I don't think I'm missing anything 
The app is derived from `appFromMachineOrName`. that's passed into `apps.BuildContext`, which initializes a bunch of things including the flaps client, and then `flaps.FromContext(ctx)` retrieves that flaps client.

At least with my understanding, this should all be fairly normal usage?

dangra: > That said, I guess it would be more reasonable to just catch the error in the second loop, and if it's a machine not found error, turn the error into something like "machine {id} not found. perhaps this machine is not in the app {app}"

Yeap, good idea. That is a good tradeoff IMO.

> I rechecked the flow and I don't think I'm missing anything
The app is derived from appFromMachineOrName. that's passed into apps.BuildContext, which initializes a bunch of things including the flaps client, and then flaps.FromContext(ctx) retrieves that flaps client.

You just taught me how it works! thx

internal/command/machine/restart.go
```diff
@@ -105,7 +104,11 @@ func runMachineRestart(ctx context.Context) error {
 	for _, machineID := range args {
 		machine, err := flapsClient.Get(ctx, machineID)
 		if err != nil {
-			return fmt.Errorf("could not get machine %s: %w", machineID, err)
+			if strings.Contains(err.Error(), "machine not found") {
+				return fmt.Errorf("could not get machine %s. perhaps this machine doesn't exist, or isn't part of the app '%s'", machineID, app.Name)
```
tvdfly, line 108: with #1674 being merged (just a few hours ago), it's now possible to get the status code from these errors. I think something like this would work:

```suggestion
			var flapsErr *flaps.FlapsError
			if errors.As(err, &flapsErr) && flapsErr.ResponseStatusCode == http.StatusNotFound {
				return fmt.Errorf("machine %s not found in app '%s'", machineID, app.Name)
			} else if flapsErr != nil {
				return fmt.Errorf("error getting machine %s in app '%s': %w", machineID, app.Name, flapsErr)
```

please feel empowered to merge with this or without this. we can always come back to update this kind of error handling later or in a separate pr.

alichay: I'll leave this as an enhancement for later, just to get this out the door. that API is really nice to have though, thanks!

PR #1712, opened 2023-02-14
Jphenow/alert paid plan region

jphenow:


internal/prompt/prompt.go
```diff
@@ -346,12 +400,16 @@ func SelectRegion(ctx context.Context, msg string, regions []api.Region, default
...
 
-func MultiSelectRegion(ctx context.Context, msg string, regions []api.Region, currentRegions []string) (selectedRegions []api.Region, err error) {
+func MultiSelectRegion(ctx context.Context, msg string, paid []api.Region, regions []api.Region, currentRegions []string) (selectedRegions []api.Region, err error) {
 	var options []string
 
 	var currentIndices []int
 	var indices []int
 
+	if isInteractive(ctx) && len(paid) > 0 {
+		fmt.Printf("Some regions require a paid plan (%s).\nSee https://fly.io/plans to set up a plan.\n\n", strings.Join(lo.Map(paid, func(r api.Region, _ int) string { return r.Code }), ", "))
```
tvdfly, line 410: how about using iostreams stuff? that will handle various terminal differences. it also will allow colors/markup if we want to add that (now or in the future).

I think something like this would work. I haven't tested, so might be slightly wrong.

```suggestion
		io := iostreams.FromContext(ctx)
		fmt.Fprintf(io.ErrOut, "Some regions require a paid plan (%s).\nSee https://fly.io/plans to set up a plan.\n\n", strings.Join(lo.Map(paid, func(r api.Region, _ int) string { return r.Code }), ", "))
```

it's also nice to have this stuff to print to stderr, so if there is output from the command (especially json output) it can go to stdout and be safely parsed without having to filter the prompts/warnings. we don't do that consistently throughout flyctl today; doing it for new code is a good start, though.

jsierles (approved):

PR #1718, opened 2023-02-15
Don't always enable virtual terminal input/processing under Windows, and clean up when exiting SSH

### LINKED ISSUE #1657: Windows arrow keys print key escape codes
**Describe the bug**
Arrow keys (and possibly others) don't work on windows, instead just printing their escape codes.

* Operating system: Windows 11
* `fly version`: `fly.exe v0.0.451 windows/amd64 Commit: 5a5b5b1f BuildDate: 2023-01-29T21:49:28Z`

**Command output:**

```text
$ fly launch
Creating app in C:\Users\legor
Scanning source code
? Choose an app name (leave blank to generate one): [D[A[C[B
```

ndarilek:
Fixes #1657

### DIFF
--- a/internal/command/ssh/console_windows.go
+++ b/internal/command/ssh/console_windows.go
@@ -1,30 +1,71 @@
-//go:build windows
-// +build windows
-
 package ssh
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/Azure/go-ansiterm/winterm"
 )
 
-func init() {
-	stdoutFd := os.Stdout.Fd()
-	if mode, err := winterm.GetConsoleMode(stdoutFd); err == nil {
-		err := winterm.SetConsoleMode(stdoutFd, mode|winterm.ENABLE_VIRTUAL_TERMINAL_PROCESSING)
-		if err != nil {
-			fmt.Fprintln(os.Stderr, "failed to enable virtual terminal processing for stdout:", err)
-			return
-		}
-	}
+// Windows consoles need a bit of magic to correctly handle ANSI escape
+// sequences.
+func setupConsole() (uint32, uint32, uint32, error) {
+	var (
+		currentStdinMode  uint32
+		currentStdoutMode uint32
+		currentStderrMode uint32
+	)
 
 	stdinFd := os.Stdin.Fd()
-	if mode, err := winterm.GetConsoleMode(stdinFd); err == nil {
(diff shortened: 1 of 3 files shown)


internal/command/ssh/console_windows.go
```diff
@@ -1,30 +1,71 @@
-//go:build windows
-// +build windows
```
tvdfly, line 2: did you meant to remove these?

internal/command/ssh/console_windows.go
```diff
@@ -1,30 +1,71 @@
...
 package ssh
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/Azure/go-ansiterm/winterm"
 )
 
-func init() {
```
tvdfly, line 13: init() is a magic method that runs when the package is imported... do we need to replicate that behavior now that it won't happen implicitly?

internal/command/ssh/console_windows.go
ndarilek: I'll add them back for clarity and consistency. I took them out because they were also implied by the filename, but that is a bit more brittle and magical.

internal/command/ssh/console_windows.go
ndarilek: So admittedly I don't fully understand the lower-level console output code, particularly the Windows magical bits, but that initial `init` was enabling virtual terminal processing under Windows, in the belief that there were no consequences for doing so.

Unfortunately, that universal hack broke prompt input such that arrows spat out their control characters. Further, I just started using PowerShell more heavily, and after running an SSH console I was no longer able to arrow up through command history. When I removed that original hack on a whim, arrows were handled correctly in prompts, and I was able to navigate command history again.

Since the original hack was only needed for `ssh console`, I now wrap that console invocation in calls that enable virtual terminal processing/input, then restore the original values when the console exits.

Not sure if there's a better pattern for this in Go. Please let me know if there is--I can't decide whether my dislike of this code is due to Go or my lack of knowledge. :)

I've tested spinning up an SSH console and launching a new app in the same terminal. In all cases, including arrowing through command history and prompt input, my terminal behaved as expected.

tvdfly: oh perfect, if you've tested and things are working that's good enough for me. Thanks!

PR #1733, opened 2023-02-19
Improve long help message for `secrets list`

scorphus:
The long message appears broken on the terminal. This CL not only removes the broken indentation but also amends the alignment.

Before:

```
$ flyctl help secrets list
List the secrets available to the application. It shows each
                secret's name, a digest of the its value and the time the secret was last set.
                The actual value of the secret is only available to the application.

[snip]
```

After:

```
$ ./bin/flyctl help secrets list
List the secrets available to the application. It shows each secret's
name, a digest of its value and the time the secret was last set. The
actual value of the secret is only available to the application.

[snip]
```

Update: I'm happy making any changes or any further amendments throughout the code. The initial intent here was only to fix the broken text but then I couldn't leave the extra `the` in there and consequently had to fix all other instances of that message, to keep them identical [😊](# "trying to follow the good boy scout rule")

### DIFF
--- a/docstrings/gen.go
+++ b/docstrings/gen.go
@@ -681,9 +681,9 @@ are read from stdin as name=value`,
 		}
 	case "secrets.list":
 		return KeyStrings{"list", "Lists the secrets available to the app",
-			`List the secrets available to the application. It shows each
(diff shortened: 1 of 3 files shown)


dangra:
thanks!

PR #1740, opened 2023-02-20
Fixed `services list` for machine apps

### LINKED ISSUE #1716: `fly services list` failing on Machines apps
fly v0.0.457 darwin/arm64 Commit: f58a2e68 BuildDate: 2023-02-15T01:31:03Z

```
$ fly services list -a appname         
```

Error from @lillianberryfly's local build from master:
```
panic: interface conversion: interface {} is nil, not *flaps.Client
```

billyb2:
I was printing the pointer's memory address, and not the port number itself Should fix #1716

### DIFF
--- a/internal/command/services/machines.go
+++ b/internal/command/services/machines.go
@@ -61,7 +61,7 @@ func showMachineServiceInfo(ctx context.Context, app *api.AppInfo) error {
 
 			fields := []string{
 				strings.ToUpper(protocol),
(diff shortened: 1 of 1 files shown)


tvdfly:
how did you repro #1716? the " panic: interface conversion: interface {} is nil, not *flaps.Client" error looks like an issue with the flaps client instead of port.

billyb2:
I think it was fixed in 7c7ddff5f7dafb41e96fb909528bd7c981b34a68 (thank you by the way).

tvdfly:
LGTM, thx!

Issue #1741, opened 2023-02-20
flyctl does not support [mount] configuration


tvdfly:
Apps v1 allows developers to use `[mount]` or `[mounts]` (singular or plural) to specify the volumes to mount to instances.

Apps v2 only supports the `[mounts]` version.

Update flyctl to support both version for apps v2.

PR #1743, opened 2023-02-20
[appsv2] Convert fly.toml's [mount] to [mounts]

dangra:
Fixes #1741

PR #1744, opened 2023-02-20
add backwards compat for mounts in apps v2

gwuah:
this should resolve https://github.com/superfly/flyctl/issues/1741


dangra:
Hi @gwuah ! Seems we step over each other on this one. My bad because I didn't assign the original ticket to myself before working on it. I don't mind closing https://github.com/superfly/flyctl/pull/1743 but please take a look at it. Changing full-reference.toml is not desirable, I think old-format.toml is a better place to do it.

gwuah:
oh damn @dangra , i didn't notice you had already worked on it. 
And yes, you're right. I'll update the PR and re-request a review.

internal/appv2/patches.go
```diff
@@ -243,6 +244,18 @@ func ensureArrayOfMap(raw any) ([]map[string]any, error) {
 	return out, nil
 }
 
+func patchMounts(cfg map[string]any) (map[string]any, error) {
+
```
dangra, line 248: I find it weird but lately common that people adds a newline after the function signature and so many blank lines between bodies which are so small already. Is this caused by some editor (like vscode), some go formatting tool or just added manually?

dangra:
> oh damn @dangra , i didn't notice you had already worked on it.
no prob! I was delighted to see you found the patches.go file and existing test cases. Your PR was very close to mine and I am away until tomorrow.

internal/appv2/patches.go
```diff
@@ -243,6 +244,18 @@ func ensureArrayOfMap(raw any) ([]map[string]any, error) {
...
+	if mount, ok := cfg["mount"]; ok {
```
tvdfly, line 249: what's the behavior if `[mount]` and `[mounts]` are both defined?

dangra: `mount` overrides `mounts` and IMO that is fine, very weird to see both on same file.

internal/appv2/patches.go
gwuah: honestly i can't tell for sure if i did that unintentionally or vscode auto inserted those spaces. 
I've cleaned it up. Can i get a review(✅) on this?

internal/appv2/patches.go
tvdfly: yeah, probably not a valid config file if both are defined. what if we errored in that case that way the developers has to update the config file to use only one?

gwuah: @tvdfly i could easily add that validation. but, I was [following a convo](https://flyio.discourse.team/t/fly-toml-mounts-v-mount/1905) you were having with Michael on discourse and it turns out we accept both the singular and plural form for all the sections. and so if we are going to do it for **mount**, then we might as well have the validation on all the sections?

i think we can just leave it as is. what do you think?

tvdfly: oh gotcha. it looks like we're overwriting checks with check when both are specified in the nomad platform, so that seems like a good choice here.

tvdfly (approved):

PR #1749, opened 2023-02-21
WIP: Automate management of a log shipper VM

jsierles:
WIP: This PR adds the `flyctl logs ship` command which sets up a [log shipper](https://github.com/superfly/fly-log-shipper) VM. Initially, this command should offer automatic log shipping via an integrated [Logtail](https://betterstack.com/logtail) account, and instructions for adding other log providers.

This will be accompanied by a backend PR.

### DIFF
--- a/internal/command/logs/ship.go
+++ b/internal/command/logs/ship.go
@@ -1,0 +1,103 @@
+package logs
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/flaps"
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/command/orgs"
+	"github.com/superfly/flyctl/internal/flag"
+	"github.com/superfly/flyctl/iostreams"
+)
+
+func newShip() *cobra.Command {
+
+	const (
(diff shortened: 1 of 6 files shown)


internal/command/logs/ship.go
```diff
@@ -0,0 +1,113 @@
...
+func runShip(ctx context.Context) (err error) {
+	client := client.FromContext(ctx).API()
+	io := iostreams.FromContext(ctx)
+	selectedOrg, err := orgs.OrgFromFirstArgOrSelect(ctx)
+
+	if err != nil {
+		return err
+	}
+
+	appName := selectedOrg.Slug + "-auto-log-shipper"
```
tvdfly, line 45: how about adding an optional flag for this, so folks have the option of choosing the app name? I like this for the default name.

internal/command/logs/ship.go
```diff
@@ -0,0 +1,113 @@
...
+
+		createdApp, err := client.CreateApp(ctx, input)
+		app = client.AppToCompact(createdApp)
+
+		if err != nil {
+			return err
+		}
+	}
+
+	fmt.Fprintf(io.Out, "Setting up secrets for %s\n", app.Name)
```
tvdfly, line 66: how about using stderr here? that would allow stdout to support the `--json` flag in the future for returning the app information.

tvdfly (approved):

jsierles:
Closing in favor of a simpler PR that doesn't touch apps deployment.

PR #1750, opened 2023-02-21
`fly apps restart`: only restart machines managed by Apps platform

### LINKED ISSUE #1567: Apps v2: fly apps restart
Apps v2: We're pushing to migrate the Fly Apps platform to machines instead of nomad to improve developer experience. The first big step will be making new apps use the machines platform instead of nomad. There's a bunch of small steps to get there, and this is one of them...

`fly apps restart` only restarts machines that are managed as part of the Fly Apps Platform. It ignores any other machines that may have been launched in the app.

Make sure a release is created in the graphql backend to record the restart happened.

If #1551 is still in progress, update that PR with these changes.

alichay:
Would fix #1567
This is a work in progress.

`fly apps restart` should only restart the machines that are orchestrated by Apps. This has had *some* testing but not rigorous enough for release.

To be done:
 * test that this works
 * what should happen if restarting an app with a stopped machine?

### DIFF
--- a/internal/command/apps/restart.go
+++ b/internal/command/apps/restart.go
@@ -3,8 +3,8 @@ package apps
 import (
 	"context"
 	"fmt"
-
 	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/flaps"
 
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/client"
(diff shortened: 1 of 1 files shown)


tvdfly (commented):
this looks good.
I think we should call the createRelease() mutation as part of this. That could be done here or separately.

alichay:
I split this off from createRelease() because that's causing a bit of discussion, so I figured i'd split the uncontroversial from the controversial

anyway, I finally got time to test this all and it seems to work how it should
i wasn't initially sure how this would handle an app with a manually stopped machine, but it just starts it up which seems right

tvdfly (approved):
thx!

PR #1757, opened 2023-02-22
Update fly config save for apps v2

tvdfly:
update the gql schema, add a gql query to get the current release, and use the config definition from that release to build out the config that gets written.

The exciting news is: this can also synthesize a fly.toml from existing machines. That's going to help folks that have been using machines and need/want to keep using `fly deploy`, `fly status`, etc. `fly config save` will spit out a fly.toml that will as close as possible replicate the config that is found on the existing machines.


cmd/config.go
```diff
@@ -169,3 +182,48 @@ func writeAppConfig(path string, appConfig *flyctl.AppConfig) error {
 
 	return nil
 }
+
+func saveAppV2Config(ctx context.Context, apiClient *api.Client, appName, path string) error {
+	appConfig, err := getAppV2Config(ctx, apiClient, appName)
+	// FIXME: handle situation where no releases are found, make up a fly config from the app data we have + the machine config (will be fun!)
```
tvdfly, line 188: doing this turned into a slightly larger task than I expected. i'm going to make that a separate pr.

tvdfly:
(not ready to merge yet, needs more testing)

PR #1759, opened 2023-02-22
`fly machine update`: fix hanging in certain edge cases

### LINKED ISSUE #1751: fly m update fails due to bug in calling wait endpoint
Call something like `flyctl m update abcdef123123 --image registry.fly.io/myapp:deployment-0ABCDDFADFADSFA`, and it will hang forever.

LOG_LEVEL=debug will show an error message calling the wait endpoint.
```
{
"error": "instance_id required when waiting for stopped machine"
}
```

Things we can do to improve: 1) start passing in instance id in this code, and 2) do some better error handling when calling wait (we should probably cast errors to FlapsError and look at the status code before deciding whether or not to retry the wait call—not all errors mean keep waiting).

alichay:
Would fix #1751 

* add flaps error handling to `WaitForStartOrStop` for `400` responses
* make `internal/machine/update.go`'s `Update` pass a fully-formed machine to `WaitForStartOrStop` rather than just a machine ID, because in some cases (scheduled machines?) it might need an instance ID.

### DIFF
--- a/internal/machine/update.go
+++ b/internal/machine/update.go
@@ -13,15 +13,18 @@ import (
 
 func Update(ctx context.Context, m *api.Machine, input *api.LaunchMachineInput) error {
 	var (
-		flapsClient = flaps.FromContext(ctx)
-		io          = iostreams.FromContext(ctx)
-		colorize    = io.ColorScheme()
+		flapsClient    = flaps.FromContext(ctx)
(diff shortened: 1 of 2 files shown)


billyb2:
Looks good on my end!

tvdfly (approved):

PR #1762, opened 2023-02-22
Better machine validation + allow performance presets

### LINKED ISSUE #1648: Apps v2 - (pre-5) `fly machine update` allows invalid(?) CPU/RAM combos
Playing around with changing Machine sizing. flyctl doesn't seem to do any validation of Machine scaling updates.

It's happy to let me try to update a shared-cpu-1x 256MB machine to an 8x CPU, until it gets an error from flaps:

```
$ fly m update 90800d5b346e87 --cpus 8            
...
Configuration changes to be applied to machine: 90800d5b346e87 (polished-resonance-3972)

        ... // 52 identical lines
                "guest": {
                        "cpu_kind": "shared",
-                       "cpus": 1,
+                       "cpus": 8,
                        "memory_mb": 256
                }
        }
  
? Apply changes? Yes
Updating machine 90800d5b346e87
Error could not stop machine 90800d5b346e87: failed to update VM 90800d5b346e87: invalid config.guest.memory_mb, minimum required 2048 MiB
```

I can get a shared-cpu-16x with 4096MB RAM right past `fly m update`:

```
$ fly m update 90800d5b346e87 --cpus 16 --memory 4096
```

After the update, `fly m status -d` shows
```
 ...
  "guest": {
    "cpu_kind": "shared",
    "cpus": 16,
    "memory_mb": 4096
  }
```

billyb2:
Should fix #1648 

This just adds some client side checks when updating machines to ensure that machines are valid. 

Relevant discussion:
https://flyio.discourse.team/t/applying-limits-to-virtual-machines-on-the-cli/1843/8

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -599,7 +599,7 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 				machine_type = "performance"
 
 			} else {
-				return machineConf, fmt.Errorf("invalid machine preset requested, '%s', expected 'shared' or 'performance'", guestSize)
+				return machineConf, fmt.Errorf("invalid machine preset requested, '%s', expected to start with 'shared' or 'performance'", guestSize)
 
 			}
 
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -588,10 +588,24 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 
 	if guestSize := flag.GetString(ctx, "size"); guestSize != "" {
 		guest, ok := api.MachinePresets[guestSize]
+
 		if !ok {
+			var machine_type string
+
+			if strings.HasPrefix(guestSize, "shared") {
+				machine_type = "shared"
+
+			} else if strings.HasPrefix(guestSize, "performance") {
+				machine_type = "performance"
+
+			} else {
+				return machineConf, fmt.Errorf("invalid machine preset requested, '%s', expected 'shared' or 'performance'", guestSize)
+
+			}
+
 			validSizes := []string{}
 			for size := range api.MachinePresets {
-				if strings.HasPrefix(size, "shared") {
+				if strings.HasPrefix(size, machine_type) {
 					validSizes = append(validSizes, size)
 				}
 			}
--- a/internal/machine/update.go
+++ b/internal/machine/update.go
@@ -18,6 +18,72 @@ func Update(ctx context.Context, m *api.Machine, input *api.LaunchMachineInput) 
 		colorize    = io.ColorScheme()
 	)
 
+	if input != nil && input.Config != nil && input.Config.Guest != nil {
+		// Check that there's a valid number of CPUs
+		var validNumCpus []int
+
+		if input.Config.Guest.CPUKind == "shared" {
+			validNumCpus = append(validNumCpus, 1, 2, 4, 8)
+
+		} else if input.Config.Guest.CPUKind == "performance" {
+			validNumCpus = append(validNumCpus, 1, 2, 4, 8, 16, 64)
+
+		}
+
+		validCpuNum := false
+
+		for _, num := range validNumCpus {
+			if num == input.Config.Guest.CPUs {
+				validCpuNum = true
+				break
+
+			}
+		}
+
+		if !validCpuNum {
+			return fmt.Errorf("invalid config: invalid number of CPUs for %s guest. Valid numbers are %v", input.Config.Guest.CPUKind, validNumCpus)
+
+		}
+
+		// Check that the amount of memory is evenly divisible by 256 MiB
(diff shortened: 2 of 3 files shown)


internal/command/machine/run.go
```diff
@@ -588,10 +588,24 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
...
+			var machine_type string
+
+			if strings.HasPrefix(guestSize, "shared") {
+				machine_type = "shared"
+
+			} else if strings.HasPrefix(guestSize, "performance") {
+				machine_type = "performance"
+
+			} else {
+				return machineConf, fmt.Errorf("invalid machine preset requested, '%s', expected 'shared' or 'performance'", guestSize)
```
DAlperin, line 602: This error is slightly misleading, we expected it to _start with_ 'shared' or 'performance' right?

internal/machine/update.go
```diff
@@ -18,6 +18,82 @@ func Update(ctx context.Context, m *api.Machine, input *api.LaunchMachineInput)
...
+			}
+		}
+
+		if !validCpuNum {
+			return fmt.Errorf("invalid config: invalid number of CPUs for %s guest. Valid numbers are %v\nView more information here: https://fly.io/docs/about/pricing/#machines", input.Config.Guest.CPUKind, validNumCpus)
+
+		}
+
+		if input.Config.Guest.CPUKind == "shared" {
+			if input.Config.Guest.MemoryMB%256 != 0 {
```
DAlperin, line 49: Channeling jp here: these can be one if, not nested :)

internal/machine/update.go
```diff
@@ -18,6 +18,82 @@ func Update(ctx context.Context, m *api.Machine, input *api.LaunchMachineInput)
...
+				return fmt.Errorf("invalid config: invalid memory size; must be in 256 MiB increment\nView more information here: https://fly.io/docs/about/pricing/#machines")
+
+			}
+
+		} else if input.Config.Guest.CPUKind == "performance" {
+			if input.Config.Guest.MemoryMB%1024 != 0 {
```
DAlperin, line 55: here too

DAlperin (approved):

internal/command/machine/run.go
billyb2: You're right, I'll fix that now

internal/machine/update.go
billyb2: roger roger

internal/machine/update.go
```diff
@@ -18,6 +18,76 @@ func Update(ctx context.Context, m *api.Machine, input *api.LaunchMachineInput)
...
+			}
+		}
+
+		if !validCpuNum {
+			return fmt.Errorf("invalid config: invalid number of CPUs for %s guest. Valid numbers are %v\nView more information here: https://fly.io/docs/about/pricing/#machines", input.Config.Guest.CPUKind, validNumCpus)
+
+		}
+
+		if input.Config.Guest.CPUKind == "shared" && input.Config.Guest.MemoryMB%256 != 0 {
+			return fmt.Errorf("invalid config: invalid memory size; must be in 256 MiB increment\nView more information here: https://fly.io/docs/about/pricing/#machines")
```
tvdfly, line 49: how about adding a suggestion like:

```diff
-			return fmt.Errorf("invalid config: invalid memory size; must be in 256 MiB increment\nView more information here: https://fly.io/docs/about/pricing/#machines")
+			return fmt.Errorf("invalid config: invalid memory size %d; must be in 256 MiB increment (%d would work)\nView more information here: https://fly.io/docs/about/pricing/#machines", input.Config.Guest.MemoryMB, input.Config.Guest.MemoryMB - (input.Config.Guest.MemoryMB%256))
```

internal/machine/update.go
```diff
@@ -18,6 +18,76 @@ func Update(ctx context.Context, m *api.Machine, input *api.LaunchMachineInput)
...
+				maxMemory = input.Config.Guest.CPUs * api.MAX_MEMORY_MB_PER_CPU
+			}
+
+			if input.Config.Guest.MemoryMB > maxMemory {
+				return fmt.Errorf("invalid config: for machines with %d CPUs, the maximum amount of memory is %d MiB\nView more information here: https://fly.io/docs/about/pricing/#machines", machinePreset.CPUs, maxMemory)
+
+			}
+
+		} else {
+			// this shouldn't happen
```
tvdfly, line 85: if getting here is a bug, reporting to sentry is a good way to ensure we see that the bug is happening.

internal/machine/update.go
billyb2: That's a great idea, I'll add this in a new PR

internal/machine/update.go
billyb2: Actually now that you point this out, this shouldn't ever happen, because of previous checks. I'll make a new PR that takes that into account.

PR #1763, opened 2023-02-22
`fly apps restart`: tag a release when restarting AppsV2 apps

alichay:
This makes `fly apps restart` tag a release for V2 apps, like would happen during a deploy, or when a nomad app was restarted.

This would make the move from nomad -> machines more familiar, but it would also act as a debugging tool; sometimes, things just start acting differently after a reboot, and keeping track of this can be helpful. That said, there could be an argument made that maybe piggybacking off of the release system to give users this info isn't the right approach.

Regardless, I already had this branch finished, so it might be helpful to have a proof of concept working :)

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -501,7 +501,7 @@ func (md *machineDeployment) setStrategy(passedInStrategy string) error {
 	return nil
 }
 
-func (md *machineDeployment) createReleaseInBackend(ctx context.Context) error {
+func CreateReleaseInBackend(ctx context.Context, client *api.Client, app *appv2.Config, strategy, image string) (*gql.MachinesCreateReleaseResponse, error) {
 	_ = `# @genqlient
 	mutation MachinesCreateRelease($input:CreateReleaseInput!) {
 		createRelease(input:$input) {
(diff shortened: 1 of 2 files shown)


internal/command/deploy/machines.go
```diff
@@ -544,7 +544,7 @@ func (md *machineDeployment) setStrategy(passedInStrategy string) error {
 	return nil
 }
 
-func (md *machineDeployment) createReleaseInBackend(ctx context.Context) error {
+func CreateReleaseInBackend(ctx context.Context, client *api.Client, app *appv2.Config, strategy, image string) (*gql.MachinesCreateReleaseResponse, error) {
```
tvdfly, line 547: how about moving this to the api package, since we're making it more general use?

tvdfly: I usually try to avoid naming things `app`, since we have an `app` package that gets used in a bunch of places. Not a big deal here, but that's why we have appConfig in a bunch of place.

alichay: moving it to the `api` package solves the second one cleanly too! :)

api/resource_releases.go
```diff
@@ -105,3 +108,31 @@ func (c *Client) GetAppReleaseNomad(ctx context.Context, appName string, id stri
 
 	return data.App.Release, nil
 }
+
+func (c *Client) CreateReleaseMachines(ctx context.Context, input CreateReleaseInput) (*CreatedRelease, error) {
+	query := `
```
tvdfly, line 113: thanks for moving this here! No need to make a change; this is just an fyi...

It's ok to use the genqlient code in the api package, too. It's the newer way to do graphql queries and avoids the tedium of having to write out input/output the structs by hand. We haven't backported genqlient to all the existing gql queries that use this pattern (and likely won't backport them).

No change needed here. Next time, feel free to use genqlient though!

alichay: Good to know! I searched for it, and found literally nowhere else using it so I figured I might as well port it over to how everything else does it.

Only now realizing that I searched for `genqclient` not `genqlient`!

tvdfly:
I get an error. I think you need to merge master to fix the restart call?

```
% fly apps restart tvd-test-apps-v2
Restarting machine 732874435c9685
Error could not stop machine 732874435c9685: failed to restart VM 732874435c9685: machine ID 732874435c9685 lease currently held by t@fly.io, expires at 2023-03-09T22:37:09Z
```

PR #1764, opened 2023-02-23
`fly apps releases`: use new GraphQL queries for AppsV2 apps

alichay:
use `releasesUnprocessed`/`releaseUnprocessed` queries for AppsV2 apps

### DIFF
--- a/api/resource_releases.go
+++ b/api/resource_releases.go
@@ -105,31 +105,3 @@ func (c *Client) GetAppReleaseNomad(ctx context.Context, appName string, id stri
 
 	return data.App.Release, nil
 }
-
-func (c *Client) GetAppReleaseMachines(ctx context.Context, appName string, id string) (*Release, error) {
-	query := `
-		query ($appName: String!, $releaseId: ID!) {
-			app(name: $appName) {
-				release: releaseUnprocessed(id: $releaseId) {
-					id
-					status
-					evaluationId
-					createdAt
-					deploymentStrategy
-				}
-			}
-		}
-	`
-
-	req := c.NewRequest(query)
-
-	req.Var("appName", appName)
-	req.Var("releaseId", id)
-
-	data, err := c.RunWithContext(ctx, req)
-	if err != nil {
-		return nil, err
-	}
-
-	return data.App.Release, nil
-}
--- a/api/resource_releases.go
+++ b/api/resource_releases.go
@@ -2,7 +2,7 @@ package api
 
 import "context"
 
-func (c *Client) GetAppReleases(ctx context.Context, appName string, limit int) ([]Release, error) {
+func (c *Client) GetAppReleasesNomad(ctx context.Context, appName string, limit int) ([]Release, error) {
 	query := `
 		query ($appName: String!, $limit: Int!) {
 			app(name: $appName) {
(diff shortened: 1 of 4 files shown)


api/resource_releases.go
```diff
@@ -40,7 +40,45 @@ func (c *Client) GetAppReleases(ctx context.Context, appName string, limit int)
 	return data.App.Releases.Nodes, nil
 }
 
-func (c *Client) GetAppRelease(ctx context.Context, appName string, id string) (*Release, error) {
+func (c *Client) GetAppReleasesMachines(ctx context.Context, appName string, limit int) ([]Release, error) {
+	query := `
+		query ($appName: String!, $limit: Int!) {
+			app(name: $appName) {
+				releases: releasesUnprocessed(first: $limit) {
```
tvdfly, line 47: nice! I didn't know graphql supported renaming the field. TIL

api/resource_releases.go
```diff
@@ -67,3 +105,31 @@ func (c *Client) GetAppRelease(ctx context.Context, appName string, id string) (
 
 	return data.App.Release, nil
 }
+
+func (c *Client) GetAppReleaseMachines(ctx context.Context, appName string, id string) (*Release, error) {
+	query := `
+		query ($appName: String!, $releaseId: ID!) {
+			app(name: $appName) {
+				release: releaseUnprocessed(id: $releaseId) {
```
tvdfly, line 113: does this one work? I don't think I wired that up. it would be easy to do so.

internal/command/apps/releases.go
```diff
@@ -47,9 +47,23 @@ including type, when, success/fail and which user triggered the release.
...
+		releases []api.Release
+	)
+
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("failed retrieving app %s: %w", appName, err)
+	}
+
+	if app.PlatformVersion == "machines" {
+		releases, err = client.GetAppReleasesMachines(ctx, appName, 25)
```
tvdfly, line 62: please double check that the ordering is the same. this *might* return the first 25 releases instead of the most recent 25. we can fix it here or in web (probably makes sense to do in web if they are different).

api/resource_releases.go
alichay: me neither! just wanted to find a way to _not_ cram new stuff into the huge query data struct haha

api/resource_releases.go
alichay: d'oh, you're right. I saw the data type in the graphql schema update, then the symmetry in the request code and it never even occurred to me.

you could wire it up or I could just remove this code path for now. it's unused so either's fine imo

I think it might be wise to just hook it up, but if that's more than seriously like a 15 minute job then maybe that's unnecessary

internal/command/apps/releases.go
alichay: desired behavior is *most recent* 25, right? (this is what it does)

api/resource_releases.go
tvdfly: oh if it's not used, how about removing from here? thx!

internal/command/apps/releases.go
tvdfly: yep, most recent 25 is desired behavior. awesome!

alichay:
ok, i think this is looking good now!

tvdfly (approved):

PR #1765, opened 2023-02-23
Support [[mounts]] in fly.toml

dangra:
Seen at https://community.fly.io/t/problem-exposing-postgresql-database-to-external-world/10869/7?u=dangra


gwuah (approved):
LGTM

PR #1768, opened 2023-02-23
Fixed a crash when saying no to creating an app

### LINKED ISSUE #1705: `fly m run` should stop gracefully if we say "N" to creating a new app
Tested in the Apps v2 branch.

If there's no `fly.toml` in the directory and I say `N` to creating a new app, flyctl stops with an error. This is almost ok since you can't go very far with no app specified, but should get cleaned up to either stop with an informative error or move toward specifying an existing app.

With v0.0.455-pre-1:

```
$ LOG_LEVEL=debug fly m run . --region yul
...
Update available 0.0.455-pre-1 -> v0.0.456.
Run "fly version update" to upgrade.
DEBUG client initialized.
DEBUG no app config found at /Users/chris/FlyTests/feb9/log-hello/fly.toml; skipped.
? Running a machine without specifying an app will create one for you, is this what you want? No
```
flyctl built Feb 12 from the apps-v2 branch

```
$ LOG_LEVEL=debug DEV=1 ~/flyctl/bin/flyctl m run . --region yul
...
? Running a machine without specifying an app will create one for you, is this what you want? No
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 pc=0x1038c7e78]
goroutine 1 [running]:
github.com/superfly/flyctl/internal/command/machine.runMachineRun({0x10405b728, 0x14000e46570})
        /Users/chris/flyctl/internal/command/machine/run.go:206 +0x328
...
```

At the time, line 206 was the second of these:
```
	input := api.LaunchMachineInput{
		AppID:  app.Name,
```

billyb2:
Specifically when running running `machine run`

Should fix #1705

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -180,6 +180,11 @@ func runMachineRun(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+
+		if app == nil {
+			return nil
+		}
+
 	} else {
 		app, err = client.GetAppCompact(ctx, appName)
 		if err != nil && strings.Contains(err.Error(), "Could not find App") {


tvdfly (approved):
Does this show an error message when the createApp() doesn't work? If not, how about adding one?

billyb2:
@tvdfly it does, yea

PR #1777, opened 2023-02-24
Don't mount volumes to releaseCommandMachine

### LINKED ISSUE #1752: fly deploy panics when mount and release command are configured
with this fly.toml:
```toml
app = "tvd-cfg-repro-2023-02-21"

[mount]
destination = "/mydata"

[deploy]
release_command = "/release_cmd.sh foo bar baz 123 --six"
```

`fly deploy` will panic with these commands:

```
% fly version
fly v0.0.463 darwin/arm64 Commit: be3d531e BuildDate: 2023-02-21T16:42:54Z

% fly apps create tvd-cfg-repro-2023-02-21
New app created: tvd-cfg-repro-2023-02-21

% fly vol create -r sea --size 1 tvd_cfg_repro_2023_02_21_vol
        ID: vol_1q85vg72yz1vzdxe
...

% curl -vs -X POST "https://api.machines.dev/v1/apps/tvd-cfg-repro-2023-02-21/machines" \
    -H "Authorization: Bearer ${FLY_API_TOKEN}" \
    -H "content-type: application/json" \
    --data-binary '{
    "region": "sea",
    "config": {
        "image": "nginx",
        "mounts": [{
            "volume": "vol_1q85vg72yz1vzdxe",
            "path": "/testdata"
        }]
    }
}'

% fly deploy --image nginx
==> Verifying app config
--> Verified app config
==> Building image
image found: img_19gm46k36kzvx0jk
WARN Found 1 machines that are not part of the Fly Apps Platform:
tvd-cfg-repro-2023-02-21 machines
ID              NAME                    STATE   REGION  IMAGE                   IP ADDRESS                      VOLUME                      CREATED                 LAST UPDATED
39080e6e2c3487  aged-glitter-9049       started sea     library/nginx:latest    fdaa:0:bfcd:a7b:f9:fb09:1c2a:2  vol_1q85vg72yz1vzdxe        2023-02-22T06:44:21Z    2023-02-22T06:44:27Z

? Migrate 1 existing machines into Fly Apps Platform? Yes
INFO Migrating 1 machines to the Fly Apps Platform
Running tvd-cfg-repro-2023-02-21 release_command: /release_cmd.sh foo bar baz 123 --six
```

if I run with a dev build, i get a panic.

This seems to be happening because we try to attach a volume to the release command machine and there is no volume to attach. We shouldn't attach volumes to release commands, at least for now to fix this panic.

billyb2:
From my understanding after reading the code, lunchInput is just the machine used for the release command. We aren't supposed to mount volumes to them anyway, so this should be fine. Should fix #1752

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -624,11 +624,6 @@ func (md *machineDeployment) resolveUpdatedMachineConfig(origMachineRaw *api.Mac
 
 	if origMachineRaw.Config.Mounts != nil {
 		launchInput.Config.Mounts = origMachineRaw.Config.Mounts
-	} else if md.appConfig.Mounts != nil {
-		launchInput.Config.Mounts = []api.MachineMount{{
-			Path:   md.volumeDestination,
-			Volume: md.volumes[0].ID,
-		}}
 	}
 
 	if len(launchInput.Config.Mounts) == 1 && launchInput.Config.Mounts[0].Path != md.volumeDestination {


internal/command/deploy/machines.go
```diff
@@ -624,11 +624,6 @@ func (md *machineDeployment) resolveUpdatedMachineConfig(origMachineRaw *api.Mac
 
 	if origMachineRaw.Config.Mounts != nil {
 		launchInput.Config.Mounts = origMachineRaw.Config.Mounts
-	} else if md.appConfig.Mounts != nil {
```
tvdfly, line 627: I think this will work? resolveUpdatedMachineConfig() is used in the regular deployment as well, but those should all hit the first branch of the if statement (and I'm pretty certain we validate all the machines have mounts before getting to this part).

I'd recommend making some machines with volumes, deploying, adding a release command to the fly.toml and ensuring it still works. This script might be useful for prepping the volumes and machines:

```bash
#!/usr/bin/env bash
set -eo pipefail
set -x
FLY_API_BASE=https://api.machines.dev
app=tvd-test-apps-v2
volname=tvd_appsv2_test_vol
tag=test-machine-$(date '+%Y-%m-%dT%H%M%S')
count=1
for reg in $(fly platform regions --json | jq -r '.[].Code' | grep -v fra | grep -v maa | shuf -n ${count}) ; do
    vid=$(fly vol create -r ${reg} --size 1 --json ${volname} | jq -r '.id')
    curl -vs --fail -X POST "${FLY_API_BASE}/v1/apps/${app}/machines" \
    -H "Authorization: Bearer ${FLY_API_TOKEN}" \
    -H "content-type: application/json" \
    --data-binary '{
    "region": "'${reg}'",
    "config": {
        "image": "nginx",
        "guest": {
            "cpu_kind": "shared",
            "cpus": 1,
            "memory_mb": 512
        },
        "services": [{
            "protocol": "tcp",
            "internal_port": 80,
            "concurrency": {
                "type": "requests",
                "soft_limit": 2000,
                "hard_limit": 2500
            },
            "ports": [{
                "port": 443,
                "handlers": ["tls", "http"]
            }]
        }],
        "checks": {
            "vm": {
                "type": "http",
                "port": 80,
                "method": "GET",
                "path": "/",
                "interval": "15s",
                "timeout": "10s"
            }
        },
        "mounts": [{
            "volume": "'${vid}'",
            "path": "/testdata"
        }]
    }
}'
done

```

billyb2: Hi tvd,
 So I made a new fly app using the hello-fly template. I used the script above to create a machine with an attached volume. When i run fly deploy, it asks me to confirm converting the existing machine to the fly apps platform (if i say no it actually crashes flyctl, but that's gonna be a separate issue when I get around to it). If i say yes, the app deploys successfully. The release command does seem to run as well. If I try running with `echo 'hi'` as the command, it runs successfully, but if I use a command that will purposely fail (running `exit 1` as an example), the deployment fails. The machines are also correctly mounted to the volumes.

Here's the app config I used:
[hello-fly.zip](https://github.com/superfly/flyctl/files/10841884/hello-fly.zip)

tvdfly (approved):

PR #1778, opened 2023-02-24
Fixed crash when saving config

billyb2:
When an AppV2 has no machines, then it won't be able to load the config, since there aren't any machines to load it from.

Ignore the name of the branch, it was originally meant to fix a different error, but I encountered this one in the mean time.
More discussion here:
https://flyio.discourse.team/t/internal-error-setting-secret-for-machines-apps/1932/2


tvdfly:
Thank you!

PR #1780, opened 2023-02-24
Fixed a crash with setting secrets on not deployed machines apps

billyb2:


DAlperin (approved):

billyb2:
Not a crash, just a 500 on web bc we're trying to do something that doesn't make sense.

PR #1792, opened 2023-02-27
Set releaseCommandMachine to an empty list by default

billyb2:
By default, releaseCommandMachine isn't set at all. If you have a machine with mounts, but the mounts aren't set in fly.toml, for whatever reason, releaseCommandMachine is never populated, and as such, deploys crash. An easy fix is to just set it to an empty list by default.

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -139,6 +139,7 @@ func NewMachineDeployment(ctx context.Context, args MachineDeploymentArgs) (Mach
 		leaseTimeout:               leaseTimeout,
 		leaseDelayBetween:          leaseDelayBetween,
 		releaseCommand:             releaseCmd,
+		releaseCommandMachine:      machine.NewMachineSet(flapsClient, io, nil),
 	}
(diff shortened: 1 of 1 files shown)


tvdfly (commented):
What fails if releaseCommandMachine is nil?

billyb2:
@tvdfly I don't know why I wrote this so vaguely, lol. Here are the steps to reproduce (with a zip file of the directory I'm using linked below).
1. Create a volume
2. Create a machine that has said volume mounted to it
3. Ensure that your fly.toml does *not* have the volume mounted
4. Create the app, but don't deploy (in this case I used fly launch --force-machines)
5. Deploy the app, but say no to migrating the machine to the fly apps platform
6. In this case, the crash happens at `internal/command/deploy/machines.go:375` when calling `IsEmpty()`, since releaseCommandMachine is nil

billyb2:
Oh, I saw that we stopped asking users about migrating machines. That also fixes this bug in a sense, lol. I'll close this for now, since it no longer has a purpose

Issue #1770, opened 2023-02-23
fly machine run doesn't respect `[build]` block in `fly.toml`


mrkurt:
https://community.fly.io/t/flyctl-machine-run-does-not-respect-build-dockerfile-in-config/10961

---
My config toml has

```
[build]
dockerfile = "Containerfile.foo"
```

Relying on config fails:

```
$ flyctl machine run . -c fly-foo.toml --name bar --schedule hourly
Error app does not have a Dockerfile or buildpacks configured. See https://fly.io/docs/reference/configuration/#the-build-section
```

Specifying the containerfile explicitly works:

```
$ flyctl machine run . -c fly-foo.toml --dockerfile Containerfile.foo --name bar --schedule hourly
```

PR #1793, opened 2023-02-27
consider dockerfile set in fly.toml when running machines

gwuah:
this fixes https://github.com/superfly/flyctl/issues/1770

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -340,6 +340,7 @@ func determineImage(ctx context.Context, appName string, imageOrPath string) (im
 	var (
 		client = client.FromContext(ctx).API()
 		io     = iostreams.FromContext(ctx)
+		cfg    = app.ConfigFromContext(ctx)
 	)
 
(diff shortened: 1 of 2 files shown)


dangra (approved):

internal/command/machine/run.go
```diff
@@ -336,10 +336,18 @@ func parseKVFlag(ctx context.Context, flagName string, initialMap map[string]str
 	return parsed, nil
 }
 
+func safelyGetDockerFilePathFromConfig(cfg *app.Config) string {
```
tvdfly, line 339: How about making this a method on `*app.Config`, so people could call `cfg.GetDockerfile()` and it will return a string with the path (or an empty string if unset)? I could see this being used in other places.

tvdfly (approved):

internal/command/machine/run.go
gwuah: A similar method already existed on appConfig so I just used that.

PR #1794, opened 2023-02-28
Unhide `--image-ref`

catflydotio:
Since it's part of our current instructions for using Timescale, it's better if `fly pg create -h` shows it.


davissp14 (approved):
Thanks!

PR #1795, opened 2023-02-28
Fix `fly launch` not saving the chosen region

### LINKED ISSUE #1798: Apps V2 - Region mistmatch between app region pool and first machine
So @catflydotio [mentioned in the forum](https://community.fly.io/t/fly-apps-on-machine-prerelease/10219/25?u=pier) that UX confusions are fair game for Github issues.

When creating a new v2 app with `fly launch --force-machines` the app region selection is still shown but then the machine itself is not created in that region.

I created a v2 app in QRO but the initial machine was created in DFW which is confusing.

@catflydotio clarified that machines don't use the region pool. Makes sense but as a user I don't understand why the first machine was created in DFW. Not only I actually live in QRO but I selected QRO as the region for my app.

What I think would make more sense is that the user should have the option to decide where the first machine will be created. 

The first machine could be deployed in the same region that was selected in the app region selection (even though machines ignore the region pool). Or maybe there should be a new menu afterwards just for that first machine?

Also, when using `fly launch --force-machines` what's the point of having an app region pool at all if all VMs will be running with machines?

alichay:
The primary_region should be saved in `fly.toml` for v2 apps. I don't believe this is part of the V1 `fly.toml` file, so there's some weird special casing going on here.

todo: 
- does this affect `fly config save`?
  - should it?
- does this break anything else?

(it's possible that other places that read fly.toml won't like that it's now perfectly OK for it to have `primary_region`. Granted, I think it was supposed to be that way already, but you can never be too safe!)

Fixes #1798

### DIFF
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -95,6 +95,12 @@ func run(ctx context.Context) (err error) {
 	colorize := io.ColorScheme()
 	workingDir := flag.GetString(ctx, "path")
 
+	deployArgs := deploy.DeployWithConfigArgs{
+		ForceNomad:    flag.GetBool(ctx, "force-nomad"),
+		ForceMachines: flag.GetBool(ctx, "force-machines"),
+		ForceYes:      flag.GetBool(ctx, "now"),
+	}
+
 	// Determine the working directory
 	if absDir, err := filepath.Abs(workingDir); err == nil {
 		workingDir = absDir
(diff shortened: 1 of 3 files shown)


tvdfly (commented):
I am slightly inclined to special case this for the appv2 package, and ignore it for app package. We really only need the region in config for appv2 (nomad gets it in other ways).  That would probably mean rewriting the fly.toml after we deserialize appv2 the first time, and I think that's ok.

alichay:
> I am slightly inclined to special case this for the appv2 package, and ignore it for app package. We really only need the region in config for appv2 (nomad gets it in other ways). That would probably mean rewriting the fly.toml after we deserialize appv2 the first time, and I think that's ok.

Thankfully I didn't have to change anything deep in appsv2, it already handled that! I avoided doing it this way because it seemed like a hassle finding a way to get the app config into a v2 app config without regressing UX by making it spam file written everywhere, so thanks for the FromDefinition tip!

internal/command/launch/launch.go
```diff
@@ -535,9 +537,28 @@ func run(ctx context.Context) (err error) {
...
+	var v2AppConfig *appv2.Config
+	if deployArgs.ForceMachines {
+
+		v2AppConfig, err = appv2.FromDefinition(api.DefinitionPtr(appConfig.Definition))
+		if err != nil {
+			return fmt.Errorf("invalid config: %w", err)
+		}
+		v2AppConfig.AppName = appConfig.AppName
+
+		appConfig.PrimaryRegion = region.Code
```
tvdfly, line 550: do we need to set this here? I can't tell if it get used later on.

tvdfly (approved):
code looks good, thx! please make sure to run testflight before merging. changes to fly launch can easily break in subtle ways.

internal/command/launch/launch.go
alichay: If you meant `v2AppConfig.AppName`, that's used when the config is written to disk.
If you meant `appConfig.PrimaryRegion`, I believe it's used in deploy, if you pass `--deploy`.

PR #1796, opened 2023-03-01
Add grace period to machine checks

dangra:
Prep for flyd support


api/client.go
```diff
@@ -121,7 +122,7 @@ func GetAccessToken(ctx context.Context, email, password, otp string) (token str
 	defer func() {
 		err := res.Body.Close()
 		if err != nil {
-			terminal.Errorf("error closing response body: %v", err)
+			fmt.Println("Error closing response body: %v", err)
```
tvdfly, line 125: send to stderr?

tvdfly (approved):

api/client.go
dangra: something like c2f3440e

PR #1797, opened 2023-03-01
Remove confusing msg from fly deploy

tvdfly:
Remove the "Migrate %d existing machines into Fly Apps Platform?" confirmation from `fly deploy`.

Previously developers got that message if they were deploying to machines that that were missing the apps v2 metadata. This has just been confusing for developers with little upside.

The original idea was it would give folks a warning if they were deploying and had a heterogenous cluster that they didn't want to trample over. The reality is that `fly deploy` has always applied to all machines, so this confirmation was just confusing.

I debated adding in a check for different images/other configs that would be material, but decided it wasn't worth it. Most developers will use `fly deploy` to update all machines in their app, so we should just do that for them without prompting. Developers that need to update a subset of machines will likely be using `fly m update` or the machines api already anyway (and those are the right answer).

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -20,8 +20,6 @@ import (
 	"github.com/superfly/flyctl/internal/cmdutil"
 	"github.com/superfly/flyctl/internal/logger"
 	"github.com/superfly/flyctl/internal/machine"
-	"github.com/superfly/flyctl/internal/prompt"
-	"github.com/superfly/flyctl/internal/render"
 	"github.com/superfly/flyctl/iostreams"
 	"github.com/superfly/flyctl/terminal"
 )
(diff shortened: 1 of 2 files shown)


gwuah (approved):

internal/command/deploy/machines.go
```diff
@@ -317,49 +315,11 @@ func (md *machineDeployment) setMachinesForDeployment(ctx context.Context) error
...
-					machine.ImageRefWithVersion(),
-					machine.PrivateIP,
-					volName,
-					machine.CreatedAt,
-					machine.UpdatedAt,
-				})
-			}
-			terminal.Warnf("Found %d machines that are not part of the Fly Apps Platform:\n", len(machines))
-			_ = render.Table(iostreams.FromContext(ctx).Out, fmt.Sprintf("%s machines", md.app.Name), rows, "ID", "Name", "State", "Region", "Image", "IP Address", "Volume", "Created", "Last Updated")
-			if !md.autoConfirmAppsV2Migration {
```
gwuah, line 347: Should we remove the `autoConfirmAppsV2Migration` field from the machine deployment struct?

tvdfly: nice catch, thx! will remove

PR #1799, opened 2023-03-01
deprecate `fly domains`, remove `fly domains register`

### LINKED ISSUE #1791: flyctl domains should be deprecated
We recently had a forum user [wondering](https://community.fly.io/t/domain-registration-is-not-enabled-on-this-organization/11031) why `flyctl domains` wouldn't work for them. The reason is that it doesn't work for anybody but the people who tested the feature as a preview.

`fly domains` is removed from docs and hidden in flyctl's list of commands we don't hate, but it's still not deprecated. I can get a couple of steps into trying to register a domain without any indication that it's not an available feature.

alichay:
Would fix #1791

### DIFF
--- a/cmd/domains.go
+++ b/cmd/domains.go
@@ -160,7 +160,5 @@ func runDomainsCreate(cmdCtx *cmdctx.CmdContext) error {
 
 func runDomainsRegister(_ *cmdctx.CmdContext) error {
 
-	fmt.Println("This command is no longer supported.")
-
-	return nil
+	return fmt.Errorf("This command is no longer supported.\n")
 }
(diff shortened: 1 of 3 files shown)


cmd/domains.go
```diff
@@ -157,68 +158,9 @@ func runDomainsCreate(cmdCtx *cmdctx.CmdContext) error {
...
-	_, err = cmdCtx.Client.API().CreateAndRegisterDomain(org.ID, name)
-	if err != nil {
-		return err
-	}
+func runDomainsRegister(_ *cmdctx.CmdContext) error {
 
-	fmt.Println("Registration started")
+	fmt.Println("This command is no longer supported.")
 
 	return nil
```
tvdfly, line 165: I think if we return an error here that will cause flyctl to exit with a non-zero failed exit code, which might be nice in this case (since we are in effect failing to register).

tvdfly (approved):

Issue #1760, opened 2023-02-22
fly machine run with a schedule seems to hang


alichay:
### **Describe the bug**

Calling `fly machine run` and passing the machine a schedule causes flyctl to wait a seemingly indefinite time.
It _might_ be waiting for the scheduled event to pass? Regardless, this can't be right...

To reproduce, just run `fly m run . -a [app name] --schedule hourly`

For what it's worth, I'm running these commands in a directory with [these files](https://gist.github.com/alichay/f3d4e102c66d8ea6ad3e0bff9e1d27cf).

### **Command output:**

```text
--> Pushing image done
Image: registry.fly.io/[app name]:deployment-[deployment id]
Image size: 41 MB

Success! A machine has been successfully launched in app [app name], waiting for it to be started
 Machine ID: [machine id]
 Instance ID: [instance id]
 State: created
```
but then it just hangs.

With `LOG_LEVEL=debug`, it says
```
Success! A machine has been successfully launched in app [app name], waiting for it to be started
 Machine ID: [machine id]
 Instance ID: [instance id]
DEBUG --> GET http://[fdaa:1:3c2e::3]:4280/v1/apps/test-appsv2/machines/[machine id]/wait?instance_id=[instance id]&state=started&timeout=60

DEBUG <-- 408 http://[fdaa:1:3c2e::3]:4280/v1/apps/test-appsv2/machines/[machine id]/wait?instance_id=[instance id]&state=started&timeout=60 (56.95s)

{
  "error": "machine failed to reach desired state, started, currently stopped"
}

DEBUG --> GET http://[fdaa:1:3c2e::3]:4280/v1/apps/test-appsv2/machines/[machine id]/wait?instance_id=[instance id]&state=started&timeout=60
```

This might be related to #1751

matttpt:
Scheduled machines have been updated to start immediately (whereas previously they waited for the full schedule period before starting the first time). This is what `flyctl` currently [expects](https://github.com/superfly/flyctl/blob/bec833f4ab258f18c29f1ff5d5e301801ec6bea0/internal/command/machine/run.go#L277), so I'm thinking that that change has resolved this?

PR #1801, opened 2023-03-01
Fix `Context` error handling in `WaitForStartOrStop`

matttpt:
I ran across this today while learning about/experimenting with machines. This is my first foray into flyctl, so here goes ...

flyctl will not time out when waiting for a machine to start or stop (in `WaitForStartOrStop`). When the net/http request made in the flaps client is canceled due to `waitCtx` being canceled or timing out, it returns a `net.Error`, which does not wrap `context.Canceled` or `context.DeadlineExceeded`. As a result, the general-case error-handling runs: flyctl waits for a backoff period to expire and then tries again. This repeats forever. Furthermore, since (AFAIK) `Context` cancellation is used to handle SIGINT and SIGTERM, ^C doesn't work.

This compounds #1760.

To handle this correctly, we can check the error associated with `waitCtx` as opposed to the one returned from the flaps client.

### DIFF
--- a/internal/machine/wait.go
+++ b/internal/machine/wait.go
@@ -36,19 +36,21 @@ func WaitForStartOrStop(ctx context.Context, machine *api.Machine, action string
 	}
 	for {
 		err := flapsClient.Wait(waitCtx, machine, waitOnAction, 60*time.Second)
+		if err == nil {
+			return nil
+		}
+
 		switch {
-		case errors.Is(err, context.Canceled):
+		case errors.Is(waitCtx.Err(), context.Canceled):
 			return err
-		case errors.Is(err, context.DeadlineExceeded):
+		case errors.Is(waitCtx.Err(), context.DeadlineExceeded):
 			return fmt.Errorf("timeout reached waiting for machine to %s %w", waitOnAction, err)
-		case err != nil:
+		default:
 			var flapsErr *flaps.FlapsError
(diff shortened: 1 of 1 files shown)


tvdfly (approved):
code seems reasonable, thx!

how did you test this?

matttpt:
I ran across this while trying to reproduce #1760 and have been testing it by running `flyctl machines run --schedule hourly ...` with `LOG_LEVEL=debug`. (flyctl waits for the machine to be started, which doesn't happen.) I also made sure that `make test` passed. If I should also be dealing with testflight, let me know.

This issue still seemed odd to me, since I didn't expect the Go standard library to return errors that didn't wrap the `context` errors. Looking a little deeper, I think this is actually because the underlying HTTP transport being used is [rehttp](https://github.com/PuerkitoBio/rehttp), and in fact it's rehttp that [doesn't wrap context errors](https://github.com/PuerkitoBio/rehttp/blob/67bd845012d9ff38adec6910ae3f3d16d6d31164/rehttp.go#L341). Maybe this needs to be addressed in [api/http.go](https://github.com/superfly/flyctl/blob/master/api/http.go) (where rehttp is used) instead? I poked around and there's a few other spots checking for `context.Canceled` or `context.DeadlineExceeded` that I'm guessing might be affected ...

dangra (approved):

PR #1803, opened 2023-03-02
add autostart machines option

senyosimpson:
Adds a configuration option, `autostart_machines` that enables you to toggle on/off automatically starting machines

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -354,11 +354,12 @@ func (mp *MachinePort) HasNonHttpPorts() bool {
 }
 
 type MachineService struct {
-	Protocol     string                     `json:"protocol,omitempty" toml:"protocol,omitempty"`
-	InternalPort int                        `json:"internal_port,omitempty" toml:"internal_port,omitempty"`
-	Ports        []MachinePort              `json:"ports,omitempty" toml:"ports,omitempty"`
-	Checks       []MachineCheck             `json:"checks,omitempty" toml:"checks,omitempty"`
-	Concurrency  *MachineServiceConcurrency `json:"concurrency,omitempty" toml:"concurrency"`
+	Protocol          string                     `json:"protocol,omitempty" toml:"protocol,omitempty"`
+	InternalPort      int                        `json:"internal_port,omitempty" toml:"internal_port,omitempty"`
+	Ports             []MachinePort              `json:"ports,omitempty" toml:"ports,omitempty"`
+	Checks            []MachineCheck             `json:"checks,omitempty" toml:"checks,omitempty"`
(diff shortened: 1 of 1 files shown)


api/machine_types.go
```diff
@@ -354,11 +354,11 @@ func (mp *MachinePort) HasNonHttpPorts() bool {
 }
 
 type MachineService struct {
-	Protocol     string                     `json:"protocol,omitempty" toml:"protocol,omitempty"`
```
senyosimpson, line 357: oh, mistake here. Let me put these `omitempty`'s back

api/machine_types.go
```diff
@@ -354,11 +354,12 @@ func (mp *MachinePort) HasNonHttpPorts() bool {
...
-	InternalPort int                        `json:"internal_port,omitempty" toml:"internal_port,omitempty"`
-	Ports        []MachinePort              `json:"ports,omitempty" toml:"ports,omitempty"`
-	Checks       []MachineCheck             `json:"checks,omitempty" toml:"checks,omitempty"`
-	Concurrency  *MachineServiceConcurrency `json:"concurrency,omitempty" toml:"concurrency"`
+	Protocol          string                     `json:"protocol,omitempty" toml:"protocol,omitempty"`
+	InternalPort      int                        `json:"internal_port,omitempty" toml:"internal_port,omitempty"`
+	Ports             []MachinePort              `json:"ports,omitempty" toml:"ports,omitempty"`
+	Checks            []MachineCheck             `json:"checks,omitempty" toml:"checks,omitempty"`
+	Concurrency       *MachineServiceConcurrency `json:"concurrency,omitempty" toml:"concurrency"`
+	AutostartMachines bool                       `json:"autostart_machines,omitempty" toml:"autostart_machines,omitempty"`
```
tvdfly, line 362: should the default be off or on? I think the default is "on" today (before this change), and I imagine we want to preserve that.

If so, how about making this `NoAutostartMachines`? that way the default when unset will be `false` (golang's default for bool type).

```suggestion
	NoAutostartMachines bool                       `json:"no_autostart_machines,omitempty" toml:"no_autostart_machines,omitempty"`
```

senyosimpson: Updating to work in line with changes in `web` (see comment here that's inline with your thinking https://github.com/superfly/web/pull/1439#issuecomment-1428315679)

tvdfly (approved):
LGTM

For now, flyctl will always create machines without this setting (I think). We'll have someone update `fly deploy` and whatnot to start setting it from fly.toml/etc and add support for `fly machine ...` sub cmds.

senyosimpson:
No problem! I have to update some other code bases before it comes into full effect in any case

PR #1739, opened 2023-02-20
Added patch for services.internal_port

billyb2:

PR #1809, opened 2023-03-02
Changed Where Configs Are Grabbed From For Appsv2

billyb2:
While working on #1739 , @tvdfly ran into a weird bug where the change didn't seem to be working at all. After some digging, we found out that it's because, for apps v2, we're loading a config from the nomad backend. It was decided that retrieving the config for apps v2 from release records would be best.

This PR is NOT ready to be merged yet, I need to do quite a bit of testing to make sure I don't break anything. Even right now I can name one or two bugs. This is just a draft PR if anyone would like to criticize my code.


billyb2:
Testing deploys and launches:
- [x]  `launch --local-only --force-machines -o personal`
  - `ord`
  - `deploy now? Yes`
  -  [x] Deployed v2
- [x]  `launch --local-only --force-nomad -o personal`
  - `ord`
  - `deploy now? Yes`
  - [x] Deployed v1
- [x]  `deploy --local-only --force-machines`
  - [x] Got config v2 locally
  - [x] Deployed V2
- [ ]  `deploy --local-only --force-machines`
  - [ ] Got config v2 remotely
  - [ ] Deployed V2
- [ ]  `deploy --local-only --force-nomad`
  - [ ] Got config v1 remotely
  - [ ] Deployed V1
- [ ]  `deploy --local-only --force-nomad`
  - [ ] Got config v1 locally
  - [ ] Deployed V1

billyb2:
Misc tests:
- [x] `app.NameFromContext for appsv1`
- [ ] `app.NameFromContext for appsv1`: *FAILS* currently

billyb2:
You can deploy app v2 config to nomad apps, no bueno

internal/command/deploy/deploy.go
```diff
@@ -226,6 +246,108 @@ func DeployWithConfig(ctx context.Context, appConfig *app.Config, args DeployWit
 	return err
 }
 
+func DeployWithConfigV2(ctx context.Context, appConfig *appv2.Config, args DeployWithConfigArgs) (err error) {
```
dangra, line 249: Can we move all this new V2 functions into another file (i.e. `deploy_v2.go`) ?

cmd/config.go
```diff
@@ -215,11 +215,20 @@ func writeAppConfig(path string, appConfig *flyctl.AppConfig) error {
 	return nil
 }
 
-func saveAppV2Config(ctx context.Context, apiClient *api.Client, appCompact *api.AppCompact, path string) error {
-	appConfig, err := getAppV2ConfigFromReleases(ctx, apiClient, appCompact.Name)
+func GetRemoteAppV2Config(ctx context.Context, apiClient *api.Client, appCompact *api.AppCompact) (appConfig *appv2.Config, err error) {
```
dangra, line 218: Anything in this file is legacy, if it deserves to survive it should be moved somewhere else

internal/command/command.go
```diff
@@ -512,18 +512,33 @@ var errRequireAppName = fmt.Errorf("we couldn't find a fly.toml nor an app speci
 // application name via command line arguments, the environment or an application
 // config file (fly.toml). It embeds LoadAppConfigIfPresent.
 func RequireAppName(ctx context.Context) (context.Context, error) {
-	ctx, err := LoadAppConfigIfPresent(ctx)
+	platform := appv2.AppsV1Platform
+	newCtx, err := LoadAppConfigIfPresent(ctx)
```
dangra, line 516: This function basically parses fly.toml into a `map[string]any` which covers pretty much everything. What would make this fail for a v2 app?

internal/command/deploy/deploy.go
```diff
@@ -107,15 +110,31 @@ func New() (cmd *cobra.Command) {
...
-		ForceNomad:    flag.GetBool(ctx, "force-nomad"),
-		ForceMachines: flag.GetBool(ctx, "force-machines"),
-		ForceYes:      flag.GetBool(ctx, "auto-confirm"),
-	})
+	if err == nil {
+		return DeployWithConfig(ctx, appConfig, DeployWithConfigArgs{
+			ForceNomad:    flag.GetBool(ctx, "force-nomad"),
+			ForceMachines: flag.GetBool(ctx, "force-machines"),
+			ForceYes:      flag.GetBool(ctx, "auto-confirm"),
+		})
```
dangra, line 119: you can take the shortcut and write this stuff like:

```go
if appConfig, err := determineAppConfig(cxt); err == nil {
  return DeployWithConfig(ctx, appConfig, ...)
} else if err != wrongAppVersionErr {
   return err
}

cfg, err := determineAppV2Config(ctx)
if err != nil {
  return err
}
return DeployWithConfigV2(ctx, cfg, ...)
```

internal/command/deploy/deploy.go
```diff
@@ -247,11 +266,41 @@ func useMachines(appConfig *app.Config, appCompact *api.AppCompact, args DeployW
 	}
 }
 
+// gets an app's name, whether its for apps v1 or v2
+func determineAppName(ctx context.Context) string {
+	appNameFromContext := app.NameFromContext(ctx)
+
+	if appNameFromContext == "" {
+		appNameFromContext = appv2.NameFromContext(ctx)
+
```
dangra, line 275: what is the point of so many blank lines?

internal/command/deploy/deploy.go
```diff
@@ -247,11 +266,41 @@ func useMachines(appConfig *app.Config, appCompact *api.AppCompact, args DeployW
...
 // determineAppConfig fetches the app config from a local file, or in its absence, from the API
 func determineAppConfig(ctx context.Context) (cfg *app.Config, err error) {
 	tb := render.NewTextBlock(ctx, "Verifying app config")
 	client := client.FromContext(ctx).API()
-	appNameFromContext := app.NameFromContext(ctx)
+
+	appNameFromContext := determineAppName(ctx)
+	appCompact, err := client.GetAppCompact(ctx, appNameFromContext)
+
```
dangra, line 288: nitpick but again, no need for a blank line right after to check the err.

internal/command/deploy/deploy.go
```diff
@@ -247,11 +266,41 @@ func useMachines(appConfig *app.Config, appCompact *api.AppCompact, args DeployW
...
+	if err != nil {
+		return nil, err
+
+	}
+
+	if appCompact.PlatformVersion == appv2.AppsV2Platform {
+		return nil, wrongAppVersionErr
+
```
dangra, line 296: I'm starting to believe this is your code editor or a anxious ENTER key 😁

internal/command/deploy/deployv2.go
```diff
@@ -0,0 +1,388 @@
...
+		// statics are not supported in Apps v2 yet
+		return false, nil
+	case args.ForceYes:
+		// if running automated, stay on nomad platform for now
+		return false, nil
+	default:
+		// choose nomad for now if not otherwise specified
+		return false, nil
+	}
+}
```
alichay, line 215: Just a note that this will probably create a merge conflict with #1802. not a problem, just making a paper trail so it's easier to track down later :)

internal/command/deploy/deployv2.go
```diff
@@ -0,0 +1,388 @@
...
+	// Start deployment of the determined image
+	client := client.FromContext(ctx).API()
+
+	release, releaseCommand, err := client.DeployImage(ctx, input)
+	if err == nil {
+		tb.Donef("release v%d created\n", release.Version)
+	}
+
+	return release, releaseCommand, err
+}
```
alichay, line 388: It's really pedantic, but CI wants there to be a \n at the end of the file I think

internal/command/deploy/deployv2.go
billyb2: Thank you :)

internal/command/deploy/deployv2.go
billyb2: Lol i saw

dangra (commented):
As this PR is a draft I will only comment on it.

tl;dr I understand the motivation but hate my deepest heart that we have to do duplicate pretty much all functions because the appConfig objects are different.

Ignore the styling comments I made, my real issue with this PR is the duplicate code but I see no other way than what you did already until we merge app configs for v1 and v2 apps.

alichay:
I skimmed some of the functions that looked copy-pasted or moved (things that I recognized), so my review isn't *too* thorough, but everything is looking pretty good.

billyb2:
> As this PR is a draft I will only comment on it.
> 
> tl;dr I understand the motivation but hate my deepest heart that we have to do duplicate pretty much all functions because the appConfig objects are different.
> 
> Ignore the styling comments I made, my real issue with this PR is the duplicate code but I see no other way than what you did already until we merge app configs for v1 and v2 apps.

I actually really hated this too. I talked about possibly *just* using app v2 config stuff, but it seems like it'd be a bad idea due to differences. Once app v1 and v2 configs got merged, this would be much cleaner.

alichay:
> hate my deepest heart that we have to do duplicate pretty much all functions because the appConfig objects are different

unfortunately the alternative is making shared functions and then surrounding every other line with `if appsv2 {} else {}` so this isn't much worse. hopefully with the movement towards v2 we can reduce a lot of the churn in the codebase and focus on v2, leaving the v1 implementations stable and not needing to go back and forth too much.

just see this as diverging and making one half of the codebase future-proof :)

internal/command/deploy/deploy.go
billyb2: Honestly a reflex from writing Rust and C code, definitely not in line w the code base's code style. I'll clean these up where I can find them.

dangra:
May I ask for a weekend delay as last chance to merge v1 and v2 appConfig parsing? If I didn't by Monday we continue on this path 🙏

billyb2:
> May I ask for a weekend delay as last chance to merge v1 and v2 appConfig parsing? If I didn't by Monday we continue on this path pray

that's totally fine!

cmd/config.go
billyb2: Do you mind if i ask where it should be moved? Maybe `internal/config/config.go`?

dangra: Oh, that is another usual confusion, we have to rename things :)

`internal/config/*` is for flyctl configuration (`~/.fly/*` files), not app configuration (aka fly.toml). 

under `internal/appv2/*`, i.e: `internal/appv2/remoteconfig.go`

tvdfly: don't worry about moving stuff around in this pr. it's easier to review changes without code moving files. we can make a separate pr for moving stuff around.

internal/command/command.go
tvdfly: What wasn't working before the changes to this file?

internal/command/deploy/deployv2.go
alichay: Now that #1802 has merged, this'll have to be updated. Here's what the new version looks like :)
https://github.com/superfly/flyctl/blob/ad1f3de704c1a01ee588b0ca9839987953d37a1d/internal/command/deploy/deploy.go#L228-L244

internal/command/command.go
billyb2: I think this was while I was trying to fix a bug with loading the config for apps v2, it doesn't actually change any behavior but it's also unnecesssary. I'll revert it for simplicity.

dangra:
@billyb2 @alichay https://github.com/superfly/flyctl/pull/1830 is ready for a review, happy to talk about it specially because it touches many places.

billyb2:
I really think my PR is ready for review. The scope of the PR was expanded a little to change where all remote configs are grabbed from for apps v2.

dangra (approved):
So good to see FromRemoteApp in use 💪

billyb2:
> So good to see FromRemoteApp in use muscle

It was super useful!

internal/appconfig/validation.go
```diff
@@ -1,14 +1,53 @@
...
+
+	apiClient := client.FromContext(ctx).API()
+	appCompact, err := apiClient.GetAppCompact(ctx, appNameFromContext)
+
+	if err != nil {
+		return err
+	}
+
+	if appCompact.PlatformVersion == MachinesPlatform {
+		return cfg.validateLocally()
```
dangra, line 26: Look at `internal/command/config/validate.go`  much of it could be moved or replaced by this. 

Also there are differences on how we validate machines configs.

dangra (changes requested):

internal/appconfig/validation.go
```diff
@@ -1,14 +1,53 @@
...
+	} else {
+		parsedCfg, err := apiClient.ParseConfig(ctx, appNameFromContext, cfg.SanitizedDefinition())
+		if err != nil {
+			return err
+		}
+
+		if !parsedCfg.Valid {
+			fmt.Println()
+			if len(parsedCfg.Errors) > 0 {
+				terminal.Errorf("\nConfiguration errors in %s:\n\n", cfg.ConfigFilePath())
```
dangra, line 37: wdyt if if we avoid terminal output in this method, and instead return an error suggesting `fly config validate` to see details.

billyb2: I totally agree, i just moved it into that function since that's how we did things in the places I got the function

internal/appconfig/validation.go
dangra: my take would be to validation into `Config.Validate()`, and let it returns enough data so runValidate can render the errors if any. Definitively not make runValidate public.

matttpt (commented):
General question: in `deploy.go` and `machines.go`, we set the `AppName` field after calling `FromRemoteApp`. In this PR, the same thing is removed from `clone.go`. What is actually responsible for setting this field?

In any case, `machines clone` worked fine for me, so it doesn't seem to be causing a problem.

internal/command/deploy/deploy.go
```diff
@@ -256,42 +255,26 @@ func determineAppConfig(ctx context.Context) (cfg *appconfig.Config, err error)
...
+			return
 		}
 
-		cfg.AppName = basicApp.Name
-
-		if err := cfg.SetPlatformVersion(basicApp.PlatformVersion); err != nil {
+		cfg.AppName = appCompact.Name
+		if err := cfg.SetPlatformVersion(appCompact.PlatformVersion); err != nil {
 			return cfg, err
 		}
```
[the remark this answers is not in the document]
billyb2: Ah you're right, I'll fix that

billyb2:
> General question: in `deploy.go` and `machines.go`, we set the `AppName` field after calling `FromRemoteApp`. In this PR, the same thing is removed from `clone.go`. What is actually responsible for setting this field?
> 
> In any case, `machines clone` worked fine for me, so it doesn't seem to be causing a problem.

I'm actually not sure that we need to do this at all (but i could be wrong, this is done everywhere in the code base). I'll fix this now.

internal/command/machine/clone.go
```diff
@@ -262,7 +262,15 @@ func getAppConfig(ctx context.Context, appName string) (*appconfig.Config, error
...
 	}
 
-	err, _ := cfg.Validate(ctx)
+	client := client.FromContext(ctx)
+	appBasic, err := client.API().GetAppBasic(ctx, appName)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg.AppName = appBasic.Name
```
dangra, line 271: Not sure we need this and it is an extra roundtrip to get the application details from Web API. 
If we want to be sure cfg returned by FromRemoteApp has a name, so set it there inside the function where it already has all the context of the requested app name

internal/command/deploy/deploy.go
```diff
@@ -267,9 +267,6 @@ func determineAppConfig(ctx context.Context) (cfg *appconfig.Config, err error)
...
 
 		cfg.AppName = appCompact.Name
-		if err := cfg.SetPlatformVersion(appCompact.PlatformVersion); err != nil {
```
dangra: good call, this is done by FromRemoteApp already

internal/command/deploy/deploy.go
```diff
@@ -267,9 +267,6 @@ func determineAppConfig(ctx context.Context) (cfg *appconfig.Config, err error)
 		}
 
 		cfg.AppName = appCompact.Name
```
dangra, line 269: this can be moved to FromRemoteApp and no need to query GetAppCompact here

internal/command/config/validate.go
```diff
@@ -31,56 +26,11 @@ ensure it is correct and meaningful to the platform.`
...
-		platformVersion = app.PlatformVersion
-	case strings.Contains(err.Error(), "Could not find App"):
-		fmt.Fprintf(io.Out, "WARNING: Failed to fetch platform version: %s\n", err)
-	default:
-		return err
-	}
+	err, extra_info := cfg.Validate(ctx)
 
-	fmt.Fprintf(io.Out, "Validating %s (%s)\n", cfg.ConfigFilePath(), platformVersion)
+	fmt.Println(extra_info)
```
dangra, line 32: I don't recall the details but we have to use `fmt.FPrintln(io.Out, ...)`  here

internal/command/deploy/deploy.go
```diff
@@ -250,48 +249,20 @@ func useMachines(ctx context.Context, appConfig *appconfig.Config, appCompact *a
...
 		}
 
-		cfg.AppName = basicApp.Name
-
-		if err := cfg.SetPlatformVersion(basicApp.PlatformVersion); err != nil {
-			return cfg, err
-		}
 	} else {
-		parsedCfg, err := client.ParseConfig(ctx, appNameFromContext, cfg.SanitizedDefinition())
+		err, _ = cfg.Validate(ctx)
```
dangra, line 263: This is looking great!  my last take on it would be that cfg validation should happen as last step before return from this function.  A line above where we print the config was verified to be valid.

billyb2: Do you mean moving the cfg.Validate line to literally right before?

dangra: I mean, before this line https://github.com/superfly/flyctl/blob/e363285490718b10adf35c87e976d503f7f3af69/internal/command/deploy/deploy.go#L289

dangra: the point is that any change we do to the config, like updating envvars, region and whatever comes next, has to be considered for validation

billyb2: oh i see what you mean! yea i understand

dangra (changes requested):
Grabbing appConfig could be an expensive operation as it may have to do external calls. Considering we only need it when `--process-group` flag is passed, I think we should move this:
 
https://github.com/superfly/flyctl/blob/29ce1a67a0597c51e739964843c578bbf6451b20/internal/command/machine/clone.go#L102-L105

inside this conditional:

https://github.com/superfly/flyctl/blob/29ce1a67a0597c51e739964843c578bbf6451b20/internal/command/machine/clone.go#L125

That is going to be another issue with FromRemoteApp() call, it needs flaps client in the context and right now it crashes because flapsClient isn't there.

billyb2:
> Grabbing appConfig could be an expensive operation as it may have to do external calls. Considering we only need it when `--process-group` flag is passed, I think we should move this:
> 
> https://github.com/superfly/flyctl/blob/29ce1a67a0597c51e739964843c578bbf6451b20/internal/command/machine/clone.go#L102-L105
> 
> inside this conditional:
> 
> https://github.com/superfly/flyctl/blob/29ce1a67a0597c51e739964843c578bbf6451b20/internal/command/machine/clone.go#L125
> 
> That is going to fix another common issue with FromRemoteApp() call, it needs flaps client in the context and right now it crashes because flapsClient isn't there.

I fully agree

dangra (commented):
It looks good! but I found a bug 🐞 

```
$ fly apps create dangra-t3 --machines -o personal
New app created: dangra-t3

$ fly m run nginx:latest -a dangra-t3
Searching for image 'nginx:latest' remotely...
image found: img_e1zd4mkn80yp02yw
Image: registry-1.docker.io/library/nginx:latest
Image size: 57 MB

Success! A machine has been successfully launched in app dangra-t3, waiting for it to be started
 Machine ID: 3287e52a027685
 Instance ID: 01GV3BVXRTD0XD370ZTKKGCG2Z
 State: created


$ fly -a dangra-t3 m clone 3287e52a027685 --process-group app
Cloning machine 3287e52a027685 into region scl
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/superfly/flyctl/internal/command/machine.runMachineClone({0x1c83168, 0xc000b93560})
        /home/daniel/src/flyctl/internal/command/machine/clone.go:138 +0x61d
github.com/superfly/flyctl/internal/command.newRunE.func1(0xc000751680, {0xc0000ae140?, 0x5?, 0x5?})
        /home/daniel/src/flyctl/internal/command/command.go:110 +0x17e
github.com/spf13/cobra.(*Command).execute(0xc000751680, {0xc0000ae0f0, 0x5, 0x5})
        /home/daniel/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0xc000b86500)
        /home/daniel/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:974 +0x3bd
github.com/spf13/cobra.(*Command).ExecuteContextC(...)
        /home/daniel/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:911
github.com/superfly/flyctl/internal/cli.Run({0x1c83da8?, 0xc0003480c0?}, 0xc000716000, {0xc00012e010, 0x7, 0x7})
        /home/daniel/src/flyctl/internal/cli/cli.go:36 +0x21e
main.run()
        /home/daniel/src/flyctl/main.go:45 +0x117
main.main()
        /home/daniel/src/flyctl/main.go:26 +0x1e
```

billyb2:
@dangra, would a nil check and a `make` be enough? I'm trying to think ofo if there's a way to always ensure that the Metadata field is allocated

dangra:
> @dangra, would a nil check and a `make` be enough? I'm trying to think ofo if there's a way to always ensure that the Metadata field is allocated

yes, that should do it.

dangra (approved):
LGTM!

PR #1813, opened 2023-03-02
hide fly restart command (replaced by fly apps restart)

catflydotio:


alichay (approved):
out with the old!

catflydotio:
(it was already out, but it was still showing up in help so this'll help close https://github.com/superfly/docs/issues/551)

PR #1814, opened 2023-03-02
`fly m update`: extra cpu and memory validation, fix invalid suggestions

### LINKED ISSUE #1806: `fly machine update` gives confusing responses for invalid values
```
lillian ~ % fly m update 591857266c3983 --memory 128

Configuration changes to be applied to machine: 591857266c3983 (teststatic-prod)

  	... // 23 identical lines
  			"cpu_kind": "shared",
  			"cpus": 1,
- 			"memory_mb": 256,
+ 			"memory_mb": 128,
  			"kernel_args": [
  				"MY_ARG=value"
  	... // 9 identical lines
  	
? Apply changes? Yes
Error invalid config: invalid memory size 128; must be in 256 MiB increment (0 would work)
View more information here: https://fly.io/docs/about/pricing/#machines
```

"0 would work"... I don't think I can give 0 MiB memory to my machine 🤔 

```
lillian ~ % fly m update 591857266c3983 --memory 0  

Error no config changes found
```

Nope, doesn't work, but this error message doesn't make much sense either.

alichay:
It used to give you a really useless error message if you specified zero CPUs or zero MB of ram, and if you specified ram in the range of 0-255MB it would suggest 0MB as an option.

Closes #1806

### DIFF
--- a/internal/flag/context.go
+++ b/internal/flag/context.go
@@ -71,6 +71,13 @@ func GetBool(ctx context.Context, name string) bool {
 	}
 }
 
+// IsSpecified returns whether a flag has been specified at all or not.
+// This is useful, for example, when differentiating between 0/"" and unspecified.
+func IsSpecified(ctx context.Context, name string) bool {
+	flag := FromContext(ctx).Lookup(name)
+	return flag != nil && flag.Changed
+}
+
 // GetOrg is shorthand for GetString(ctx, OrgName).
 func GetOrg(ctx context.Context) string {
 	return GetString(ctx, OrgName)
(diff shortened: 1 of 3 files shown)


internal/flag/context.go
```diff
@@ -71,6 +71,13 @@ func GetBool(ctx context.Context, name string) bool {
 	}
 }
 
+// IsSpecified returns whether a flag has been specified at all or not.
+// This is useful, for example, when differentiating between 0/"" and unspecified.
+func IsSpecified(ctx context.Context, name string) bool {
```
dangra, line 76: 👏

dangra (approved):

PR #1818, opened 2023-03-02
Pass environment variables to `nixpacks` with --env

matttpt:
As far as I can tell, `nixpacks` doesn't actually read environment variables for the build from the system, but rather accepts them through the `--env` option. E.g., `NIXPACKS_NO_MUSL=1 nixpacks build .` does not work, but `nixpacks build . --env NIXPACKS_NO_MUSL=1` does. This can also be observed by checking the "variables" field in the JSON emitted by `nixpacks plan`. (Finally, I took a brief look at the`nixpacks` source and didn't find it reading environment variables anywhere.)

This keeps the existing behavior (forwarding `NIXPACKS_*` environment variables to `nixpacks`), since that's what existing documentation says will work, but uses the `--env` option to do so instead.

Applicable forum thread: https://community.fly.io/t/ring-rust-dependency-build-failure-with-nixpacks/11121


dangra (approved):
nice!

jeromegn:
NIce. I wonder if this changed over time.

PR #1822, opened 2023-03-03
Fixed `config show` for apps v2

### LINKED ISSUE #1821: `fly config show` shows wrong app config
To reproduce:

* deploy a V2 app (`fly launch --force-machines`) using a custom `fly.toml` -- say change the internal port and remove health checks
* check that app is reachable at that port
* run `fly config show` 
In the displayed output, `internal_port` is 8080 and the `[[services.tcp_checks]]` stanza is back.

This looks a lot like https://github.com/superfly/flyctl/issues/1630, except that now `fly config save` works fine and it's only `fly config show` that's still a problem.

billyb2:
Should fix #1821

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -69,12 +69,7 @@ func runShowConfig(cmdCtx *cmdctx.CmdContext) error {
 			return err
 		}
 
-		definition, err := appConfig.ToDefinition()
-		if err != nil {
-			return err
-		}
-
-		cmdCtx.WriteJSON(definition)
+		cmdCtx.WriteJSON(appConfig)
 	default:
 		return fmt.Errorf("likely a bug, unknown platform version %s for app %s", appCompact.PlatformVersion, appCompact.Name)
 	}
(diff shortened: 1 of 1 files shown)


catflydotio (approved):

cmd/config.go
```diff
@@ -47,15 +47,38 @@ func newConfigCommand(client *client.Client) *Command {
...
+	case "machines":
+		appConfig, err := getAppV2ConfigFromReleases(ctx, apiClient, appCompact.Name)
+		if appConfig == nil {
+			appConfig, err = getAppV2ConfigFromMachines(ctx, apiClient, appCompact)
+		}
+		if err != nil {
+			return err
+		}
+
+		definition, err := appConfig.ToDefinition()
```
dangra, line 72: no need to convert the config into a definition, it should marshal fine when called as `cmxCtx.WriteJSON(appConfig)`

dangra (approved):

Issue #1710, opened 2023-02-14
`fly secrets import` doesn't allow spaces in values


grantnorwood:
## Describe the bug

When using `fly secrets import`, secret values containing spaces cause the command to fail.  This seems highly related to https://github.com/superfly/flyctl/issues/589.

_Questions:_
- Does the `fly secrets import` simply not support spaces in values?
- Should we loop through the [`secretsArray`](https://github.com/superfly/flyctl/blob/master/internal/command/secrets/import.go#L55) and be more flexible about properly handling spaces & double quotes?
- _**Do you need a volunteer to fix this?**_  I'm not a go developer, but I'm happy to try!

* Operating system: mac, linux
* `fly version`: `v0.0.456 darwin/amd64`

** Paste your `fly.toml`

```toml
# paste your config file here
```

## Command output

I attempted several ways to use spaces in values when importing secrets, with various newlines & quotes, and they all failed in similar ways.  The example below of munged secret values includes the output from a GitHub Action.

### Yaml

```yaml
secrets: |-
  MY_ENVIRONMENT="${{ env.MY_ENVIRONMENT }}" # = staging
  TEST_VAR_1="${{ env.TEST_VAR_1 }}" # = test value 1
```

### GitHub Action shell script

```shell
# input passed from yaml `secrets`
if [ -n "$INPUT_SECRETS" ]; then
  echo $INPUT_SECRETS | flyctl secrets import --app "$app"
fi
```

### `flyctl` output
```shell
+ flyctl secrets import --app some-app
+ echo 'MY_ENVIRONMENT="staging"' 'TEST_VAR_1="test' value '1"'
```

The way that the current cli version works by attempting to automatically surround values in quotes is not working particularly well for me, and seems unable to support spaces in values.

I am, however, able to successfully import secrets if I don't use any spaces or double quotes at all.  But that is specifically breaking a very common use case.

Am I simply using this incorrectly?  I forked the repo to try and spin up the cli, but I'm not 💯 sure what I'm doing 🙂  If there are some helpful developer docs for getting started, I'd love some links.  The `building.md` didn't seem to have what I'm looking for, and I wasn't sure if I needed to run the `install.sh` before building the docker image, do I need to rebuild and restart after each change, etc ... 

Thanks, y'all!

tvdfly:
Thanks for reporting this issue! This looks like the same issue as #589.

[...]

PR #1824, opened 2023-03-03
Fix secrets import on files with \r\n as newlines

dangra:
I faced this issue while trying to port an app to V2 by exporting its secrets using `env` and importing with `fly secrets import <envfile`

### DIFF
--- a/internal/command/secrets/import.go
+++ b/internal/command/secrets/import.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
-	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/client"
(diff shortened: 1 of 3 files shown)


alichay (approved):
this looks good and with a little bit of spot checking it doesn't seem to regress.

i actually got the `master` flyctl to panic with a fairly trivial testcase that had nothing to do with `\r\n`, so this will be pretty nice to have!

alichay:
As an alternative, this could possibly use [go-envparse](https://pkg.go.dev/github.com/hashicorp/go-envparse)?

It's mentioned by T [here](https://github.com/superfly/flyctl/issues/1710#issuecomment-1429548100) as a possibility and it would also solve `export X=Y` and `X="quoted string"`

dangra:
> As an alternative, this could possibly use [go-envparse](https://pkg.go.dev/github.com/hashicorp/go-envparse)?

go-envparse looks very good indeed, not sure if it is possible to avoid the interpolation part. Do we want that?

dangra:
we also have the weird case of triple `"` for multiline values but should be possible to ditch them. I also wanted to add support for null terminated lines so it is easy to import files generated by `env -0`

alichay:
> not sure if it is possible to avoid the interpolation part. Do we want that?

Probably not?

> we also have the weird case of triple `"` for multiline values but should be possible to ditch them. I also wanted to add support for null terminated lines so it is easy to import files generated by `env -0`

Yep, makes sense!

PR #1827, opened 2023-03-03
Test stringified service internal port

dangra:


alichay:
What's the reason for removing `internal/appv2/testdata/services.toml`?

billyb2 (approved):
Other than removing the file, LGTM 👍🏾

dangra:
> What's the reason for removing internal/appv2/testdata/services.toml?

@alichay it is not used in tests. My initial attempt was to add a test case there but somehow I added that file and not use it. The regular services format is well tested in full-reference.toml

Issue #1820, opened 2023-03-02
Apps V2 - `fly launch --image` doesn't fill in `[build]` section in `fly.toml`


catflydotio:
`flyctl v0.0.0-1677791596+dev darwin/arm64 Commit: 3c041ba5dcb5b8966fb2e8b5c62287991f648097 (master) BuildDate: 2023-03-02T21:13:16Z`

This prevents `fly deploy`.

To reproduce:

`$ ~/flyctl/bin/flyctl launch --image nginx --force-machines` 

`fly.toml` will be generated but without any `[build]` section.

Optional step: try to deploy

`$ ~/flyctl/bin/flyctl deploy --force-machines`

Get error `Error failed to fetch an image or build from source: app does not have a Dockerfile or buildpacks configured. See https://fly.io/docs/reference/configuration/#the-build-section`

dangra:
Fixed by #1830 (https://github.com/superfly/flyctl/pull/1830#discussion_r1128076350)

PR #1830, opened 2023-03-03
Work towards unifying fly.toml parsing

dangra:
The whole point of this PR is to parse fly.toml into a single struct called Config under `internal/appv2` that we can use for v1 and v2 apps.

In the case of a parsing error because there are so many unknowns variations of fly.toml, there is always the option to access `Config.RawDefinition` field that keeps a pristine `map[string]any` with config content.

An error to parse fly.toml won't cause a hard failure, instead the error is stored at `Config.parseError` for further reference. This is important because we don't want to hard fail for V1 Apps, `fly launch` and `fly deploy` will use RawDefinition to interact with Nomad apps.

When deploying Machines apps, it is important to check this error as it means the configuration file is invalid.

TODO:
- [x] Remove LoadAppV2ConfigIfPresent
- [x] Check parseError before attempting deploy on V2
- [x] Fix https://github.com/superfly/flyctl/issues/1820


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/briandowns/spinner v1.19.0
 	github.com/buildpacks/pack v0.21.0
+	github.com/cavaliergopher/grab/v3 v3.0.1
```
dangra, line 15: This change is isolated in [this commit](https://github.com/superfly/flyctl/pull/1830/commits/9532519491336edef9d96c8d95b23463a1189c0c), it should have been part of https://github.com/superfly/flyctl/pull/1817

alichay (approved):
This looks really promising. 
I did some spot checking, making sure that I couldn't make the change to `DefinitionPtr` crash with a malformed fly.toml.

I also tested platform version-specific changes, such as `deploy.strategy = 'bluegreen'` for Nomad and `primary_region` for V2. I was surprised to find that these didn't show up in `fly config save/show`, but it turns out it's the same on latest flyctl, so I guess these just don't show up there ever (might be something for me to look into later, that seems wrong imo)

anyway, I looked at potential pain points of the transition, and ran it through some basic tests, and it all seems good. hopefully this can get the testflight treatment, too :)

(very excited to see the constant `app := app.Whatever(); if app == nil {appv2 = appv2.Whatever();}` gone)

------
my only question here is: wouldn't it make sense to consolidate this all under `app` rather than `appv2`? `appv2` as a name feels kind of cluttered, but it was necessary when there were two similar packages. Now, though, it feels a little bit clunky imo. It's a tiny little nitpick, though

internal/command/launch/launch.go
```diff
@@ -561,7 +553,7 @@ func run(ctx context.Context) (err error) {
 	var v2AppConfig *appv2.Config
 	if deployArgs.ForceMachines {
 
-		v2AppConfig, err = appv2.FromDefinition(api.DefinitionPtr(appConfig.Definition))
+		v2AppConfig, err = appv2.FromDefinition(api.DefinitionPtr(appConfig.RawDefinition))
```
matttpt, line 556: When `v2AppConfig` is created, its `platformVersion` isn't set, and `WriteToFile` will serialize it as a v1 configuration using its `RawDefinition`. However, `v2AppConfig`'s `RawDefinition` is an empty map, so the resulting `fly.toml` is missing almost everything:

```toml
# fly.toml file generated for a-v2-app on 2023-03-06T17:27:17-05:00

app = "a-v2-app"
primary_region = "ord"
```

tvdfly:
> When deploying Machines apps, it is important to check this error as it means the configuration file is invalid.

This strikes me as the kind of thing that flyctl developers are going to forget to do 3 months (or 3 minutes) from now and will introduce bugs. How can we make it so developers do the right thing?

Maybe the default method returns an error and there is a separate ParseOnlyForNomadBecauseIPromiseIWillCheckTheErrorWhereNeeded()` method that can be used where we need it?

dangra:
> I did some spot checking, making sure that I couldn't make the change to `DefinitionPtr` crash with a malformed fly.toml.

Awesome. thanks for this.

> I also tested platform version-specific changes, such as `deploy.strategy = 'bluegreen'` for Nomad and `primary_region` for V2. I was surprised to find that these didn't show up in `fly config save/show`, but it turns out it's the same on latest flyctl, so I guess these just don't show up there ever (might be something for me to look into later, that seems wrong imo)

I think @matttpt spotted the bug already, will commit a fix in a moment

> wouldn't it make sense to consolidate this all under `app` rather than `appv2`? `appv2` as a name feels kind of cluttered, but it was necessary when there were two similar packages. Now, though, it feels a little bit clunky imo. It's a tiny little nitpick, though

I thought of moving it back to `internal/app` as last step but there is one catch, `app` is also a common variable name within the codebase to reference `api.App` and `api.AppCompact`

Alternatively, we could use `internal/appconfig` to reflects it refers to fly.toml.
 
On another note, there is another similarly named package for flyctl config at `internal/config`, it usually confuses newcomers. In any case, I think `internal/appconfig` is different enough.

internal/command/launch/launch.go
dangra: Hey, I think I addressed it in today's fixes.

1. Retain RawDefinition on FromDefinition call https://github.com/superfly/flyctl/pull/1830/commits/d61015b8b634c9b654217a9d5bbce11a82e693fe

2. https://github.com/superfly/flyctl/pull/1830/commits/8da4db84f412740e988151628ab85fe4c4a1d2f5

matttpt: Awesome! Works correctly for me now, and also looks like #1820 is fixed.

dangra:
> Maybe the default method returns an error and there is a separate ParseOnlyForNomadBecauseIPromiseIWillCheckTheErrorWhereNeeded() method that can be used where we need it?

I tried this approach but realized we load the config file in almost all commands https://github.com/superfly/flyctl/blob/d5c82a3f871681f269c685d571924b3ad6b4a57c/internal/command/command.go#L457, returning an error won't solve the issue as the only option is to fail hard or ignore it. 

In this PR opted to show a warning when this happens https://github.com/superfly/flyctl/blob/one-appconfig/internal/command/command.go#L465

And this is guard for `fly deploy` https://github.com/superfly/flyctl/pull/1830/commits/fafa8c0f3f4d8f23fe5651a9bb9591906ba49cdb

billyb2 (approved):
Other than that, LGTM

internal/appv2/setters.go
```diff
@@ -0,0 +1,249 @@
...
+				"tls_skip_verify": false,
+			}}
+		}
+	}
+}
+
+func (c *Config) SetConcurrency(soft int, hard int) {
+	c.v1SetConcurrency(soft, hard)
+	if len(c.Services) > 0 {
+		service := &c.Services[0]
```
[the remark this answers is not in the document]
dangra: I dunno know but it was like that before so I kept it. This functions are called only at launch and sourced by scanners. I guess it is a limitation of the original system. It deserves some extra love but not on this PR please :)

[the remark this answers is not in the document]
billyb2: okie dokie

matttpt:
Sorry to add yet another thing, but I think `flyctl machine clone` might be broken now. I'm getting:

```
Error failed to get app config: Your fly.toml referenced an attribute 'app' but it is unknown. Double check the accepted attributes in fly.toml https://fly.io/docs/reference/configuration
```

I think this is because it it's trying to validate the app configuration on the backend, and `ToDefinition` no longer strips things from the definition. Not sure what to do about it yet, but I just wanted to get this out here.

dangra:
@matttpt thanks for the bug report. b0b50fec should fix it

dangra:
> @matttpt thanks for the bug report. [b0b50fe](https://github.com/superfly/flyctl/commit/b0b50fece1984690b4832f401712c36174a702b6) should fix it

that fixes it to same level it was before but I realized the command is broken anyways, it is grabbing the app config from Web, it should do something like https://github.com/superfly/flyctl/blob/d5c82a3f871681f269c685d571924b3ad6b4a57c/cmd/config.go#L219-L222 but that is for another PR.

PR #1834, opened 2023-03-06
Require app name when necessary for certain `machines` and `secrets` commands

matttpt:
A user on the [forum](https://community.fly.io/t/flyctl-secrets-list-error-could-not-find-app/11220) was running into an "Error Could not find App" message. I quickly looked into it, and it turns out that several `flyctl machines` and `flyctl secrets` subcommands need an app name to work, but use the `LoadAppNameIfNotPresent` preparer instead of `RequireAppName`. As a result, if they're invoked without an app name being present, they'll make an API call looking for an app with an empty name. This inevitably fails ("Could not find App").

I'm not sure if this is what was going wrong for the user, but in any case we avoid the API call and get a correct/actionable error message ("we couldn't find a fly.toml nor an app specified by the -a flag") consistent with the rest of `flyctl` by using `RequireAppName`.

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -31,7 +31,7 @@ func newClone() *cobra.Command {
 
 	cmd := command.New(usage, short, long, runMachineClone,
 		command.RequireSession,
-		command.LoadAppNameIfPresent,
+		command.RequireAppName,
 		command.LoadAppConfigIfPresent,
 	)
 
--- a/internal/command/machine/leases.go
+++ b/internal/command/machine/leases.go
@@ -49,7 +49,7 @@ func newLeaseView() *cobra.Command {
 
 	cmd := command.New(usage, short, long, runLeaseView,
 		command.RequireSession,
-		command.LoadAppNameIfPresent,
+		command.RequireAppName,
 	)
 
 	// at least one arg is required but we can accept a list of machine ids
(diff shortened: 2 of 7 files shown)


internal/command/machine/clone.go
```diff
@@ -31,7 +31,7 @@ func newClone() *cobra.Command {
 
 	cmd := command.New(usage, short, long, runMachineClone,
 		command.RequireSession,
-		command.LoadAppNameIfPresent,
+		command.RequireAppName,
```
dangra, line 34: @catflydotio and @alichay you may want to review this.

Looks like a different UX considering https://github.com/superfly/flyctl/pull/1708

alichay: Thanks for the heads up!

For the most part, this all seems fine to me. `secrets` is, correct me if i'm wrong, a command that operates on Apps rather than machines, so it has to take an app. it makes sense for `clone` to need an explicit app provided because it's creating a new machine in that app, and `list` can't infer anything.

Personally, I would like to see the `lease` commands behave like `restart`, where it infers the app from the first machine ID.

whether we settle on this explicit style or `restart`'s implicit style, though, we should probably agree on one style for commands that take machine ids. I just feel like inferring the app from the machine ID makes things easier for users :)

dangra (approved):

PR #1835, opened 2023-03-06
feat: Add warning text for paid feature

shortdiv:
Add warning text with a prompt for users to confirm when accessing a paid feature

Screen Shot 2023-03-06 at 3 11 06 PM

### DIFF
--- a/internal/command/ips/allocate.go
+++ b/internal/command/ips/allocate.go
@@ -31,6 +31,10 @@ func newAllocatev4() *cobra.Command {
 			Description: "Allocates a shared IPv4",
 			Default:     false,
 		},
+		flag.Bool{
+			Name:        "yes",
+			Description: "Auto-confirm IPv4 allocation",
+		},
 		flag.App(),
 		flag.AppConfig(),
 		flag.Region(),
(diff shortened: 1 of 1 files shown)


tvdfly (commented):
How about making a flag that can auto accept this? like `--yes` or `--auto-confirm`.

Alternatively, we could detect if we're in a non-interactive shell and assume yes since that's the existing behavior.

shortdiv:
Oh, I like both of those ideas! I think creating a generic --yes flag would probably be the way to go with the least room for error

shortdiv:
Updated!

internal/command/ips/allocate.go
```diff
@@ -70,6 +76,15 @@ func runAllocateIPAddressV4(ctx context.Context) error {
 	if flag.GetBool(ctx, "shared") {
 		addrType = "shared_v4"
 	}
+	if !flag.GetBool(ctx, "yes") {
+		confirmIPV4, err := prompt.Confirm(ctx, "Looks like you're accessing a paid feature. Dedicated IPv4 addresses now costs $2/mo. Are you ok with this?")
+		if err != nil {
```
dangra, line 81: Standard way to check for non-interactive input is `prompt.IsNonInteractive(err)`

shortdiv: Oh neat til

dangra (approved):

PR #1848, opened 2023-03-07
fix: Update wording highlight force flag overrides state

shortdiv:
Screen Shot 2023-03-07 at 11 45 14 AM

### DIFF
--- a/internal/command/machine/destroy.go
+++ b/internal/command/machine/destroy.go
@@ -17,7 +17,7 @@ import (
 
 func newDestroy() *cobra.Command {
 	const (
-		short = "Destroy a Fly machine"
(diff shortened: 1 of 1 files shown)


matttpt:
As part of this, maybe it's worth it to update [this](https://github.com/superfly/flyctl/blob/d5c82a3f871681f269c685d571924b3ad6b4a57c/internal/command/machine/destroy.go#L88) `switch` to write a message for all non-stopped states?

shortdiv:
Great idea @matttpt, I've updated this with a catch all so the messaging is clearer

tvdfly (approved):
thanks!

dangra:
@shortdiv all it needs is a rebase to pass pre-commit checks

shortdiv:
Rebased!

dangra (approved):

Issue #1846, opened 2023-03-07
Apps v2 - machines commands should make it easy to choose a machine to act on


catflydotio:
We should make it easy to select a machine when using flyctl commands that take a machine ID.

Copy-pasting machine IDs from earlier terminal output is tedious for the sighted and nimble-fingered. It's really inefficient with a screen reader, according to @ndarilek. I have no idea what it's like with other adaptive devices.

Can we add a flag to these, like `-s` in `fly ssh console`, to provoke a selector that displays enough info to choose a machine to operate on?

Right now we are leaning on a few `fly machine` commands for routine things in apps, like scaling, so this is a fairly high-visibility pain point.

matttpt:
#1852 is a proof-of-concept implementing this for `fly machines clone` and `fly machines destroy`. If this is on the right track in terms of UX, then I don't think it will be too much more work to implement it for the remaining commands.

redjonzaci:
Since all `fly machines` subcommands now have a `--select` flag, can we close this?

matttpt:
@redjonzaci thanks for the reminder!

PR #1852, opened 2023-03-07
Add `--select` to `machines` commands

matttpt:
This addresses #1846 by implementing a `--select` option for `flyctl machines` subcommands that take a machine ID (or several). When `--select` is used, no machine IDs may be specified as arguments, and an app name must be provided (through fly.toml, `FLY_APP`, or `-a`). A prompt is displayed to allow the user to select machines from the specified app. For commands that accept more than one machine ID, the prompt allows multiple selections.

Originally, I was going to include `-s` as a short form, like `flyctl ssh console` does. However, `run` and `update` already use it as a shorthand for `--size`.

### DIFF
--- a/internal/command/machine/select.go
+++ b/internal/command/machine/select.go
@@ -2,6 +2,7 @@ package machine
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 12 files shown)


ndarilek (approved):
Very nice! This is a huge improvement from an accessibility perspective.

internal/command/machine/select.go
```diff
@@ -0,0 +1,53 @@
...
+		machineOptions = append(machineOptions, fmt.Sprintf("%s %s", machine.ID, machine.Name))
+	}
+
+	prompt := &survey.Select{
+		Message:  "Select machine:",
+		Options:  machineOptions,
+		PageSize: 15,
+	}
+	var selected int
+	if err := survey.AskOne(prompt, &selected); err != nil {
```
tvdfly, line 48: let's use our prompt package instead of a 3p package. we have handled a bunch of terminal/platform things in our package.

I think prompt.Select() is the thing to use here.

alichay (approved):
assuming this has been tested, it looks great :)

PR #1853, opened 2023-03-07
Update flaps client to use api.machines.dev by default

tvdfly:
The `FLY_FLAPS_BASE_URL` env var may be used to configure alternative urls. Setting `FLY_FLAPS_BASE_URL=peer` will use usermode wireguard like we have done previously.

### DIFF
--- a/flaps/flaps.go
+++ b/flaps/flaps.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -21,6 +22,7 @@ import (
 	"github.com/superfly/flyctl/client"
 	"github.com/superfly/flyctl/flyctl"
 	"github.com/superfly/flyctl/internal/buildinfo"
+	"github.com/superfly/flyctl/internal/config"
 	"github.com/superfly/flyctl/internal/logger"
 	"github.com/superfly/flyctl/terminal"
 )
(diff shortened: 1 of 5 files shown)


tvdfly:
`fly config save` panics with this change. likely missing context initialization. working to fix that now

dangra:
> `fly config save` panics with this change. likely missing context initialization. working to fix that now

Oh, I ported it to internal/command in https://github.com/superfly/flyctl/pull/1855

flaps/flaps.go
```diff
@@ -31,13 +33,36 @@ const headerFlyRequestId = "fly-request-id"
...
 }
 
 func New(ctx context.Context, app *api.AppCompact) (*Client, error) {
+	cfg := config.FromContext(ctx)
+	if strings.TrimSpace(strings.ToLower(cfg.FlapsBaseURL)) == "peer" {
+		return newWithUsermodeWireguard(ctx, app)
+	}
+	flapsUrl, err := url.Parse(cfg.FlapsBaseURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid FLAPS_BASE_URL '%s' with error: %w", cfg.FlapsBaseURL, err)
```
gwuah, line 49: I think this should be 
```suggestion
		return nil, fmt.Errorf("invalid FLY_FLAPS_BASE_URL '%s' with error: %w", cfg.FlapsBaseURL, err)
```

gwuah (approved):
looks good, just had one comment related to an error message.

gwuah (approved):

PR #1854, opened 2023-03-08
Warn users about volumes

jaredonline:
This warns users about only having a single volume with a given volumeName for a given app. If the user creates a single volume, it warns them. If the user deletes volumes such that they will have less than 2 after the delete, it warns them.

### DIFF
--- a/internal/command/volumes/create.go
+++ b/internal/command/volumes/create.go
@@ -146,14 +146,10 @@ func confirmVolumeCreate(ctx context.Context, appName string) (bool, error) {
 
 	switch confirmed, err := prompt.Confirm(ctx, "Do you still want to use the volumes feature?"); {
 	case err == nil:
-		if !confirmed {
-			return false, nil
-		}
+		return confirmed, nil
 	case prompt.IsNonInteractive(err):
 		return false, prompt.NonInteractiveError("yes flag must be specified when not running interactively")
 	default:
 		return false, err
 	}
-
-	return true, nil
 }
--- a/internal/command/volumes/create.go
+++ b/internal/command/volumes/create.go
@@ -80,45 +80,14 @@ func runCreate(ctx context.Context) error {
 		return err
 	}
 
-	// If the Yes flag has been supplied we skip the warning entirely, so no
-	// need to query for the set of volumes
-	if !flag.GetYes(ctx) {
-		// fetch the set of volumes for this app. If > 0 we skip the prompt
-		var volumes []api.Volume
-		if volumes, err = client.GetVolumes(ctx, appName); err != nil {
-			return err
-		}
-
-		var matches int32
-		for _, volume := range volumes {
-			if volume.Name == volumeName {
-				matches++
-			}
-		}
-
-		if matches == 0 {
-			var confirmed bool
-			io := iostreams.FromContext(ctx)
-			colorize := io.ColorScheme()
-
-			const msg = "Warning! Individual volumes are pinned to individual hosts. You should create two or more volumes per application. You will have downtime if you only create one. If you proceed, make sure to create a second volume with fly volume create --require-unique-zone to ensure a second volume exists on a separate worker"
-			fmt.Fprintln(io.ErrOut, colorize.Red(msg))
-
-			switch confirmed, err = prompt.Confirm(ctx, "Do you still want to use the volumes feature?"); {
-			case err == nil:
-				if !confirmed {
-					return nil
-				}
-			case prompt.IsNonInteractive(err):
-				return prompt.NonInteractiveError("yes flag must be specified when not running interactively")
-			default:
-				return err
-			}
-		}
+	var confirm bool
+	if confirm, err = confirmVolumeCreate(ctx, appName); err != nil {
+		return err
+	} else if !confirm {
+		return nil
 	}
 
 	var region *api.Region
-
 	if region, err = prompt.Region(ctx, !app.Organization.PaidPlan, prompt.RegionParams{
 		Message: "",
 	}); err != nil {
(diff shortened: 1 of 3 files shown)


internal/command/volumes/create.go
```diff
@@ -79,6 +80,43 @@ func runCreate(ctx context.Context) error {
...
+				matches++
+			}
+		}
+
+		if matches == 0 {
+			var confirmed bool
+			io := iostreams.FromContext(ctx)
+			colorize := io.ColorScheme()
+
+			const msg = "Warning! Individual volumes are pinned to individual hosts. You should create two or more volumes per application. You will have downtime if you only create one."
```
tvdfly, line 104: I like to include instructions for the developer to solve the issue... maybe something like "make sure to create a second volume with `fly volume create --require-unique-zone` to ensure a second volume exists on a separate worker`. 

alternatively, maybe we make it so they can just create two volumes that will exist on different workers? E.g., `--count 2 ` would create two volumes in the same region and ensure they are on different hosts? that could be done separately / in addition to updating the warning msg.

jaredonline: That makes sense - I'll work the additional messaging in.

I'll look into adding the `--count 2` flag in a separate PR if that makes sense?

jaredonline:
@tvdfly I've just updated the error message. Let me know what you think [=

jaredonline:
@tvdfly ping

internal/command/volumes/create.go
```diff
@@ -79,6 +80,43 @@ func runCreate(ctx context.Context) error {
...
+				if !confirmed {
+					return nil
+				}
+			case prompt.IsNonInteractive(err):
+				return prompt.NonInteractiveError("yes flag must be specified when not running interactively")
+			default:
+				return err
+			}
+		}
+	}
```
dangra, line 118: this function block is not part of the main flow and it is quite large, can we move that to another function that summarizes the intention instead of inlining the whole thing in runCreate?

internal/command/volumes/create.go
```diff
@@ -79,6 +80,43 @@ func runCreate(ctx context.Context) error {
...
+
+		if matches == 0 {
+			var confirmed bool
+			io := iostreams.FromContext(ctx)
+			colorize := io.ColorScheme()
+
+			const msg = "Warning! Individual volumes are pinned to individual hosts. You should create two or more volumes per application. You will have downtime if you only create one. If you proceed, make sure to create a second volume with fly volume create --require-unique-zone to ensure a second volume exists on a separate worker"
+			fmt.Fprintln(io.ErrOut, colorize.Red(msg))
+
+			switch confirmed, err = prompt.Confirm(ctx, "Do you still want to use the volumes feature?"); {
```
dangra, line 107: Do we really want to bug the user with a confirmation after that long warning message? I can accept it but seems like we are overreacting :)

internal/command/volumes/destroy.go
```diff
@@ -44,7 +45,31 @@ func runDestroy(ctx context.Context) error {
...
+		for _, v := range volumes {
+			if v.Name == volume.Name {
+				matches++
+			}
+		}
+
+		var msg = "Deleting a volume is not reversible."
+		if matches <= 2 {
+			msg = fmt.Sprintf("Warning! Individual volumes are pinned to individual hosts. You should create two or more volumes per application. Deleting this volume will leave you with %d volume(s) for this application, and it is not reversible.", matches-1)
+		}
```
dangra, line 72: same thing, even the code to grab the volumes and matches can be factored out

internal/command/volumes/create.go
catflydotio: I could go either way on the confirmation. It does emphasise the gravity of the decision. :)

This is such a complicated idea for a few words! They need to set up replication too, if they need data availability. Would it be easier to say: 

"Warning! Individual volumes are pinned to individual hosts. You should create two or more volumes per application. You will have downtime if you only create one. Learn more at https://fly.io/docs/reference/volumes/"

I know you couldn't have done this before because the messaging was tied up in the [apps v2 PR](https://github.com/superfly/docs/pull/627). I brought some over to main, to plug the gap.

jaredonline: Linking to the docs makes a ton of sense - this way as things change we don't need to keep this message up to date.

internal/command/volumes/create.go
```diff
@@ -80,45 +80,14 @@ func runCreate(ctx context.Context) error {
...
-				return prompt.NonInteractiveError("yes flag must be specified when not running interactively")
-			default:
-				return err
-			}
-		}
+	var confirm bool
+	if confirm, err = confirmVolumeCreate(ctx, appName); err != nil {
+		return err
+	} else if !confirm {
+		return nil
```
dangra, line 87: no need to declare `confirm` like this if you don't plan to use it outside of `if` block.

it works with:
```go
if confirm, err := confirmVolumeCreate(ctx, appName); err != nil {
    return err
} else if !confirm {
    return nil
}
```

in fact, you can write it all as (but it may confuse the reader)
```go
if confirm, err := confirmVolumeCreate(ctx, appName); err != nil || !confirm {
    return err
}
```

internal/command/volumes/create.go
```diff
@@ -153,3 +122,42 @@ func runCreate(ctx context.Context) error {
 
 	return printVolume(out, volume)
 }
+
+func confirmVolumeCreate(ctx context.Context, appName string) (bool, error) {
+	var volumeName = flag.FirstArg(ctx)
+
+	// If the Yes flag has been supplied we skip the warning entirely, so no
+	// need to query for the set of volumes
+	if !flag.GetYes(ctx) {
```
dangra, line 131: Avoid the nested ifs in case like this, see [reduce nesting, reduce complexity](https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html), this is the same principle used for [shortcircuit'ing error flow](https://google.github.io/styleguide/go/decisions#indent-error-flow)

internal/command/volumes/create.go
jaredonline: I stuck with the `if/else if` for clarity, but otherwise changed the declaration. Thank you!

internal/command/volumes/create.go
jaredonline: Thank you! Done [=

internal/command/volumes/destroy.go
```diff
@@ -68,3 +61,48 @@ func runDestroy(ctx context.Context) error {
...
+	if matches <= 2 {
+		msg = fmt.Sprintf("Warning! Individual volumes are pinned to individual hosts. You should create two or more volumes per application. Deleting this volume will leave you with %d volume(s) for this application, and it is not reversible.  Learn more at https://fly.io/docs/reference/volumes/", matches-1)
+	}
+	fmt.Fprintln(io.ErrOut, colorize.Red(msg))
+
+	switch confirmed, err := prompt.Confirm(ctx, "Are you sure you want to destroy this volume?"); {
+	case err == nil:
+		if !confirmed {
+			return false, nil
+		}
```
dangra, line 100: this is the same as `return confirmed, nil`

internal/command/volumes/destroy.go
```diff
@@ -37,26 +38,18 @@ number to operate. This can be found through the volumes list command`
...
-		io       = iostreams.FromContext(ctx)
-		colorize = io.ColorScheme()
-		client   = client.FromContext(ctx).API()
-		volID    = flag.FirstArg(ctx)
+		io     = iostreams.FromContext(ctx)
+		client = client.FromContext(ctx).API()
+		volID  = flag.FirstArg(ctx)
+
+		confirm bool
+		err     error
```
dangra, line 46: no need to declare vars here, in Go you usually do`if confirm, err := ...` and keep going

internal/command/volumes/create.go
```diff
@@ -115,3 +121,39 @@ func runCreate(ctx context.Context) error {
...
+	colorize := io.ColorScheme()
+
+	const msg = "Warning! Individual volumes are pinned to individual hosts. You should create two or more volumes per application. You will have downtime if you only create one. Learn more at https://fly.io/docs/reference/volumes/"
+	fmt.Fprintln(io.ErrOut, colorize.Red(msg))
+
+	switch confirmed, err := prompt.Confirm(ctx, "Do you still want to use the volumes feature?"); {
+	case err == nil:
+		if !confirmed {
+			return false, nil
+		}
```
dangra, line 151: same thing about `return confirmed, nil`

dangra (approved):
👍

PR #1855, opened 2023-03-08
Port 'fly config' to internal/command/config

dangra:
More than the porting, the cherry is the function `appconfig.FromRemoteApp()` that abstract grabbing application configuration from Nomad and Machines apps


internal/command/config/save.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return err
+		}
+		if !confirmation {
+			return nil
+		}
+	}
+
+	return cfg.WriteToDisk(ctx, configfilename)
+}
```
gwuah, line 54: what do you think about checking for the existence of the config file ***before*** making the network_trip(FromRemoteApp)?

Because if it already exists, we're not going to use the response from the network trip anyway.

gwuah (approved):
Tested all 4 commands on my nomad & machines app and they work fine.

tvdfly (approved):

PR #1856, opened 2023-03-08
Process Groups for Apps V2! 🎉

### LINKED ISSUE #1772: Improve process group handling on `fly launch` for apps v2
On the first deployment, `fly deploy` will create one machine and it always assigns it to the app group. That's fine when there are no process groups defined, or only an app process group. We can do better when other process groups are defined, though!

A better DX would be to create one machine for each process group, or something similar.

Here's what fly launch then fly deploy looks like today (note that no worker machines are launched):
```
% cat fly.toml
app = "tvd-repro-2023-02-23"
[build]
  image = "nginx"
[processes]
  app = "nginx -g 'daemon off;'"
  worker = "tail -F /dev/null"

% fly launch --force-machines
An existing fly.toml file was found for app tvd-repro-2023-02-23
? Would you like to copy its configuration to the new app? Yes
Creating app in /Users/tvd/src/superfly/test-emporium/cfg-repro-2023-02-21
Scanning source code
? Choose an app name (leave blank to generate one): tvd-repro-2023-02-23
...
Created app tvd-repro-2023-02-23 in organization tvd-testorg
Admin URL: https://fly.io/apps/tvd-repro-2023-02-23
Hostname: tvd-repro-2023-02-23.fly.dev
Wrote config file fly.tom

% fly deploy --force-machines
==> Verifying app config
--> Verified app config
==> Building image
image found: img_19gm46k36kzvx0jk
WARN The running flyctl agent (v0.0.0-1677183437+dev) is older than the current flyctl (v0.0.0-1677185143+dev).
WARN The out-of-date agent will be shut down along with existing wireguard connections. The new agent will start automatically as needed.
No machines in tvd-repro-2023-02-23 app, launching one new machine
  Machine 39080573fd9087 [app] update finished: success
  Finished deploying
```

alichay:
Would close #1772, probably others too

My current (completely non-exhaustive) test suite is deploying with the following fly.toml files:

> I expect one "app" machine)
```toml
[[services]]
  http_checks = []
  internal_port = 8080
  protocol = "tcp"
  script_checks = []
```

> I expect one "web" machine and one "bar_web" machine
```toml
[processes]
web = "nginx -g 'daemon off;'"
bar_web = "bash -c /sleep_forever.sh"

[[services]]
  processes = ["web"] # this service only applies to the web process
  http_checks = []
  internal_port = 8080
  protocol = "tcp"
  script_checks = []
```

> I expect one "web" machine *with the same Machine ID*
```toml
[processes]
web = "nginx -g 'daemon off;'"

[[services]]
  processes = ["web"] # this service only applies to the web process
  http_checks = []
  internal_port = 8080
  protocol = "tcp"
  script_checks = []
```

So far, this passes, but I'm **sure** there's edge cases right now

---
TODO:

- [x] Make `NewMachineDeployment` gather process configs, rather than being passed them.
- [x] Test secrets commands (what happens if I set secrets on an app with multiple process groups)
- [x] Test interactive deploys (i've been using a shell script to test this noninteractively so far)
- [x] Cleanup the myriad of TODOs
- [x] Confirm that the behavior I was going for is intended (I believe it is, but double checking can't hurt!)
- [x] Actual, slightly more rigorous testing :)
- [x] Pass through the code and make names consistent. I ended up with like four different terms for process groups by the end of this.
- [x] Use `getProcessGroup`


internal/command/deploy/deploy.go
```diff
@@ -166,9 +166,16 @@ func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, args Dep
...
+
+		processConfigs, err := appConfig.GetProcessConfigs()
+		if err != nil {
+			return err
+		}
+
 		md, err := NewMachineDeployment(ctx, MachineDeploymentArgs{
 			AppCompact:        appCompact,
 			DeploymentImage:   img,
+			ProcessConfigs:    &processConfigs,
```
dangra, line 178: I'm curious why did you prefer to pass `processConfigs` from the "outside" instead of grabbing them from the context's appConfig

alichay: no particular reason! that makes a lot more sense though, and looking at it with fresh eyes I think the current method would (maybe?) break when `secrets` deploys because it doesn't set that. I'll switch it up tomorrow morning :)

tvdfly (commented):
What DX is implemented with this change? like what happens in the different add/remove/update scenarios and how to does that compare to v1 behavior?

alichay:
> What DX is implemented with this change? like what happens in the different add/remove/update scenarios

I was basing the dx mostly on this https://flyio.discourse.team/t/apps-v2-processes-and-machines-everybody-gets-a-process/1646/2

So, if you make a change that adds a new process group, it prompts you, saying 
```
 * create 1 "examplegroup" machine
 * create 1 "othergroup" machine
? Accept these changes?  
 ```

If you make a change that removes a process group, it prompts with
```
 * remove 3 "removedgrp" machines
? Accept these changes?
```

Unfortunately, right now, if you type N, it exits, but by that point it's already flagged a release and built the image. (more on that in a followup comment!)

> how to does that compare to v1 behavior?

Honestly not sure, I haven't used V1 much at all. I planned on going back and checking that V1 acted similarly after I cleaned up some of the mess, because even if it turns out a little different the building blocks are there now

**Comment by alichay** [comment]
So I have a question:
Right now, it prompts in the deploy process, asking whether or not you want to accept the changes. I did it this way because that's when the existing machines are leased, which is clean and makes sense.

The problem is that, by this point, the image was already built (which maybe took a while) and a release was created. This needs to be fixed. This is all done before the machines are leased.
I assume that deploy is split up this way so that the machines aren't leased for a potentially lengthy build process.

That leads me to the two options I see:

 * Lease the machines before the image is build or the release is flagged. That allows me to place the checks earlier, and opt out before anything is done to the app. The problem here is that the machines are leased for a long time.
 * Lease the machines, do the checks and create/remove machines, revoke the lease, build the image/release, and then re-lease the machines and continue down the old deploy path.

Between these two, the first seems like the better option, but I wanted some input before I committed down that path. (or, if there's an even better third option, I'm all ears!)

Thanks!

**Comment by tvdfly** [comment]
I personally prefer we don't use prompts in the deploy process. It makes the DX a little awkward, since fly deploy is something that gets run often and in the overwhelming majority of runs will not have a prompt. It feels disruptive to me when it prompts. (I think I was the one that recommended a prompt before, so this is me evolving my perspective aka changing my mind!)

Some alternative ideas to using a prompt:
1. output a warning and make the changes (the developer updated fly.toml, so flyctl should do what they said)
2. a variation on that is output a warning, wait like 10 or 30 seconds so the developer has a chance to read the warning and CTRL+C if it's not what they want, then make the change
3. require the `--auto-confirm` flag or maybe a new `--delete-process-group-machines` flag in order to delete machines when a process group goes away. if the flag is missing, don't delete the machine(s) and return an error.

3 is the "safest" in the sense that it will be difficult for developers to delete machines with accidental/unintentional fly.toml changes. I'm tempted to start with 1 and then we can do something different if we find that's causing issues.

**Review by matttpt — COMMENTED** [review]
Overall this looks really good to me!

I'm wary of borrowing `machines destroy`'s code. Is there a reason not to use the flaps client's `Destroy` directly in `internal/command/deploy/machines.go`? (The deletion hook, I guess?)

If it's the right thing to borrow it (at least for now), it might be helpful to make `machine.Destroy` take an `*api.Machine` instead of a machine ID, and to move the `flapsClient.get` call back into `runMachineDestroy`. I think the deploy code already has access to an `*api.Machine` to pass in, so that would avoid pulling down another copy for each machine it's going to destroy. Also (full disclosure), this is a selfish suggestion, since it would essentially remove a conflict between these changes and #1899.

File: internal/command/deploy/machines.go [hunk 2]
```diff
@@ -207,14 +211,129 @@ func (md *machineDeployment) runReleaseCommand(ctx context.Context) error {
...
+			Metadata: map[string]string{
+				api.MachineConfigMetadataKeyFlyProcessGroup: groupName,
+			},
+		},
+	}
+	launchInput := md.resolveUpdatedMachineConfig(machBase, false)
+	newMachineRaw, err := md.flapsClient.Launch(ctx, *launchInput)
+	newMachine := machine.NewLeasableMachine(md.flapsClient, md.io, newMachineRaw)
+	if err != nil {
+		return fmt.Errorf("error creating a new machine machine: %w", err)
```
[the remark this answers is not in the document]
alichay: > The deletion hook, I guess?

Yeah. I saw that and figured it was probably important to make sure behavior stayed consistent no matter how a machine was deleted. Right now it's just one hook, but I don't want to introduce a gap where the behavior can slip further apart over time.

A lot of commands export public helpers. It's clunky, but it helps to work around go's limitation against reference cycles.

> I think the deploy code already has access to an *api.Machine to pass in, so that would avoid pulling down another copy for each machine it's going to destroy

Good call

> it would essentially remove a conflict between these changes and https://github.com/superfly/flyctl/pull/1899.

I'll wait for that pr to land, then update this to do exactly that :)

internal/command/deploy/machines.go
```diff
@@ -207,14 +211,129 @@ func (md *machineDeployment) runReleaseCommand(ctx context.Context) error {
...
+	fmt.Fprintf(md.io.Out, "No machines in group '%s', launching one new machine\n", md.colorize.Bold(groupName))
+	machBase := &api.Machine{
+		Region: md.appConfig.PrimaryRegion,
+		Config: &api.MachineConfig{
+			Metadata: map[string]string{
+				api.MachineConfigMetadataKeyFlyProcessGroup: groupName,
+			},
+		},
+	}
+	launchInput := md.resolveUpdatedMachineConfig(machBase, false)
```
dangra, line 290: I know ideally we have to test `spawnMachineGroup` but for now I think it can do with a test case for resolveUpdatedMachineConfig and the same input passed here.

alichay: I had a whole little test suite for this feature that I wrote in rust, so I just ported it to preflight.

I couldn't figure out a good way to test that part of the code without making it brittle (splitting things into tiny utility functions) or easy to fall out of sync (just copying bits of code), so it just makes more sense to me to do a whole end-to-end test rather than just the config resolver.

dangra: sgtm 👍

alichay:
preflight passes, tests pass, and I'm feeling pretty good about this :)

internal/command/deploy/machines.go
```diff
@@ -539,19 +617,33 @@ func (md *machineDeployment) resolveUpdatedMachineConfig(origMachineRaw *api.Mac
 			Config: &api.MachineConfig{},
 		}
 	}
+	newConfig := machine.CloneConfig(origMachineRaw.Config)
+	if newConfig.Metadata == nil {
+		newConfig.Metadata = map[string]string{}
+	}
```
dangra, line 623: wdyt if this is moved to line 634 (right before setting metadata), it can simplify that code too and remove a nested if.

alichay:
ok yeah there were some sloppy mistakes from trying to get this out the door. Fixed!

dangra (approved):
Approved!

From this point it is _pedantic mode_ review, feel free to ignore my following comments 🙏

internal/command/deploy/machines.go
```diff
@@ -539,18 +617,34 @@ func (md *machineDeployment) resolveUpdatedMachineConfig(origMachineRaw *api.Mac
...
 		OrgSlug: md.app.Organization.ID,
-		Config:  lo.Ternary(md.restartOnly, origMachineRaw.Config, &api.MachineConfig{}),
+		Config:  newConfig,
 		Region:  origMachineRaw.Region,
 	}
 
-	launchInput.Config.Metadata = md.defaultMachineMetadata()
-	for k, v := range origMachineRaw.Config.Metadata {
-		if !isFlyAppsPlatformMetadata(k) {
+	launchInput.Config.Statics = nil
```
dangra, line 630: it's suppose to be pedantic comment but this is a bug, because it removes statics when it is is in "restart only" mode. I'd move it right above the block that process statics:

```
for _, s := range md.appConfig.Statics {
		launchInput.Config.Statics = append(launchInput.Config.Statics, &api.Static{
			GuestPath: s.GuestPath,
			UrlPrefix: s.UrlPrefix,
		})
	}
```

internal/command/deploy/machines.go
```diff
@@ -539,18 +617,34 @@ func (md *machineDeployment) resolveUpdatedMachineConfig(origMachineRaw *api.Mac
...
+		}
+	}
+	// Patch any holes in the config with defaults
+	defaultMetadata := md.defaultMachineMetadata()
+	for k, v := range defaultMetadata {
+		_, has := launchInput.Config.Metadata[k]
+		if !has {
 			launchInput.Config.Metadata[k] = v
 		}
 	}
```
dangra, line 650: all this stuff can be rewritten as :

```go
launchInput.Config.Metadata = lo.Assign(
  lo.OmitBy(launchInput.Config.Metadata, func(k, v string) bool {
    return isFlyAppsPlatformMetadata(k)
  }),
  md.defaultMachineMetadata(),
) 
```

internal/command/deploy/machines.go
alichay: I had it moved up so that it would clear for release commands too, I forgot about restartOnly. There's quite a few distinct states to keep track of here...

internal/command/deploy/machines.go
alichay: oh that's sick, I haven't gotten a chance to use `lo` much but that's great

internal/command/deploy/machines.go
dangra: indeed, it is asking for a redo but without tests it is too risky to refactor this logic  now

dangra (approved):

internal/command/secrets/secrets.go
```diff
@@ -67,7 +68,14 @@ func deployForSecrets(ctx context.Context, app *api.AppCompact, release *api.Rel
...
 
 	if app.PlatformVersion == "machines" {
-		ctx, err = command.LoadAppConfigIfPresent(ctx)
+		// It would be confusing for setting secrets to deploy the current fly.toml file.
+		// Instead, we always grab the currently deployed app config
+		cfg, err := appconfig.FromRemoteApp(ctx, app.Name)
+		if err != nil {
+			return err
+		}
+		ctx = appconfig.WithConfig(ctx, cfg)
```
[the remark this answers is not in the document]
gwuah: @alichay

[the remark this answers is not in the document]
alichay: I can't say whether this behavior would still happen - this feature went through a *lot* of iteration - but there was an issue where if you had changed the process groups in `fly.toml`, but not yet deployed them, setting secrets would deploy that change. This pulls the currently live configuration, meaning that `secrets` only affects secrets.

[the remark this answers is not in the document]
gwuah: > issue where if you had changed the process groups in fly.toml, but not yet deployed them, setting secrets would deploy that change

the sheer amount of mines in flyctl 💣

PR #1858, opened 2023-03-08
Split main launch function to reduce code complexity

dangra:
A first iteration on splitting launch command into smaller pieces.

It includes bugfixes for the use of `deployArgs.ForceMachines` instead of `shouldUseMachines` . It also ensure primary region is set and respected no matter what.


billyb2 (approved):
Generally LGTM!

Issue #1816, opened 2023-03-02
Apps V2 - `fly m status` doesn't show the machine's process group


catflydotio:
flyctl v0.0.0-1677786761+dev darwin/arm64 Commit: 3c041ba5dcb5b8966fb2e8b5c62287991f648097 (master) BuildDate: 2023-03-02T19:52:41Z

```
$ ~/flyctl/bin/flyctl status                 
...         

Machines
ID              PROCESS VERSION REGION  STATE   HEALTH CHECKS   LAST UPDATED         
217811122f1389  app     1       yyz     started                 2023-03-02T19:56:03Z
```

```
$ ~/flyctl/bin/flyctl m status 217811122f1389
Machine ID: 217811122f1389
Instance ID: 01GTHYAPECTQ6M2YMR3HGPF4KJ
State: started

VM
  ID            = 217811122f1389                  
  Instance ID   = 01GTHYAPECTQ6M2YMR3HGPF4KJ      
  State         = started                         
  Image         = library/nginx:latest            
  Name          = quiet-cherry-2286               
  Private IP    = fdaa:0:3b99:a7b:f2:c300:3db4:2  
  Region        = yyz                             
  Process Group =                                 
  Memory        = 256                             
  CPUs          = 1                               
  Created       = 2023-03-02T19:56:02Z            
  Updated       = 2023-03-02T19:56:03Z            
  Command       =                                 

Event Logs
STATE   EVENT   SOURCE  TIMESTAMP                       INFO 
started start   flyd    2023-03-02T14:56:03.009-05:00
created launch  user    2023-03-02T14:56:02.398-05:00
```

PR #1859, opened 2023-03-08
Extract machines process group from right key

gwuah:
this fixes https://github.com/superfly/flyctl/issues/1816

### DIFF
--- a/internal/command/machine/status.go
+++ b/internal/command/machine/status.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/alecthomas/chroma/quick"
 	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/flaps"
 	"github.com/superfly/flyctl/internal/appconfig"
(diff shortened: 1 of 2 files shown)


internal/command/machine/status.go
```diff
@@ -46,6 +47,29 @@ func newStatus() *cobra.Command {
...
+	// - machines with only 'process_group'
+	// - machines with both 'process_group' and 'fly_process_group'
+	// - machines with only 'fly_process_group'
+
+	processGroup := m.Config.Metadata["process_group"]
+	flyProcessGroup := m.Config.Metadata[api.MachineConfigMetadataKeyFlyProcessGroup]
+
+	if processGroup != "" && flyProcessGroup != "" {
+		return flyProcessGroup
+	}
```
dangra, line 64: Temoving these lines will have the same effect because of below `if`.
Such a small function could use a quick test case added.

dangra (approved):
nice!

billyb2 (commented):
LGTM!

PR #1860, opened 2023-03-08
check against all indicators of a machine being 'release' machines'

gwuah:
in the current flyctl codebase there are 2 indicators of a machine being a "release_machine"
- metadata 'fly_process_group' having a value of `fly_app_release_command`
- metadata 'process_group' having a value of `release_command`

this PR adds a method on Machine struct to check for both indicators when trying to confirm if a machine is a release machine.

note: this pull request was inspired by https://github.com/superfly/flyctl/pull/1859 (i separated them because this change touches postgres and I wanted folks from the pg team to have a look too)

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -74,7 +74,7 @@ func (m *Machine) IsFlyAppsPlatform() bool {
 }
 
 func (m *Machine) IsFlyAppsReleaseCommand() bool {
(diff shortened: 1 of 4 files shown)


dangra (approved):
Good call.  

I wish we could get in the habit of adding unit tests specially on functions or methods that doesn't require calls to external services.

gwuah:
yes that would go a long way in preventing tiny regressions.

dangra (approved):

PR #1863, opened 2023-03-08
Set app name in context in Turboku

### LINKED ISSUE #1857: Turboku failing: Could not find App
**Describe the bug**
I'm trying to use https://fly.io/launch/heroku to migrate from Heroku to Fly.io unsuccessfully. 

**Command output:**

```text
Preparing to run: `/flyctl turboku <heroku name> <id> --org personal --region lax --name <fly.io name> --now` as root
Using heroku app: <heroku name>
Selected fly region: lax
New app created: <fly.io name>
Secrets are staged for the first deployment
Changed to new app directory  <fly.io name>
Procfile created
Dockerfile created
Wrote config file
Error Could not find App
```

The Heroku app is a simple Express.js backend.

matttpt:
`DeployWithConfig` requires this. I think that this should fix #1857, though I don't have a real Heroku app to test it against.

### DIFF
--- a/internal/command/turboku/turboku.go
+++ b/internal/command/turboku/turboku.go
@@ -181,9 +181,11 @@ func run(ctx context.Context) error {
 	fmt.Fprintf(io.Out, "Changed to new app directory %s\n", createdApp.Name)
 
 	// Generate an app config to write to fly.toml
-	appConfig := appconfig.NewConfig()
+	appConfig, err := appconfig.FromDefinition(&createdApp.Config.Definition)
+	if err != nil {
+		return fmt.Errorf("failed to get new app configuration: %w", err)
+	}
 
-	appConfig.RawDefinition = createdApp.Config.Definition
 	procfile := ""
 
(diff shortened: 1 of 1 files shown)


internal/command/turboku/turboku.go
```diff
@@ -215,6 +215,7 @@ func run(ctx context.Context) error {
 	fmt.Fprintln(io.Out, "Dockerfile created")
 
 	appConfig.AppName = createdApp.Name
+	ctx = appconfig.WithName(ctx, appConfig.AppName)
```
dangra, line 218: I'd add `ctx = appconfig.WithConfig(ctx, appConfig)`

and change line 186 from 
```
        appConfig := appConfig.NewConfig()
	appConfig.RawDefinition = createdApp.Config.Definition
```
to
```
     appConfig, err := appconfig.FromDefinition(&createdApp.Config.Definition)
```

This is what we are doing for `fly launch`

dangra (approved):

PR #1864, opened 2023-03-08
send lease nonce to restart API call

jipperinbham:


flaps/flaps.go
```diff
@@ -208,7 +208,12 @@ func (f *Client) Stop(ctx context.Context, in api.StopMachineInput) (err error)
 	return
 }
 
-func (f *Client) Restart(ctx context.Context, in api.RestartMachineInput) (err error) {
+func (f *Client) Restart(ctx context.Context, in api.RestartMachineInput, nonce string) (err error) {
```
alichay, line 211: I don't think you pass the headers to the API call here

alichay (changes requested):

alichay (approved):

PR #1868, opened 2023-03-09
Fix/machine run hang

shortdiv:
This PR addresses some comments about machine run hanging but also adds some features for better observability
- Abort a machine run when it does not reach the desired state AND restart policy is set to no
- Add restart flag so users can properly run a machine with a specific restart policy
- Spinner to show a pending process is running

Separately, I noticed that the [-c flag in machine run is a bit confusing](https://community.fly.io/t/fly-machines-run-command-hangs-forever/10269). The config flag expects a path to a file but I've seen folks sending json blobs through it. To my knowledge, the cli doesn't accept a json blob, only the api does. I think it would be useful to have a way for users to send json via the cli similar to how they would with the api. I'll open an issue for this.

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -732,8 +732,12 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
 		machineConf.Restart.Policy = api.MachineRestartPolicyNo
 	case "on-fail":
 		machineConf.Restart.Policy = api.MachineRestartPolicyOnFailure
-	default:
+	case "always":
 		machineConf.Restart.Policy = api.MachineRestartPolicyAlways
+	case "":
+		machineConf.Restart.Policy = api.MachineRestartPolicyAlways
+	default:
+		return machineConf, errors.New("invalid restart provided")
 	}
 
 	// `machine update` and `machine run` both use `determineMachineConfig`` to populate
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -170,7 +170,7 @@ func newRun() *cobra.Command {
 		},
 		flag.String{
 			Name:        "restart",
-			Description: "Configure restart policy, for a machine",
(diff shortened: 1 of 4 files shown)


internal/command/machine/run.go
```diff
@@ -716,6 +726,16 @@ func determineMachineConfig(ctx context.Context, initialMachineConf api.MachineC
...
 	}
 
+	// default restart policy to always unless otherwise specified
+	switch flag.GetString(ctx, "restart") {
+	case "no":
+		machineConf.Restart.Policy = api.MachineRestartPolicyNo
+	case "on-fail":
+		machineConf.Restart.Policy = api.MachineRestartPolicyOnFailure
+	default:
+		machineConf.Restart.Policy = api.MachineRestartPolicyAlways
```
dangra, line 736: wdyt if it fails in case the value is not one of the allowed? been loose can lead to gotchas when someone mistype any of the other values.

shortdiv: Good idea, updated!

dangra (approved):
LGTM!

dangra:
A note about:

> Separately, I noticed that the [-c flag in machine run is a bit confusing](https://community.fly.io/t/fly-machines-run-command-hangs-forever/10269). The config flag expects a path to a file but I've seen folks sending json blobs through it. To my knowledge, the cli doesn't accept a json blob, only the api does. I think it would be useful to have a way for users to send json via the cli similar to how they would with the api. I'll open an issue for this.

The `-c` (aka `--config`) flag points to the application configuration file (fly.toml) it has nothing to do with non-managed machines started manually with commands like `fly machines run`. I am saying this because we probably need a different flag to pass json blob or a path to it.

dangra:
merge?

shortdiv:
Yeah definitely, I recently learnt about https://github.com/superfly/flyctl/pull/1694 to address the json blob issue :D

PR #1869, opened 2023-03-09
Port "fly scale" to internal/command/scale

dangra:


internal/command/scale/count.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+func newScaleCount() *cobra.Command {
+	const (
+		short = "Change an app's VM count to the given value"
+		long  = `Change an app's VM count to the given value.
+
+For pricing, see https://fly.io/docs/about/pricing/`
+	)
+	cmd := command.New("count", short, long, runScaleCount,
```
matttpt, line 24: FYI in case this wasn't intentional: previously it also [showed](https://github.com/superfly/flyctl/blob/master/docstrings/gen.go#L640) the argument in the usage. Same thing for `memory` and `vm`.

internal/command/scale/count.go
```diff
@@ -0,0 +1,93 @@
...
+		command.RequireAppName,
+		failOnMachinesApp,
+	)
+	cmd.Args = cobra.MinimumNArgs(1)
+	flag.Add(cmd,
+		flag.App(),
+		flag.AppConfig(),
+		flag.Int{Name: "max-per-region", Description: "Max number of VMs per region", Default: -1},
+	)
+	cmd.AddCommand()
```
matttpt, line 35: [AFAIK](https://github.com/spf13/cobra/blob/main/command.go#L1255) calling this with no arguments doesn't do anything. Safe to remove this (and the same thing in `memory`, `show`, and `vm`)?

matttpt (commented):
Two tiny things in comments above, otherwise LGTM!

internal/command/scale/count.go
dangra: good catch. fixing

internal/command/scale/count.go
dangra: it wasn't but I have to figure how to do it right now :) thanks

dangra: fixed! 296147e9

internal/command/scale/count.go
dangra: 296147e9 fixed it

matttpt (approved):

gwuah (approved):
LG

dangra:
thanks @matttpt @gwuah

PR #1871, opened 2023-03-09
add machine autostop config opt

senyosimpson:


tvdfly (commented):
what does this new field do/control? Like what happens when it's true and false?

Issue #1623, opened 2023-01-26
`fly launch` thinks `App is not running` and deploys over a running app


catflydotio:
`fly v0.0.451-pre-4 darwin/arm64 Commit: a52f9a0c BuildDate: 2023-01-20T14:37:55Z`

Obviously I'm reckless about where I deploy from. I just deployed an nginx image over my running Flask app. 

```
20:48:40 *[main][~/FlyTests/Jan25/hello-gunicorn-flask]$ fly open
opening http://jan25.fly.dev ...
20:48:45 *[main][~/FlyTests/Jan25/hello-gunicorn-flask]$ fly launch --image nginx --internal-port 80
An existing fly.toml file was found for app jan25
App is not running, deploy...
==> Building image
Searching for image 'nginx' remotely...
image found: img_wd57v5nge95v38o0
Deploying jan25 app with rolling strategy
  Waiting for e286065f969386 to become healthy: 0/1
^C^C^CINFO detected canceled context and allowing 500ms to release machine leases
```

I reproduced this to make sure I was interacting with the Flask app just before flyctl claimed "App is not running, deploy...".

catflydotio:
No longer an issue on v0.0.472.

catflydotio:
I've had this happen again in v0.0.476

catflydotio:
This means that if a user copies a fly.toml from a running app to use in a new app, the old app gets stomped on.

v0.0.477

```
8:46:12 *[main][~/FlyTests/march2/hello-vim-nano]$ fly status -a floo
App
  Name     = floo                  
  Owner    = personal              
  Hostname = floo.fly.dev          
  Image    = library/nginx:latest  
  Platform = machines              

Machines
ID              PROCESS VERSION REGION  STATE   HEALTH CHECKS           LAST UPDATED         
148eddeef77989  web     3       yyz     started                         2023-03-07T00:50:06Z
6e82d411c77287  app     3       yyz     started 1 total, 1 passing      2023-03-07T00:49:57Z
e2865641be9786  worker  3       yyz     started                         2023-03-07T00:50:02Z

8:46:18 *[main][~/FlyTests/march2/hello-vim-nano]$ fly launch
An existing fly.toml file was found for app floo
App is not running, deploy...
==> Building image
```

dangra:
@catflydotio please see https://github.com/superfly/flyctl/pull/1872#discussion_r1131418836

PR #1730, opened 2023-02-17
launch: Do not scan sources if config was copied from existing fly.toml

dangra:
When trying to launch a project with an existing fly.toml, the scanner will kick in even if the configuration was copied from existing toml file.

```
n$ fly launch --force-machines --name test-admin -o personal
An existing fly.toml file was found for app test-admin
? Would you like to copy its configuration to the new app? Yes
Creating app in /Users/daniel/src/flyio/fly-admin
Scanning source code
Detected a Phoenix app
? Choose a region for deployment: Ashburn, Virginia (US) (iad)
Created app test-admin in organization personal
Admin URL: https://fly.io/apps/test-admin
Hostname: test-admin.fly.dev
Set secrets on test-admin: SECRET_KEY_BASE
? Would you like to set up a Postgresql database now? No
? Would you like to set up an Upstash Redis database now? No
Error mix not found in $PATH - make sure app dependencies are installed and try again
```

After this change it simply skips the scanner which is not needed at deploy time
```
n$ fly launch --force-machines --name test-admin -o personal
An existing fly.toml file was found for app test-admin
? Would you like to copy its configuration to the new app? Yes
Creating app in /Users/daniel/src/flyio/fly-admin
? Choose a region for deployment: Ashburn, Virginia (US) (iad)
Created app test-admin in organization personal
Admin URL: https://fly.io/apps/test-admin
Hostname: test-admin.fly.dev
? Would you like to set up a Postgresql database now? No
? Would you like to set up an Upstash Redis database now? No
Wrote config file fly.toml
? Would you like to deploy now? No
Your app is ready! Deploy with `flyctl deploy`
```

PR #1872, opened 2023-03-09
Properly check if machine has already deployed before redeploying

gwuah:
Context
Machines don't have any concept of `allocations` (web always returns an empty array)
As a result, it's easily to redeploy an existing machines app with `fly launch`
This pull requests properly checks if a machine has already been deployed to prevent unwanted auto-deploys.
 
 
 fixes https://github.com/superfly/flyctl/issues/1623

### DIFF
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/flaps"
 	"github.com/superfly/flyctl/internal/appconfig"
 	"github.com/superfly/flyctl/internal/build/imgsrc"
 	"github.com/superfly/flyctl/internal/command"
@@ -78,6 +79,7 @@ func run(ctx context.Context) (err error) {
 	io := iostreams.FromContext(ctx)
 	client := client.FromContext(ctx).API()
 	workingDir := flag.GetString(ctx, "path")
+	cfg := appconfig.ConfigFromContext(ctx)
 
 	deployArgs := deploy.DeployWithConfigArgs{
 		ForceNomad:    flag.GetBool(ctx, "force-nomad"),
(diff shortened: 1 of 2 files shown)


internal/command/launch/launch.go
```diff
@@ -90,21 +92,16 @@ func run(ctx context.Context) (err error) {
...
-			return err
-		}
+	importedConfig, appConfig := false, appconfig.NewConfig()
 
+	// a fly toml already exists
+	if cfg != nil {
 		if cfg.AppName != "" {
 			fmt.Fprintln(io.Out, "An existing fly.toml file was found for app", cfg.AppName)
-			if deployExisting, err := shouldDeployExistingApp(ctx, cfg.AppName); err != nil {
+			if running, err := isAppRunning(ctx, cfg); err != nil {
```
dangra, line 102: IMHO we should remove this shortcut to fast deploy completely. It is evil and unexpected DX, we already ask people if they want to deploy at the end of `fly launch` and otherwise advise to  do it later with `fly deploy`

catflydotio: I agree; I like the idea of associating `fly launch` firmly with new-app setup magic, and `fly deploy` with deploying changes.

The fact that `fly launch` does this is, I believe, in response to the regular occurrence of users launching repeatedly when their first launch fails (which they can do because by default we'll give them new apps with generated names). It might not be the best possible solution though.

gwuah: I'm 100% with you @dangra  (I held the same opinion while fixing it)
But what's the consensus here? 

Should I rip it out? Or leave as is (@catflydotio 👁️ )

dangra: The concern about creating apps on every `fly launch` is valid tho.

We have to do something about it, 🧠 dumping options next:

1. When offering to select a new name and no `--app` is passed, offer the current appConfig.AppName
2. but show a message saying the app exists, "blah blah... should continue?"
3. If YES, we won't recreate the app but must check its platformVersion matches `fly launch` expectation.
4. ... no idea, i think that is all...

dangra: In order of preference:
1. `--generate-name` generate a random name if set
2. `--name` use this if set (IMO it should error if used in conjunction with `--generate-name`)
3. If interactive, show dialog to enter a new name, if fly.toml (aka appConfig) has a app name, use it as default value but let the user modify and confirm it
4. if non-interactive, fail as it does now.

Always check if app exists, and offer to proceed or stop there, and `--yes` to proceed no matter what

catflydotio: (Whatever behaviour you settle on, it should just be clear to users what it is and what they need to do.)

dangra: As it is right and once we remove the shortcut-unless-deployed, the scanner hooks are going to run no matter what. 

I wasn't sure we want to scan and run hooks on every `fly launch` retry but according to https://github.com/superfly/flyctl/pull/1730#issuecomment-1435264845 it is important not to skip them.

So, my proposal is to remove the shortcut, make it easy to reuse the app name from fly.toml, yet check if it exists and ask for confirmation to ~screw~ deploy into that existent app assuming all the consequences.

catflydotio: This makes sense to me.

dangra: @gwuah buying? do you think it is possible to implement this flow instead?

gwuah: yeah @dangra i should be able to pull this off. sounds good, but i'll give it some extra thought too.
(sorry for the late feedback, i've been immersing myself into networking @ fly)

gwuah: I've created a new pull request [here](https://github.com/superfly/flyctl/pull/1908).

For purposes of fixing the bug, we can merge this current pull request, and then rethink the user experience in https://github.com/superfly/flyctl/pull/1908

dangra: Not sure, since this discussion thread, we have changed the flow so much that this PR is not a bugfix anymore

gwuah:
See https://github.com/superfly/flyctl/pull/1908

PR #1878, opened 2023-03-09
Fixed crash for PG apps with no machines

### LINKED ISSUE #1873: fly status panics on pg cluster with no machines
https://flyio.sentry.io/issues/3239322312/events/fbaec7d673fd4f9abea45a90ba7aeaff/?project=4492967

billyb2:
Fixes #1873

### DIFF
--- a/internal/command/status/machines.go
+++ b/internal/command/status/machines.go
@@ -189,11 +189,16 @@ func renderPGStatus(ctx context.Context, app *api.AppCompact, machines []*api.Ma
 		client   = client.FromContext(ctx).API()
 	)
 
-	if postgres.IsFlex(machines[0]) {
-		yes, note := isQuorumMet(machines)
-		if !yes {
-			fmt.Fprintf(io.Out, colorize.Yellow(note))
+	if len(machines) > 0 {
+		if postgres.IsFlex(machines[0]) {
+			yes, note := isQuorumMet(machines)
+			if !yes {
+				fmt.Fprintf(io.Out, colorize.Yellow(note))
+			}
 		}
+	} else {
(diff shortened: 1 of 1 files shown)


alichay (approved):

alichay:
It might be nice to do something like what happens when you do `fly m list` on an empty app?
Definitely not a blocker, though
```
/tmp/pgf
❯ ~/Work/flyctl/bin/flyctl status -a ali-pg-test
ID	STATE	ROLE	REGION	HEALTH CHECKS	IMAGE	CREATED	UPDATED

/tmp/pgf
❯ ~/Work/flyctl/bin/flyctl m list -a ali-pg-test
No machines are available on this app ali-pg-test
```

internal/command/status/machines.go
```diff
@@ -189,11 +189,16 @@ func renderPGStatus(ctx context.Context, app *api.AppCompact, machines []*api.Ma
...
+	if len(machines) > 0 {
+		if postgres.IsFlex(machines[0]) {
+			yes, note := isQuorumMet(machines)
+			if !yes {
+				fmt.Fprintf(io.Out, colorize.Yellow(note))
+			}
 		}
+	} else {
+		fmt.Fprintf(io.Out, "No machines are available on this app %s\n", app.Name)
+		return
```
dangra, line 201: no need to change now but this is a common anti pattern in go code, it is much easier to follow the flow as:

```
if len(machines) == 0 {
  return
} 

.. continue the normal flow here.
```

billyb2: I had no idea, though when you say it it makes since. I'll keep that in mind, thank you

PR #1879, opened 2023-03-10
Added support for disabling autostart

### LINKED ISSUE #1804: Add support for disabling machine auto start
With https://github.com/superfly/flyctl/pull/1803, machines can now be configured to auto start (the default) or not when fly-proxy receives requests.

Ensure `fly deploy` copies this setting between machines. That is if a machine has a disable_machine_autostart config, make sure it doesn't change with the deployment.

Update `fly machine ...` sub commands to support this setting.

For now, we can probably skip adding support for this to fly.toml. At least until the feature is deployed across the fly platform (right now it's under active development, but getting close!).

billyb2:
Should fix #1804

### DIFF
--- a/internal/command/machine/list.go
+++ b/internal/command/machine/list.go
@@ -109,6 +109,8 @@ func runMachineList(ctx context.Context) (err error) {
 				volName = machine.Config.Mounts[0].Volume
 			}
 
+			fmt.Println(machine.Config.DisableMachineAutostart)
+
 			appPlatform := ""
 			machineProcessGroup := ""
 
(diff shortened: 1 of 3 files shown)


internal/command/machine/list.go
```diff
@@ -109,6 +109,8 @@ func runMachineList(ctx context.Context) (err error) {
 				volName = machine.Config.Mounts[0].Volume
 			}
 
+			fmt.Println(machine.Config.DisableMachineAutostart)
```
dangra, line 112: @billyb2 this slipped through

billyb2: No :(

PR #1880, opened 2023-03-10
Only prompt user to pick from adminned organizations for admin-only actions

btoews:
There are a few commands that can only be run by organization admins. When prompting the user to select an organization for these commands, we should only show organizations that they are admins on.

### DIFF
--- a/api/resource_organizations.go
+++ b/api/resource_organizations.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/superfly/flyctl/gql"
+	"github.com/superfly/graphql"
 )
 
 type OrganizationType string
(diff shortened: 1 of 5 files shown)


internal/command/orgs/orgs.go
```diff
@@ -78,23 +78,23 @@ func slugFromFirstArgOrSelect(ctx context.Context) (slug string, err error) {
...
+	if orgs, err = client.GetOrganizations(ctx, filters...); err != nil {
 		return
 	}
 	sort.OrganizationsByTypeAndName(orgs)
 
 	var org *api.Organization
 	if org, err = prompt.SelectOrg(ctx, orgs); prompt.IsNonInteractive(err) {
 		err = errSlugArgMustBeSpecified
-	} else {
+	} else if err == nil {
```
dangra, line 89: was it crashing on errors before?

btoews: It panicked if you ^C'ed at the prompt, yeah.

api/resource_organizations.go
```diff
@@ -14,10 +15,22 @@ const (
...
+	admin bool
+}
+
+func (f *organizationFilter) apply(req *graphql.Request) {
+	req.Var("admin", f.admin)
+}
+
+type OrganizationFilter func(*organizationFilter)
+
+var AdminOnly OrganizationFilter = func(f *organizationFilter) { f.admin = true }
```
dangra, line 28: my 2c, I find this to be an overcomplicated way of writing:

```go
func AdminOnly(f *organizationFilter) {
  f.admin = true
}
```

I understand you want to highlight AdminOnly fits OrganizationFilter type, but that is going to be catch by the compiler anyways, isn't it?

internal/command/orgs/orgs.go
dangra: nice

dangra (approved):
looks good

api/resource_organizations.go
btoews: Yeah. It's mostly for docs:

image

I could go either way.

dangra: fair point :) TIL

PR #1886, opened 2023-03-11
convert ruby apps from buildpacks to Dockerfile

rubys:


dangra:
@rubys I am not sure what to say about these changes. No doubt they look good but how have you tested them? will it break older ruby apps out there? does it matter?

At some point I would like to add test cases for scanners,  having a example fly.toml or files in a ruby project to test this as starting point would be awesome for my future self

rubys:
This pull request was motivated by https://community.fly.io/t/issue-deploying-new-sinatra-app/11355

I have tested the change.

A simple test case.  Run this in an empty directory:

```
bundle init
bundle add rackup
echo 'run { [200, {}, ["Hello World"]] }' > config.ru
fly launch
fly deploy
```

If you run this with the current released flyctl, it will incorrectly detect a Rails application, and things won't go well.  If you look at the Gemfile you will see  a comment that confuses the scanner.

If you run this with a flyctl that you built with this branch, it will correctly detect a Ruby app.  Pressing enter for all of the flyctl launch prompts is fine (no db, redis are required) will suffice.  

I plan to write a Ruby Dispatch article showing how to deploy a number of Ruby frameworks, including a pure Rack app (see above), Sinatra, and Camping.

dangra (approved):

PR #1888, opened 2023-03-13
Upgraded toml to fix bug for parsing Experimental

billyb2:
Also added a test case to avoid this bug in the future.

### DIFF
--- a/test/preflight/apps_v2_integration_test.go
+++ b/test/preflight/apps_v2_integration_test.go
@@ -374,3 +374,28 @@ func TestAppsV2ConfigSave_PostgresHA(t *testing.T) {
     timeout = "10s"
     path = "/flycheck/vm"`)
 }
+
+func TestAppsV2Config_ParseExperimental(t *testing.T) {
+	var (
+		err            error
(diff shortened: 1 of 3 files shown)


test/preflight/apps_v2_integration_test.go
```diff
@@ -374,3 +374,28 @@ func TestAppsV2ConfigSave_PostgresHA(t *testing.T) {
...
+	[experimental]
+	  auto_rollback = true
+	`
+
+	err = os.WriteFile(configFilePath, []byte(config), 0644)
+	if err != nil {
+		f.Fatalf("Failed to write config: %s", err)
+	}
+
+	result := f.Fly("launch --force-machines --local-only --name %s --region ord --copy-config", appName)
```
dangra, line 396: does the `--local-only` makes any difference?

dangra (approved):

dangra:
thanks!

test/preflight/apps_v2_integration_test.go
billyb2: Nope, a mistake with copying the command. i'll remove this

PR #1889, opened 2023-03-13
Set required v2 metadata on PG machines

davissp14:


flypg/launcher.go
```diff
@@ -304,6 +304,7 @@ func (l *Launcher) getPostgresConfig(config *CreateClusterInput) *api.MachineCon
 	// Metadata
 	machineConfig.Metadata = map[string]string{
 		"managed-by-fly-deploy": "true",
+		"fly_platform_version":  "v2",
```
dangra, line 307: what about doing it like  https://github.com/superfly/flyctl/blob/ae564072a863e385c112042b997d8dce25bf4168/internal/command/deploy/machines.go#L618

dangra: in fact, that function has this also https://github.com/superfly/flyctl/blob/ae564072a863e385c112042b997d8dce25bf4168/internal/command/deploy/machines.go#L624

dangra (approved):

PR #1893, opened 2023-03-13
`fly doctor`: get remote app config if it can't be loaded, return errors

### LINKED ISSUE #1874: fly doctor panics in some cases
https://flyio.sentry.io/issues/3246214906/events/49b19a0d11d5414ab0cc01a0555c6d51/?project=4492967

alichay:
Would fix #1874 

Note that this may break JSON formatting on errors. Other commands do it a lot, but doctor seemed to try hard to maintain *not* throwing out error messages and this does break that.

I'm open to making it display the output in a different way, maybe a `{"error": "somemsg"}` kind of thing, but swallowing the errors and silently failing seems a little strange.

### DIFF
--- a/internal/command/doctor/app_checks.go
+++ b/internal/command/doctor/app_checks.go
@@ -19,7 +19,6 @@ import (
 	"github.com/superfly/flyctl/internal/command/apps"
 	"github.com/superfly/flyctl/internal/state"
 	"github.com/superfly/flyctl/iostreams"
-	"github.com/superfly/flyctl/terminal"
 )
 
 type AppChecker struct {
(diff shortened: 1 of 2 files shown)


internal/command/doctor/app_checks.go
```diff
@@ -71,12 +64,20 @@ func NewAppChecker(ctx context.Context, jsonOutput bool, color *iostreams.ColorS
...
-		return nil
+		return nil, nil
 	}
 
 	ac.app = appCompact
 	ac.appConfig = appconfig.ConfigFromContext(ctx)
-	return ac
+
+	if ac.appConfig == nil {
+		ac.appConfig, err = appconfig.FromRemoteApp(ctx, ac.app.Name)
```
dangra, line 74: is flapsClient in the context? is common for it to be missing and the command crashes when trying to access the configuration for a non-deployed app.

the easiest way to check is something like:
```
fly apps create --machines my-app-name
fly doctor -a my-app-name
```

dangra: It doesn't crash:

```
$ fly apps create dangra-t9 -o personal
New app created: dangra-t9

$ fly doctor -a dangra-t9
Testing authentication token... PASSED
Testing flyctl agent... PASSED
Testing local Docker instance... PASSED
Pinging WireGuard gateway (give us a sec)... PASSED
dangra-t9 app has not been deployed yet. Skipping app checks. Deploy using `flyctl deploy`.
```

dangra (approved):

PR #1896, opened 2023-03-14
Add support for pnpm in Laravel Dockerfile

osbre:
pnpm is a very popular package manager now, I think it would be great to have it here.

- https://github.com/pnpm/pnpm
- https://pnpm.io/continuous-integration

I use it in all of my projects, and many people too.


dangra:
@osbre thanks for your contribution! 

At some point in the future I'll ask for the smallest app example to test this change, that will come once we add the plumbing to test flyctl's propject scanners

PR #1897, opened 2023-03-14
`fly env`: init flaps client before pulling remote config

alichay:
pretty self-explanatory, fixes the first bug in this forum thread: https://community.fly.io/t/machine-environment-errors/11421/2

I wanted to tackle the other bug from that thread in the same PR, but fixing secret deployment is going to be a little bit more involved so I want to go ahead and get this simple fix out.

### DIFF
--- a/internal/command/config/env.go
+++ b/internal/command/config/env.go
@@ -8,9 +8,9 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/flaps"
 	"github.com/superfly/flyctl/internal/appconfig"
 	"github.com/superfly/flyctl/internal/command"
-	"github.com/superfly/flyctl/internal/command/apps"
 	"github.com/superfly/flyctl/internal/flag"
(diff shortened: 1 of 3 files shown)


internal/command/config/env.go
```diff
@@ -47,6 +48,16 @@ func runEnv(ctx context.Context) error {
 		return err
 	}
 
+	app, err := apiClient.GetAppCompact(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("failed to get app: %s", err)
+	}
+
+	ctx, err = apps.BuildContext(ctx, app)
```
matttpt, line 56: I think using `flaps.New` and `flaps.NewContext` will do equally well here? It's just that `apps.BuildContext` always starts up an agent for Wireguard, but we default to `api.machines.dev` now. (`flaps.New` itself will start an agent when it needs it.)

alichay: you're so right. thanks!

matttpt:
Tiny nitpick (I sort of feel silly for even bringing it up): use `%w` instead of `%s` with `fmt.Errorf`?

Otherwise LGTM!

matttpt (approved):

Issue #1849, opened 2023-03-07
Unify `fly machines <command> <mid>` commands


alichay:
Some of the `fly m <command> <mid>` commands, like restart, work by looking up the first machine ID's app, and then using that app for the rest of the machines. This makes it easier to work with individual machines, where specifying the app each time can just feel like a hassle.

E.g. doing `fly m restart 0123 4567 8910` would look up the App associated with `0123`, and then act as if that app was passed in via `-a`.

Some commands, like the `fly m lease` family, don't work the same way. It'd be nice to make these all consistent.

At the same time, some of these commands feel like they *should* require an explicit app. Off the top of my head, `clone` actually adds a machine to the app. Conceptually, this feels like it should require an explicit app because it's not just affecting that single machine anymore.

(born out of conversation in #1834)

matttpt:
Done in #1899.

PR #1899, opened 2023-03-14
Consolidate selection and setup in machine commands

matttpt:
This PR was born out of a quick discussion in Slack. A number of machine commands use `apps.BuildContext`, which always sets up an agent for Wireguard. After the switch to `api.machines.dev` by default, this is usually not necessary. Furthermore, going to `api.machines.dev` means that we generally don't need to get an `AppCompact` from the API, since `flaps.New` needs it only for Wireguard mode.

Furthermore, the inconsistencies in how `fly machines <command> <mid>` works are behind #1849 and make it more complicated to wrap up `--select` for these commands (#1852).

This consolidates all app/machine selection and setup logic (including the creation of the flaps client for the app selected) into a new file, `select.go`. In doing so, it eliminates calls to `apps.BuildContext` and in most cases removes the API round-trip to get an `AppCompact`, which should speed things up.

Finally, by consolidating this logic in one place, we get consistent behavior for subcommands that take machine IDs. With the exception of clone, which still requires an app to be specified, these subcommands will now use the app of the first machine specified. All subsequent machines are required to be in the same app. Notably, this
    
* loosens the requirements for `leases` and
* **strengthens** the requirements for `start` and `stop`.

**If tightening up `start` and `stop` is probably going to break people's stuff,** I will revise to special-case these two.

(BTW, I've also pushed a full draft of #1852 based on a slightly older version of these changes.)

### DIFF
--- a/internal/command/machine/select.go
+++ b/internal/command/machine/select.go
@@ -2,7 +2,6 @@ package machine
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 
@@ -12,20 +11,8 @@ import (
 	"github.com/superfly/flyctl/internal/appconfig"
 )
 
-func selectOneMachine(ctx context.Context, machineID string) (*api.Machine, context.Context, error) {
-	return selectOneMachineImpl(ctx, machineID, nil)
-}
-
-func selectOneMachineFromApp(ctx context.Context, machineID string, app *api.AppCompact) (*api.Machine, context.Context, error) {
-	if app == nil {
-		return nil, nil, errors.New("tried to select a machine from a nil app; this is a bug")
-	}
-	return selectOneMachineImpl(ctx, machineID, app)
-}
-
-func selectOneMachineImpl(ctx context.Context, machineID string, app *api.AppCompact) (*api.Machine, context.Context, error) {
+func selectOneMachine(ctx context.Context, app *api.AppCompact, machineID string) (*api.Machine, context.Context, error) {
 	var err error
-
 	if app != nil {
 		ctx, err = buildContextFromApp(ctx, app)
 	} else {
--- a/internal/command/machine/select.go
+++ b/internal/command/machine/select.go
@@ -1,0 +1,121 @@
+package machine
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/flaps"
+	"github.com/superfly/flyctl/internal/appconfig"
+)
+
+func selectOneMachine(ctx context.Context, machineID string) (*api.Machine, context.Context, error) {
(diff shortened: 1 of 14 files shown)


internal/command/machine/select.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func selectOneMachineImpl(ctx context.Context, machineID string, app *api.AppCompact) (*api.Machine, context.Context, error) {
+	var err error
+
+	if app != nil {
+		ctx, err = buildContextFromApp(ctx, app)
+	} else {
+		ctx, err = buildContextFromAppNameOrMachineID(ctx, machineID)
+	}
```
alichay, line 33: I'm not sure I see the need for three functions here. Since `selectOneMachineImpl` already handles whether app is nil or not, it could just be one function, or if you want the enforcement that app is not nil in some cases, two functions.

internal/command/machine/start.go
```diff
@@ -57,28 +62,18 @@ func Start(ctx context.Context, machineID string) (err error) {
...
 	if err != nil {
 		switch {
-		case strings.Contains(err.Error(), "not found") && appName != "":
-			return fmt.Errorf("machine %s was not found in app %s", machineID, appName)
+		case strings.Contains(err.Error(), "not found"):
+			return fmt.Errorf("machine %s was not found in app '%s'", machineID, appName)
 		default:
 			return fmt.Errorf("could not start machine %s: %w", machineID, err)
 		}
 	}
```
alichay, line 73: Could this reuse `rewriteMachineNotFoundErrors`?

internal/command/machine/stop.go
```diff
@@ -65,21 +70,11 @@ func Stop(ctx context.Context, machineID string) (err error) {
...
-	}
-
-	err = flapsClient.Stop(ctx, machineStopInput)
+	err = flaps.FromContext(ctx).Stop(ctx, machineStopInput)
 	if err != nil {
 		switch {
-		case strings.Contains(err.Error(), "not found") && appName != "":
-			return fmt.Errorf("machine %s was not found in app %s", machineID, appName)
+		case strings.Contains(err.Error(), "not found"):
+			return fmt.Errorf("machine %s was not found in app '%s'", machineID, appName)
```
alichay, line 77: ditto last comment re: using `rewriteMachineNotFoundErrors`

alichay:
by the way, I love this.

dangra (commented):
I can't add to what @alichay already said but would like to reiterate the "love this changes" part

internal/command/machine/start.go
matttpt: Just updated `start`, `stop`, `destroy` and `kill` to use it.

alichay (approved):
yeah, this looks pretty good to me! probably a good idea to cut a prerel after merging and make sure nothing blows up bc this touches a lot, but it think it's good to go :)

matttpt:
Thanks @alichay and @dangra. Will absolutely cut a prerel after the merge!

PR #1904, opened 2023-03-15
Don't trash fly.toml if `fly launch` is copying configuration.

### LINKED ISSUE #1900: Apps v2 - `fly launch` overwriting `fly.toml` even when asked to use existing config
(App is a Dockerfile app and flyctl is v0.0.487)

To reproduce:

1. Start with a custom `fly.toml` in the working directory. Say internal port is 8043 instead of 8080.

2.:
```
$ fly launch
An existing fly.toml file was found for app garbage
? Would you like to copy its configuration to the new app? Yes
etc
```

3. Check fly.toml before deploy: `internal_port` has reverted to 8080.

ndarilek:
Fixes #1900

### DIFF
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -94,6 +94,7 @@ func run(ctx context.Context) (err error) {
 	appConfig := appconfig.NewConfig()
 
 	configFilePath := filepath.Join(workingDir, appconfig.DefaultConfigFileName)
+	copyConfig := false
 	if exists, _ := appconfig.ConfigFileExistsAtPath(configFilePath); exists {
 		cfg, err := appconfig.LoadConfig(configFilePath)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


dangra:
I'm not sure it is that simple, fly.toml could be a template or demo app and may not contain the app name.
Do we want to update fly.toml in that case?

There is also the issue with in-memory app config, that is going to be used for deployments if you opt to deploy within `fly launch`.

I think we need to think of what is the expected behaviour for the scanners and what to do with source info when services and other settings are already declared in fly.toml

/cc @catflydotio

catflydotio:
If I recall correctly, as of (say) a few months ago, "copying the configuration" from a local `fly.toml` meant that file was used to set the `appConfig.Definition` in memory, near the beginning, and then `fly.toml` always got overwritten from there in the end, even if there wasn't any meaningful change to it.

I'm out of my depth at this point. There's been a lot of activity on configs and I haven't had an eye on it. But it feels like @dangra's right and we need to zoom out to be sure the whole evolution of the app (and machine) config works together.

ndarilek:
Ah, somehow I suspected it wouldn't be this simple. :)

PR #1908, opened 2023-03-16
Rework `fly launch` UX

gwuah:
Title is descriptive enough but that should also fix https://github.com/superfly/flyctl/issues/1623

### DIFF
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -78,6 +78,12 @@ func run(ctx context.Context) (err error) {
 	io := iostreams.FromContext(ctx)
 	client := client.FromContext(ctx).API()
 	workingDir := flag.GetString(ctx, "path")
+	existingConfig := appconfig.ConfigFromContext(ctx)
+	generateName := flag.GetBool(ctx, "generate-name")
+	copyConfig := flag.GetBool(ctx, "copy-config")
+	name := strings.TrimSpace(flag.GetString(ctx, "name"))
+	appConfig := appconfig.NewConfig()
+	launchIntoExistingApp := false
 
 	deployArgs := deploy.DeployWithConfigArgs{
 		ForceNomad:    flag.GetBool(ctx, "force-nomad"),
@@ -90,42 +96,26 @@ func run(ctx context.Context) (err error) {
 		workingDir = absDir
 	}
 
-	var importedConfig bool
-	appConfig := appconfig.NewConfig()
-
 	configFilePath := filepath.Join(workingDir, appconfig.DefaultConfigFileName)
-	if exists, _ := appconfig.ConfigFileExistsAtPath(configFilePath); exists {
-		cfg, err := appconfig.LoadConfig(configFilePath)
-		if err != nil {
-			return err
-		}
 
-		if cfg.AppName != "" {
-			fmt.Fprintln(io.Out, "An existing fly.toml file was found for app", cfg.AppName)
-			if deployExisting, err := shouldDeployExistingApp(ctx, cfg.AppName); err != nil {
-				return err
-			} else if deployExisting {
-				fmt.Fprintln(io.Out, "App is not running, deploy...")
-				ctx = appconfig.WithName(ctx, cfg.AppName)
-				return deploy.DeployWithConfig(ctx, cfg, deployArgs)
-			}
+	if existingConfig != nil {
+		if existingConfig.AppName != "" {
+			fmt.Fprintln(io.Out, "An existing fly.toml file was found for app", existingConfig.AppName)
 		} else {
 			fmt.Fprintln(io.Out, "An existing fly.toml file was found")
 		}
 
-		copyConfig := false
-		if flag.GetBool(ctx, "copy-config") {
-			copyConfig = true
-		} else {
+		if !copyConfig {
 			copy, err := prompt.Confirm(ctx, "Would you like to copy its configuration to the new app?")
-			if copy && err == nil {
-				copyConfig = true
+			if err != nil {
+				return err
 			}
+
+			copyConfig = copy
 		}
 
 		if copyConfig {
-			appConfig = cfg
-			importedConfig = true
+			appConfig = existingConfig
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


internal/command/launch/launch.go
```diff
@@ -210,17 +200,43 @@ func run(ctx context.Context) (err error) {
...
 				return err
-			} else {
-				appConfig.AppName = inputName
 			}
-		} else {
-			appConfig.AppName = appName
+
+			if yes {
+				ctx = appconfig.WithName(ctx, cfg.AppName)
+				return deploy.DeployWithConfig(ctx, cfg, deployArgs)
```
dangra, line 234: The whole point of https://github.com/superfly/flyctl/pull/1872#discussion_r1131418836 was to remove this "shortcut to deploy".

`--yes` will continue with normal `fly launch` even if the app exists, see https://github.com/superfly/flyctl/pull/1872#discussion_r1131656793. `fly launch` will offer the user to deploy at the end of the flow as it normally does.

gwuah:
@dangra We're looking good. Done some testing on several cases and everything looks good. Would love for you to build my branch locally and run some tests. Just to ensure we've got the right flow.

There are alot of flags in the code path and it irks me a bit (they make it hard to reason about the flow) but I can't think of a better way to bypass them. (aside wrapping chunks in functions with very descriptive names and calling those sequentially)

Atm, the only thing left is the `--yes` flag, and the only reason why I haven't added it is because it's yet another flag.

gwuah:
yeah that’s a good catch, will rewire that.

internal/command/launch/launch.go
```diff
@@ -210,25 +200,59 @@ func run(ctx context.Context) (err error) {
...
-			} else {
-				appConfig.AppName = inputName
 			}
-		} else {
-			appConfig.AppName = appName
+			if !launchIntoExistingApp {
+				return nil
+			}
+
+			org = &api.Organization{
```
dangra, line 236: hmm, just use `org = app.Organization` ?

gwuah: i spent quite sometime on this actually.  so the type of `app.Organization` is `api.OrganizationBasic`, so I'm unable to assign directly to org, which has a type `api.Organization`.

I had a few options
- use a different query to check if app exists so I can get the organization type. (I didn't take this approach because the other app queries load more information than needed)
- manually create an api.Organization using api.OrganizationBasic (what I did)

I understand how this could cause problems if someone were to use "org" downstream expecting some other fields to exist.

internal/command/launch/launch.go
```diff
@@ -437,9 +432,9 @@ func shouldAppUseMachinesPlatform(ctx context.Context, orgSlug string) (bool, er
 	return orgDefault, nil
 }
 
-func appExists(ctx context.Context, cfg *appconfig.Config) (bool, *api.AppBasic, error) {
+func appExists(ctx context.Context, cfg *appconfig.Config) (bool, *api.App, error) {
 	client := client.FromContext(ctx).API()
-	app, err := client.GetAppBasic(ctx, cfg.AppName)
+	app, err := client.GetApp(ctx, cfg.AppName)
```
dangra, line 437: oh no, this call is way more expensive. if we have to do this then better to revert and use the previous attempt

internal/command/launch/launch.go
dangra: sorry I didn't read in time. You were right before.

internal/command/launch/launch.go
gwuah: ok i've reverted the commit.

dangra (approved):
LGTM!

catflydotio:
Here's the UX I think this is all about. When in an interactive terminal:

* `fly launch` always prompts for an app name, unless you use the `--name` flag to prepopulate that. The prompt is different if there's an app name specified inside a `fly.toml` file in the cwd.
    * if no name is found, and you don't enter one, it generates a new name
    * if a name is found, and you don't enter one, it proceeds with the name found in `fly.toml`.
    * if you do enter one, it could still be a name of an app that exists

* once the user has specified a name one way or another, flyctl checks to see if that app exists in your org (it doesn't check yet whether the name is available to use--it could still be taken by another org, which is caught later after you choose a region)
    * if this org has an app with that name, it asks you whether you "want to launch into that app". 
   
Is that right so far?

Here's where I'm still curious about some things (I haven't delved into the code to try to figure it out). I don't know what the definition of "launch into" is:

* I think this acts on the existing app (does not destroy the app and recreate it) -- because logs didn't disappear.
* I don't know whether I should expect IPs, certs, volumes, secrets to stay the same
* does it run launch scanners from scratch?

dangra:
> `fly launch` always prompts for an app name, unless you use the `--name` flag to prepopulate that. 

correct, and it shouldn't prompt with `--generate-name` flag 

> The prompt is different if there's an app name specified inside a `fly.toml` file in the cwd.
> 
> * if no name is found, and you don't enter one, it generates a new name
> * if a name is found, and you don't enter one, it proceeds with the name found in `fly.toml`.
> * if you do enter one, it could still be a name of an app that exists

that's correct.



> once the user has specified a name one way or another, flyctl checks to see if that app exists in your org (it doesn't check yet whether the name is available to use--it could still be taken by another org, which is caught later after you choose a region)

This is correct too.  To reiterate it, if the app exists but belongs to an organization the user has no access, it will fail later, while trying to create the app by that name.

While reading your comment my mind paused for a moment thinking about what to do with the region when:
* the user reuses the configuration and `primary_region` is set in fly.toml
* or if the app exists, it probably has a preferred region set too  

> * if this org has an app with that name, it asks you whether you "want to launch into that app".

Yes, although there is no concept of "this org", it gets the app data by name and fetches org details with it. 


> Here's where I'm still curious about some things (I haven't delved into the code to try to figure it out). I don't know what the definition of "launch into" is:

To be clear, "what `fly launch` is" hasn't changed on this PR except for what happens to existing apps whether they are *deployed* or not. Before this PR, you could run fly-launch on an existing non-deployed app and it reused IPs, volumes, certs and whatever other resources were associated with the app.

> * I think this acts on the existing app (does not destroy the app and recreate it) -- because logs didn't disappear.
Right, it doesn't recreate the app. It reuses it.

> * I don't know whether I should expect IPs, certs, volumes, secrets to stay the same

I think they should stay the same and `fly launch` should attempt to be idempotent, creating only the resources that are missing and reusing as much as it can of the existings.

> * does it run launch scanners from scratch?

It does but I don't think this is fully correct. I naively tried to disable them on https://github.com/superfly/flyctl/pull/1730 but it  needs more discussion and thinking. 

We can discuss here or in the forum, it should be a matter of sitting down, brain storm the list of use cases, extract what makes most sense from user POV while keeping it stupidly simple to avoid unexpected edge cases. It surely needs a few iterations. easy right? 😄 

Brain dumping for fun and profit:

1. The user wants to deploy his hand crafted app to Fly.io for the first time.
    * She has no idea about fly.toml, Dockerfile, or any other resource (ips, volumes) needed to run the app
2. The user generated an app using a _known framework_ (rails, nextjs, phoenix,...) and wants to deploy to Fly for the first time.
   * Same as previous case in term of non existent fly.toml
   * The framework may have generated a (in)valid Dockerfile
   * To support the Framework, the scanner runs scripts to generate local files ([smells badly](https://github.com/superfly/flyctl/blob/fe2aa5534b8f978eda71a70d95b99a06e1ce7827/scanner/phoenix.go#L56-L71))
3. The user found a  _fly-ready_ app (i.e. live_beats, [wordpress-sqlite](https://github.com/fly-apps/wordpress-sqlite#running-this-on-fly)) and expects to launch a personal copy of it.
    * "fly-ready" meaning fly.toml and Dockerfile exists
    *  Having to install a language development environment (phoenix, elixir, go, rails/ruby, ...) is detrimental 

Is it too simplistic to think in terms of this three variables: Known framework?, fly.toml exists?, Dockerfile exists?. 
* Launch an app for known framework, no fly.toml nor Dockerfile.
* Launch an app for known framework, no flytoml but existing Dockerfile
* Launch an app for known framework, fly.toml and Dockerfile exists
* Launch an app for known framework, fly.toml exists, no Dockerfile (weird case?)
* Launch an app for unknown framework, no fly.toml, no Dockerfile
* Launch an app for unknown framework, no fly.toml, Dockerfile exists
* Launch an app for unknown framework, fly.toml and Dockerfile exists
* Launch an app for unknown framework, fly.toml exists, no Dockerfile (weird again)
 
Notes:
* If fly.toml exists but user decides to ignore it, it is like "no fly.toml" case
* Rerunning `fly launch` will likely fall in the "fly.toml and Dockerfile" exists case
* While I was writing the list realized it could be better expressed as a table. 
* Operations on existing resources (ips, volumes, certs, ...) should be idempotent (do case by case analysis?)
* Framework hooks that require language specific development environments are a problem for "fly-ready" apps.
* Adding app services like databases (postgres, redis, ...) should be an idempotent operation too
* Move this discussion to the community forum

PR #1922, opened 2023-03-17
try linting api module for blocked deps

dangra:
The main motivation for this PR is to block imports from github.com/superfly/flyctl and watch new dependencies closer
image


alichay:
glad you were able to figure this out! I tried for like a couple hours but I could **not** get depguard to work in `api`

dangra:
> glad you were able to figure this out! I tried for like an hour or so but I could **not** get depguard to work in `api`

still tuning but I think I am close now

.golangci.yml
```diff
@@ -61,7 +61,8 @@ run:
   # can use regexp here: generated.*, regexp is applied on full path;
   # default value is empty list, but default dirs are skipped independently
   # from this option's value (see skip-dirs-use-default).
-  #skip-dirs:
+  skip-dirs:
+    - api/
```
alichay, line 65: Because `disable-all` is true in `api/.golangci.yml` anyway, couldn't we continue to run the main linters on `api`, and then only enable `gomodguard` in `api`? That way there's less duplicated configuration

this is totally a nitpick though

alichay (approved):

.golangci.yml
dangra: I think the best part of having different golangci.yml is that we can enable more linters or be more picky on what is accepted for the api vs flyctl.

the last bit was adding it to pre-commit at https://github.com/superfly/flyctl/pull/1922/commits/ae74695b130dbaed1cae46e7ce96587653acaacc

dangra: it's not a strong argument just a preference, happy to revisit it if not useful

gwuah (approved):
Looks good, but I have one request. Can we squash all the commits on this pull request into one commit?

Edit: 
I just realized "squash and merge" is disabled for this repo, so you'll have to squash locally and force push.

dangra:
> Can we squash all the commits on this pull request into one commit?

sure thing!

you will still see some new commits, I am still fleshing out some details

.golangci.yml
dangra: @alichay I'm about to merge but don't want without having a confirmation you are fine with it.

dangra:
@gwuah I rebased and squashed the commits, only left the main ones and 3 more that were what I used to validate it all works as expected.

gwuah (approved):
🍿

.golangci.yml
dangra: saw your approval! sorry

PR #1923, opened 2023-03-17
Don't use "Bearer" authorization scheme for new-style tokens

btoews:
Our new style of authentication tokens use `FlyV1` instead of the `Bearer` scheme in the `Authorization` header. This PR checks to see whether the `FLY_ACCESS_TOKEN` is prefixed with `FlyV1` before deciding to add `Bearer`.

### DIFF
--- a/api/auth.go
+++ b/api/auth.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 )
 
 // CLISessionAuth holds access information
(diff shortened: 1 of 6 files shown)


helpers/authentication.go
```diff
@@ -0,0 +1,15 @@
+package helpers
+
+import (
+	"fmt"
+	"strings"
+)
+
+const flyv1Scheme = "FlyV1"
+
+func AuthorizationHeader(token string) string {
```
dangra, line 10: this function is asking for a test case 🙏

dangra (approved):

dangra:
Oh, flyctl/api can't import flyctl/* and we use standard lib for testing in api module

btoews:
I'll move the helper into `flyctl/api` then if that's okay

dangra (approved):

PR #1928, opened 2023-03-20
fix: remove paid feature prompt when allocating shared IPv4

aofei:
According to https://community.fly.io/t/announcement-shared-anycast-ipv4/9384, I believe the "Shared Anycast IPv4" is a free feature, so `flyctl ips allocate-v4 --shared` should not print any paid feature warning (It does nothing but make people nervous).


dangra:
so true! thanks.

PR #1933, opened 2023-03-20
`bump_version.sh`: check that repository is up-to-date

alichay:
Nag people before allowing them to make a release if there seems to be unmerged commits in the remote repository


billyb2 (approved):
This is a good change

PR #1940, opened 2023-03-22
Warn if multiple build strategies are found

benwaffle:
We see some people in the community forum getting confused when
troubleshooting their build configuration. (e.g. changing a dockerfile
having no effect). This PR prints a warning when multiple different
build strategies are detected.

### DIFF
--- a/internal/appconfig/validation.go
+++ b/internal/appconfig/validation.go
@@ -31,6 +31,11 @@ func (cfg *Config) Validate(ctx context.Context) (err error, extra_info string) 
 		return err, extra_info
 	}
 
+	buildStrats := cfg.buildStrategies()
+	if len(buildStrats) > 1 {
+		return fmt.Errorf("More than one build configuration found: [%s]", strings.Join(buildStrats, ", ")), extra_info
+	}
+
 	switch platformVersion {
 	case MachinesPlatform:
 		err := cfg.EnsureV2Config()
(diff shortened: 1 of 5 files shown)


benwaffle:
Screenshot from 2023-03-22 16-14-21

internal/command/deploy/deploy.go
```diff
@@ -445,18 +447,18 @@ func mergeBuildArgs(ctx context.Context, args map[string]string) (map[string]str
 	return args, nil
 }
 
-func fetchImageRef(ctx context.Context, cfg *appconfig.Config) (ref string, err error) {
```
benwaffle, line 448: `err` was never set in this function

dangra (approved):
So, conflicting options are possible when fly.toml has some build options but `fly deploy` is called with a set of flags that changes how we want to build right? 

looks good anyways.

benwaffle:
@dangra 
Even just something like this in fly.toml is enough:
```
[build]
image = "nginx"
dockerfile = "api/Dockerfile"
builder = "heroku/buildpacks:20"
```

dangra:
@benwaffle sounds like something we could warn (or fail) on appConfig validation step

benwaffle:
@dangra I moved most of the checks to the validation stage (so bad confs will fail now). Just warning about an ignored dockerfile in the build stage.

internal/appconfig/validation.go
```diff
@@ -31,6 +31,11 @@ func (cfg *Config) Validate(ctx context.Context) (err error, extra_info string)
 		return err, extra_info
 	}
 
+	buildStrats := cfg.BuildStrategies()
+	if len(buildStrats) > 1 {
+		return fmt.Errorf("More than one build configuration found: [%s]", strings.Join(buildStrats, ", ")), extra_info
```
dangra, line 36: I like it, much cleaner but failing hard here could be too much at first for users, we are going to break people's pipelines without notice.

We need some way of warning them and send an event to sentry to know how many are there : https://github.com/superfly/flyctl/blob/287d00975f600fa22ce6eeebda2f21ed8829f956/internal/appconfig/toplevelcheck.go#L33

dangra (approved):
Looks good!  BuildStrategies() function can benefit from adding test cases tho.

benwaffle:
now looks like this:
image

benwaffle:
@dangra ok, I'm done with all the updates. Take another look.

internal/appconfig/validation.go
```diff
@@ -18,7 +19,7 @@ func (cfg *Config) Validate(ctx context.Context) (err error, extra_info string)
 		return errors.New("App config file not found"), ""
 	}
 
-	platformVersion := NomadPlatform
+	platformVersion := cfg.platformVersion
```
dangra, line 22: here be 🐉 but we can ⚔️ them

The only reason we call `GetAppBasic()` below is to get the platform version when it is unknown, but if it is already known we shouldn't be making an extra roundtrip to Web API.

But when it is unknown, meaning `cfg.PlatformVersion == nil`, we have to default to NomadPlatform to be backwards compatible in case the app doesn't exists.

internal/appconfig/validation.go
```diff
@@ -64,3 +73,30 @@ func (cfg *Config) Validate(ctx context.Context) (err error, extra_info string)
 	}
 
 }
+
+func (cfg *Config) BuildStrategies() []string {
+	strategies := []string{}
+
+	if cfg.Build == nil {
```
dangra, line 80: This is not tested. 

A common trap here is to forget to check for nil cfg, like in `cfg == nil || cfg.Build == nil`

benwaffle: interesting, TIL

internal/appconfig/validation.go
benwaffle: updated

dangra (approved):

PR #1941, opened 2023-03-22
Fix deepsource warnings

benwaffle:


matttpt:
Maybe use `//skipcq: GO-S2307` like [this](https://github.com/superfly/flyctl/commit/9ac1511dfe328bdb7f986e54097f2e2bfbcfa4d5)? Might keep the code a little cleaner.

benwaffle:
@matttpt thanks, updated

matttpt (approved):
Looks good. Apparently they [addressed](https://discuss.deepsource.io/t/go-s2307-false-positives/527) `io.ReadCloser` false positives, so maybe it regressed.

benwaffle:
Good find. I posted a comment letting them know

PR #1944, opened 2023-03-23
Don't clobber access_token in config with one from env

btoews:
Running `flyctl` with the `FLY_ACCESS_TOKEN` environment variable set results in it being written to the config file. This is disruptive when trying to run a single command with a token. I tried a few ways of fixing this and I think we ultimately don't want to load this environment variable into the viper config and want to consistently use `GetAPIToken()` for getting the appropriate token for use when running a command. I double checked that nothing else is relying on fetching `ConfigAPIToken` directly from viper.

### DIFF
--- a/flyctl/flyctl.go
+++ b/flyctl/flyctl.go
@@ -55,6 +55,11 @@ func initConfigDir() error {
 	return nil
 }
 
+// viper keys that shouldn't be loaded from the environment
+var noEnvKeys = map[string]bool{
+	ConfigAPIToken: true,
+}
+
 func initViper() {
 	if err := loadConfig(); err != nil {
 		fmt.Println("Error loading config", err)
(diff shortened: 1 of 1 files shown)


flyctl/flyctl.go
```diff
@@ -68,7 +73,14 @@ func initViper() {
 	viper.BindEnv(ConfigGQLErrorLogging, "GQLErrorLogging")
 
 	viper.SetEnvPrefix("FLY")
-	viper.AutomaticEnv()
+
+	for _, key := range viper.AllKeys() {
+		fmt.Printf("loading %s from env\n", key)
```
matttpt, line 78: I'm guessing this line is for debugging and can be removed?

matttpt:
Out of curiosity, what command were you running that exhibited this behavior?

flyctl/flyctl.go
btoews: Doh

btoews:
I was noticing it with `flyctl deploy`, but I think it's actually happening in the agent.

btoews:
It happens any time `SaveConfig()` is called. Looking at the call graph, there are a few places:

image

matttpt (approved):

PR #1945, opened 2023-03-23
flyctl auth token --deploy

btoews:


matttpt (commented):
Just curious, why include `tokens deploy` as an additional hidden command?

LGTM though, I'll come back and approve once Actions comes back to life!

btoews:
> Just curious, why include tokens deploy as an additional hidden command?

I don't have to do that. It just seemed like a nice shortcut for `fly tokens create deploy`, since `create` is the only option currently.

btoews:
Actions seem better now.

matttpt (approved):

PR #1946, opened 2023-03-23
Prevent docker scanner from overriding port set in fly.toml

gwuah:
this should fix https://github.com/superfly/flyctl/issues/1900 & https://community.fly.io/t/how-to-deploy-wordpress-on-fly/6274/10?u=griff


catflydotio:
Tried it out and it seems to work! 
* if I choose to copy config from a `fly.toml` with a custom port `<tomlport>`:
    * if my Dockerfile has `EXPOSE <dockerport>`, the app gets `internal_port = <dockerport>`
    * if my Dockerfile has no EXPOSE, the app gets internal_port = `<tomlport>`

* if I choose NOT to copy the config from `fly.toml`
    * if my Dockerfile has `EXPOSE <dockerport>`, the app gets `internal_port = <dockerport>`
    * if my Dockerfile has no EXPOSE, the app gets internal_port = 8080 (our default)

The fact that you split the logic up into different places instead of just changing dockerfile.go: does this indicate bigger plans?

gwuah:
> The fact that you split the logic up into different places instead of just changing dockerfile.go: does this indicate bigger plans?

Bigger plans? No. 
Making the changes this way was the simplest way i could fix this regression without touching too many things. But now that you mentioned "only dockerfile.go",  I have another idea. Imma need you to help test again though 🤪

internal/appconfig/config.go
```diff
@@ -147,3 +147,10 @@ func (c *Config) DockerBuildTarget() string {
 	}
 	return c.Build.DockerBuildTarget
 }
+
+func (c Config) GetInternalPort() int {
```
dangra, line 151: Getter doesn't "Get" it :) please see https://go.dev/doc/effective_go#Getters

internal/appconfig/config.go
```diff
@@ -147,3 +147,10 @@ func (c *Config) DockerBuildTarget() string {
...
+	if len(c.Services) > 0 {
+		return c.Services[0].InternalPort
```
dangra, line 153: there is a trick we need to account for, the "HttpService" field may not be nil and if present it is our reference service to get and set the internal port.

scanner/dockerfile.go
```diff
@@ -7,34 +7,45 @@ import (
...
-				} else if port == 8080 {
-					break
-				}
+	dockerfile, err := os.ReadFile(s.DockerfilePath)
+	if err != nil {
+		// just maintaining existing behaviour from old code.
+		return s, nil
+	}
+
+	re := regexp.MustCompile(`(?m)^EXPOSE\s+(?P<port>\d+)`)
```
dangra, line 31: it is better placed as a package constant at the top of the file. It will fail to compile if the regex is invalid

internal/appconfig/config.go
dangra: Oh, and there is also a SetInternalPort() that needs to take HttpService into account

internal/appconfig/config.go
gwuah: I only prefixed it with "Get" because I wanted to have one function that validates and safely extracts the port.
what would you suggest i name it, cos i really cant think of a better name.

scanner/dockerfile.go
gwuah: i'm not sure i follow. I can't set it to a constant either because that variable("re"), is actually mutated later on by the regexp pacakage.  I've moved it to the top of the file though.

dangra: My bad, you are right. I said "constant" where the intention was to declare it as a package private variable at the top of the file. 👍

internal/appconfig/config.go
dangra: just `InternalPort()` or `ServiceInternalPort` if you prefer.

internal/appconfig/config.go
gwuah: >there is a trick we need to account for, the "HttpService" field may not be nil and if present it is our reference service to get and set the internal port.

just so I understand what you mean @dangra 

are you saying `http_service` takes precedence over `service`? 👇 
```go
if c.HttpService != nil {
	return c.HttpService.InternalPort
}
	
if len(c.Services) > 0 {
	return c.Services[0].InternalPort
}
```

or are you saying in need to read from both services and http_service?

```go	
if len(c.Services) > 0 {
	return c.Services[0].InternalPort
}

if c.HttpService != nil {
	return c.HttpService.InternalPort
}
```

dangra: btw, getters and setters are easy to test and some already have test cases. Would be cool to use this opportunity to add test cases for InternalPort(), there one already for SetInternalPort() but doesn't consider HttpService.

catflydotio:
Naively I thought you could just change `dockerfile.go` so that it sets any `s.Port` iff there's a valid EXPOSE statement in the Dockerfile, leaving it untouched otherwise. Then I thought the existing appConfig would be untouched by that scanner's output in `launch.go`, thus keeping the services that were defined in either 1) our default starter config, or 2) the local `fly.toml` if the user said yes to copying config from that.

For my own edification I'd love to know if that wouldn't work!

internal/appconfig/config.go
gwuah: >Oh, and there is also a SetInternalPort() that needs to take HttpService into account

@dangra can you explain this though? this seems like new behaviour and im not too confident about introducing it in this pr. (for dockerfiles)

gwuah:
>Naively I thought you could just change dockerfile.go so that it sets an s.Port iff there's a valid EXPOSE statement in the Dockerfile, leaving it untouched otherwise. Then I thought the existing appConfig would be untouched by that scanner's output in launch.go, thus keeping the services that were defined in either 1) our default starter config, or 2) the local fly.toml if the user said yes to copying config from that.

I see where you're coming from. Only problem is, the fly.toml gets overwritten([1](https://github.com/superfly/flyctl/blob/9964513cd1a180f3c127a7950b20920f4724d6f2/internal/command/launch/launch.go#L339), [2](https://github.com/superfly/flyctl/blob/9964513cd1a180f3c127a7950b20920f4724d6f2/internal/command/launch/launch.go#L357)) anyway. ie. both fly.toml and appconfig gets overwritten.

This is actually why i was recommending showing a change diff to users, in my discourse post.

internal/appconfig/config.go
dangra: Oh, I was thinking on something like this:

```diff
diff --git a/internal/appconfig/setters.go b/internal/appconfig/setters.go
index efdbab52..d30baf8e 100644
--- a/internal/appconfig/setters.go
+++ b/internal/appconfig/setters.go
@@ -16,7 +16,9 @@ import (

 func (c *Config) SetInternalPort(port int) {
        c.v1SetInternalPort(port)
-       if len(c.Services) > 0 {
+       if c.HttpService != nil {
+               c.HttpService.InternalPort = port
+       } else if len(c.Services) > 0 {
                c.Services[0].InternalPort = port
        }
 }
```

I understand if you prefer not to in this PR, we will need it for V2 apps later anyways

catflydotio:
So I believe how it used to work was (approximately; I never quite got clear in my head everything that happened with Web) 

* you'd populate an initial `appConfig` with either a default config from Web, or a config read in from `fly.toml`.
* scanners would do their thing and contribute some subset of things into `srcInfo`
* a bunch of functions would run to change `appConfig` properties individually, only if `srcInfo` contained a value for them (that looks like it still happens in  [setAppconfigFromSrcinfo()](https://github.com/superfly/flyctl/blob/74f5be4147cb6b08201561b41521db1c91dde8ea/internal/command/launch/srcinfo.go#L219))
* `fly.toml` was always overwritten using the values in the final `appConfig` -- but even after that it might have no changed values relative to what it had before the scanners ran.

I'm too tired to figure out if this proves anything, but: If I remove the port-setting logic from dockerfile.go, the port I set in `fly.toml` is preserved.

catflydotio:
I apologise if I'm being ridiculous. Your changes seemed to work -- I just want to make sure I get it!

catflydotio:
(Also, don't let it block you that understanding hasn't caught up 😆 )

gwuah:
@catflydotio No you're not being ridiculous. I'm glad you're pushing this. You actually got the [flow](https://github.com/superfly/flyctl/pull/1946#issuecomment-1483633374) right. Infact, that's still the current flow.

The cause of the regression of the "default port" that was introduced in dockerfiles.go.

There are 2 ways we can fix this regression.
- Either by removing the "default port" from dockerfile.go and only setting a port when we find one in the dockerfile ( this is what you're suggesting) 
- Keeping the default port, but also making sure we don't override any existing port (this is what im doing)

My approach kinda wraps around yours. In the sense that, we don't always override, but we can also keep the fallback port which im sure is useful in certain cases. Your approach would have required only changing dockerfiles.go, but mine requires changes in 4 files.

So in summary, your approach would work. It's even more cleaner, but we'll loose the fallback_port(8080) feature.

catflydotio:
The last missing point for me is that I thought that port 8080 was always supplied by the default config at the start, so if no scanners grab the project, the downloaded `fly.toml` is a good default that would work for a basic standard app that needs an http service, and easy to customise. It's very possible that I've missed a change that would interfere with that--but it feels like that would have complicated things a lot!

internal/appconfig/config.go
gwuah: Ok I'll add it in a new PR. (so i can easily revert incase it breaks something)

gwuah:
> The last missing point for me is that I thought that port 8080 was always supplied by the default config at the start, so if no scanners grab the project, the downloaded `fly.toml` is a good default that would work for a basic standard app that needs an http service, and easy to customise. It's very possible that I've missed a change that would interfere with that--but it feels like that would have complicated things a lot!

@catflydotio You're right. You've been right the whole time. The config we load from web handles the "default port" feature.
Here's what I'll do. I've done some work here, so I'll merge this for now. and then create a new PR to cleanup and simplify the fix for this regression. thanks for following through with this.

catflydotio:
@dangra  My arguments here may be starting to go in the wrong direction if we're not going to have a starter fly.toml for v2 apps. I haven't spent much time figuring out if it changes anything for this particular case, just noticed your internal post about that today -- just wanted to flag it up.

PR #1948, opened 2023-03-24
Set a default node version to satisfy node-build

jsierles:
This should fix internal CI where `node-build` does not understand `latest` tag. This would only affect environments where node is not installed locally.


rubys:
Good catch!

PR #1949, opened 2023-03-24
New command: fly migrate-to-v2

tvdfly:
Work in progress, mainly missing some new gql mutations. The way this works is:

* Gets the app config, scaling info, and current allocs
* If there are volumes or autoscaling configs, fail (for now)
* Get a lock on the app (new gql mutation: not yet implemented)
* Create a release
* Scale nomad to zero and disable any nomad scaling (new gql mutation: not yet implemented)
* Update platform_version to machines (new gql mutation: not yet implemented)
* Create machines based on the existing allocs (eventually this will also make machines based on autoscale config)
* Acquires leases on the machines
* Unlock the app
* Release the machine leases
* Run a deployment over the machines
* Clean up the nomad job and any other nomad config (new gql mutation: not yet implemented)

The command is hidden for now, since we still need to make and integrate the gql mutations.

I have done zero testing except to check this compiles.


tvdfly:
I'm fixing the merge conflict, in part with: https://github.com/superfly/flyctl/pull/1951

internal/appconfig/config.go
```diff
@@ -147,3 +147,11 @@ func (c *Config) DockerBuildTarget() string {
 	}
 	return c.Build.DockerBuildTarget
 }
+
+func (c *Config) MountsDestination() string {
+	if c.Mounts == nil {
+		return ""
+	} else {
+		return c.Mounts.Destination
+	}
```
dangra, line 156: wdyt if we follow the same pattern than the other getters right above this one?

```suggestion
	if c == nil || c.Mounts == nil {
		return ""
	} 
	return c.Mounts.Destination
```

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -772,3 +790,17 @@ func determineAppConfigForMachines(ctx context.Context) (*appconfig.Config, erro
...
+func determineVmSpecs(vmSize api.VMSize) (*api.MachineGuest, error) {
+
+	preset := strings.Replace(vmSize.Name, "dedicated-cpu", "performance", 1)
+
+	guest := &api.MachineGuest{}
+	err := guest.SetSize(preset)
+
+	if err != nil {
+		return nil, fmt.Errorf("nomad VM definition incompatible with machines API: %w", err)
+	}
```
dangra, line 841: I think we also want to set `guest.MemoryMB = vmSize.MemoryMB` because that is something `fly scale vm --memory ` and `fly scale memory` commands allows to do.

alichay: oh! at some point I refactored machineGuest stuff to use .SetSize and I totally forgot that memory wasn't in there. thanks!

PR #1955, opened 2023-03-27
Warn about `http_service` `fly.toml` section for Nomad apps

matttpt:
The existing behavior is to silently ignore this section, which can cause a confusing user experience. I think that's what happened [here](https://community.fly.io/t/inbound-connections-to-80-443-get-reset-on-ipv6-only-app/11567) on the forum.

Following existing code, I made this a warning with a Sentry exception with an eye toward eventually making this fail validation.


dangra (approved):
loved the quick response

PR #1956, opened 2023-03-27
Close the docker client to close the connections

jeromegn:


dangra:
@jeromegn we have been adding ` //skipcq: GO-S2307` to `defer Close()` whenever it makes sense to make DeepSource analysis happy

jeromegn:
Thank you!

Issue #988, opened 2022-05-27
During ssh, when the remote is terminated, ssh hangs without remedy


robacarp:
**Describe the bug**

If I'm trying to diagnose a failing deploy, I have a small window of time to ssh to the vm before it gets terminated. Using `fly ssh connect` opens ssh, but when the vm is terminated I need to send QUIT to the local fly process in order to recover the terminal -- and then it dumps a stacktrace into my window.

Vanilla SSH natively provides for this with an escape sequence but it doesn't work when using `fly ssh connect`. I'd like to be able to send `~.` to my terminal in order to cause the local flyctl/ssh process to exit cleanly.

* fly v0.0.330 darwin/arm64 Commit: 851f549 BuildDate: 2022-05-26T17:59:56Z

robacarp:
Additionally, if an ssh command is slow or failing to connect, it cannot be interrupted during the "connecting to [ip::addr]" phase. Sending `^c` simply prints the control character until the spinner repaints. The process must be similarly QUIT-ted.

jsierles:
Thanks for the report. This is a known issue we just haven't gotten around to yet.

matttpt:
#1958 should fix the second part of this (^C unable to interrupt connection attempts).

redjonzaci:
@jsierles can we close this?

PR #1958, opened 2023-03-27
Allow connection attempts over agent tunnels to be cancelled

matttpt:
In particular, this allows users to ctrl-C their way out of a `fly ssh console` connection attempt when SSH on the instance isn't responding (something mentioned in #988).

### DIFF
--- a/agent/client.go
+++ b/agent/client.go
@@ -620,24 +620,35 @@ func (d *dialer) DialContext(ctx context.Context, network, addr string) (conn ne
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


gwuah (approved):
thx for the fix!

gwuah:
can you close all the linked & related issues to this bug when you merge this?

matttpt:
I'll drop a note in #988 saying that this aspect of it has been fixed. It also talks about hangs when the VM is terminated after the SSH connection has already been established, though, so unfortunately I don't think this PR totally closes it.

PR #1960, opened 2023-03-28
Fix agent deadlock when all wireguard peers are deleted

benwaffle:
...by upgrading wireguard-go. See https://community.fly.io/t/fly-ssh-console-just-hangs/11704

Upstream fix is https://github.com/WireGuard/wireguard-go/commit/fde0a9525ae5be7f6356a410f1bfc52ecc720f85


gwuah (approved):
This looks good, however I'm not entirely sure what the process is around updating dependencies and golang versions. Maybe you want to confirm in slack before merging this. 


Thanks for fixing this!

dangra (approved):
Upgrading to Go 1.20 is fine, it was about time to do it !

PR #1973, opened 2023-03-30
Fly proxy ctrl

senyosimpson:
Adds the `fly_proxy` machines option which is used to configure behaviour of the proxy specifically for machines

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -446,7 +446,7 @@ type MachineConfig struct {
 	AutoDestroy             bool                    `json:"auto_destroy,omitempty"`
 	DNS                     *DNSConfig              `json:"dns,omitempty"`
 	Statics                 []*Static               `json:"statics,omitempty"`
-	DisableMachineAutostart *bool                    `json:"disable_machine_autostart,omitempty"`
+	DisableMachineAutostart *bool                   `json:"disable_machine_autostart,omitempty"`
 	FlyProxy                *MachineFlyProxy        `json:"fly_proxy,omitempty"`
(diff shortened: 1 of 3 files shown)


dangra:
@senyosimpson does it supersede #1871  ?

senyosimpson:
> @senyosimpson does it supersede #1871 ?

yeah! i'll need a deprecation plan for that

dangra:
> > @senyosimpson does it supersede #1871 ?
> 
> yeah! i'll need a deprecation plan for that

it's not merged, we can just close it. I guess the deprecation plan is more for flaps/flyd side.

senyosimpson:
> > > @senyosimpson does it supersede #1871 ?
> > 
> > 
> > yeah! i'll need a deprecation plan for that
> 
> it's not merged, we can just close it. I guess the deprecation plan is more for flaps/flyd side.

ah yep correct

api/machine_types.go
```diff
@@ -379,6 +379,11 @@ type MachineServiceConcurrency struct {
 	SoftLimit int    `json:"soft_limit,omitempty" toml:"soft_limit,omitempty"`
 }
 
+type MachineFlyProxy struct {
+	AutostartMachine bool `json:"autostart_machine"`
+	AutostopMachine  bool `json:"autostop_machine"`
```
dangra, line 384: same thing here, `*bool` is the only way we can know when a flag is not set and left to server defaults.

dangra (approved):
pre-approving in case I am not around

Issue #1838, opened 2023-03-07
Apps v2 -`fly services list` doesn't show services after cloning a Machine


catflydotio:
This might have something to do with process groups too?

To reproduce:

```
$ fly launch --image nginx --force-machines
...
? Would you like to deploy now? Yes
==> Building image
Searching for image 'nginx' remotely...
image found: img_e1zd4mkn80yp02yw
Provisioning ips for gah
  Dedicated ipv6: 2a09:8280:1::f553
  Shared ipv4: 66.241.125.38
  Add a dedicated ipv4 with: fly ips allocate-v4
No machines in gah app, launching one new machine
  Machine 3d8d9900ce2289 [app] update finished: success
  Finished deploying

```

Check the status; health check is passing
```
$ fly status
App
  Name     = gah                   
  Owner    = personal              
  Hostname = gah.fly.dev           
  Image    = library/nginx:latest  
  Platform = machines              

Machines
ID              PROCESS VERSION REGION  STATE   HEALTH CHECKS           LAST UPDATED         
3d8d9900ce2289  app     1       yyz     started 1 total, 1 passing      2023-03-07T01:14:56Z
```
`fly services list` is fine:

```
$ fly services list
Services
PROTOCOL        PORTS                   FORCE HTTPS 
TCP             80 => 80 [HTTP]         True       
                443 => 80 [TLS,HTTP]    False      
```

Add some process groups to config:

```
[processes]
  app = ""
  web = "nginx -g 'daemon off;'"
  worker = "tail -F /dev/null"
```

(I got an error trying to deploy without adding the `"app"` line to `fly.toml`)

Clone the machine to one of the process groups            
```
$ fly machine clone --process-group web 3d8d9900ce2289
Cloning machine 3d8d9900ce2289 into region yyz
INFO Setting process group to web for new machine and updating cmd, services, and checks
Provisioning a new machine with image registry-1.docker.io/library/nginx:latest@sha256:942ae2dfd73088b54d7151a3c3fd5af038a51c50029bfcfd21f1e650d9579967...
  Machine 4d891de2f6e587 has been created...
  Waiting for machine 4d891de2f6e587 to start...
No health checks found
Machine has been successfully cloned!
```

Now check on services again:
```
$ fly services list                                   
Services
PROTOCOL        PORTS   FORCE HTTPS 

20:18:13 [~/FlyTests/march2/nginx2]$ 
```

(Services are still actually OK: can visit the app in the browser and `fly status` gives
```
[...]

PR #1975, opened 2023-03-30
Add process groups to fly.toml on `fly launch`

gwuah:
This pull request spawned from https://github.com/superfly/flyctl/issues/1838 (I got an error trying to deploy without adding the "app" line to fly.toml). 

Causes: 
1. we weren't setting platform_version. or better put, we were [setting](https://github.com/superfly/flyctl/blob/cc4c187fa25a0f87f5a6235a122e6d0ea70986ee/internal/command/launch/launch.go#L278) it, but [overriding](https://github.com/superfly/flyctl/blob/cc4c187fa25a0f87f5a6235a122e6d0ea70986ee/internal/command/launch/launch.go#L305) it almost immediately. 
And so this [block](https://github.com/superfly/flyctl/blob/cc4c187fa25a0f87f5a6235a122e6d0ea70986ee/internal/appconfig/serde.go#L95) downstream, never run.

2. Even for nomad apps, we weren't adding it to rawData before writing to file. 

This pull request addresses both issues.

### DIFF
--- a/internal/appconfig/serde.go
+++ b/internal/appconfig/serde.go
@@ -113,6 +113,9 @@ func (c *Config) marshalTOML(w io.Writer) error {
 	if c.HttpService != nil {
 		rawData["http_service"] = c.HttpService
 	}
+	if len(c.Processes) > 0 {
+		rawData["processes"] = c.Processes
+	}
 
 	if len(rawData) > 0 {
 		// roundtrip through json encoder to convert float64 numbers to json.Number, otherwise numbers are floats in toml
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -296,6 +296,12 @@ func run(ctx context.Context) (err error) {
 			newCfg.Build = appConfig.Build
 			newCfg.PrimaryRegion = appConfig.PrimaryRegion
 			appConfig = newCfg
+
+			// remove auto-rollback from machine fly.tomls
+			if shouldUseMachines {
+				appConfig.Experimental = nil
+			}
+
 		} else {
 			appConfig.AppName = createdApp.Name
 		}
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -273,13 +273,6 @@ func run(ctx context.Context) (err error) {
 		return err
 	}
 
-	if shouldUseMachines && copyConfig {
-		// Check imported fly.toml is a valid V2 config before creating the app
-		if err := appConfig.SetMachinesPlatform(); err != nil {
-			return fmt.Errorf("Can not use configuration for Apps V2, check fly.toml: %w", err)
-		}
-	}
-
 	if !launchIntoExistingApp {
 		input := api.CreateAppInput{
 			Name:            appConfig.AppName,
@@ -307,6 +300,22 @@ func run(ctx context.Context) (err error) {
 			appConfig.AppName = createdApp.Name
 		}
 		fmt.Fprintf(io.Out, "Created app '%s' in organization '%s'\n", appConfig.AppName, org.Slug)
+	}
+
+	if shouldUseMachines {
+		// Check imported fly.toml is a valid V2 config before creating the app
+		if err := appConfig.SetMachinesPlatform(); err != nil {
+			return fmt.Errorf("Can not use configuration for Apps V2, check fly.toml: %w", err)
+		}
+
+		// remove auto-rollback from machine fly.tomls
+		appConfig.Experimental = nil
+
+		// set default process group
+		// instead of using "", get the entrypoint.sh from the image and stash it there. eg. "nginx -g daemon off" for nginx images
+		if len(appConfig.Processes) == 0 {
+			appConfig.SetProcess(appConfig.DefaultProcessName(), "")
+		}
 	}
 
 	fmt.Fprintf(io.Out, "Admin URL: https://fly.io/apps/%s\n", appConfig.AppName)
@@ -348,11 +357,6 @@ func run(ctx context.Context) (err error) {
 	// Override internal port if requested using --internal-port flag
 	if n := flag.GetInt(ctx, "internal-port"); n > 0 {
 		appConfig.SetInternalPort(n)
-	}
-
-	// remove auto-rollback from machine fly.tomls
-	if shouldUseMachines {
-		appConfig.Experimental = nil
 	}
 
 	// Finally write application configuration to fly.toml


internal/appconfig/serde.go
```diff
@@ -113,6 +113,9 @@ func (c *Config) marshalTOML(w io.Writer) error {
 	if c.HttpService != nil {
 		rawData["http_service"] = c.HttpService
 	}
+	if len(c.Processes) > 0 {
+		rawData["processes"] = c.Processes
+	}
```
dangra, line 119: I think we have a bug somewhere else because `SetProcess()`  should set `"processes"` in https://github.com/superfly/flyctl/blob/cc4c187fa25a0f87f5a6235a122e6d0ea70986ee/internal/appconfig/setters.go#L209-L227 and `SanitizedDefinition()` doesn't remove it.

internal/command/launch/launch.go
```diff
@@ -309,6 +302,22 @@ func run(ctx context.Context) (err error) {
...
+		}
+
+		// remove auto-rollback from machine fly.tomls
+		appConfig.Experimental = nil
+
+		// set default process group
+		// instead of using "", get the entrypoint.sh from the image and stash it there. eg. "nginx -g daemon off" for nginx images
+		if len(appConfig.Processes) == 0 {
+			appConfig.SetProcess(appConfig.DefaultProcessName(), "")
+		}
```
dangra, line 318: this can't be right either 🤔 

the moment we have to force storing a process here means we are not doing the right thing when it is missing on reader side. There are tons of existing fly.tomls without `[processes]` section and this change won't fix the issue for them

internal/command/launch/launch.go
```diff
@@ -309,6 +302,22 @@ func run(ctx context.Context) (err error) {
 		fmt.Fprintf(io.Out, "Created app '%s' in organization '%s'\n", appConfig.AppName, org.Slug)
 	}
 
+	if shouldUseMachines {
+		// Check imported fly.toml is a valid V2 config before creating the app
+		if err := appConfig.SetMachinesPlatform(); err != nil {
```
dangra, line 307: Nice catch! Moving this here is a win.

internal/command/launch/launch.go
```diff
@@ -309,6 +302,22 @@ func run(ctx context.Context) (err error) {
...
+			return fmt.Errorf("Can not use configuration for Apps V2, check fly.toml: %w", err)
+		}
+
+		// remove auto-rollback from machine fly.tomls
+		appConfig.Experimental = nil
```
dangra, line 312: There is a subtle difference on nullifying the experimental section here instead of where it was before.

Before it was after the call to `setAppconfigFromSrcinfo()` which among other things sets some experimental features like docker entrypoint and docker command.

Honestly, I still think your change is better because we initial intention for experimental was to remove "AutoRollback" sent by Web when fetching app config for a new app. In summary, I'd move it to https://github.com/superfly/flyctl/blob/42c0ac74d8b687ac16e4e6d111300466d0c886cb/internal/command/launch/launch.go#L295-L297 so new apps won't get experimental by default but existent doesn't remove it silently.

Additionally, we probably need a warning for `fly deploy` to v2 apps when experimental features are set.

internal/command/launch/launch.go
gwuah: this felt wrong anyway to be fair lmao

internal/appconfig/serde.go
dangra: > the reason why i don't think it's a bug is because "SetProcess" isn't called in this path.
> Path => fly launch --image nginx --force-machines (with deploy)

Ok, but this is a machine's only path and since you fixed the bug with `SetMachinesPlatform()` there is no need to `rawData["processes"] = c.Processes` here because that is Nomad only path, right?

> when we are creating machines for an appv2 deployment, we default to the "app" process group, but we forget to update the fly.toml file 

We don't forget, we assume a missing `processes` section means a single process group named `app`. That is the same assumption Nomad does for v1 apps.

In cases where a process group is not defined, we have to fallback to [Config.DefaultProcessName()]( 
https://github.com/superfly/flyctl/blob/cc4c187fa25a0f87f5a6235a122e6d0ea70986ee/internal/appconfig/processconfig.go#L89-L93)

internal/command/launch/launch.go
gwuah: Ok. I'll make this change now.

internal/appconfig/serde.go
gwuah: >Ok, but this is a machine's only path and since you fixed the bug with SetMachinesPlatform() there is no need to rawData["processes"] = c.Processes here because that is Nomad only path, right?

Yes, you are right. I have to delete that line. 

>We don't forget, we assume a missing processes section means a single process group named app. That is the same assumption Nomad does for v1 apps.

**tldr**: we use a default process group("app") when you don't provide one, but fly.toml has no idea about that default process group.

Right. But our deployment doesn't match what is in the fly.toml. We have a machine with progress_group "app", but our fly.toml has no idea about app and it's causing certains errors. 
Here's how to reproduce the error
1. in an empty folder, run `fly launch --image nginx --force-machines` and deploy it
2. add the section below to your fly.toml
```
[processes]
  web = "nginx -g 'daemon off;'"
  ```
3. run `fly deploy` 
4. you'll get an [error](https://github.com/superfly/flyctl/blob/cc4c187fa25a0f87f5a6235a122e6d0ea70986ee/internal/appconfig/processconfig.go#L71).


I can fix this error in the code, but i'm not 100% satisfied with that solution. (normal users can use the app process group too, so we don't want to make too many assumptions on it)

Comment (dangra) [reply internal/appconfig/serde.go:119]: @catflydotio we will need a third 👁️  to solve this :)

@guawh dumping here the repro to visualize it for later: 

```
$ fly launch --image nginx --force-machines
Creating app in /Users/daniel/src/apps/flyctl1965
Using image nginx
? Choose an app name (leave blank to generate one): f1965
? Select Organization: Daniel Graña (personal)
Some regions require a paid plan (fra, maa).
See https://fly.io/plans to set up a plan.

? Choose a region for deployment: Santiago, Chile (scl)
Created app 'f1965' in organization 'personal'
Admin URL: https://fly.io/apps/f1965
Hostname: f1965.fly.dev
? Would you like to set up a Postgresql database now? No
? Would you like to set up an Upstash Redis database now? No
Wrote config file fly.toml
? Would you like to deploy now? Yes
==> Building image
image found: img_e1zd4m90oynv02yw
Provisioning ips for f1965
  Dedicated ipv6: 2a09:8280:1::f:52a
  Shared ipv4: 66.241.125.64
  Add a dedicated ipv4 with: fly ips allocate-v4
 * create 1 "app" machine
No machines in group 'app', launching one new machine
  Machine 4d891d15a23d87 [app] update finished: success
  Finished deploying
```

fly.toml at this point:
```toml
# fly.toml file generated for f1965 on 2023-03-30T15:09:10-03:00

app = "f1965"
kill_signal = "SIGINT"
kill_timeout = 5
mounts = []
primary_region = "scl"
processes = []

[build]
  image = "nginx"

[[services]]
  internal_port = 8080
  processes = ["app"]
  protocol = "tcp"
  [services.concurrency]
    hard_limit = 25
    soft_limit = 20
    type = "connections"

  [[services.ports]]
    force_https = true
    handlers = ["http"]
    port = 80

  [[services.ports]]
    handlers = ["tls", "http"]
    port = 443
```

Updated `processes` to:
```toml
[processes]
web = "nginx -g 'daemon off;'"
```

and run `fly deploy`:
```
$ fly deploy
==> Verifying app config
Validating /Users/daniel/src/apps/flyctl1965/fly.toml
Platform: machines
✓ Configuration is valid
--> Verified app config
==> Building image
image found: img_e1zd4m90oynv02yw
Error error service specifies 'app' as one of its processes, but no processes are defined with that name; update fly.toml [processes] to include a app process
```

The error states missing the process definition for `app`  because **there is a service that refers to it** 
```
$ grep \"app fly.toml  -B3

[[services]]
  internal_port = 8080
  processes = ["app"]
```

If we edit that line and replace `app` by `web`, it works:
```
$ fly deploy
==> Verifying app config
Validating /Users/daniel/src/apps/flyctl1965/fly.toml
Platform: machines
✓ Configuration is valid
--> Verified app config
==> Building image
image found: img_e1zd4m90oynv02yw
 * destroy 1 "app" machine
 * create 1 "web" machine
machine 4d891d15a23d87 was found and is currently in started state, attempting to destroy...
No machines in group 'web', launching one new machine
  Machine 73287413b13185 [web] update finished: success
  Finished deploying
Deploying f1965 app with rolling strategy
  Finished deploying
```

dangra: Now that I reread everything I think I get what you are suggesting to do. 

Add an explicit `app` process to fly.toml so the user doesn't fall in this common gotcha when adding a process by himself.

I think `app` is so known as the default process name that we can show a better error message asking the user if he added a processes section and forgot to add an entry for `app`

dangra: A bit side topic, we could caught this mistakes on appconfig validation
https://github.com/superfly/flyctl/blob/cc4c187fa25a0f87f5a6235a122e6d0ea70986ee/internal/appconfig/validation.go#L56

gwuah: im wondering if we should have a short call.

dangra:
superseded by https://github.com/superfly/flyctl/pull/2079

PR #1976, opened 2023-03-30
Show all running services when app has more than 1 machine

gwuah:
This should fix https://github.com/superfly/flyctl/issues/1838

### DIFF
--- a/internal/command/services/machines.go
+++ b/internal/command/services/machines.go
@@ -47,24 +47,27 @@ func ShowMachineServiceInfo(ctx context.Context, app *api.AppInfo) error {
 
 	services := [][]string{}
 
-	for _, service := range machines[0].Config.Services {
-		for i, port := range service.Ports {
-			protocol := service.Protocol
-			if i > 0 {
-				protocol = ""
+	for _, machine := range machines {
+		for _, service := range machine.Config.Services {
+			for i, port := range service.Ports {
+				protocol := service.Protocol
+				if i > 0 {
+					protocol = ""
+				}
+
+				handlers := []string{}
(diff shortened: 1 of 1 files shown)


dangra (commented):
How does it look when there are multiple machines with same service?

gwuah:
> the number of machines that serves that service
Sorry but I'm confused. I thought machines had a 1 to M relation with services?
So only 1 machine can service M amount of services?

I think 1 & 2 are good additions.

dangra:
> Sorry but I'm confused. I thought machines had a 1 to M relation with services? So only 1 machine can service M amount of services?

hmm no, it is a "N to M" relation:
* A machine can serve M services
* A service can be served from N machines
* The machine and service must be linked by the same process group
* Machine's regions determine the regions from where a service can be served

internal/command/services/machines.go
```diff
@@ -46,30 +48,70 @@ func ShowMachineServiceInfo(ctx context.Context, app *api.AppInfo) error {
...
+	_ = render.Table(io.Out, "Services", services, "Protocol", "Ports", "Handlers", "Force HTTPS", "Process Group", "Regions", "Machines")
 
 	return nil
 }
+
+func getServiceKey(protocol, ports, forcehttps, handlers string) string {
+	return fmt.Sprintf("%s-%s-%s-%s", protocol, ports, forcehttps, handlers)
+}
+
+func removeDuplicates(values []string) []string {
```
dangra, line 105: I think you are looking for [lo.Uniq](https://github.com/samber/lo#uniq)

dangra (approved):

PR #1977, opened 2023-03-30
`migrate-to-v2`: add rollback support

alichay:


internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -167,34 +184,107 @@ func NewV2PlatformMigrator(ctx context.Context, appName string) (V2PlatformMigra
...
+	finalPlatformVersion := "nomad"
 
-	tb.Detail("Locking app to prevent changes during the migration")
-	err = m.lockAppForMigration(ctx)
-	if err != nil {
-		return err
+	fmt.Printf("Recovery info: %v\n", m.recovery)
+
+	if m.recovery.machinesCreated {
+		// TODO: Unsure how to recover here.
```
gwuah, line 199: > If you already have machines, we can just scale those down and bring nomad back up, right?
this is from slack. would you want to implement it before we merge this?
or that's a different convo?

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -167,34 +184,107 @@ func NewV2PlatformMigrator(ctx context.Context, appName string) (V2PlatformMigra
...
+	// Hook into Ctrl+C so that aborting the migration
+	// leaves the app in a stable, unlocked, non-detached state
+	{
+		signalCh := make(chan os.Signal)
+		abortSignals := []os.Signal{os.Interrupt}
+		if runtime.GOOS != "windows" {
+			abortSignals = append(abortSignals, syscall.SIGTERM)
+		}
+		// Prevent ctx from being cancelled, we need it to do recovery operations
+		signal.Reset(abortSignals...)
```
gwuah, line 269: [q] does this thing un-cancel the context?

gwuah (approved):
looks good. got like 2 questions but that shouldn't block merging.

gwuah (approved):

Issue #1971, opened 2023-03-30
flyctl platform status --json opens a web browser.


gedw99:
```
fly version
fly v0.0.500 darwin/amd64 Commit: f71fe0ef BuildDate: 2023-03-29T16:56:42Z
flyctl platform status --json 
```

So that we can monitor using tooling please. Opening a web page is not helpful.

alichay:
Fixed in version `0.0.505`, thanks for reporting!

gedw99:
thanks you @alichay 

works a treat :)

PR #1978, opened 2023-03-30
Fix/status json

shortdiv:
Addressing #1971

This command now gives a quick status check on overall fly health in json form. I'll probably extend status with subcommands to get specific status updates on incidents, and scheduled maintenance as well


internal/command/platform/status.go
```diff
@@ -2,31 +2,135 @@ package platform
...
+		client, _ := createClient(ctx, url)
+		req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/%s", client.baseUrl, getStatusEndpoint), nil)
+		if err != nil {
+			log.Fatal(err)
+		}
+		var res *http.Response
+
+		if res, err = client.httpClient.Do(req); err != nil {
+			return err
+		}
```
dangra, line 118: Looks like you can replace the whole thing with:

```suggestion
		httpClient, err := api.NewHTTPClient(logger.MaybeFromContext(ctx), http.DefaultTransport)
		if err != nil {
		    return err
		}
		res, err := httpClient.Get(url+getStatusEndpoint)
		if err != nil {
		    return err
		}
```

No need for `createClient` function and if concatenating url paths looks odd (it does), there is https://pkg.go.dev/net/url#JoinPath

internal/command/platform/status.go
```diff
@@ -2,31 +2,135 @@ package platform
...
+		defer res.Body.Close()
+
+		if res.StatusCode != 200 {
+			err = api.ErrorFromResp(res)
+			return fmt.Errorf("failed to retrieve status: %w", err)
+		}
+
+		var result = StatusPage{}
```
dangra, line 126: we can use a generic map here because the only goal is to pass it down to `render.JSON()`
```suggestion
		var result = map[string]any{}
```

internal/command/platform/status.go
```diff
@@ -2,31 +2,135 @@ package platform
...
+		if err = json.NewDecoder(res.Body).Decode(&result); err != nil {
+			fmt.Println(err)
```
dangra, line 128: no need to print the error, just return err and it will be handled by the caller

internal/command/platform/status.go
```diff
@@ -2,31 +2,135 @@ package platform
...
 	const (
-		long = `Show current Fly platform status in a browser
+		long = `Show current Fly platform status in a browser or via json with the json flag
 `
-		short = "Show current platform status"
+		short = "Show current platform status with an optional filter for maintenance or incidents in json mode (eg. incidents, maintenance)"
 	)
 
-	cmd = command.New("status", short, long, runStatus)
+	cmd = command.New("status [kind](optional)", short, long, runStatus)
```
dangra, line 31: Sorry for the nitpick but in [BNF grammar](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form) the square brackets mean the parameter is optional, somehow I found the extra `(optional)` text confusing but it could be just me.

internal/command/platform/status.go
```diff
@@ -2,31 +2,135 @@ package platform
...
+		req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/%s", client.baseUrl, getStatusEndpoint), nil)
+		if err != nil {
+			log.Fatal(err)
+		}
+		var res *http.Response
+
+		if res, err = client.httpClient.Do(req); err != nil {
+			return err
+		}
+		defer res.Body.Close()
```
dangra, line 119: DeppSource analysis will complain about this line, for read only ops we ignore the check like in:

https://github.com/superfly/flyctl/blob/89bd22db756bd903249e9eecf16f49dfa26d97a1/scanner/helpers.go#L32

internal/command/platform/status.go
```diff
@@ -2,31 +2,135 @@ package platform
...
+		cfg               = config.FromContext(ctx)
+		getStatusEndpoint string
+		getStatusKind     = flag.FirstArg(ctx)
+	)
+
+	switch getStatusKind {
+	case "incidents":
+		getStatusEndpoint = "api/v2/incidents/unresolved.json"
+	case "maintenance":
+		getStatusEndpoint = "api/v2/scheduled-maintenances/active.json"
```
dangra, line 103: what do you think if it accepts `status` or empty string and return an error otherwise?

internal/command/platform/status.go
```diff
@@ -2,31 +2,135 @@ package platform
...
+
+	return &Client{
+		baseUrl:    cleanedURL,
+		httpClient: httpClient,
+	}, nil
 }
 
 func runStatus(ctx context.Context) error {
-	const url = "https://status.fly.io/"
+	const url = "https://status.fly.io"
```
dangra, line 92: this is a good candidate for a package level constant

internal/command/platform/status.go
shortdiv: Oh neat!

internal/command/platform/status.go
shortdiv: til, I wasn't sure if the square brackets meant optional

internal/command/platform/status.go
shortdiv: Do you mean for the kind part of the command so `fly platform status status` or `fly platform status ` and those including maintenance and incidents?

dangra: Yes. Although `fly platform status status` reads weird

dangra: The whole point is that `fly platform status unknown-word` should return an error, not the status.

shortdiv: ah ok gotcha

dangra (changes requested):
It's looking good! only part missing is the removal of the unused types (StatusPage, Incident,...) and `createClient` function.

dangra (approved):

PR #1983, opened 2023-03-31
V2 fly scale count

dangra:
*NOTE: Scaling machines with mounts is not supported yet.*

```
$ fly scale show
VM Resources for app: scalecmd

Groups
NAME    COUNT   KIND            CPUS    MEMORY  REGIONS
app     3       shared          4       1024 MB mia,scl(2)
other   3       shared          4       1024 MB ord(2),scl
task    4       performance     1       2048 MB iad,mia,ord,scl
```

## Scale up a new region for default group:
```
$ fly scale count 3 --region ord -y
App 'scalecmd' is going to be scaled according to this plan:
  +3 machines for group 'app' on region 'ord' with size 'shared-cpu-4x'
Executing scale plan
  Created 6e8257da00de78 group:app region:ord size:shared-cpu-4x
  Created 5683566bd32938 group:app region:ord size:shared-cpu-4x
  Created 6e82947a232168 group:app region:ord size:shared-cpu-4x

$ fly scale show
VM Resources for app: scalecmd

Groups
NAME    COUNT   KIND            CPUS    MEMORY  REGIONS
app     6       shared          4       1024 MB mia,ord(3),scl(2)
other   3       shared          4       1024 MB ord(2),scl
task    4       performance     1       2048 MB iad,mia,ord,scl
```

## Scale up to 3 instances and one max per region
```
$ fly scale count app=3 --max-per-region 1 -y
App 'scalecmd' is going to be scaled according to this plan:
  -1 machines for group 'app' on region 'scl' with size 'shared-cpu-4x'
  -2 machines for group 'app' on region 'ord' with size 'shared-cpu-4x'
Executing scale plan
  Destroyed 73d8d97df9d989 group:app region:scl size:shared-cpu-4x
  Destroyed 6e82947a232168 group:app region:ord size:shared-cpu-4x
  Destroyed 6e8257da00de78 group:app region:ord size:shared-cpu-4x

$ fly scale show
VM Resources for app: scalecmd

Groups
NAME    COUNT   KIND            CPUS    MEMORY  REGIONS
app     3       shared          4       1024 MB mia,ord,scl
other   3       shared          4       1024 MB ord(2),scl
task    4       performance     1       2048 MB iad,mia,ord,scl
```

## Max per region works on different groups at the same time
```
$ fly scale count other=1 task=2 --max-per-region 1 -y
App 'scalecmd' is going to be scaled according to this plan:
  -2 machines for group 'other' on region 'ord' with size 'shared-cpu-4x'
  -1 machines for group 'task' on region 'mia' with size 'performance-1x'
  -1 machines for group 'task' on region 'iad' with size 'performance-1x'
Executing scale plan
  Destroyed 5683939cd077d8 group:other region:ord size:shared-cpu-4x
  Destroyed e784921cee0d78 group:other region:ord size:shared-cpu-4x
  Destroyed 9080d13f699287 group:task region:mia size:performance-1x
  Destroyed 328744eef01e78 group:task region:iad size:performance-1x

$ fly scale show
VM Resources for app: scalecmd

Groups
NAME    COUNT   KIND            CPUS    MEMORY  REGIONS
app     3       shared          4       1024 MB mia,ord,scl
other   1       shared          4       1024 MB scl
task    2       performance     1       2048 MB ord,scl
```

## Scaling to zero and back up on multiple regions
```
$ fly scale count app=0
App 'scalecmd' is going to be scaled according to this plan:
  -1 machines for group 'app' on region 'mia' with size 'shared-cpu-4x'
  -1 machines for group 'app' on region 'ord' with size 'shared-cpu-4x'
  -1 machines for group 'app' on region 'scl' with size 'shared-cpu-4x'
? Scale app scalecmd? Yes
Executing scale plan
  Destroyed 3d8d549ae50489 group:app region:mia size:shared-cpu-4x
  Destroyed 5683566bd32938 group:app region:ord size:shared-cpu-4x
  Destroyed e286369be9e286 group:app region:scl size:shared-cpu-4x

$ fly scale count app=5 --region scl,ord,iad
App 'scalecmd' is going to be scaled according to this plan:
  +1 machines for group 'app' on region 'iad' with size 'shared-cpu-4x'
  +2 machines for group 'app' on region 'scl' with size 'shared-cpu-4x'
  +2 machines for group 'app' on region 'ord' with size 'shared-cpu-4x'
? Scale app scalecmd? Yes
Executing scale plan
  Created 91857709a43328 group:app region:iad size:shared-cpu-4x
  Created 24d89106f25487 group:app region:scl size:shared-cpu-4x
  Created e78473df2dd983 group:app region:scl size:shared-cpu-4x
  Created 1781969c53ede8 group:app region:ord size:shared-cpu-4x
  Created 908017ea11d008 group:app region:ord size:shared-cpu-4x

$ fly scale show
VM Resources for app: scalecmd

Groups
NAME    COUNT   KIND            CPUS    MEMORY  REGIONS
app     5       shared          4       1024 MB iad,ord(2),scl(2)
other   1       shared          4       1024 MB scl
task    2       performance     1       2048 MB ord,scl
```

### DIFF
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -31,7 +31,7 @@ func runMachinesScaleCount(ctx context.Context, appName string, expectedGroupCou
 	}
 	ctx = appconfig.WithConfig(ctx, appConfig)
 
-	machines, err := mach.AppV2ListActive(ctx)
+	machines, _, err := flapsClient.ListFlyAppsMachines(ctx)
 	if err != nil {
 		return err
 	}
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -2,6 +2,7 @@ package scale
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/samber/lo"
@@ -41,6 +42,10 @@ func runMachinesScaleCount(ctx context.Context, appName string, expectedGroupCou
 	if err != nil {
 		return err
 	}
+	// Only machines that are part of apps-v2 platform
+	machines = lo.Filter(machines, func(m *api.Machine, _ int) bool {
+		return m.IsFlyAppsPlatform()
+	})
 
 	machines, releaseFunc, err := mach.AcquireLeases(ctx, machines)
 	defer releaseFunc(ctx, machines)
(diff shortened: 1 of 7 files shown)


internal/command/scale/count_machines.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	ctx = appconfig.WithConfig(ctx, appConfig)
+
+	machines, err := mach.AppV2ListActive(ctx)
+	if err != nil {
+		return err
+	}
+
+	if len(machines) == 0 {
+		fmt.Fprintf(io.Out, "There are not active machines for this app. Did you ever `fly deploy`? if not, do it now\n")
```
alichay, line 40: this almost comes off as kind of aggro to me, perhaps something along the lines of "There are no active machines for this app. Have you run 'fly deploy' on this application yet?"

internal/machine/query.go
```diff
@@ -22,3 +22,15 @@ func ListActive(ctx context.Context) ([]*api.Machine, error) {
...
+func AppV2ListActive(ctx context.Context) ([]*api.Machine, error) {
+	machines, err := ListActive(ctx)
+	if err != nil {
+		return nil, err
+	}
+	machines = lo.Filter(machines, func(m *api.Machine, _ int) bool {
+		return m.IsFlyAppsPlatform()
+	})
+	return machines, nil
+}
```
alichay, line 36: Does this do the same thing as `flaps.ListFlyAppsMachines`?

internal/command/scale/count_machines.go
tvdfly:
```suggestion
		fmt.Fprintf(io.Out, "There are no active machines for this app. Run `fly deploy` to create one, and then re-run this command.\n")
```

tvdfly: oh actually, let's return this message as the error message. It will get printed for the developer and I like it when errors tell developer how to correct them.

tvdfly (commented):
Nice work!

Should we validate some of the parameters passed in? For example, if a developer tries to do:

```
fly scale count 99
```

And they only have one machine, might be helpful to triple check that they want 99 machines.  We could do something simple like: if requested count > 10x current count, add an extra prompt.

My perspective is flyctl shouldn't prevent developers from trying to create machines. We'll rely on our apis to manage any limits. The validation in flyctl is more to help developers identify and avoid accidents/mistakes.

Have we tested the failure mode when a developer hits the machine limit for their org? I suspect they would get as many machines as possible and then it would fail, and that seems like a reasonable DX. I can also test it later if someone else doesn't get there first :-)

dangra:
> Should we validate some of the parameters passed in? For example, if a developer tries to do:
> 
> ```
> fly scale count 99
> ```
> 
> And they only have one machine, might be helpful to triple check that they want 99 machines. We could do something simple like: if requested count > 10x current count, add an extra prompt.

```
$ fly scale count 99
App 'scalecmd' is going to be scaled according to this plan:
 +32 machines for group 'app' on region 'scl' with size 'shared-cpu-4x'
 +31 machines for group 'app' on region 'iad' with size 'shared-cpu-4x'
 +31 machines for group 'app' on region 'ord' with size 'shared-cpu-4x'
? Scale app scalecmd? (y/N)
```

Right now it asks for confirmation unless `--yes` flag is passed. Do you mean that it should ask for another confirmation because the increase is so off (10x) that is suspicious?  that could work. Another option is to emphasize the counts with colors (green, yellow, red) depending on how big is the jump.

> My perspective is flyctl shouldn't prevent developers from trying to create machines. We'll rely on our apis to manage any limits. The validation in flyctl is more to help developers identify and avoid accidents/mistakes.

I agree.



> Have we tested the failure mode when a developer hits the machine limit for their org? I suspect they would get as many machines as possible and then it would fail, and that seems like a reasonable DX.

Right, that it is what would happen.

internal/machine/query.go
dangra: Yes, kind of. I think flaps and `internal/machine/query.go` needs so factoring to keep concerns separate but you are right that is basically doing the same.

there is also https://github.com/superfly/flyctl/blob/68223a9c393f91de16eddb74fc7be70890c99cef/internal/machine/query.go#L11

I didn't want to change either of them until having a better picture of how these functions are used, so added a new function, sorry.

internal/command/scale/count_machines.go
dangra: Thanks for looking into those messages! English is not my first lang and sometimes the tone doesn't work the way I think. I will make the wording changes asap.

internal/machine/query.go
dangra: addressed in 005d53726d0ad71594d1a4f2d60cba6a1de1e13a

internal/command/scale/count_machines.go
dangra: Fixed by 84ca42c2

alichay (approved):
good stuff!

PR #1986, opened 2023-03-31
Normalize SSH behavior

matttpt:
I'm not committed to everything in this PR, but I wanted to get it out here for discussion.

I think there's a general expectation for `fly ssh console` to work similarly to OpenSSH, which by default will allocate a PTY only for login sessions. See [these](https://community.fly.io/t/ssh-console-c-outputs-r/7774) forum [posts](https://community.fly.io/t/fly-ssh-console-c-cmd-fails/11555) for examples. Actually, even Fly's official docs include a [script](https://fly.io/docs/apps/migrate-to-v2/#copy-any-secrets-you-need-from-your-existing-app) that uses `tr` to remove carriage returns from `ssh console` output, because PTY output is getting piped into the next command, not the raw stdout of the command run over SSH.

I've tried to make `fly ssh console` work more like OpenSSH. If we get an explicit command with `ssh console -C`, it won't allocate a PTY now. Using `--pty` will force PTY allocation. (This is like `-t` for `ssh`. Unfortunately we can’t actually use `-t` because it’s already used for providing an access token.) I've made `sh`, `bash`, `/bin/sh`, and `/bin/bash` exceptions for the time being, in case people are used to using `-C /bin/bash` and getting a PTY allocated. It'll print a warning when it does this, though.

Additional notes:
- When `flyctl`'s stdin reaches EOF, we need to close the stdin of the command over SSH. This is fixed. (Amusingly Hallpass doesn’t close the process's stdin when SSH stdin closes, so it has to be fixed there too.)
- On Unix systems, `flyctl` will now forward the `TERM` environment variable to SSH.
- All this is just for `ssh console`. I believe I've avoided doing anything that will change how Postgres commands that use SSH work. They'll still allocate a PTY with `TERM=xterm` like before.
- **Possible alternative:** `fly ssh exec` or something like that for running ad-hoc commands without a PTY?

### DIFF
--- a/ssh/io.go
+++ b/ssh/io.go
@@ -2,6 +2,7 @@ package ssh
 
 import (
 	"context"
+	"errors"
 	"io"
 	"runtime"
 	"sync"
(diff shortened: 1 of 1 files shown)


dangra (approved):
I'm buying 💰

PR #1988, opened 2023-03-31
handle ctrl-c signals

gwuah:
attempt to unlock before killing the process. this helps ensure we don't leave an app in a weird state.


alichay:
superseded by #1989 :)

PR #1994, opened 2023-04-03
Fix pre release

gwuah:


gwuah:
don't worry this won't be merged. not worth it.

PR #1996, opened 2023-04-03
Restrict prereleases to prerelease branch

gwuah:


alichay:
If we're restricting this, should we go ahead and restrict full releases to master?

gwuah:
Makes sense I guess.

alichay (approved):

alichay:
thanks for hardening all of this up!

PR #1998, opened 2023-04-04
Fix deployment image release bug

tvdfly:
Before this change, the image was not set on releases when there were no machines and no image was built.

After this change, in cases where the image was previously not set it now gets set to the image from the previous release.


alichay (approved):
great catch!

PR #1999, opened 2023-04-04
Remove '--volume' flag from 'fly machine update'

pborzenkov:
`flyd` doesn't allow to attach volumes to already created machines.
So remove the flag to avoid confusion.


pborzenkov:
Scratch that. It looks like it should allow to change the volumes during `update`, so removing the flag is incorrect.

PR #2007, opened 2023-04-04
Longer default expiry for deploy tokens

btoews:


jphenow (approved):
Only thing I'd do then maybe is update recommendations on help output and docs. But I think this makes sense as a default for what most people will expect from "didn't actively set an expiration"

PR #2009, opened 2023-04-04
Clarify deploy error when app name is missing from config or has not been created

shortdiv:
Currently fly deploy on a dir with fly.toml but missing app name gives a `couldn't find a fly.toml` error, which is misleading.
Screen Shot 2023-04-04 at 4 52 58 PM

This PR updates the error messaging for greater clarity to the user.

appnotadded
appnamenotfound

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -536,7 +536,7 @@ func RequireAppName(ctx context.Context) (context.Context, error) {
 	}
 
 	if name == "" {
-		err := fmt.Errorf("we found a fly.toml file but it looks like the app has not been created, did you create this app with fly apps create?", buildinfo.Name())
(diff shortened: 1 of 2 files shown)


tvdfly (approved):

internal/command/command.go
```diff
@@ -536,7 +536,8 @@ func RequireAppName(ctx context.Context) (context.Context, error) {
 	}
 
 	if name == "" {
-		return nil, errRequireAppName
+		err := fmt.Errorf("the config for your app is missing an app name, did you add it in the fly.toml file or via -a? %s", buildinfo.Name())
```
tvdfly, line 539: nit (take it or leave it):

```suggestion
		err := fmt.Errorf("the config for your app is missing an app name, add an app_name field to the fly.toml file or specify with the -a flag`")
```

PR #2012, opened 2023-04-05
Migrate nomad pg to v2

DAlperin:
needs https://github.com/fly-apps/postgres-ha/pull/99
Is also untested


internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -294,6 +313,14 @@ func (m *v2PlatformMigrator) Migrate(ctx context.Context) (err error) {
 		}()
 	}
 
+	if m.isPostgres {
+		tb.Detail("Upgrading postgres image")
+		err := m.updateNomadImage(ctx)
```
alichay, line 318: could this be something like `updateNomadPgImage` if it's postgres-specific?

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -303,6 +330,11 @@ func (m *v2PlatformMigrator) Migrate(ctx context.Context) (err error) {
 		return abortedErr
 	}
 
+	err = m.setNomadPgReadonly(ctx)
+	if err != nil {
+		return err
+	}
```
alichay, line 336: This looks like it should be gated to only postgres migrations

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -360,6 +392,18 @@ func (m *v2PlatformMigrator) Migrate(ctx context.Context) (err error) {
...
+
+	dbUIDs, err := m.getPgDBUids(ctx, inRegionDbs)
+	if err != nil {
+		return err
+	}
+
+	err = m.waitForPGSync(ctx, dbUIDs)
+	if err != nil {
+		return err
+	}
```
alichay, line 405: Ditto re: being gated to only postgres migrations

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -682,6 +979,7 @@ func (m *v2PlatformMigrator) resolveMachineFromAlloc(alloc *api.AllocationStatus
 	launchInput.Config.Image = m.img
 	launchInput.Config.Env = lo.Assign(m.appConfig.Env)
 	if m.appConfig.PrimaryRegion != "" {
+		m.primaryRegion = m.appConfig.PrimaryRegion
```
alichay, line 982: Nit: I feel like this spiritually belongs in determinePrimaryRegion
```diff
func (m *v2PlatformMigrator) determinePrimaryRegion(ctx context.Context) error {
+    defer func(){
+        m.primaryRegion = m.appConfig.PrimaryRegion
+    }()
```

Only mentioning because this only runs once machines are being created, and that might lead to a confusing dependency chain if something tries to use `m.primaryRegion` before machines are created

alichay (commented):
This is all from the perspective of "how does this impact apps migrations," I'm wholly unqualified to comment on the postgres work itself

and, this diff's not _that_ bad!

internal/command/deploy/machines.go
```diff
@@ -90,6 +91,9 @@ func NewMachineDeployment(ctx context.Context, args MachineDeploymentArgs) (Mach
 	if err != nil {
 		return nil, err
 	}
+	if args.NewVolumeName != "" {
+		appConfig.Mounts.Source = args.NewVolumeName
```
alichay, line 95: Could `appConfig.Mounts` ever be nil?

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -206,11 +214,22 @@ func NewV2PlatformMigrator(ctx context.Context, appName string) (V2PlatformMigra
...
+		consul, err := apiClient.EnablePostgresConsul(ctx, appCompact.Name)
+		if err != nil {
+			return nil, err
+		}
+		migrator.pgConsulUrl = consul.ConsulURL
+	}
+	//err = migrator.validate(ctx)
+	//if err != nil {
+	//	return nil, err
+	//}
```
alichay, line 232: assuming this and the other commented out block in here are in in-development-thing-i-don't-need-to-worry-about land and won't end up affecting app migrations

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -421,6 +446,29 @@ func (m *v2PlatformMigrator) Migrate(ctx context.Context) (err error) {
...
+			return err
+		}
+
+		tb.Detailf("We are checking the following DB UIDs %+v", dbUIDs)
+
+		err = m.waitForPGSync(ctx, dbUIDs)
+		if err != nil {
+			return err
+		}
+	}
```
alichay, line 470: do we want this after the app unlock? it seems like we'd want this before that if at all possible, but I'm not entirely qualified to make that judgement :)

alichay (commented):
no nasty side effects as far as I can tell. reviewed & tested normal deploys/migrations. seems fine!

rugwirobaker:
Folks will have to upgrade their image first right?:
```
DEBUG flypg will connect to: http://fdaa:0:d1:a7b:276d:1:4bab:2:5500
DEBUG --> POST http://fdaa:0:d1:a7b:276d:1:4bab:2:5500/commands/admin/readonly/enable
DEBUG <-- 404 http://fdaa:0:d1:a7b:276d:1:4bab:2:5500/commands/admin/readonly/enable (4.65s)
DEBUG 404 page not found
```

alichay:
> Folks will have to upgrade their image first right?:

There's code sketched out to do this, just not implemented fully yet
https://github.com/superfly/flyctl/blob/e32a04a10074ce56b48ecbf94fd2ce5e47f6a11d/internal/command/migrate_to_v2/migrate_to_v2.go#L344-L350

internal/command/migrate_to_v2/migrate_to_v2.go
DAlperin: I want it after the release is done, which currently happens after the lock so I'm just inheriting. If we can move the release up it would be sweet to move this up with it

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -455,10 +546,290 @@ func (m *v2PlatformMigrator) Migrate(ctx context.Context) (err error) {
...
+	// Find all volumes
+	for _, vol := range app.Volumes.Nodes {
+		if strings.Contains(vol.Name, "machines") {
+			continue
+		}
+		if _, ok := regionsToVols[vol.Region]; ok {
+			regionsToVols[vol.Region] = append(regionsToVols[vol.Region], vol)
+		} else {
+			regionsToVols[vol.Region] = []api.Volume{vol}
+		}
```
alichay, line 629: this can be written like so:
```suggestion
		regionsToVols[vol.Region] = append(regionsToVols[vol.Region], vol)
```

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -455,10 +546,290 @@ func (m *v2PlatformMigrator) Migrate(ctx context.Context) (err error) {
...
+				AppID:     app.ID,
+				Name:      fmt.Sprintf("%s_machines", vol.Name),
+				Region:    region,
+				SizeGb:    vol.SizeGb,
+				Encrypted: vol.Encrypted,
+			}
+			if len(vol.Snapshots.Nodes) > 0 {
+				//TODO(dov) figure out if this assumption about ordering is correct
+				input.SnapshotID = &vol.Snapshots.Nodes[len(vol.Snapshots.Nodes)-1].ID
+			}
```
alichay, line 647: is it to ship this with these unresolved?

alichay (approved):
couple questions, couple assignments that CI caught that seem pointless. aside from those, looks pretty good!

internal/appconfig/validation.go
```diff
@@ -45,7 +45,7 @@ func (cfg *Config) Validate(ctx context.Context) (err error, extra_info string)
 	case NomadPlatform:
 		platErr, platExtra := cfg.ValidateForNomadPlatform(ctx)
 		return platErr, extra_info + platExtra
-	case "":
+	case "", "detached":
```
dangra, line 48: https://github.com/superfly/flyctl/blob/c20baa0ac5cbf21d2f6220f1d1262811c2c016fb/internal/appconfig/config.go#L16

internal/command/deploy/machines.go
dangra: appConfig is fetched from the context, it is simpler to override this at the caller's appConfig instance instead of patching it here as a side effect.

internal/command/migrate_to_v2/migrate_to_v2.go
DAlperin: Yeah

dangra (approved):

PR #2018, opened 2023-04-05
Serialize nomad config

dangra:
The initial intention of this PR was to fix `fly config show` for nomads that whose configuration couldn't be parsed as V2 compatible.  Notably old PG nomad clusters.

The win here is that appconfig.Config implements json.Marshaler interface now so can be serialized no matter what platform is used

### DIFF
--- a/internal/appconfig/serde.go
+++ b/internal/appconfig/serde.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/BurntSushi/toml"
+	"github.com/samber/lo"
 	"github.com/superfly/flyctl/helpers"
 	"github.com/superfly/flyctl/iostreams"
 )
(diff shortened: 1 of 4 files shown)


alichay:
this changes the json output to toml, but wasn't the issue that it wasn't printing the rawDefinition? (or am I mistaken)

dangra:
> this changes the json output to toml, but wasn't the issue that it wasn't printing the rawDefinition? (or am I mistaken)

right, but the PR is a draft so @DAlperin can test it out locally.  

It is not final because of the format change.

The "what is serialized" comes from https://github.com/superfly/flyctl/blob/master/internal/appconfig/serde.go#L95

alichay:
oh, marshalToml has magic to it aside from just literally calling into a library. my bad!

dangra:
@alichay when you have time this is ready for review. thanks

internal/appconfig/serde.go
```diff
@@ -57,56 +71,66 @@ func (c *Config) WriteToDisk(ctx context.Context, path string) (err error) {
...
-	cfg.RawDefinition = rawDefinition
-	return cfg, nil
 }
 
-func (c *Config) marshalTOML(w io.Writer) error {
+// marshalTOML serializes the configuration to TOML format
+// NOTES:
+//   * It can't be called `MarshalTOML` because toml libraries don't support marshaler interface on root values
+//   * Needs to reimplements most of MarshalJSON to enforce order of fields
+//   * Instead of this, you usually need one WriteTo(), WriteToFile() or WriteToDsik()
```
alichay, line 94: typo: dsik

alichay (approved):
looks good to me, I didn't get a chance to look it over too, too close but it all looks right

PR #2019, opened 2023-04-05
Jphenow/process group regions

jphenow:
* [appsv2] Show regions based on machine placement

## appv2

```
-> % flocal regions list --app appsv2
Regions [web]: den
Regions [worker]: ord
```

## Origin

Inability to see per-process-group regions made this harder: https://community.fly.io/t/different-numbers-shown-on-the-logs-compared-to-the-instance/11923/11

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -340,9 +340,6 @@ type App struct {
 	ImageDetails                ImageVersion
 	LatestImageDetails          ImageVersion
 
-	// TODO: seems wrong
-	Machines struct{ Nodes []GqlMachine }
-
 	PlatformVersion string
 }
 
--- a/api/types.go
+++ b/api/types.go
@@ -340,6 +340,9 @@ type App struct {
 	ImageDetails                ImageVersion
 	LatestImageDetails          ImageVersion
 
+	// TODO: seems wrong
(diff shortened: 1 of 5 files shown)


api/types.go
```diff
@@ -340,6 +340,9 @@ type App struct {
 	ImageDetails                ImageVersion
 	LatestImageDetails          ImageVersion
 
+	// TODO: seems wrong
+	Machines struct{ Nodes []GqlMachine }
```
jphenow, line 344: Less plugged into this work recently - should I just make a type scoped to this module?

jphenow:
@dangra it sounded like this might relate to some region things you were talking to Kurt about - let me know if this is way off the mark?

jphenow:
note to self, need to filter the machines we're querying for by state probably.

jphenow:
Update: regions in appsv1 are kind of a mess in general between a simple app, an app with multi-process, and then if you put volumes in the mix its a whole thing. I backed out of the appsv1 portion of this and this just adds a regions list per-group for appsv2.

dangra (approved):
Looks good. This command should be ported to `internal/command/regions/` but that can be done in another PR.

PR #2022, opened 2023-04-05
`deploy`: make release command machines pull amount of ram from app machines

alichay:


dangra (approved):

internal/command/deploy/machines.go
```diff
@@ -495,6 +503,38 @@ func (md *machineDeployment) configureLaunchInputForReleaseCommand(launchInput *
...
+		} else {
+			// No "main" app, sort lexicographically and just pick the first group
+			pgNames := lo.Keys(md.processConfigs)
+			if len(pgNames) > 0 {
+				slices.Sort(pgNames)
+				if leastRamApp = leastRamInGroup(pgNames[0]); leastRamApp != 0 {
+					desiredMem = leastRamApp
+				}
+			}
+		}
```
alichay, line 535: So basically, this looks at each machine in "app" and takes the least amount of ram.
If there's no "app" process group, it sorts the list of process groups and runs the same algorithm through the first group.

Fallback is 256 mb

internal/command/deploy/machines.go
```diff
@@ -495,6 +504,46 @@ func (md *machineDeployment) configureLaunchInputForReleaseCommand(launchInput *
...
+					}
+				}
+			}
+			return guest
+		}
+
+		if maxRamInMachs := maxRamInGroup("app"); maxRamInMachs != nil {
+			desiredGuest = *maxRamInMachs
+		} else {
+			// No "main" app, sort lexicographically and just pick the first group
```
dangra, line 536: could simplify if uses https://github.com/superfly/flyctl/blob/f92741d971911c1f542265c5e9dd696611e6e2c1/internal/appconfig/processconfig.go#L93

alichay: Oh awesome, didn't know this existed. that'll make it a lot less ugly, thanks!

internal/command/deploy/machines.go
```diff
@@ -495,6 +504,46 @@ func (md *machineDeployment) configureLaunchInputForReleaseCommand(launchInput *
...
+			)
+			for _, lm := range md.machineSet.GetMachines() {
+				mach := lm.Machine()
+				if mach.Config != nil && mach.Config.Env != nil && mach.ProcessGroup() == group && mach.Config.Guest != nil {
+					if mach.Config.Guest.MemoryMB > maxRam {
+						guest = mach.Config.Guest
+					}
+				}
+			}
+			return guest
```
dangra, line 530: A good candidate for https://github.com/samber/lo#reduce

internal/command/deploy/machines.go
```diff
@@ -495,6 +504,46 @@ func (md *machineDeployment) configureLaunchInputForReleaseCommand(launchInput *
 	if _, present := launchInput.Config.Env["RELEASE_COMMAND"]; !present {
 		launchInput.Config.Env["RELEASE_COMMAND"] = "1"
 	}
+	if launchInput.Config.Guest == nil {
+		launchInput.Config.Guest = &api.MachineGuest{}
+	}
+	// The pointer dereferences around the MachineGuest logic are awkward, but it's done this way
+	// to ensure that the struct gets copied - that way, if something touches the relcmd machine,
+	// it doesn't affect the machine it stole its config from
```
dangra, line 512: what about having a util to deep copy any type? 

A wrapper of [copier.DeepCopy](https://github.com/superfly/flyctl/blob/f92741d971911c1f542265c5e9dd696611e6e2c1/internal/machine/config.go#L65) but for generic type

alichay: ugh i got nerd sniped

dangra (approved):
Looks good! 

I was expecting Clone to deal only with `*T` but yours implementation looks complete ! 

```go
func Clone[T any](src *T) (*T, error) {
  if src == nil {
    return nil, nil
  }
  var dst *T
  options := copier.Option{IgnoreEmpty: true, DeepCopy: true}
  err := copier.CopyWithOption(dst, src, options)
  return dst, err
}
```

internal/command/deploy/machines.go
```diff
@@ -481,6 +482,14 @@ func (md *machineDeployment) createOrUpdateReleaseCmdMachine(ctx context.Context
 	}
 }
 
+func defaultReleaseMachineGuest() *api.MachineGuest {
```
dangra, line 485: now that we have clone, this can be fed from a top level variable + init() and we avoid the panic call

PR #2025, opened 2023-04-06
Don't replace empty restart policy of existing machines

matttpt:
#1914 made it so that `m update` would change the restart policy to the machine default ("always") only if it were currently empty. However, Apps v2 deployments use the empty policy by default—IIRC @catflydotio's take on Discourse was that this is a fine choice—and now I see `fly m update` on Apps v2 machines wanting to change that to "always" unprompted.

So, to try to get it right the second time:

- Make it easy to distinguish between `m run` and `m update` in `determineMachineConfig`.
- Don't replace the empty restart policy of existing machines (by checking whether we're in `m run` or `m update`).
- Make `--restart ''` work in case people want to reset the restart policy to empty/unspecified.

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -257,7 +257,12 @@ func runMachineRun(ctx context.Context) error {
 		return fmt.Errorf("to update an existing machine, use 'flyctl machine update'")
 	}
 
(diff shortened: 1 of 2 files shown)


catflydotio:
I was trying to figure out how empty the policy the other day! 

One thought: should `fly m run` actually choose a "non-default" policy by default? I think there was some discussion over that _somewhere_.

dangra (approved):

PR #2028, opened 2023-04-06
Test process names for machines, nomad and detached platform

dangra:


alichay (approved):
oh, this is much cleaner!

PR #2031, opened 2023-04-07
Config.ToMachineConfig() and machineDeployment improvements

dangra:
True to be told I went into a rabbit hole on #1983 trying to reuse machineDeployment functionality for scaling apps horizontally (fly scale count)

The initial motivation was to create reusable functions to convert appConfig (fly.toml) to machine config so when scaling groups from zero it worked the same as deploying from scratch.

That motivated the port of functionality found in `machineDeployment.resolveUpdatedMachineConfig` as a method of `appconfig.Config` named `ToMachineConfig` and `ToReleaseMachineConfig` that could be reused by scaling code.

The code is much more organized and better tested, it also handles (and test) changes to configured mounts, like for example replacing a machine if the volume changes completely. 

I need to draw a line, submit the changes now so I can go back to `fly scale count` asap.


internal/appconfig/testdata/tomachine.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[env]
+  FOO = "BAR"
+
+[metrics]
+  port = 9999
+  path = "/metrics"
+
+[http_service]
+  interal_port = 8080
```
matttpt, line 15:
```suggestion
  internal_port = 8080
```

internal/appconfig/machines_test.go
```diff
@@ -0,0 +1,112 @@
...
+	want := &api.MachineConfig{
+		Env: map[string]string{
+			"FOO":            "BAR",
+			"PRIMARY_REGION": "mia",
+		},
+
+		Services: []api.MachineService{
+			{
+				Protocol:     "tcp",
+				InternalPort: 0,
```
matttpt, line 24:
```suggestion
				InternalPort: 8080,
```

internal/appconfig/testdata/tomachine.toml
dangra: fixed 8ee7868e

internal/appconfig/machines_test.go
dangra: fixed by 8ee7868e

alichay (approved):
thank you for modularizing this! it's been getting more and more unwieldy lately

PR #2032, opened 2023-04-07
`version update`: print what version you updated to

### LINKED ISSUE #1444: Don't prompt to update when updating
Not a problem, but it's a little redundant to prompt you to update when running the update command. 😃 

```sh
$ flyctl version update 
Update available 0.0.398 -> v0.0.429.
Run "flyctl version update" to upgrade.
/usr/local/bin/bash -c
Running automatic update [brew upgrade flyctl]
Running `brew update --auto-update`...
[...]
```

### LINKED ISSUE #2017: display the new flyctl version number after `fly version update`
flyctl doesn't indicate which version we've just updated to:

```
$ fly version update
Update available 0.0.504-pre-3 -> v0.0.505.
Run "fly version update" to upgrade.
/bin/zsh -c
Running automatic update [curl -L "https://fly.io/install.sh" | sh]
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1475    0  1475    0     0   3635      0 --:--:-- --:--:-- --:--:--  3660
######################################################################## 100.0%
set channel to shell
flyctl was installed successfully to /Users/chris/.fly/bin/flyctl
Run 'flyctl --help' to get started
```

There's already issue #1444 about the redundant "update available" message.

I always run `fly version` after an update to see what I ended up getting:

```
$ fly version
fly v0.0.506 darwin/arm64 Commit: 137323a9 BuildDate: 2023-04-05T08:38:51Z
```

alichay:
Closes #2017 
Closes #1444 

This is output in the format:
`Updated flyctl v[current version] -> v[latest version]`

### DIFF
--- a/internal/command/version/update.go
+++ b/internal/command/version/update.go
@@ -46,5 +46,10 @@ func runUpdate(ctx context.Context) error {
 	}
 
 	io := iostreams.FromContext(ctx)
-	return update.UpgradeInPlace(ctx, io, release.Prerelease)
+
+	if err = update.UpgradeInPlace(ctx, io, release.Prerelease); err != nil {
+		return err
+	}
+	fmt.Fprintf(io.Out, "Updated flyctl v%s -> v%s\n", buildinfo.Version(), latest)
+	return nil
 }
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -349,7 +349,41 @@ func startQueryingForNewRelease(ctx context.Context) (context.Context, error) {
 	return ctx, nil
 }
 
+// shouldIgnore allows a preparer to disable itself for specific commands
+// E.g. `shouldIgnore([][]string{{"version", "update"}, {"machine", "status"}})`
+// would return true for "fly version update" and "fly machine status"
+func shouldIgnore(ctx context.Context, cmds [][]string) bool {
+	cmd := FromContext(ctx)
+	for _, ignoredCmd := range cmds {
+		match := true
+		currentCmd := cmd
+		// The shape of the ignoredCmd slice is something like ["version", "update"],
+		// but we're walking up the tree from the end, so we have to iterate that in reverse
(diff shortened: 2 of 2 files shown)


dangra (approved):

internal/command/version/update.go
```diff
@@ -46,5 +46,10 @@ func runUpdate(ctx context.Context) error {
 	}
 
 	io := iostreams.FromContext(ctx)
-	return update.UpgradeInPlace(ctx, io, release.Prerelease)
+
+	if err = update.UpgradeInPlace(ctx, io, release.Prerelease); err != nil {
+		return err
+	}
+	fmt.Fprintf(io.Out, "Updated flyctl v%s -> v%s\n", buildinfo.Version(), latest)
```
tvdfly, line 53: I think this can show the wrong version for latest in some cases, like right after a release has happened. The reason is we query our web api to detect if there is a new version, and the web api sometimes has a slightly stale cached version.

I've seen cases like this:
```
$ fly version update 
Update available 0.0.450 -> v0.0.459.
...

$ fly version
fly v0.0.463 darwin/arm64 Commit: c44008d3 BuildDate: 2023-04-05T23:45:38Z
```

An alternate way to do this (and there are more!) would be run `fly version` in a subprocess so it would use the new binary. I didn't test this, but something along these lines might work:

```go
if output, err := exec.CommandContext(ctx, os.Args[0], "version").CombinedOutput(); err != nil {
    fmt.Fprintf(io.Out, "%s\n", output)
}
```

alichay: I didn't end up calling into the new binary in this pr because I know windows can be... _fussy_ about binaries being moved, and I didn't have a windows machine handy to test on.

didn't think about the possibility there could be caching issues, though. thanks!

PR #2039, opened 2023-04-10
support using a specific signal to stop a machine

jipperinbham:

PR #2040, opened 2023-04-10
use empty struct since we don't care about the signal value

jipperinbham:
Follow-up to https://github.com/superfly/flyctl/pull/2039 which resulted in a build failure once merged due to the `syscall` package and architecture support. Since we're not using the value in the map, it can be an empty struct for now.

### DIFF
--- a/internal/command/machine/stop.go
+++ b/internal/command/machine/stop.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


dangra (approved):

dangra (commented):
to be honest, I wouldn't use a `map[string]struct{}`, I know it is suppose to be faster than scanning an array (although the array is quite small and it doesn't matter here) and more memory efficient than `map[string]bool`, but it reads quite ugly if you don't know what is going on.

jipperinbham:
> but it reads quite ugly if you don't know what is going on
for me, if I see `struct{}` used as the value in a map, I know it's not used for anything vs a more concrete type like `bool`

PR #2043, opened 2023-04-10
Move all IP allocation to first deploy

dangra:
Ready for review. https://flyio.discourse.team/t/xxx/2429/11

### DIFF
--- a/internal/appconfig/patches.go
+++ b/internal/appconfig/patches.go
@@ -172,14 +172,22 @@ func patchServices(cfg map[string]any) (map[string]any, error) {
 			return nil, fmt.Errorf("Error processing services: %w", err)
 		}
 
-		for idx, service := range services {
+		var newServices []map[string]any
+		for _, service := range services {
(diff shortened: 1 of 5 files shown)


internal/appconfig/patches.go
```diff
@@ -172,14 +172,22 @@ func patchServices(cfg map[string]any) (map[string]any, error) {
...
 				return nil, err
 			}
-			services[idx] = service
+			if len(service) != 0 {
+				newServices = append(newServices, service)
+			}
+		}
+
+		if len(services) > 0 {
+			cfg["services"] = newServices
```
alichay, line 187: Could be misunderstanding, but shouldn't this be `len(newServices)`?

dangra: you are right but that left me thinking if we need the length check at all 🤔

dangra: nice catch! fixed by c484058b

alichay (approved):

Issue #2044, opened 2023-04-10
Request from Homebrew maintainers


p-linnane:
Hello 👋 . I'm a maintainer for the [Homebrew](https://brew.sh) project. I am reaching out due to the large number of PR's we are receving from your release process. In the last 7 days we have received 11 version bumps for flyctl, with there being 4 PR's today alone. Often these releases are made before the previous one has been merged, resulting in conflicts and wasted CI resources. Our project has limited CI resources, so we would like to request that you remove the automated version bump from your release process.

https://github.com/Homebrew/homebrew-core/pulls?q=is%3Apr+flyctl

Thank you

tvdfly:
Hi @p-linnane, thank you for reaching out! We'll remove that now.

We would love to find a way to work within the Homebrew maintainers while enabling our Homebrew users to get timely updates. If we submitted at most one PR to update per day, would that be acceptable?

We've considered moving to a cask that autoupdates to circumvent this issue entirely. Is that a reasonable approach? Do you have any links or references you could share for how other projects have handled multiple daily releases?

Thanks!

tvdfly:
We have taken these actions:

1. We will no longer open Homebrew PRs if an existing PR is open for flyctl
2. We will only attempt to open a Homebrew release once per day

Please let us know if this is still inadequate and we'll be happy to take additional actions.

p-linnane:
Hello. Thank you very much for your quick attention to this! That should be perfectly fine moving forward. I appreciate your willingness to work with us on this. Cheers!

PR #2046, opened 2023-04-11
Update homebrew release once per day

tvdfly:
Previously we updated the homebrew release every time a new release was created. This overwhelmed the Homebrew maintainers. https://github.com/superfly/flyctl/issues/2044

With this change, we run a github action once per day at 17:23 UTC and update homebrew with the latest release. This will fail if the previous homebrew release has not yet been merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,18 +74,6 @@ jobs:
       - name: Run tests
         run: make test
 
-  release_homebrew:
-    needs: test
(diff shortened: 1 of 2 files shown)


alichay (approved):
assuming this wouldn't try to bump if there's no new version

tvdfly:
@alichay yep, the force = false will cause this job to fail if there is an already open pr.

https://github.com/dawidd6/action-homebrew-bump-formula/blob/v3/README.md?plain=1#L46-L47

PR #2050, opened 2023-04-11
Update release status during machine deployments

matttpt:
This will:

* set the release status to `running` when starting to deploy, and
* set the release status to `completed` or `failed` as appropriate after deployment finishes.

Handling of interrupted deployments will come later.

Opening as a draft since this depends on a new mutation in `web` that will need to be merged first.


matttpt:
The mutation is in `web` now!

I've confirmed that the status gets set to `running` correctly at the start, and that `complete` or `failed` show up as appropriate (the last one I induced by killing a long-running release command to make a deployment fail). Preflight tests pass, though admittedly `TestAppsV2Example` did fail the first time I tried it (I think it's unrelated; it's been happening to me on and off all week).

alichay (approved):

PR #2054, opened 2023-04-11
adds a new field, standbys, on the MachineConfig

jipperinbham:
It's unclear how/where this will be presented through the CLI but we need the field added in order for the API to properly support it.

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -453,6 +453,11 @@ type MachineConfig struct {
 	FlyProxy    *MachineFlyProxy `json:"fly_proxy,omitempty"`
 	Processes   []MachineProcess `json:"processes,omitempty"`
 
+	// Not used by flyctl yet
+	// Standbys enable a machine to be a standby for another. In the event of a hardware failure,
+	// the standby machine will be started.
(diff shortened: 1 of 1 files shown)


api/machine_types.go
```diff
@@ -453,6 +453,11 @@ type MachineConfig struct {
 	FlyProxy    *MachineFlyProxy `json:"fly_proxy,omitempty"`
 	Processes   []MachineProcess `json:"processes,omitempty"`
 
+	// Not used by flyctl yet
+	// Standbys enable a machine to be a standby for another. In the event of a hardware failure,
+	// the standby machine will be started.
+	Standbys []string `json:"standbys,omitempty"`
```
dangra, line 459: propagate the field on deploys by adding it to this list https://github.com/superfly/flyctl/blob/9eb8fa4c8846d333ef674f6e9a4bdeabf4b1a29a/internal/command/deploy/machines_launchinput.go#L51-L57

I am working on doing that by default but for now it has to be added there

dangra: this is not an issue once https://github.com/superfly/flyctl/pull/2070 is merged

dangra (approved):

PR #2055, opened 2023-04-11
Make log shippers operate on a per-app basis

jsierles:
This PR makes automated log shippers use separate Vector source and sink configurations for each app and logging endpoint. This is good for security, and makes it easier to not accidentally fill up the log provider with logs you aren't aware of.

This PR takes advantage of the Machines Exec API for configuring Vector without restarting the VM. This ensures that logs are not dropped when sources are added and removed from the shipper.


internal/command/logs/shipper_setup.go
```diff
@@ -80,12 +78,21 @@ func runSetup(ctx context.Context) (err error) {
 		fmt.Fprintf(io.ErrOut, "Provisioning a log shipper VM in the app named %s\n", shipperApp.Name)
 	}
 
-	// Fetch or create the org-specific Logtail integration
+	// Fetch or create the Logtail integration for the app
 
+	var addOnName = appName + "-log-shipper"
 	getAddOnResponse, err := gql.GetAddOn(ctx, client, addOnName)
 
 	if err != nil {
```
btoews, line 86: Would it be worth checking for a more specific error here?

internal/command/logs/shipper_setup.go
```diff
@@ -147,40 +123,55 @@ func runSetup(ctx context.Context) (err error) {
...
+		Image: "flyio/log-shipper:auto-d5a96e6",
 	}
 
 	machines, err := flapsClient.List(ctx, "")
 
+	if err != nil {
+		return err
+	}
+
 	launchInput := api.LaunchMachineInput{
```
btoews, line 135: Should this and `machineConf ` be defined in the `else` block starting on L145?

internal/command/logs/shipper_setup.go
```diff
@@ -147,40 +123,55 @@ func runSetup(ctx context.Context) (err error) {
...
 
-	fmt.Fprintf(io.Out, "Launched machine %s\n in the %s region", machine.ID, launchInput.Region)
+		if err != nil {
+			return err
+		}
 
+		fmt.Fprintf(io.Out, "Launched log shipper VM %s\n in the %s region", machine.ID, launchInput.Region)
+	}
+
+	cmd := []string{"/add-logger.sh", targetApp.Name, "logtail", "'" + tokenResponse.CreateLimitedAccessToken.LimitedAccessToken.TokenHeader + "'", logtailToken}
```
btoews, line 164: It would be a bit more robust to use something like [shellescape](https://pkg.go.dev/github.com/alessio/shellescape) for this.

internal/command/logs/shipper_setup.go
```diff
@@ -147,40 +123,55 @@ func runSetup(ctx context.Context) (err error) {
...
+	fmt.Println(tokenResponse.CreateLimitedAccessToken.LimitedAccessToken.TokenHeader)
```
btoews, line 165: Is this for debugging, or should it stay?

btoews (approved):
A few notes, but I dig it overall!

internal/command/logs/shipper_setup.go
```diff
@@ -96,46 +103,15 @@ func runSetup(ctx context.Context) (err error) {
 	} else {
 		logtailToken = getAddOnResponse.AddOn.Token
 	}
-	// Fetch a macaroon token whose access is limited to reading app logs
-	tokenResponse, err := gql.CreateLimitedAccessToken(ctx, client, targetOrg.Slug+"-logs", targetOrg.Id, "read_organization_apps", &gql.LimitedAccessTokenOptions{
+	// Fetch a macaroon token whose access is limited to reading this app's logs
+	tokenResponse, err := gql.CreateLimitedAccessToken(ctx, client, appName+"-logs", targetOrg.Id, "read_organization_apps", &gql.LimitedAccessTokenOptions{
 		"app_id": targetApp.Name,
```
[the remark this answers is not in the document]
jsierles: Doh good catch.

internal/command/logs/shipper_setup.go
jsierles: Done!

internal/command/logs/shipper_setup.go
jsierles: byebye

internal/command/logs/shipper_setup.go
jsierles: Will address in a follow-up as the gqclient library is a bit weird about errors.

PR #2057, opened 2023-04-11
Prompt app create on deploy if none is found

shortdiv:
There's some friction on app deploy if a user runs deploy without first creating an app. This reduces a step towards deploying an app, by removing the need to run a separate app creation command. 

appcreateondeploy

### DIFF
--- a/internal/command/apps/create.go
+++ b/internal/command/apps/create.go
@@ -78,7 +78,7 @@ func RunCreate(ctx context.Context) (err error) {
 	var name string
 	switch {
 	case aName != "" && fName != "" && areNamesClashing([]string{aName, fName, ctxName}):
-		err = fmt.Errorf("app names specified %s, %s %s, only one may be specified",
+		err = fmt.Errorf("app names specified via command argument %s, via flag %s and via fly.toml %s. Only one may be specified",
 			aName, fName, ctxName)
 
 		return
--- a/internal/command/apps/create.go
+++ b/internal/command/apps/create.go
@@ -77,9 +77,9 @@ func RunCreate(ctx context.Context) (err error) {
 
 	var name string
 	switch {
-	case aName != "" && fName != "" && aName != fName:
-		err = fmt.Errorf("two app names specified %s and %s, only one may be specified",
-			aName, fName)
+	case aName != "" && fName != "" && areNamesClashing([]string{aName, fName, ctxName}):
+		err = fmt.Errorf("app names specified %s, %s %s, only one may be specified",
+			aName, fName, ctxName)
 
 		return
 	case ctxName != "":
@@ -126,6 +126,16 @@ func RunCreate(ctx context.Context) (err error) {
 	}
 
 	return err
+}
+
+func areNamesClashing(a []string) bool {
+	keys := make(map[string]bool)
+	for _, name := range a {
+		if _, ok := keys[name]; !ok {
+			keys[name] = true
+		}
+	}
+	return len(keys) != len(a)
 }
 
 func shouldAppUseMachinesPlatform(ctx context.Context, apiClient *api.Client, orgSlug string) (bool, error) {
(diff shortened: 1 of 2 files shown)


internal/command/apps/create.go
```diff
@@ -80,6 +82,8 @@ func RunCreate(ctx context.Context) (err error) {
 			aName, fName)
 
 		return
+	case ctxName != "":
+		name = ctxName
```
dangra, line 86: I like it but this is a different thing right? it creates the name passed by `-a` flag or found in local fly.toml, is it?

internal/command/deploy/deploy.go
```diff
@@ -124,7 +126,17 @@ func run(ctx context.Context) error {
...
-		return fmt.Errorf("the app name %s could not be found, did you create the app or misspell it in the fly.toml file or via -a?", appConfig.AppName)
+		confirmation, err := prompt.Confirm(ctx,
+			fmt.Sprintf("the app name %s could not be found, do you want to create a new app with that name?", appName))
+		if err != nil {
+			return err
+		}
+		if !confirmation {
+			return fmt.Errorf("the app name %s could not be found, it looks like the app has either not been created or is misspelt", appName)
+		}
+		if err = apps.RunCreate(ctx); err != nil {
```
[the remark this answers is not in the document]
shortdiv: Yeah, I'm not sure the best way to go about it tbh. I was initially wondering if there was a way to call a command from the context of another command i.e. fly deploy -> fly apps create. But it looked more complicated

internal/command/apps/create.go
dangra: should it check that the ctxName doesn't differ from the other names `aName` and `fName` ?

the context name can come from FLY_APP envvar, `-a` and local fly.toml. 

it can make `fly apps create myappname` fail or pick the wrong name if run in any of those conditions.

shortdiv: oh interesting! I hadn't thought of that, I'll add that to the above conditional

alichay:
i'm a little worried that this will further muddy the waters between launch and deploy. how do you feel about just prompting the user to run `fly launch` instead?

shortdiv:
> i'm a little worried that this will further muddy the waters between launch and deploy. how do you feel about just prompting the user to run `fly launch` instead?

Good point @alichay, I think that would work so we don't have to recreate logic as well

shortdiv:
Ok I thought this through a little more and I think prompting a user to launch an app if one is not found when they run deploy isn't the best workflow. The assumption when a user runs fly deploy is that they want to run a deploy manually. We already prompt them to run apps create when an app is not found prior to running deploy, this just removes that one step. I don't believe that really confuses launch and deploy too much, since launch handholds you a lot more compared to deploy.

internal/command/apps/create.go
```diff
@@ -71,15 +72,18 @@ func RunCreate(ctx context.Context) (err error) {
...
 		apiClient     = client.FromContext(ctx).API()
+		ctxName       = appconfig.NameFromContext(ctx)
 	)
 
 	var name string
 	switch {
-	case aName != "" && fName != "" && aName != fName:
-		err = fmt.Errorf("two app names specified %s and %s, only one may be specified",
-			aName, fName)
+	case aName != "" && fName != "" && areNamesClashing([]string{aName, fName, ctxName}):
```
dangra, line 80: I'd pass the list of names to [lo.Compact](https://github.com/samber/lo#compact) and fail if its length is not equal 1.

dangra:
@shortdiv are you still working on this? I thought you were ditching it in favor or redirecting the user to run `fly launch`

shortdiv:
Oh, I'm not, based on this -> https://github.com/superfly/flyctl/pull/2057#issuecomment-1504086735, I saw this as not confusing launch from deploy. But open to feedback/criticism on whether this is the right approach

PR #2058, opened 2023-04-11
`migrate-to-v2`: support apps with volumes

alichay:
Use volume forking to migrate apps with volumes.

Creates copies of *every volume in an app*, named "[oldname]_machines", and attaches them to the new machines.
When migrating apps with volumes, there will be a short downtime because nomad has to be scaled down before the volumes can be forked.

For apps without volumes, we maintain zero-downtime migrations because nomad can be scaled down after machines are created.


alichay:
Blocked on https://github.com/superfly/web/pull/1740

alichay (commented):
pointing out some of my questions that I scattered in the codebase:

internal/appconfig/setters.go
```diff
@@ -247,3 +247,15 @@ func (c *Config) SetVolumes(volumes []Volume) {
 		}
 	}
 }
+
+func (c *Config) Volumes() []Volume {
```
dangra, line 251: totally a nit pick but this is not a setter, we have been placing v1/v2 compatibility getters in config.go mainly.

alichay: that's what I get for relying on IDE features and not using my eyeballs. thanks!

dangra (approved):

Issue #2036, opened 2023-04-08
apps v2 `fly deploy` hangs on release_command


tvdfly:
I ran `fly deploy` on an app that has a release_command. The release command succeeded, but `fly deploy` hung forever waiting for the machine to have the started state.

This was the output from fly deploy:

```
--> Pushing image done
image: [registry.fly.io/tvd-analytics:deployment-01GXDTF8DQM3NX1D369K1RCVB6](http://registry.fly.io/tvd-analytics:deployment-01GXDTF8DQM3NX1D369K1RCVB6)
image size: 132 MB
Running tvd-analytics release_command: /release_cmd.sh
  Waiting for 5683976f7d7e8e to have state: started
```

logs show that the release command finished successfully:

```
2023-04-07T12:19:02.661 app[5683976f7d7e8e] sea [info] Migrations successful!
2023-04-07T12:19:02.766 app[5683976f7d7e8e] sea [info] Starting clean up.
2023-04-07T12:19:03.767 app[5683976f7d7e8e] sea [info] [ 4.137571] reboot: Restarting system
2023-04-07T12:19:04.663 runner[5683976f7d7e8e] sea [info] machine restart policy set to 'no', not restarting
```

The release_command machine was successfully destroyed:

```
% fly m status 5683976f7d7e8e
Machine ID: 5683976f7d7e8e
Instance ID: 01GXDTFDX2SVFAW1V6YQPG5761
State: destroyed

VM
  ID            = 5683976f7d7e8e
  Instance ID   = 01GXDTFDX2SVFAW1V6YQPG5761
  State         = destroyed
  Image         = tvd-analytics:deployment-01GXDTF8DQM3NX1D369K1RCVB6
  Name          = restless-breeze-5622
  Private IP    =
  Region        = sea
  Process Group = fly_app_release_command
  CPU Kind      = shared
  vCPUs         = 2
  Memory        = 512
  Created       = 2023-04-07T12:18:50Z
  Updated       = 2023-04-07T12:19:04Z
  Command       = /release_cmd.sh

Event Logs
STATE           EVENT   SOURCE  TIMESTAMP                       INFO
destroyed       destroy flyd    2023-04-07T05:19:04.786-07:00
destroying      destroy flyd    2023-04-07T05:19:04.664-07:00
stopped         exit    flyd    2023-04-07T05:19:04.663-07:00   exit_code=0,oom_killed=false,requested_stop=false
[...]

PR #2059, opened 2023-04-11
Handle release command machines that terminate quickly

matttpt:
Addresses a possible cause of #2036. If a release command runs and exits (and is auto-destroyed) before we call the wait endpoint, it will get stuck in an error loop waiting for it to start. I *think* it's alright just not to wait for start and to rely on the following wait for machine destruction. Preflight tests pass (though I'm still a bit suspicious of this fix, honestly—thoughts?).

### DIFF
--- a/internal/command/deploy/machines_releasecommand.go
+++ b/internal/command/deploy/machines_releasecommand.go
@@ -2,12 +2,15 @@ package deploy
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"strconv"
 	"time"
 
 	"github.com/samber/lo"
 	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/flaps"
 	"github.com/superfly/flyctl/helpers"
 	"github.com/superfly/flyctl/internal/machine"
 )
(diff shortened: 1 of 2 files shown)


tvdfly (commented):
This might be fine? But we should probably fix WaitForState() to handle the error that we're getting from the wait endpoint. That error could happen in other situations, like a machine gets destroyed after a deployment starts.

There's probably a response code we're not handling in here: https://github.com/superfly/flyctl/blob/master/internal/machine/leasable_machine.go#L162-L167

matttpt:
When the machine does not exist, `WaitForState` receives a 404. This new version will catch them and return them to the caller, so the various users of `WaitForState` should avoid an error loop if the machine gets unexpectedly destroyed.

The release command code will catch these and treat them as successes. So if the release command exits quickly, before we can call the wait endpoint, it still works. Note that if the release command machine was actually destroyed unexpectedly, the following `WaitForEventTypeAfterType` call will detect the non-zero exit code and cause the deployment to fail (I have tested this).

FWIW, on the whole I think some of the machine wait code could use a closer look to double-check its error handling, but I don't want to expand the scope of this one PR too much.

tvdfly (approved):
This looks good, thanks!

Please make sure to test out running `fly deploy` from this branch on an app with a release command before merging. If you have already done that, no need to re-test.

PR #2062, opened 2023-04-12
fix(scanner/templates/remix): bring in line with `remix_prisma` template

MichaelDeBoey:


dangra:
DISCLAIMER: Sorry for the dumb question but I am illiterate about Remix or Prisma. 

Are you trying to copy over the working template of remix_prisma to the more general remix template?
Why do you leave the prisma stuff commented instead of removing?
Is the expectation that the user can uncomment the lines if decides to use Prisma later?

MichaelDeBoey:
@dangra The non-Prisma template already has some things from the Prisma template commented out. Don't know the reasoning behind it, but I think your reasoning of being able to just uncomment is right.

So I figured, I'd update the non-Prisma template to be in line with the Prisma template & also just comment the Prisma specific things

dangra:
> So I figured, I'd update the non-Prisma template to be in line with the Prisma template & also just comment the Prisma specific things

What was the motivation to update the template? Did you find a case that was breaking or simply that the template wasn't working for a new app?

Just trying to understand that we are not making changes for the sake of it and this is actually solving someone's problem :)

MichaelDeBoey:
@dangra I was looking into the templates when I was following the Remix Jokes tutorial (and updating it) & I'm planning to contribute some updates we did in our stacks to these templates as well.

Just wanted to have them consistent to start with before I create extra PRs

rubys:
I'd like to go a step further, and am willing to contribute code to make it happen.  I'd like to detect whether nor not prisms is used and conditionally include the prism parts.  The Node and Django templates do this now, as did Rails before it moved to using the dockerfile-rails gem.

rubys:
Now taking a look at the go code, it attempts to do this... sorta.  There are separate templates for remix and remix+prisma.  I gather from that the remix (without prisma) templates are being used for remix.  That's suboptimal.  Let's figure out what the proper check for prisms is and converge on one template.

MichaelDeBoey:
@rubys If it would be possible to just have 1 Remix template and still have all Prisma-related things disabled, that would be awesome!

I don't have Go knowledge so it's hard for me to really dive into this unfortunately 😕

rubys:
I'm comfortable with node and go, you have more experience that I do with remix and prisma.  Let's work together!

Perhaps we can go a step further, and have remix use the node template: https://github.com/superfly/flyctl/blob/master/scanner/templates/node/Dockerfile; this will immediately support remix on node releases other than 16, and yarn.  As a byproduct, this template will end up supporting prisma users that aren't using remix.  And if, in the future, pnpm or other builders are added, remix users will immediately benefit.

It looks like the remix template can handle typescript, so that's another benefit that node users will accrue.

What's the best way to check for prisma?

MichaelDeBoey:
> Let's work together!

@rubys I'm happy to help out wherever I can!

> support remix on node releases other than 16, and yarn

> And if, in the future, pnpm or other builders are added, remix users will immediately benefit.

That would be awesome, as we currently removed all `package-lock.json` references in `Dockerfile` in our default stacks to support other package managers like Yarn & pnpm (see https://github.com/remix-run/indie-stack/commit/3f2e07214e3b55d3d7bcf03face50c7131ea1a65)

> It looks like the remix template can handle typescript

I could be wrong, but I don't see any TS specific code tbh 🤔

> What's the best way to check for prisma?

If I'm correct, the current Remix scanner is looking for if there's a `prisma` folder that contains a `*.prisma` file, which contains `sqlite`.

This actually checks if Prisma is used with `sqlite`, but Prisma could also be used with other databases like PostgreSQL, MySQL, MongoDB, CockroachDB & Microsoft SQL Server (see https://www.prisma.io/docs/concepts/database-connectors).

To _actually_ check for Prisma, we could just check if `prisma` and/or `@prisma/client` is available (in `package.json`).

rubys:
> @rubys I'm happy to help out wherever I can!

Let's do this!  I'm actually returning from a vacation and have one more side trip, so I won't be home until Tuesday, but meanwhile a first installment of the changes I'm envisioning can be found here: https://github.com/superfly/flyctl/pull/2092 .  I won't merge until I get a chance to fully test it and it replaces what the current remix scanner does.

If you are on a Mac or Linux, building `flyctl` is a matter of installing go and make, and then running `make`.  Windows build instructions can be found in the [readme](https://github.com/superfly/flyctl#building-on-windows).  The result is an executable that can be found in `flyctl/bin`.

The plan once this is ready would be to remove the current remix scanner:

https://github.com/superfly/flyctl/blob/master/scanner/scanner.go#L95

> support other package managers like Yarn & pnpm

I've never used pnpm myself, but if pnpm equivalents of the few npm/yarn commands that are used by the Dockerfile can be identified, the change should be straightforward.

> I could be wrong, but I don't see any TS specific code tbh

Supporting typescript is a matter of initially installing all of the dev dependencies, running 'npm run build` or equivalent, and then removing the dev dependencies.  This is indeed more generic than typescript specific.

> If I'm correct, the current Remix scanner is looking for if there's a prisma folder that contains a *.prisma file, which contains sqlite.

This conflates two things: is prisma used (and therefore `prisma generate` and `start_with_migrations` are needed), and whether or not a volume should be mounted.  For now, I've focused on the former, and will add the latter before merging this change.

> just check if prisma and/or @prisma/client is available (in package.json).

I've added a check for `prisma` in `package.json`.

MichaelDeBoey:
> I'm actually returning from a vacation and have one more side trip, so I won't be home until Tuesday

If it's easier for you to talk through chat (in the meantime), don't hesitate to send me a DM on [Twitter](https://twitter.com/MichaelDeBoey93)/[Discord](https://discordapp.com/users/112974388369854464)!

> The plan once this is ready would be to remove the current remix scanner

I know it's not my intention with this PR & the follow-up PRs I was planning to do, but what with other frameworks like NextJS?

I think the decision for a separate scanner for these frameworks was made to not pollute the runtime (Node/Deno/Ruby/...) scanner too much with framework specific things tbh 🤔

> if pnpm equivalents of the few npm/yarn commands that are used by the Dockerfile can be identified, the change should be straightforward

We use a function in our [Remix Stacks](https://remix.run/stacks) to determine the used package manager & run all the commands in a generic way (that get's translated to package manager specific commands).

https://github.com/MichaelDeBoey/indie-stack/blob/9bde7bffec03a59907eebd1a21a6015ec37ba925/remix.init/index.js#L49-L80

Since that function is inspired with what the @nrwl folks did in `nx`, I created an issue to possible extract it as well: https://github.com/nrwl/nx/issues/11197, but I don't think that would be of any use in this repo 🤔

MichaelDeBoey:
Closing in favor of #2092

PR #2066, opened 2023-04-12
feat(scanner/remix): skip database questions on Prisma detection

MichaelDeBoey:
You currently get the questions to set up Postgresql/Upstash Redis database, but Prisma/SQLite is already detected

```sh
? Would you like to set up a Postgresql database now? No
? Would you like to set up an Upstash Redis database now? No
Wrote config file fly.toml

This launch configuration uses SQLite on a single, dedicated volume. It will not scale beyond a single VM. Look into 'fly postgres' for a more robust production database.
```


dangra:
makes sense. thanks!

PR #2069, opened 2023-04-12
Flat app configs by process group

dangra:
The whole goal of this PR is to keep the logic that converts application configuration (aka fly.toml) to machine specs considering process groups contained so it doesn't spread all over the codebase.

There unsorted list of changes happening in this PR:
 * Removes `appConfig.GetProcessConfig()` in favor flattening app configuration for a process group
 * In general all you need is `appConfig.ToMachineConfig(groupName)`
 * All the configuration quirks per process group are kept within appconfig.Config object 
 * All the processes and services validation that was happening on `GetProcessConfig` was moved to `appConfig.Validate()`
 * HTTPService is not special, it will be converted to a normal Service
 * HTTPService works when there are multiple process groups in fly.toml
 * Top level checks now support process groups, before they were global
 * `fly config validate` can force validate for machines or nomad if it can't get the platform from a remote endpoint
 * Tests!


dangra:
I have to remove a two panic() calls, but it is open for review and feedback.

alichay (approved):
everything looks right, rigorous testing, very nice deduplication!!

PR #2070, opened 2023-04-12
Propagate machine fields not managed by fly.toml

dangra:
Fields on MachineConfig type are of two kind, those that are set and updated on every deploy from information found in application configuration (aka fly.toml), and the others that can be set only using `fly machines update` command.

Until now we were cherry picking which fields of the later type propagated on `fly deploys`, in the past we missed a few like restart policies, autoscaling fields, vm size, guest, dns config, .... The recent addition of `standbys` field motivated this PR so adding new fields won't cause surprises with apps v2.

### DIFF
--- a/internal/appconfig/machines.go
+++ b/internal/appconfig/machines.go
@@ -47,7 +47,7 @@ func (c *Config) ToReleaseMachineConfig() (*api.MachineConfig, error) {
 	return mConfig, nil
 }
 
-//
+// updateMachineConfig applies configuration options from the optional MachineConfig passed in, then the base config, into a new MachineConfig
 func (c *Config) updateMachineConfig(src *api.MachineConfig) (*api.MachineConfig, error) {
 	processGroup := c.DefaultProcessName()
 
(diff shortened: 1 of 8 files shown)


internal/appconfig/machines.go
```diff
@@ -38,66 +39,83 @@ func (c *Config) ToReleaseMachineConfig() (*api.MachineConfig, error) {
...
-		machineConfig.Env["PRIMARY_REGION"] = c.PrimaryRegion
+		mConfig.Env["PRIMARY_REGION"] = c.PrimaryRegion
 	}
 
-	return machineConfig, nil
+	return mConfig, nil
 }
 
-func (c *Config) defaultMachineConfig() (*api.MachineConfig, error) {
+//
```
alichay, line 50: just suggesting some sort of comment because having a blank `//` bothers me lol
```suggestion
// updateMachineConfig applies configuration options from the optional MachineConfig passed in, then the base config, into a new MachineConfig
```

alichay (approved):

PR #2077, opened 2023-04-13
`helpers.Clone`: go back to `jinzhu/copier` so Config can be cloned

alichay:
We do magic when serializing appconfig.Config to json, which broke object cloning

### DIFF
--- a/internal/appconfig/config_test.go
+++ b/internal/appconfig/config_test.go
@@ -149,10 +149,18 @@ func TestCloneAppconfig(t *testing.T) {
 			Source:      "src",
 			Destination: "dst",
 		},
+		HttpService: &HTTPService{
+			InternalPort: 100,
+		},
 		defaultGroupName: "some-group",
 	}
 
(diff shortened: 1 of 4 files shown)


internal/appconfig/config_test.go
```diff
@@ -126,3 +127,32 @@ func TestConfigPortGetter(t *testing.T) {
...
+		Mounts: &Volume{
+			Source:      "src",
+			Destination: "dst",
+		},
+		defaultGroupName: "some-group",
+	}
+
+	cloned := helpers.Clone(config)
+
+	assert.Equal(t, config, cloned)
```
dangra, line 157: wdyt if the test modifies a few deep fields in the source before comparing, so we can be sure the cloned instance doesn't hold any references to the source

dangra (approved):

internal/appconfig/config_test.go
alichay: good call!

PR #2078, opened 2023-04-13
Fix error messages being printed twice

andreolvr:
Some error messages are being printed twice, as the error message for unknown command for example. That is happening because both cobra, the library we use for CLI, and our own logic is printing the errors. Silencing the errors from cobra did the trick.

|Before|After|
|---|---|
|Screenshot 2023-04-13 at 19 49 39|Screenshot 2023-04-13 at 19 49 58|

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -30,6 +30,7 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 	cmd.SetOut(io.Out)
 	cmd.SetErr(io.ErrOut)
 	cmd.SetArgs(args)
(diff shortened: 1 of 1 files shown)


dangra (approved):
Well done!   not sure why I never payed attention to the duplicate errors. Had to tried it twice to convince myself  :))

michaeldwan:
Heads up, I noticed that this change is printing `Run 'flyctl --help' for usage.` for every error, not just command parsing errors where `--help` is a good next step. For example, `fly apps destroy app-that-doesnt-exist` would print 
```
Error: Could not find App

Run 'flyctl --help' for usage.
```

PR #2080, opened 2023-04-14
Fix timeout handling in WaitFor*

pborzenkov:
Retry RoundTripper used by flaps client returns a custom error if it
detects a canceled context: https://github.com/PuerkitoBio/rehttp/blob/master/rehttp.go#L343

This makes timeout checks fails and wait indefinitely, ignoring even
Ctrl-C.


dangra (approved):
:shipit:

PR #2085, opened 2023-04-15
query homebrew cli for version info if installed through brew

alichay:
This should hopefully prevent `Error failed to execute new flyctl binary: signal: killed` for brew users.

In my testing, I couldn't get that to bug to happen outside of brew installations. That said, perhaps that is a sign that our approach for executing the new binary is bad? At least, for package manager installations?
I'd like to keep this feature, but I don't want to turn it into a rabbithole for a minuscule UX improvement.

### DIFF
--- a/internal/command/version/update.go
+++ b/internal/command/version/update.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/blang/semver"
+	"github.com/samber/lo"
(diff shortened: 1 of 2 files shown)


dangra (approved):
Nice trick. I am buying 💰

dangra:
DeepSource is complaining about the usual defer.Close()

Issue #1179, opened 2022-08-16
Create and attach volumes based on [mounts] content in fly.toml


jsierles:
Discussion https://flyio.discourse.team/t/paas-the-volumes-edition/922/5

Issue #729, opened 2022-01-12
fly launch should create a volume when using an existing config with `[mounts]`


mrkurt:
When I run `fly launch` in a directory with this config:

```toml
kill_signal = "SIGINT"
kill_timeout = 5
processes = []

[mounts]
  source="myapp_data"
  destination="/home/coder/project"
```

It should create the volume for me.  I believe all it needs to know is the size, at this stage we should already know the region. So if it prompts for size, then runs the equivalent of `fly volumes create myapp_data --size <size>` it'll be pretty magical.

rugwirobaker:
Isn't that the whole point of this block: **https://github.com/superfly/flyctl/blob/02c46ec337f600a46bb7c388a602776c8ab4884a/cmd/launch.go#L355-L380**

michaeldwan:
@rugwirobaker that block creates the volumes requested by the scanner. Kurt is saying that if a fly.toml file has mounts we should detect this and prompt for the size then create as needed.

dangra:
Fixed by #2086

PR #2086, opened 2023-04-15
Mounts per process group

dangra:
Brings "Mounts per process group"'s Nomad feature to machines, but also creates needed volumes on `fly launch`  

Fixes https://github.com/superfly/flyctl/issues/1179 and https://github.com/superfly/flyctl/issues/729

For instance this fly.toml (3 process, 2 mounts) just work on first `fly launch`
```toml
[build]
  image = "nginx"

[processes]
  app = ""
  other = "sleep 1h"
  backend = "sleep inf"

[http_service]
  internal_port = 80
  force_https = true
  processes = ["app"]

[[mounts]]
  source = "data"
  destination = "/data"
  processes = ["app"]

[[mounts]]
  source = "trashbin"
  destination = "/data"
  processes = ["other"]
```

It does all in one shot 🔫 
```
$ fly launch --copy-config -o personal --name groupmounts --region scl --now
Creating app in /home/daniel/tt/aa
An existing fly.toml file was found for app groupmounts
Using build strategies '[the "nginx" docker image]'. Remove [build] from fly.toml to force a rescan
App will use 'scl' region as primary
Created app 'groupmounts' in organization 'personal'
Admin URL: https://fly.io/apps/groupmounts
Hostname: groupmounts.fly.dev
Wrote config file fly.toml
==> Building image
image found: img_3xdk4x0kg574go0e
Provisioning ips for groupmounts
  Dedicated ipv6: 2a09:8280:1::15:d5b
  Shared ipv4: 66.241.124.148
  Add a dedicated ipv4 with: fly ips allocate-v4
Creating volume 'data' for process group 'app'
Creating volume 'trashbin' for process group 'other'
 * create 1 "other" machine
 * create 1 "app" machine
 * create 1 "backend" machine

No machines in group 'app', launching one new machine
  Machine 3d8d9790b75789 [app] update finished: success
No machines in group 'backend', launching one new machine
  Machine 6e82d441bd3487 [backend] update finished: success
No machines in group 'other', launching one new machine
  Machine 4d891d15a71387 [other] update finished: success
Finished launching new machines
Updating existing machines in 'groupmounts' with rolling strategy
  Finished deploying
```

### DIFF
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -39,7 +39,7 @@ type Config struct {
 	HTTPService   *HTTPService              `toml:"http_service,omitempty" json:"http_service,omitempty"`
 	Metrics       *api.MachineMetrics       `toml:"metrics,omitempty" json:"metrics,omitempty"`
 	Statics       []Static                  `toml:"statics,omitempty" json:"statics,omitempty"`
-	Mounts        *Mount                    `toml:"mounts,omitempty" json:"mounts,omitempty"`
+	Mounts        []Mount                   `toml:"mounts,omitempty" json:"mounts,omitempty"`
 	Processes     map[string]string         `toml:"processes,omitempty" json:"processes,omitempty"`
 	Checks        map[string]*ToplevelCheck `toml:"checks,omitempty" json:"checks,omitempty"`
 	Services      []Service                 `toml:"services,omitempty" json:"services,omitempty"`
@@ -196,8 +196,5 @@ func (c *Config) Volumes() []Mount {
 			return arr
 		}
 	}
-	if c.Mounts != nil {
-		return []Mount{*c.Mounts}
-	}
-	return nil
+	return c.Mounts
 }
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -73,8 +73,9 @@ type Static struct {
 }
 
 type Mount struct {
-	Source      string `toml:"source,omitempty" json:"source,omitempty"`
-	Destination string `toml:"destination" json:"destination,omitempty"`
+	Source      string   `toml:"source,omitempty" json:"source,omitempty"`
+	Destination string   `toml:"destination" json:"destination,omitempty"`
+	Processes   []string `json:"processes,omitempty" toml:"processes,omitempty"`
 }
 
 type Build struct {
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -39,7 +39,7 @@ type Config struct {
 	HTTPService   *HTTPService              `toml:"http_service,omitempty" json:"http_service,omitempty"`
 	Metrics       *api.MachineMetrics       `toml:"metrics,omitempty" json:"metrics,omitempty"`
 	Statics       []Static                  `toml:"statics,omitempty" json:"statics,omitempty"`
-	Mounts        *Volume                   `toml:"mounts,omitempty" json:"mounts,omitempty"`
+	Mounts        *Mount                    `toml:"mounts,omitempty" json:"mounts,omitempty"`
 	Processes     map[string]string         `toml:"processes,omitempty" json:"processes,omitempty"`
 	Checks        map[string]*ToplevelCheck `toml:"checks,omitempty" json:"checks,omitempty"`
 	Services      []Service                 `toml:"services,omitempty" json:"services,omitempty"`
@@ -72,7 +72,7 @@ type Static struct {
 	UrlPrefix string `toml:"url_prefix" json:"url_prefix,omitempty" validate:"required"`
 }
 
-type Volume struct {
+type Mount struct {
 	Source      string `toml:"source,omitempty" json:"source,omitempty"`
 	Destination string `toml:"destination" json:"destination,omitempty"`
 }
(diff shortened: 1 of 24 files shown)


internal/command/migrate_to_v2/volumes.go
```diff
@@ -38,15 +38,14 @@ for updates, watch https://community.fly.io for announcements about volume migra
...
 
 func (m *v2PlatformMigrator) migrateAppVolumes(ctx context.Context) error {
-
 	// NOTE: appconfig.Volume doesn't support the rare "processes" key on mounts
 	//       this is extremely rarely used, and seemingly undocumented,
 	//       but because of this rollback stores a copy of the old appconfig now
 	//       and deploys the original config if something goes wrong.
 	//       That said, once an app gets moved to V2, that mapping gets wiped.
 	// (not an issue now, because we don't even support multiple volumes on v2,
 	//  but it's worth documenting here nonetheless)
```
alichay, line 47: this comment might be (partially) outdated now?

internal/appconfig/setters.go
```diff
@@ -261,13 +261,13 @@ func (c *Config) SetStatics(statics []Static) {
 	}
 }
 
-func (c *Config) SetVolumes(volumes []Volume) {
+func (c *Config) SetVolumes(volumes []Mount) {
 	c.RawDefinition["mounts"] = volumes
 	// FIXME: "mounts" section is confusing, it is plural but only allows one mount
```
alichay, line 266: Fixme is outdated now, and maybe the function could be `SetMounts`?

internal/appconfig/from_machine_set.go
```diff
@@ -108,10 +108,10 @@ func fromAppAndOneMachine(appCompact *api.AppCompact, m machine.LeasableMachine,
 		})
 	}
 	if len(m.Machine().Config.Mounts) > 0 {
-		mounts = &Volume{
-			Destination: m.Machine().Config.Mounts[0].Path,
-		}
+		mount := m.Machine().Config.Mounts[0]
+		mounts = append(mounts, Mount{Source: mount.Name, Destination: mount.Path})
```
alichay, line 112: this is a total nit, but imo setting the slice to a new value seems more clear than appending to a nil slice
gave me the impression that the machine _could_ end up with multiple mounts, even if in reality it couldn't
```suggestion
		mounts = []Mount{{Source: mount.Name, Destination: mount.Path}}
```

internal/appconfig/config.go
```diff
@@ -190,14 +191,11 @@ func (cfg *Config) BuildStrategies() []string {
 	return strategies
 }
 
-func (c *Config) Volumes() []Volume {
+func (c *Config) Volumes() []Mount {
```
alichay, line 194: Should this be called `Mounts` now?

internal/appconfig/setters.go
```diff
@@ -261,13 +261,13 @@ func (c *Config) SetStatics(statics []Static) {
...
 	if len(volumes) > 0 {
```
alichay, line 267: Should this just set `c.Mounts` to `volumes` directly instead of pulling the first one?

alichay (commented):
so sorry I did not intend to leave like eight hundred nitpick comments

internal/command/migrate_to_v2/volumes.go
dangra: looks like so but it should be working the same as before. I didn't stop to look what it was truly doing to be honest

internal/appconfig/setters.go
dangra: good idea 👍

dangra: done db8b4e13

internal/command/migrate_to_v2/volumes.go
alichay: So, this is 1am brain, so absolutely don't take my word for this, but I think because of the interactions between the config flattening stuff you did and adding the `processes` field to mounts, this might Just Work:tm: now? Not sure, I can see about it on monday when I start finishing up all the volume stuff :)

internal/appconfig/config.go
dangra: I'm afraid it can't because there is a field with the same name.

it could be `GetMounts` but not sure we need a version that reads volumes from rawDefinition. 

What was the motivation to add it?

internal/appconfig/setters.go
dangra: good call! 34480705

internal/appconfig/config.go
alichay: The motivation was the fact that Volumes (now Mounts) wasn't an array and didn't have the processes field in the struct definition.
It should be fine to get rid of, actually, yeah

dangra: @alichay do you mind if we leave that to a follow up PR? I'd like to merge this before it diverges even more

alichay: yeah no problem

alichay (approved):

internal/appconfig/config.go
dangra: thanks

PR #2089, opened 2023-04-15
Persist --env flags to fly.toml when running "fly launch"

benwaffle:


internal/command/launch/launch.go
```diff
@@ -256,6 +258,16 @@ func run(ctx context.Context) (err error) {
 		appConfig.SetInternalPort(n)
 	}
 
+	env := flag.GetStringSlice(ctx, "env")
+	if len(env) > 0 {
+		vars, err := cmdutil.ParseKVStringsToMap(env)
```
dangra, line 263: failing for a flag parsing error at this point is a bit too late. It has created the app, the database and run scanners; all expensive operations that takes time until reaching this point of deception 😁

benwaffle: thanks, moved it earlier

dangra (approved):
great thanks.

PR #2090, opened 2023-04-16
Convert `processes = "app"` into `processes = ["app"]`

dangra:


dangra:
Docs were improved to make it clear the value must be a list of strings.

PR #2091, opened 2023-04-16
Move to Service.Autoscaled and MachineService.Autostart/Autostop

dangra:
The whole idea is that autoscaling is per service so it must be in services section.

The simplest thing to do is to enable or disable it with a single field `autoscaled`, later we can add a section to tune min/max/spare whatever else is needed.

```toml
[http_service]
 internal_port = 8080
 force_https = true
 autoscaled = true
```


internal/appconfig/machines_test.go
```diff
@@ -257,3 +255,43 @@ func TestToReleaseMachineConfig_processGroupsAndMounts(t *testing.T) {
...
+		{
+			Protocol:     "udp",
+			InternalPort: 1000,
+			Autostart:    api.Pointer(false),
+			Autostop:     api.Pointer(false),
+		},
+		{
+			Protocol:     "udp",
+			InternalPort: 2000,
+			Autostart:    nil,
```
senyosimpson, line 289: Thinking about it now, if the user doesn't specify `autoscaled`, we probably want to ensure that these fields are set to `autostart=false, autostop=false` instead of the defaults by web. Mostly because, it would be odd if you stop/suspend an instance and it just comes back up. And it won't be stopped since we have autostop off by default.

senyosimpson (approved):
One comment about defaults but looks good!

internal/appconfig/machines_test.go
dangra: My motivation not to for autostart/autostop here, was that the defaults by web are what existing apps are expecting when they don't touch autostart/autostop fields at all. 

Those defaults are meant to let fly-proxy start machines when needed but let the machines stop by themselves when they think are idling.

So basically, if you don't set `autoscaled` we are effectively on old well known legacy behavior.

senyosimpson: In this particular case, the defaults were set for customers using machines directly so that behaviour wouldn't change. For apps, the autostart/stop default means even if we stop it manually, the proxy will start it later if requests come in. So I think we should set both to false if the user does not set it explicitly

rugwirobaker: > . For apps, the autostart/stop default means even if we stop it manually, the proxy will start it later if requests come in.

Doesn't that mean commands like `fly machine stop` should also optionally set `auto_start` to false so that the proxy doesn't restart it when I don't want it too?

dangra: I think manually stopped machines are not started by fly-proxy already

dangra: Closing because of the move to `auto_start_machines` and `auto_stop_machines`. We will have to revisit once autoscaling is a thing. thanks!

PR #2092, opened 2023-04-16
Have the node scanner check for remix and prisma

rubys:
Work in progress: not fully tested, nor does it yet fully handle all of what the current remix scanner does (in particular, doesn't mount a volume when sqlite3 and prisma are detected together).

See https://github.com/superfly/flyctl/pull/2062 for details.


scanner/node.go
```diff
@@ -12,6 +12,9 @@ func configureNode(sourceDir string, config *ScannerConfig) (*SourceInfo, error)
 		return nil, nil
 	}
 
+	remix := checksPass(sourceDir, dirContains("package.json", "remix"))
```
MichaelDeBoey, line 15: Since we're deprecating the `remix` package won't be available. It would be better to check against `@remix-run/dev`

```suggestion
	remix := checksPass(sourceDir, dirContains("package.json", "@remix-run/dev"))
```

MichaelDeBoey: We could also keep what we had, as that way we don't rely on package names

```suggestion
	remix := checksPass(sourceDir, fileExists("remix.config.js"))
```

If we do this, I would suggest to also check for the `*.prisma` file again 🤔

```go
	prisma := checksPass(sourceDir, fileExists("*.prisma"))
```

scanner/templates/node/Dockerfile
```diff
@@ -5,17 +5,22 @@ ARG NODE_VERSION={{ .nodeVersion }}
 {{ if .yarn -}}
 ARG YARN_VERSION={{ .yarnVersion }}
 {{ end }}
-RUN apt-get update; apt install -y curl python-is-python3 pkg-config build-essential && \
+RUN apt-get update; apt install -y curl python-is-python3 pkg-config build-essential {{ if .prisma }}openssl {{ end }}&& \
```
MichaelDeBoey, line 8: Do we _really_ need `curl`, `python-is-python3`, `pkg-config` & `build-essential`?

We didn't use them in the `remix`/`remix_prisma` templates & I'd like to have the `Dockerfile` as clean as possible, like we had in the other templates tbh.
Especially since we will have these files (`Dockerfile` & `start_with_migrations`) in the source code of our [Remix Stacks](https://remix.run/stacks).

So basically what I think I would want once this PR is merged is getting the same `Dockerfile` as we did before with the `remix`/`remix_prisma` templates

MichaelDeBoey: Something like the following should work I think

```Docker
# base node image
FROM node:{{ .nodeVersion }}-bullseye-slim as base
```

MichaelDeBoey: Thinking more about this: should we only set major version here instead of something like `16.15.1`? 🤔

MichaelDeBoey:
The more I think about this, the more I'm leaning towards keeping a separate Remix scanner tbh.

As I mentioned in my own PR:

> I think the decision for a separate scanner for these frameworks (NextJS/Redwood/Remix/...) was made to not pollute the runtime (Node/Deno/Ruby/...) scanner too much with framework specific things tbh 🤔

What about keeping the Remix scanner, but updating it to be compatible with different package managers, supporting Prisma & (if Prisma is used) different Prisma databases?

scanner/node.go
rubys: What dirContains does is look for the string in the file, basically like grep.  So remix may produce false positives, but would match remix-run too.  I'm fine either way.

scanner/templates/node/Dockerfile
rubys: curl is needed by node-build.

The question to ponder: do any remix applications make use of any server side npm modules that may need to be compiled using node-gyp?  If so, the remaining three are required.  Not including them will lead to a bad first experience when deploying such applications on fly.io.  Also, these packages are only installed during the build phase and are not present on the deployed machine.

the next release of debian is likely to come out this summer, and after that point new versions will likely be based on bookworm.  Do we really want people's first experience with a new version of node to be in production?

rubys:
> I think the decision for a separate scanner for these frameworks (NextJS/Redwood/Remix/...) was made to not pollute the runtime (Node/Deno/Ruby/...) scanner too much with framework specific things tbh

I have a different theory.  Before I added template support, scanners would only copy files as is without modification into your application.  So any tailoring would need to be a separate set of templates.  That lead to two templates for Remix, one for sqlite and prisma, and the other nominally not being for prisma but effectively being postgres and prisma - which your pending pull request would undo and make it prisma-less but leading to a suboptimal first experience for prisma and postgres developers.

And that is before having to deal with things like npm vs yarn vs pnpm.  Or for changes in debian releases.  Or for being prepared to install packages that may require node-gyp.  Or....

scanner/node.go
MichaelDeBoey: Hmmm I think I would want to check for either `remix` or `@remix-run/dev` (since they're some sort of glob) in `package.json` as you technically could have no `remix.config.js` file and just use all basic config options 🤔

I do have a slight preference of `@remix-run/dev` though, but don't want to fight for it either 🤷‍♂️

scanner/templates/node/Dockerfile
MichaelDeBoey: > `curl` is needed by `node-build`.

But we only need `node-build` because we're building Node ourselves.
Using an already existing Node image (by doing `node:{{ .nodeVersion }}-bullseye-slim`) would have the same result without extra code in the `Dockerfile`

> do any remix applications make use of any server side npm modules that may need to be compiled using node-gyp? If so, the remaining three are required

The current Remix templates just use 

```Docker
FROM node:16-bullseye-slim as base
```

and it just works fine for most use-cases.

> Not including them will lead to a bad first experience when deploying such applications on fly.io

People own the `Dockerfile` themselves, so if people have special needs, they can always include `python-is-python3`, `pkg-config` & `build-essential` themselves.
I think the majority of people won't need it, so leaving them out by default wouldn't be a problem.

If people really need it (more than expected), we can always add something extra to the scanner and add them conditionally, just like we do with `openssl` for Prisma 🤷‍♂️

> these packages are only installed during the build phase and are not present on the deployed machine.

That's exactly my point: I don't think we need to build Node ourselves & we can just use the existing Node images

> Do we really want people's first experience with a new version of node to be in production?

The current `remix`/`remix_prisma` templates are using `bullseye`, but I think we can just use `node:{{ .nodeVersion }}-slim` as well (I'm not that much into Docker though)

---

> Thinking more about this: should we only set major version here instead of something like `16.15.1`? 🤔

Some other thoughts I have with how the current Node template is implemented & my question above: the used version is _really_ specific, so if we both have another Node version (for instance: I have `16.2.0` and you have `16.15.1`) we would both get a different base image to start with
Using a major version would make the chance of different `Dockerfile`s much less.

Could we maybe also look if we have an `.nvmrc` file & use that version first? 🤔
An `.nvmrc` file is most of the time comitted, so even if we have a patch version in there, we would both get the same `Dockerfile` if we would create one with `fly launch`

MichaelDeBoey:
> Before I added template support, scanners would only copy files as is without modification into your application.

I didn't know the history of these templates.

I still think having separate scanners/templates for each framework could be a better solution though.

Let's look at the current templates of `NextJS`, `Redwood` & `Remix` & their differences.
If we want to include all of them into scanner/template, this could become really cumbersome to deal with all small differences I think.

Some differences I already see in the scanners (which would be extra cases that need to be handled in the Node scanner):
- NextJs/Remix needs port 8080, Redwood needs Port 8910
- NextJS has extra BuildArgs (`NEXT_PUBLIC_EXAMPLE`)
- Redwood has extra env variables (`REDWOOD_DISABLE_TELEMETRY`)

Some differences I already see in the templates (which would be extra cases that need to be handled in the Node template):
- NextJS has extra env variables (`NEXT_TELEMETRY_DISABLED`) set
- NextJS has _possible_ extra ARG instructions (`NEXT_PUBLIC_*`)
- Redwood has to copy over `api` & `web` folder
- Redwood needs to run `yarn rw build api` & `yarn rw build web`
- Remix/Prisma needs to install `openssl`
- Remix/Prisma needs to run `npx prisma generate`

> And that is before having to deal with things like npm vs yarn vs pnpm. Or for changes in debian releases. Or for being prepared to install packages that may require node-gyp. Or....

Indeed, what if we want to support another package and add them conditionally, like we currently do with Prisma?
This would make the Node template really cumbersome and really hard to maintain imo

I get the idea of having everything in 1 place, as that would need only 1 change & everyone has the update, whatever framework you're using.
But the big downside is having a lot of conditionals in your scanner/template, which doesn't outweigh the benefit of only needing to change 1 file imo.

scanner/templates/node/Dockerfile
rubys: you are correct that we don't need node-build (and therefore not curl).   Yes, `node:{{ .nodeVersion }}-slim` would work fine.

And yes `.nvmrc` files should be respected.  And [.node-version](https://stackoverflow.com/questions/27425852/what-uses-respects-the-node-version-file/), and [.tool-versions](https://asdf-vm.com/manage/configuration.html#tool-versions) and perhaps others.

I'm still strongly opposed to people with "special needs" first experience with fly.io is to see an error that they don't understand, being told to update a file that they aren't familiar with, and become an expert on an operating system that they don't use daily.

rubys:
How many files would we need to update to support the next release of Debian?

I'm concerned about things like wanting to support [nextjs/prisma](https://www.prisma.io/nextjs).

Or for that matter, any package that has special needs that potentially can be used by multiple frameworks.

scanner/node.go
rubys: I probably didn't say something clearly.  But any package.json that has `@remix-run/dev` in it clearly also has the string `remix` in it.

scanner/templates/node/Dockerfile
MichaelDeBoey: > And [.node-version](https://stackoverflow.com/questions/27425852/what-uses-respects-the-node-version-file/), and [.tool-versions](https://asdf-vm.com/manage/configuration.html#tool-versions) and perhaps others.

Haven't come across them, but we could definitely add support for these files as well I think
Only "problem" I see is the order of resolution if we have both `.nvmrc` & `.node-version`, but don't see that as a blocker of not supporting those other files, just something that we need to look into how other libraries are doing it so that we have an intuitive resolution here.

> I'm still strongly opposed to people with "special needs" first experience with fly.io is to see an error that they don't understand

People would already have that currently when running `fly launch` in a Remix app, but that's not a reason to keep it that way.
My main concern is adding extra things to `Dockerfile` that most people probably don't need.

If we _know_ what packages are needing them, we could maybe add them conditionally if these packages are present in `package.json`?
That way, people not having these would not have these in their `Dockerfile`

MichaelDeBoey:
> How many files would we need to update to support the next release of Debian?

I haven't talked about Debian releases.
I could be wrong, but as far as I know, we don't need to think about them when using `node:{{ .nodeVersion }}-slim`, because it will always use the most recent stable Debian operating system release  🤔

> I'm concerned about things like wanting to support [nextjs/prisma](https://www.prisma.io/nextjs).
> Or for that matter, any package that has special needs that potentially can be used by multiple frameworks.

I definitely understand the concern of needing to copy/paste these things into each framework template if we want to support these kind of packages.

Would there maybe be a way of re-using some logic in the scanners/`Dockerfile`? 🤔

I just think that the current frameworks we support (NextJS/Redwood/Remix) have already quit a lot of small little differences that need to be handled in both the scanner & `Dockerfile` & then I'm not even talking about supporting Prisma (which could also have a lot of different databases like PostgreSQL or SQLite) or some other library where we currently don't have extra (custom) logic for in our scanners & `Dockerfile`s

Just adding all these extra differences into the mix could make the scanner & `Dockerfile` template really hard to understand & maintain I think

scanner/templates/node/Dockerfile
```diff
@@ -29,25 +24,23 @@ RUN {{ .packager }} install && {{ .packager }} run build
 RUN {{ .packager }} install
 {{ end }}
 
-FROM debian:bullseye-slim
+FROM node:${NODE_VERSION}-slim
```
MichaelDeBoey, line 27: What about building upon previous steps, like we do in the current `remix`/`remix_prisma` templates?

That way we don't need to re-set the actual base image over and over again

rubys: I agree with more steps.  mkdir and WORKDIR should be moved to base.

I don't (yet?) see the value of separate deps, production-deps, and build.  Can you explain why they are separate steps?  A see a lot of copying and repetition.  Making a single streamlined build step seems to me to be more readable?

Unrelated, but perhaps `fly launch` should generate a 64 character hex string and run `fly secrets set SESSION_SECRET`?

MichaelDeBoey: > Can you explain why they are separate steps?

I'm not the author of the original `Dockerfile`, so I don't know the thought process behind it.
According to https://github.com/superfly/flyctl/pull/569 & [Remix-related PRs](https://github.com/superfly/flyctl/pulls?q=is%3Apr+Remix), @jsierles is the original author of the file & main maintainer of the Remix templates.

> I see a lot of copying and repetition

I think this is because
- the production image only needs a clean image + the build files & production deps
- the build step needs a clean image + all dependencies
- the production deps needs  a clean image + all dependencies to remove dev dependencies

> Making a single streamlined build step seems to me to be more readable?

I'm in favor of simplifying this, but I don't think we can as we don't want to end up with unnecessary `node_modules` in the production image 🤔

> Unrelated, but perhaps `fly launch` should generate a 64 character hex string and run `fly secrets set SESSION_SECRET`?

Should `fly launch` also set this value in `.env`?
Or log it so it can be re-used there?

I guess this is something that should be done regardless the scanner that's used?

scanner/templates/node/Dockerfile
```diff
@@ -1,38 +1,53 @@
-FROM debian:bullseye as builder
+# syntax = docker/dockerfile:1
 
-ENV PATH=/usr/local/node/bin:$PATH
 ARG NODE_VERSION={{ .nodeVersion }}
+FROM node:${NODE_VERSION}-slim as base
```
MichaelDeBoey, line 4: I guess we can combine these?

```suggestion
FROM node:{{ .nodeVersion }}-slim as base
```

Still thinking we should make `.nodeVersion` a major version instead of a patch version if it doesn't come from `.nvmrc` (or any of the other mentioned files)

scanner/templates/node/Dockerfile
```diff
@@ -1,38 +1,53 @@
...
 ARG YARN_VERSION={{ .yarnVersion }}
-{{ end }}
-RUN apt-get update; apt install -y curl python-is-python3 pkg-config build-essential && \
-    curl -sL https://github.com/nodenv/node-build/archive/master.tar.gz | tar xz -C /tmp/ && \
-    /tmp/node-build-master/bin/node-build "${NODE_VERSION}" /usr/local/node && \
-{{ if .yarn -}}
-    npm install -g yarn@$YARN_VERSION && \
-{{ end -}}
-    rm -rf /tmp/node-build-master
+RUN npm install -g yarn@$YARN_VERSION
```
MichaelDeBoey, line 12: We can combine these as well I think

```suggestion
RUN npm install -g yarn@{{ .yarnVersion }}
```

Here as well, I would think having a major version would be better

scanner/templates/node/Dockerfile
```diff
@@ -1,38 +1,53 @@
...
+RUN {{ .packager }} run build
 {{ end }}
 
-FROM debian:bullseye-slim
-
-LABEL fly_launch_runtime="nodejs"
+FROM base
 
-COPY --from=builder /usr/local/node /usr/local/node
 COPY --from=builder /app /app
```
MichaelDeBoey, line 41:
```suggestion
COPY --from=build /app /app
```

Not sure if we should add everything though 🤔
As explained in https://github.com/remix-run/indie-stack/pull/90, this would include _all_ files available in the repo

The current `Dockerfile` in our `indie-stack` does the following:

```Dockerfile
COPY --from=production-deps /myapp/node_modules /myapp/node_modules
COPY --from=build /myapp/node_modules/.prisma /myapp/node_modules/.prisma

COPY --from=build /myapp/build /myapp/build
COPY --from=build /myapp/public /myapp/public
COPY --from=build /myapp/package.json /myapp/package.json
COPY --from=build /myapp/start.sh /myapp/start.sh
COPY --from=build /myapp/prisma /myapp/prisma
```

scanner/templates/node/Dockerfile
```diff
@@ -1,38 +1,53 @@
...
 
-WORKDIR /app
-ENV NODE_ENV production
-ENV PATH /usr/local/node/bin:$PATH
+{{ if .yarn -}}
+RUN yarn install --production=true
+{{ else -}}
+RUN npm prune --production
+{{ end -}}
```
MichaelDeBoey, line 47: Do we still have the `.prisma` folder inside `node_modules` after this command?

scanner/templates/node/Dockerfile
rubys: > Do we even need the LABEL?

No, but it helps us here at fly.io prioritize where we focus our efforts.  It's not perfect, but going forward it will enable us to scan for applications that use Remix.

scanner/templates/node/Dockerfile
MichaelDeBoey: > [ARG](https://docs.docker.com/engine/reference/builder/#arg)s can be overridden at built time, and we've used them elsewhere to give a hint as to things that can be adjusted.

Didn't know this.
Let's keep all `ARG`s at the top of the file then 👍

rubys: Unfortunately, Docker is weird.  Anything mentioned by the first FROM statement needs to be above that statement.  Everything else needs to be either inside either the build stage where it is referenced, or inside a stage from which it is based.

MichaelDeBoey: Weird flex 😅

scanner/templates/node/Dockerfile
```diff
@@ -1,38 +1,68 @@
...
-WORKDIR /app
-ENV NODE_ENV production
-ENV PATH /usr/local/node/bin:$PATH
+{{ if .prisma -}}
+# Entrypoint prepares the database.
+ENTRYPOINT ["/app/docker-entrypoint"]
 
+{{ end -}}
+# Start the server by default, this can be overwritten at runtime
 CMD [ "{{ .packager }}", "run", "start" ]
```
MichaelDeBoey, line 68: Can we use the same way of running the app?

We updated the dockerfile to _always_ use the script file, so that both production & local development are running the exact same thing in https://github.com/remix-run/indie-stack/pull/98

PR #2093, opened 2023-04-17
docs: add argument guide to error message

SeiwonPark:
Added argument guideline to error message as follows:

**Before**
```
Use `fly machine list` to list machines and 
`fly machine update --metadata fly_platform_version=v2` 
to update individual machines with the metadata. 
```

**After**
```
Use `fly machine list` to list machines and 
`fly machine update --metadata fly_platform_version=v2 <machine id>` 
to update individual machines with the metadata.
```

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -174,7 +174,7 @@ func (md *machineDeployment) setMachinesForDeployment(ctx context.Context) error
 		}
 		if len(activeMachines) > 0 {
 			return fmt.Errorf(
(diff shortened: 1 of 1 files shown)


kzys (approved):

kzys:
First thought that `<>` is odd, but that's what we use already.

```
% fly machine update -h
Update a machine

Usage:
  flyctl machine update <machine_id> [flags]

Flags:
  -a, --app string              Application name
      --autostart               Automatically start a stopped machine when a network request is received (default true)
      --autostop                Automatically stop a machine when there aren't network requests for it (default true)
      --build-nixpacks          Build your image with nixpacks
...
```

PR #2095, opened 2023-04-17
make http standard services test order-independant

guillaumervls:


guillaumervls:
@dangra It's to still allocate a shared IPv4 automatically whatever the order of TLS/HTTP in the array `handlers = [ "tls", "http" ]` , for users like me who have a TOML formatter that sorts strings

dangra:
sounds good. do you mind adding a few test cases so we don't break it in future changes?

guillaumervls:
Done

PR #2102, opened 2023-04-17
`deploy`: promote group-less machines into "app" group

alichay:
This makes machines with no process group set (e.g. machines migrated to the platform with `m update`) take on the group "app".

If there are two machines in "app", and you just updated a machine to use the platform, so its group is "", after a deploy you'd have three "app" machines.

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -185,6 +185,14 @@ func (md *machineDeployment) setMachinesForDeployment(ctx context.Context) error
 		}
 	}
 
+	for _, m := range machines {
+		if m.Config != nil && m.Config.Metadata != nil {
+			if m.Config.Metadata[api.MachineConfigMetadataKeyFlyProcessGroup] == "" {
+				m.Config.Metadata[api.MachineConfigMetadataKeyFlyProcessGroup] = api.MachineProcessGroupApp
+			}
+		}
+	}
+
 	md.machineSet = machine.NewMachineSet(md.flapsClient, md.io, machines)
 	var releaseCmdSet []*api.Machine
 	if releaseCmdMachine != nil {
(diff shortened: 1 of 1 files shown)


internal/command/deploy/machines.go
```diff
@@ -185,6 +185,14 @@ func (md *machineDeployment) setMachinesForDeployment(ctx context.Context) error
 		}
 	}
 
+	for _, m := range machines {
+		if m.Config != nil && m.Config.Metadata != nil {
+			if m.Config.Metadata[api.MachineConfigMetadataKeyFlyProcessGroup] == "" {
+				m.Config.Metadata[api.MachineConfigMetadataKeyFlyProcessGroup] = md.appConfig.DefaultProcessName()
```
dangra, line 191: There must be something I am not following because I don’t understand why this is needed.

If the machine's metadata says it is a v2 app machine, I am sure the later .ToMachine("")  call will force set the default process group. Isn't it? 

Maybe that I am missing some intermediary machine grouping by process group.

dangra (approved):
Approving!  I still think it smells fishy down the line but this works as intended.

alichay:
> Approving!  I still think it smells fishy down the line but this works as intended.

mmhm, long term there's gotta be a better solution, but I usually like to hotfix destructive bugs in the interim

internal/command/deploy/machines.go
alichay: I missed this comment, I merged the PR on my phone 💀

no, this does end up making a difference - maybe it shouldn't, but it does

if you create an app with `fly apps create`, `fly m run` some machines into it, then `fly m update` them into managed machines, their process group is unset, and on the first deploy they would just get deleted because they're considered `""` instead of `"app"`

there's probably something wrong here, for sure

PR #2104, opened 2023-04-17
`migrate-to-v2`: volumes general availability

alichay:
Removes the experimental flag from volume migrations, and adds support for migrating apps with multiple* volumes
<sup>*one per process group</sup>

### DIFF
--- a/internal/command/migrate_to_v2/volumes.go
+++ b/internal/command/migrate_to_v2/volumes.go
@@ -8,7 +8,6 @@ import (
 	"github.com/samber/lo"
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/internal/appconfig"
-	"github.com/superfly/flyctl/internal/flag"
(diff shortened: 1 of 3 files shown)


internal/command/migrate_to_v2/volumes.go
```diff
@@ -93,9 +101,9 @@ func (m *v2PlatformMigrator) nomadVolPath(v *api.Volume) string {
 	// so we have to account for that here
 	name := nomadVolNameToV2VolName(v.Name)
 
-	// TODO(ali): Process group-specific volumes likely change the logic here
 	for _, mount := range m.appConfig.Mounts {
-		if mount.Source == name {
+		inGroup := len(mount.Processes) == 0 || lo.Contains(mount.Processes, group)
```
dangra, line 105: nit: `strings.Contains` is in go stdlib now.

dangra (approved):
Nice!

PR #2107, opened 2023-04-17
`fly console`

matttpt:
Here's an experimental, very hacky version of `fly console`.

* `fly console`: run the `console_command` from your `fly.toml` (or `/bin/sh` if none is set) in a new `shared-cpu-1x` machine
* `fly console <machine id>`: run the `console_command` in the specified existing machine
* `fly console --select`: show a menu to decide what machine to run the `console_command` in, then do it

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -17,7 +17,7 @@ const (
 	MachineFlyPlatformVersion2                 = "v2"
 	MachineProcessGroupApp                     = "app"
 	MachineProcessGroupFlyAppReleaseCommand    = "fly_app_release_command"
-	MachineProcessGroupFlyAppEphemeralRunner   = "fly_app_ephemeral_runner"
+	MachineProcessGroupFlyAppConsole           = "fly_app_console"
 	MachineStateDestroyed                      = "destroyed"
 	MachineStateDestroying                     = "destroying"
 	MachineStateStarted                        = "started"
@@ -82,8 +82,8 @@ func (m *Machine) IsFlyAppsReleaseCommand() bool {
 	return m.IsFlyAppsPlatform() && m.IsReleaseCommandMachine()
 }
 
-func (m *Machine) IsFlyAppsEphemeralRunner() bool {
-	return m.IsFlyAppsPlatform() && m.HasProcessGroup(MachineProcessGroupFlyAppEphemeralRunner)
+func (m *Machine) IsFlyAppsConsole() bool {
+	return m.IsFlyAppsPlatform() && m.HasProcessGroup(MachineProcessGroupFlyAppConsole)
 }
 
 func (m *Machine) IsActive() bool {
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -17,6 +17,7 @@ const (
 	MachineFlyPlatformVersion2                 = "v2"
 	MachineProcessGroupApp                     = "app"
 	MachineProcessGroupFlyAppReleaseCommand    = "fly_app_release_command"
+	MachineProcessGroupFlyAppEphemeralRunner   = "fly_app_ephemeral_runner"
 	MachineStateDestroyed                      = "destroyed"
 	MachineStateDestroying                     = "destroying"
 	MachineStateStarted                        = "started"
@@ -79,6 +80,10 @@ func (m *Machine) IsFlyAppsPlatform() bool {
 
 func (m *Machine) IsFlyAppsReleaseCommand() bool {
 	return m.IsFlyAppsPlatform() && m.IsReleaseCommandMachine()
+}
+
+func (m *Machine) IsFlyAppsEphemeralRunner() bool {
+	return m.IsFlyAppsPlatform() && m.HasProcessGroup(MachineProcessGroupFlyAppEphemeralRunner)
 }
 
 func (m *Machine) IsActive() bool {
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -32,10 +32,11 @@ func NewConfig() *Config {
 // Config wraps the properties of app configuration.
 // NOTE: If you any new setting here, please also add a value for it at testdata/rull-reference.toml
 type Config struct {
-	AppName       string        `toml:"app,omitempty" json:"app,omitempty"`
-	PrimaryRegion string        `toml:"primary_region,omitempty" json:"primary_region,omitempty"`
-	KillSignal    *string       `toml:"kill_signal,omitempty" json:"kill_signal,omitempty"`
-	KillTimeout   *api.Duration `toml:"kill_timeout,omitempty" json:"kill_timeout,omitempty"`
+	AppName        string        `toml:"app,omitempty" json:"app,omitempty"`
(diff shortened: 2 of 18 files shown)


internal/appconfig/definition_test.go
```diff
@@ -161,10 +161,11 @@ func TestToDefinition(t *testing.T) {
...
 	assert.Equal(t, &api.Definition{
-		"app":            "foo",
-		"primary_region": "sea",
-		"kill_signal":    "SIGTERM",
-		"kill_timeout":   int64(3),
+		"app":             "foo",
+		"primary_region":  "sea",
+		"kill_signal":     "SIGTERM",
+		"kill_timeout":    int64(3),
+		"console_command": "/bin/bash",
```
dangra, line 168: thanks!

internal/appconfig/machines.go
```diff
@@ -49,6 +49,36 @@ func (c *Config) ToReleaseMachineConfig() (*api.MachineConfig, error) {
...
 }
 
+func (c *Config) ToConsoleMachineConfig() (*api.MachineConfig, error) {
+	mConfig := &api.MachineConfig{
+		Init: api.MachineInit{
+			// TODO: this is an ugly hack, abusing Hallpass as a
+			// stand-in for `/bin/sleep infinity`, simply because it
+			// should always be available.
+			Cmd:        []string{"dummy", "to", "keep", "machine", "running"}, // just to clear the CMD; Hallpass ignores it
+			Entrypoint: []string{"/.fly/hallpass"},
```
dangra, line 59: not a fan but you can use `Exec` instead of `Entrypoint`+`Cmd`.  "Exec" will replace both at the same time. See https://flyio.discourse.team/t/apps-v2-and-experimental-config-features/2000/8

internal/appconfig/machines.go
```diff
@@ -49,6 +49,36 @@ func (c *Config) ToReleaseMachineConfig() (*api.MachineConfig, error) {
...
+		},
+		Restart: api.MachineRestart{
+			Policy: api.MachineRestartPolicyNo,
+		},
+		AutoDestroy: true,
+		DNS: &api.DNSConfig{
+			SkipRegistration: true,
+		},
+		// TODO: should we set metadata on these machines?
```
dangra, line 68: Yes! setting the process group and platform is particularly important. Also ensuring those machines are not part of fly deploy.

internal/appconfig/machines.go
```diff
@@ -49,6 +49,36 @@ func (c *Config) ToReleaseMachineConfig() (*api.MachineConfig, error) {
...
+		Env: lo.Assign(c.Env),
+	}
+
+	// TODO: ugly hack; see above
+	mConfig.Env["SSH_LISTEN"] = "[::1]:22"
+
+	if c.PrimaryRegion != "" {
+		mConfig.Env["PRIMARY_REGION"] = c.PrimaryRegion
```
dangra, line 76: Once you add the process group metadata, it has to set FLY_PROCESS_GROUP here too

internal/appconfig/machines.go
matttpt: Thanks for the tip! Unfortunately it doesn't seem to be working for me, which sounds like a bug somewhere in flyd.

In any case I'd really like to replace this hack with something sane. I'll probably bring this up again in Discourse today.

internal/appconfig/config.go
```diff
@@ -29,21 +29,22 @@ func NewConfig() *Config {
...
+	Deploy         *Deploy                   `toml:"deploy, omitempty" json:"deploy,omitempty"`
+	Env            map[string]string         `toml:"env,omitempty" json:"env,omitempty"`
+	HTTPService    *HTTPService              `toml:"http_service,omitempty" json:"http_service,omitempty"`
+	Metrics        *api.MachineMetrics       `toml:"metrics,omitempty" json:"metrics,omitempty"`
+	Statics        []Static                  `toml:"statics,omitempty" json:"statics,omitempty"`
+	Mounts         []Mount                   `toml:"mounts,omitempty" json:"mounts,omitempty"`
+	Processes      map[string]string         `toml:"processes,omitempty" json:"processes,omitempty"`
+	Checks         map[string]*ToplevelCheck `toml:"checks,omitempty" json:"checks,omitempty"`
+	Services       []Service                 `toml:"services,omitempty" json:"services,omitempty"`
+	ConsoleCommand string                    `toml:"console_command,omitempty" json:"console_command,omitempty"`
```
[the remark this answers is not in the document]
matttpt: Good point!

matttpt:
A few notable changes since the original:

* If you want to use an existing machine, it's now `fly console --machine <id>` instead of `fly console <machine id>`. I figure that leaving the arguments open for something like a command override (down the road) is a better idea than using them for the machine ID.
* The ephemeral machine size can be configured with `--cpus` and `--memory`.
* This now sets metadata on the ephemeral runner machines.

This still uses the ugly run-Hallpass hack to run the machine without running a "real" command, but I'm marking it ready for review since everything else is where I'd like it to be for a first pass at this.

dangra:
> * If you want to use an existing machine, it's now `fly console --machine <id>`

`--select` or `-s` is another common flag option used by `fly machine` commands

internal/command/console/console.go
```diff
@@ -0,0 +1,349 @@
...
+				ID:      machine.ID,
+				Timeout: api.Duration{Duration: stopTimeout},
+			}
+			if err := flapsClient.Stop(stopCtx, stopInput, ""); err != nil {
+				terminal.Warnf("Failed to stop ephemeral runner machine: %v\n", err)
+				terminal.Warn("You may need to destroy it manually (`fly machine destroy`).")
+				return
+			}
+
+			fmt.Fprintf(io.Out, "Waiting for ephemeral runner machine %s to be destroyed ...", colorize.Bold(machine.ID))
```
dangra, line 128: Just tried the command, it was amazing but this wait at the end felt weird. It blocked my terminal with a "long" pause before it becomes usable again.

api/machine_types.go
```diff
@@ -17,6 +17,7 @@ const (
 	MachineFlyPlatformVersion2                 = "v2"
 	MachineProcessGroupApp                     = "app"
 	MachineProcessGroupFlyAppReleaseCommand    = "fly_app_release_command"
+	MachineProcessGroupFlyAppEphemeralRunner   = "fly_app_ephemeral_runner"
```
dangra, line 20: ephemeral_runner seems very generic, do you have more use cases in mind?
There was a discussion about showing the value of the process group name on the dashboard UI, having `fly_app_console` will read better than a generic `ephemeral runner`. may not..

matttpt: I deliberately left this open in case we end up implementing a `fly run`–type command in the future, since internally a few people seem enthusiastic about such a thing. (I sort of wonder if we’ll get any customer feedback along these lines once `fly console` is shipped.)

Would definitely be open to renaming now and creating another process group name if/when that happens, I was just concerned about how many special process group names we want to keep track of.

matttpt: Cool, thanks! Switched it to `fly_app_console` as suggested for now. If this ever grows into `fly run` with more process groups I think the namespace idea is a good call.

PR #2109, opened 2023-04-17
Include a reference link to docs in generated fly.toml

dangra:
```toml
# fly.toml app configuration file generated for mounts-per-group on 2023-04-17T22:03:13Z
#
# See https://fly.io/docs/reference/configuration/ for reference documentation about this file
#
app = "mounts-per-group"
primary_region = "iad"

[http_service]
  internal_port = 8080
  force_https = true
  processes = ["app"]

[[mounts]]
  source = "webcache"
  destination = "/cache"
  processes = ["app"]

[[mounts]]
  source = "video_stage"
  destination = "/stage"
  processes = ["video"]

[processes]
  app = "/start webserver"
  notifier = "/start notifier"
  video = "/start video-converter"
```

### DIFF
--- a/internal/appconfig/serde.go
+++ b/internal/appconfig/serde.go
@@ -15,6 +15,12 @@ import (
 	"github.com/superfly/flyctl/iostreams"
 )
 
+const flytomlHeader = `# fly.toml file generated for %s on %s\n\n", c.AppName, time.Now().Format(time.RFC3339))
+#
+# See https://fly.io/docs/reference/configuration/ for reference documentation about this file
+#
+`
+
 // LoadConfig loads the app config at the given path.
 func LoadConfig(path string) (cfg *Config, err error) {
 	buf, err := os.ReadFile(path)
@@ -37,7 +43,7 @@ func (c *Config) WriteTo(w io.Writer) error {
 	if err != nil {
 		return err
 	}
-	_, err = fmt.Fprintf(w, "# fly.toml file generated for %s on %s\n\n", c.AppName, time.Now().Format(time.RFC3339))
+	_, err = fmt.Fprintf(w, flytomlHeader, c.AppName, time.Now().Format(time.RFC3339))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


internal/appconfig/serde.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/superfly/flyctl/iostreams"
 )
 
+const flytomlHeader = `# fly.toml file generated for %s on %s\n\n", c.AppName, time.Now().Format(time.RFC3339))
```
alichay, line 18: there's interpolation here that got encoded into the raw string on accident

alichay (changes requested):

internal/appconfig/serde.go
catflydotio, line 18: How about 
```
fly.toml app configuration file generated for mounts-per-group on 2023-04-17T22:03:13Z
```
to make it more obvious what `fly.toml` is for?

internal/appconfig/serde.go
```diff
@@ -15,6 +15,12 @@ import (
...
+#
+# See https://fly.io/docs/reference/configuration/ for reference documentation about this file
```
catflydotio, line 20:
```suggestion
# See https://fly.io/docs/reference/configuration/ for  information about how to use this file.
```

Just looking for ways to connect more pathways in peoples' brains

internal/appconfig/serde.go
dangra: nice catch

dangra: sounds good to me. you rocks.

internal/appconfig/serde.go
dangra: that is perfect. thanks

internal/appconfig/serde.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/superfly/flyctl/iostreams"
 )
 
+const flytomlHeader = `# fly.toml app configuration file generated for %s on %s
+#
+# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
+#
+`
+
```
alichay, line 23: This is *such* a nitpick and it doesn't _really_ matter but IMO it looks nicer when there's a new line between the header comment and the content. This is a really nice change!
```suggestion
const flytomlHeader = `# fly.toml app configuration file generated for %s on %s
#
# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
#

`

```

alichay (approved):

PR #2110, opened 2023-04-17
[v2] Set FLY_PROCESS_GROUP envvar

dangra:
I wrote an app that respond differently depending on what process group it is started.
I felt like this was missing to make it simpler.


alichay (approved):
I saw you pinged Kurt about this but here's a ✅ just in case :)

PR #2112, opened 2023-04-18
Check if command not found when showing --help text

andreolvr:
Fixes https://github.com/superfly/flyctl/pull/2078#issuecomment-1509249262

Screenshot 2023-04-18 at 11 58 01
Screenshot 2023-04-18 at 11 57 43

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -50,8 +50,7 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 	default:
 		printError(io.ErrOut, cs, cmd, err)
 
-		var _, _, e = cmd.Find(args)
-
+		_, _, e := cmd.Find(args)
 		if e != nil {
 			fmt.Printf("Run '%v --help' for usage.\n", cmd.CommandPath())
 			fmt.Println()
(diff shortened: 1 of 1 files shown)


internal/cli/cli.go
```diff
@@ -49,6 +49,14 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 		return 0
 	default:
 		printError(io.ErrOut, cs, cmd, err)
+
+		var _, _, e = cmd.Find(args)
+
```
dangra, line 54: this is a nitpick, take it or leave it. 

the extra line usually reads bad. The common Go style for error checking is for it to happen right after the function call. Also using `_, e := cmd.Find(args)` is preferred to `var _, e = cmd.Find(args)`  

```go
 _, _, e := cmd.Find(args)
if e != nil {
  // handle error
}
```

andreolvr: I'll take it 😄 

https://github.com/superfly/flyctl/pull/2112/commits/01b3e8eba8398f821ed62070c77257775c2b258e

dangra (approved):

PR #2114, opened 2023-04-18
migration of mounts with empty processes list

dangra:


internal/command/migrate_to_v2/volumes.go
```diff
@@ -53,7 +56,6 @@ func (m *v2PlatformMigrator) validateVolumes(ctx context.Context) error {
 }
 
 func (m *v2PlatformMigrator) migrateAppVolumes(ctx context.Context) error {
-
 	m.appConfig.SetMounts(lo.Map(m.appConfig.Mounts, func(v appconfig.Mount, _ int) appconfig.Mount {
 		v.Source = nomadVolNameToV2VolName(v.Source)
 		return v
```
alichay, line 61: Personally, I'd prefer it ends up in here, both because this is where the other patch happens to migrate the volume config, and because modifying in a validation function is easy to miss.
```suggestion
	m.appConfig.SetMounts(lo.Map(m.appConfig.Mounts, func(v appconfig.Mount, _ int) appconfig.Mount {
	    if len(v.Processes) == 0 {
	        v.Processes = m.appConfig.ProcessNames()
	    }
		v.Source = nomadVolNameToV2VolName(v.Source)
		return v
```

internal/command/migrate_to_v2/volumes.go
```diff
@@ -23,8 +23,11 @@ func (m *v2PlatformMigrator) validateVolumes(ctx context.Context) error {
 	m.usesForkedVolumes = numMounts != 0
 
 	volsPerProcess := map[string]int{}
-	for _, m := range m.appConfig.Mounts {
-		for _, p := range m.Processes {
+	for _, mount := range m.appConfig.Mounts {
+		if len(mount.Processes) == 0 {
+			mount.Processes = m.appConfig.ProcessNames()
+		}
```
alichay, line 29:
```suggestion
```

internal/command/migrate_to_v2/volumes.go
dangra: totally makes sense. thanks

internal/command/migrate_to_v2/volumes.go
dangra: still left a call to processNames there but without the side effect because validation was missing the case of empty processes list too

dangra: double check please

alichay:
ohhhhh okay I see why you did it that way now, sorry for giving you grief!

internal/command/migrate_to_v2/volumes.go
```diff
@@ -53,9 +57,11 @@ func (m *v2PlatformMigrator) validateVolumes(ctx context.Context) error {
 }
 
 func (m *v2PlatformMigrator) migrateAppVolumes(ctx context.Context) error {
-
 	m.appConfig.SetMounts(lo.Map(m.appConfig.Mounts, func(v appconfig.Mount, _ int) appconfig.Mount {
 		v.Source = nomadVolNameToV2VolName(v.Source)
+		if len(v.Processes) == 0 {
+			v.Processes = m.appConfig.ProcessNames()
+		}
```
alichay, line 64: If processes is set after `nomadVolNameToV2VolName`, I think the check in that function would return the wrong volume. Swapping the order of these operations should fix it

dangra: good point!

dangra: hmm now it is confusing, the `nomadVolNameToV2VolName` doesn't depend on the process group, it is quite simple actually:

```go
func nomadVolNameToV2VolName(name string) string {
  return name + forkedVolSuffix
}
```

alichay: oh i'm tired sorry! you're right

alichay (approved):

PR #2115, opened 2023-04-18
Fix console modes on Windows

aarol:
This fix restores the console mods properly on Windows.

`setupConsole ` is not returning the correct stdin, stdout and stderr values due to variable shadowing.


kzys (approved):

dangra:
the happy path for this code looks so odd! thanks for the fix.

PR #2118, opened 2023-04-19
Add metrics

alichay:
This should help us understand user problems better, and make it easier to fix pain points.

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/internal/metrics"
 
 	"github.com/superfly/flyctl/iostreams"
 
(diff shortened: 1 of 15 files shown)


alichay:
The service isn't live yet, but it all works when I run it locally

internal/command/deploy/deploy.go
```diff
@@ -177,10 +177,15 @@ func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, args Dep
...
+func deployToMachines(ctx context.Context, appConfig *appconfig.Config, appCompact *api.AppCompact, img *imgsrc.DeploymentImage) (err error) {
 	// It's important to push appConfig into context because MachineDeployment will fetch it from there
 	ctx = appconfig.WithConfig(ctx, appConfig)
 
+	metrics.Started(ctx, "deploy_machines")
+	defer func() {
+		metrics.Status(ctx, "deploy_machines", err == nil)
+	}()
+
 	md, err := NewMachineDeployment(ctx, MachineDeploymentArgs{
```
dangra, line 189: what is the difference with 
```
defer metrics.Status(ctx, "deploy_machines", err == nil) 
```

dangra (approved):
Looks good!

internal/command/deploy/deploy.go
alichay: Is that the same?
I'd assume that, written like that, `err == nil` would be evaluated in-place, rather than at the end of scope, but I've never formally studied the language :)

alichay: Based on this, I *think* what I did was right
image

dangra: indeed. you are right.

PR #2123, opened 2023-04-19
Add preflight test for HA setup

dangra:


test/preflight/apps_v2_integration_test.go
```diff
@@ -419,7 +419,7 @@ func TestAppsV2Config_ProcessGroups(t *testing.T) {
 		return cmd
 	}
 
-	expectMachinesInGroups := func(machines []api.Machine, expected map[string]int) {
+	expectMachinesInGroups := func(machines []*api.Machine, expected map[string]int) {
```
kzys, line 422: Because we define methods on `*api.Machine` not `api.Machine`?

kzys (commented):
Only non-blocking comments/questions.

test/preflight/fly_launch_test.go
```diff
@@ -266,3 +268,52 @@ func TestFlyLaunch_case08(t *testing.T) {
 	require.Equal(f, 1, len(ml))
 	require.Equal(f, "shared-cpu-4x", ml[0].Config.Guest.ToSize())
 }
+
+// test default HA setup
+func TestFlyLaunch_case09(t *testing.T) {
```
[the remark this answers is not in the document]
dangra: that is a good poitn but I'm just following the pattern. Don't mind changing it later but it has been very few test additions .

test/preflight/apps_v2_integration_test.go
dangra: because that is what the main flyctl codebase does. no good reason, just that I was expecting the reference for all functions that take a list of machines

PR #2124, opened 2023-04-19
Restore `--json` flag for commands under /cmd

### LINKED ISSUE #2119: `flyctl regions list` lost `--json` support
**Describe the bug**

Since recently, `flyctl regions list --json` doesn't work anymore ("Error: unknown flag: --json"). I need to parse the output of that command in a script, thus I'd like to have JSON output back.

* Operating system: Linux (Ubuntu 22.04)
* `fly version`: v0.0.526

matttpt:
Should fix #2119. Looks like several old commands in `/cmd` lost `--json`.


dangra (approved):
Nice one.

All this commands under /cmd have to be ported to /internal/commands at some point, but restoring --json makes sense.

PR #2126, opened 2023-04-19
Add helper to calculate app URLs, and output an app's URL on deploy

ndarilek:
- Add `URL` helper for calculating an app's URL from `appconfig.Config`.
- Refactor `fly open` to use new URL helper.
- Output an app's URL on conclusion of `fly deploy`.

### DIFF
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -4,6 +4,7 @@ package appconfig
 
 import (
 	"fmt"
+	"net/url"
 	"reflect"
 
 	"github.com/superfly/flyctl/api"
@@ -196,4 +197,19 @@ func (cfg *Config) BuildStrategies() []string {
 	}
 
 	return strategies
+}
+
+func (cfg *Config) URL() (*url.URL, error) {
+	protocol := "http"
+	for _, service := range cfg.Services {
+		for _, port := range service.Ports {
+			for _, handler := range port.Handlers {
+				if handler == "tls" {
+					protocol = "https"
+					break
+				}
+			}
+		}
+	}
+	return url.Parse(protocol + "://" + cfg.AppName + ".fly.dev")
 }
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -211,5 +211,10 @@ func (cfg *Config) URL() (*url.URL, error) {
 			}
 		}
 	}
+
+	if protocol == "http" && cfg.HTTPService != nil && cfg.HTTPService.ForceHTTPS {
+		protocol = "https"
+	}
+
 	return url.Parse(protocol + "://" + cfg.AppName + ".fly.dev")
 }
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -200,21 +200,35 @@ func (cfg *Config) BuildStrategies() []string {
 }
 
 func (cfg *Config) URL() (*url.URL, error) {
+	hasHttp := false
 	protocol := "http"
-	for _, service := range cfg.Services {
-		for _, port := range service.Ports {
-			for _, handler := range port.Handlers {
-				if handler == "tls" {
-					protocol = "https"
-					break
+	if cfg.HTTPService != nil {
+		hasHttp = true
+		if cfg.HTTPService.ForceHTTPS {
+			protocol = "https"
+		}
+	} else {
+		for _, service := range cfg.Services {
+			for _, port := range service.Ports {
+				hasTls := false
+				for _, handler := range port.Handlers {
+					if handler == "tls" {
+						hasTls = true
+					} else if handler == "http" {
+						hasHttp = true
+					}
+					if *port.Port == 443 && hasHttp && hasTls {
+						protocol = "https"
+						break
+					}
 				}
 			}
 		}
 	}
 
-	if protocol == "http" && cfg.HTTPService != nil && cfg.HTTPService.ForceHTTPS {
-		protocol = "https"
+	if hasHttp {
+		return url.Parse(protocol + "://" + cfg.AppName + ".fly.dev")
+	} else {
+		return nil, nil
 	}
-
-	return url.Parse(protocol + "://" + cfg.AppName + ".fly.dev")
 }
--- a/internal/appconfig/config_test.go
+++ b/internal/appconfig/config_test.go
@@ -1,6 +1,7 @@
 package appconfig
 
 import (
+	"net/url"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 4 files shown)


internal/appconfig/config.go
```diff
@@ -197,3 +198,18 @@ func (cfg *Config) BuildStrategies() []string {
 
 	return strategies
 }
+
+func (cfg *Config) URL() (*url.URL, error) {
```
dangra, line 202: Adding test cases for this method to internal/appconfig/config_test.go will help a lot discover the corner cases of this function.

You haven't considered apps with `http_service` section defined.

ndarilek: Thanks, this is my first time really digging into these types so I appreciate the pointer to the failure cases.

As an aside, does the lint failure give any more context visually? I see there are a couple errors but Actions' UI is notoriously inaccessible and either the raw logs aren't helpful or I'm just not searching for the right keywords.

dangra: There are 2 linter errors on lines 161 and 163 for file and they are the same:
```
ineffectual assignment to err (ineffassign)
```

I think it means the assignment to the`err` variable will be ignored because it is declared in the block.

ndarilek: OK, I believe I've addressed the http_service case. I don't necessarily want to make this perfect since it's literally replacing "http://" + appName + ".fly.dev" but please let me know if I've missed any obvious cases where an app would spin up an HTTP(S) endpoint and not get caught by this.

Also not sure what the right solution for these linter errors should be. I assumed I was just re-assigning the variable and then returning it if it wasn't nil after the switch, but I'll just be super explicit and hope that makes Go happy. :)

internal/command/deploy/deploy.go
```diff
@@ -155,13 +155,26 @@ func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, args Dep
 	case err != nil:
 		return err
 	case isV2App:
-		if err := appConfig.EnsureV2Config(); err != nil {
+		if err = appConfig.EnsureV2Config(); err != nil {
```
dangra, line 158: no need to remove the `:`, for simple function calls like this it is preferred to stick to this style:

```go
if err := functionCall(); err != nil {
  return err
}
```

internal/command/deploy/deploy.go
```diff
@@ -155,13 +155,26 @@ func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, args Dep
...
 			return fmt.Errorf("Can't deploy an invalid v2 app config: %s", err)
 		}
-		return deployToMachines(ctx, appConfig, appCompact, img)
+		err = deployToMachines(ctx, appConfig, appCompact, img)
```
dangra, line 161: same as above

internal/command/deploy/deploy.go
```diff
@@ -155,13 +155,26 @@ func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, args Dep
...
 	default:
-		return deployToNomad(ctx, appConfig, appCompact, img)
+		err = deployToNomad(ctx, appConfig, appCompact, img)
+		if err != nil {
+			return err
+		}
 	}
+
+	url, err := appConfig.URL()
+	if err == nil {
```
dangra, line 173: the only extra check I would do her is `url != ""` in case the app have no http services

internal/command/deploy/deploy.go
ndarilek: Thanks for that--I wasn't sure whether to prefer re-assignment or whether I should just use `:=` everywhere. Fixed.

internal/appconfig/config_test.go
```diff
@@ -205,3 +206,35 @@ func TestHasNonHttpAndHttpsStandardServices(t *testing.T) {
...
+	cfg.Services = []Service{{Protocol: "tcp", Ports: []api.MachinePort{
+		{Port: &port80, Handlers: []string{"http"}},
+	}}}
+	url, _ := cfg.URL()
+	assert.Equal(t, http, url)
+
+	cfg = NewConfig()
+	cfg.AppName = "test"
+	cfg.Services = []Service{{Protocol: "tcp", Ports: []api.MachinePort{
+		{Port: &port443, Handlers: []string{"tls"}},
```
dangra, line 228: "tls" alone doesn't guarantee an http service, the correct handlers are `[]string{"tls", "http"}`

internal/appconfig/config.go
```diff
@@ -197,3 +198,23 @@ func (cfg *Config) BuildStrategies() []string {
...
+				}
+			}
+		}
+	}
+
+	if protocol == "http" && cfg.HTTPService != nil && cfg.HTTPService.ForceHTTPS {
+		protocol = "https"
+	}
+
+	return url.Parse(protocol + "://" + cfg.AppName + ".fly.dev")
```
dangra, line 219: truly speaking this function can return an empty string when there are no http services defined.

dangra: we know a http service is defined when any of the following are true:
1. cfg.HTTPService is not nil
2. cfg.Services have a service with port 80 and handler "http"
3. cfg.Services have a service with port 443 and handlers "tls"+"http"

the protocol, http or https, can be inferred from HTTPService.ForceHTTPS or port equal 443

otherwise it must return an empty string to signal there is no http service, so no URL to open the app.

internal/appconfig/config.go
dangra: the http service cases are not complete tho, you need to revisit that logic because it is possible that there is no http service to infer an url from.

ndarilek:
I think I may have something that captures all of this logic, but I'm getting one final test failure I don't know how to fix. Specifically I have this assertion:

`	assert.Equal(t, nil, url)`

And I'm getting this failure:

```
--- FAIL: TestURLCalculation (0.00s)                                                                               
    config_test.go:223:                                                                                            
                Error Trace:    C:\Users\nolan\Projects\Fly\flyctl\internal\appconfig\config_test.go:223           
                Error:          Not equal:                                                                         
                                expected: <nil>(<nil>)                                                             
                                actual  : *url.URL((*url.URL)(nil))                                                
 ```

Looks like there's maybe some casting I need to do to indicate that the nil value can either be a URL or nil but I'm not sure how to do that in the context of checking an assertion.

Clearly I bit off more than I could chew with this as a flight school project, so if there's no easy fix for this test case or if I still have the logic wrong, then I may shelve it and move onto something else.

Thanks a bunch.

ndarilek:
Pushed the broken test if anyone wants to take a look. Still newish to Go so I'm probably doing something dumb. :)

Thanks.

dangra:
> Clearly I bit off more than I could chew with this as a flight school project, so if there's no easy fix for this test case or if I still have the logic wrong, then I may shelve it and move onto something else.

hey! I think you have all of that very well done. I can help you pass the broken tests and you can merge then

dangra (approved):
@ndarilek all green ! hit that button 🚢

ndarilek:
Oh hey thanks, should've checked my email before making the forum post. :)

PR #2128, opened 2023-04-19
Introducing Postgres Forks

davissp14:
This introduces the `--fork-from`  flag that can be called on Postgres provision.  Functionally, the end-result is the same as snapshotting the target volume and restoring into a new application.  The primary difference being that we are actually forking the underlying storage device and copying those blocks into a new logical volume on the same host.  It's the same result, just significantly faster. 

The downside is that this feature is currently restricted to same-host forks, which means that some volumes may not be eligible due to capacity issues with the host the target volume is residing on.  The bright side is that as long as you're replicas are sync'd, there's no reason you can't just fork a volume associated with one of your replicas.   

This flag can be specified in one of two formats:

**Option 1:**  `--fork-from <app-name>` 
This will evaluate the target app and attempt to resolve the volume associated with the primary to use as a fork target.

**Option 2:** `--fork-from <app-name>:<volume-id>`
This should be used if there's a specific volume you wish to fork from.  This is useful in the event you're unable to fork the primary due to host capacity issues, or maybe you want to fork your application into a new region.

### DIFF
--- a/flypg/launcher.go
+++ b/flypg/launcher.go
@@ -52,6 +52,7 @@ type CreateClusterInput struct {
 	Manager            string
 	Autostart          bool
 	ScaleToZero        bool
+	ForkFrom           string
 }
 
 func NewLauncher(client *api.Client) *Launcher {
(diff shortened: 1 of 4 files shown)


dangra (approved):
🚀

PR #2131, opened 2023-04-20
`version update`: check `IsUnderHomebrew` before update instead of after

alichay:
I think `IsUnderHomebrew` lied because the exe was moved by brew during update


tvdfly (approved):
Thank you!

dangra (approved):

PR #2132, opened 2023-04-20
api/client: improve constructor configurability

tychoish:
I think being able to configure these options will make it easier to
reuse existing HTTP request infrastructure in our application and call
fly.io APIs from within our application.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -42,23 +42,60 @@ type Client struct {
 
 // NewClient - creates a new Client, takes an access token
 func NewClient(accessToken, name, version string, logger Logger) *Client {
-	transport := &Transport{
-		UnderlyingTransport: http.DefaultTransport,
-		Token:               accessToken,
-		Ctx:                 context.Background(),
-		EnableDebugTrace:    !slices.Contains([]string{"", "0", "false"}, os.Getenv("FLY_FORCE_TRACE")),
-		UserAgent:           fmt.Sprintf("%s/%s", name, version),
+	return NewClientFromOptions(ClientOptions{
+		AccessToken: accessToken,
+		Name:        name,
+		Version:     version,
+		Logger:      logger,
+		BaseURL:     baseURL,
+	})
+}
+
+type ClientOptions struct {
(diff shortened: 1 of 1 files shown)


api/client.go
```diff
@@ -42,23 +42,60 @@ type Client struct {
...
+}
 
-	url := fmt.Sprintf("%s/graphql", baseURL)
+func NewClientFromOptions(opts ClientOptions) *Client {
+	var transport *Transport
+	if opts.Transport == nil {
+		transport = opts.Transport
+	} else {
+		transport = &Transport{}
+	}
```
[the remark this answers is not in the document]
tychoish: Ah good catch. Sorry, I got the if statement backwards. I had/have a slight bias towards not constructing the `Transport` object twice in the case that we run the statement on line 90. 

There aren't any existing tests in this package, but I'm happy to add one.

tychoish:
> Can you also shed more light on how changing how we setup our transport helps your application?

I was mostly hoping to be able to use an `http.Client` instance that was connected to our instance. I also saw the context in the transport and wanted to pass our own in; however, I now see that this is unused and likely vestigial.

api/client.go
```diff
@@ -42,23 +42,58 @@ type Client struct {
...
+		t.EnableDebugTrace = !slices.Contains([]string{"", "0", "false"}, os.Getenv("FLY_FORCE_TRACE"))
+	}
+}
 
-	url := fmt.Sprintf("%s/graphql", baseURL)
+func NewClientFromOptions(opts ClientOptions) *Client {
+	transport := &Transport{}
+	if opts.Transport != nil {
+		transport = opts.Transport
+	}
```
dangra, line 88: a small twist suggestion:

```go
transport := opts.Transport
if transport == nil {
  transport = &Transport{}
}
```

dangra:
needs to resolve the conflicts to merge it

api/client.go
tychoish: ah good call!

dangra:
thank you!

PR #2133, opened 2023-04-20
client: remove unused context

tychoish:
I saw during #2132 that there was a context stored in this struct
which is unused (and generally inaccessible? Though this is
unrelated.)


dangra:
thanks

PR #2134, opened 2023-04-20
Don't allow to set autoscale on Apps V2

pborzenkov:
This leads to Nomad trying to autoscale a machines app, failing and
constantly increasing a release number. See this post for reference:

https://community.fly.io/t/fly-releasing-non-stop/12334/17

### DIFF
--- a/cmd/autoscaling.go
+++ b/cmd/autoscaling.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/superfly/flyctl/client"
 	"github.com/superfly/flyctl/cmdctx"
+	"github.com/superfly/flyctl/internal/appconfig"
 
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/docstrings"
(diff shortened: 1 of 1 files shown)


cmd/autoscaling.go
```diff
@@ -163,3 +185,12 @@ func printScaleConfig(cmdCtx *cmdctx.CmdContext, cfg *api.AutoscalingConfig) {
 		}
 	}
 }
+
+func printMachinesAutoscalingBanner() {
+	fmt.Printf(`
+Configuring autoscaling via 'flyctl autoscale' is supported only for apps running on Nomad platform.
+To enable autoscale for Apps V2 create the required number of machines in advance and allow fly.io
+to start and stop them on demand by setting auto_start_machine=true and auto_stop_machine=true
```
dangra, line 193: could we link to the fresh-produce post instead of reexplaining this here? Ideally we should link to docs but they are in the making.

dangra (approved):
looks good. Except for the link to the post, the rest are nits.

dangra (approved):
nice!

PR #2139, opened 2023-04-20
--standby-for flag for fly machine run|update|clone commands

dangra:


internal/command/machine/clone.go
```diff
@@ -74,6 +74,10 @@ func newClone() *cobra.Command {
 			Name:        "clear-auto-destroy",
 			Description: "Disable auto destroy setting on new machine",
 		},
+		flag.StringSlice{
+			Name:        "standby-for",
+			Description: "Comma separated list of machine ids to watch for",
```
kzys, line 79: How about mentioning `source` here?

dangra: very good point, thanks. Added by 747be88e

dangra:
@kzys needs more time or can I get a ✅  to merge and ship asap?

kzys (approved):
LGTM!

PR #2140, opened 2023-04-20
Add a new command to provision a Sentry project for an app

jsierles:


tvdfly (approved):
Overall looks good!

How about adding an extensions section to the `fly help` output?

I expected `fly ext sentry` to show help. What do you think about making it `fly ext sentry create`?

jsierles:
Yeah it needs some help. `create` is probably good as it mimics what other extensions would do. There are no other commands because we don't have access yet to do more than just `create`. Side note: `ext` is intended as holding area for new extensions, until we decide where it goes. It could also stay here.

jsierles:
Going to keep this out of `help` for now since these commands are still not useful for the public.

PR #2144, opened 2023-04-20
Force dockerfile scanner.

rubys:
See: https://community.fly.io/t/flyctl-trying-to-auto-detect-phoenix-even-though-i-specified-a-dockerfile/12394


dangra (approved):

rubys:
@alichay had a better fix.

PR #2145, opened 2023-04-21
Split dockerfile scan out so explicit launches don't use scanners

alichay:
When specifically `--dockerfile Dockerfile` was passed, `determineSourceInfo` would try to use scanners so that the Dockerfile scanner could load relevant data.

This PR a) makes this behavior work no matter the dockerfile name, and b) prevents [framework scanners from taking over when you pass in a dockerfile](https://community.fly.io/t/flyctl-trying-to-auto-detect-phoenix-even-though-i-specified-a-dockerfile/12394).


dangra:
sorry, I'm a bit tired to understand the issue, leaving ✅ but beware of https://github.com/superfly/flyctl/pull/2144

dangra (approved):

PR #2146, opened 2023-04-21
api: avoid logging if the logging is not configured

tychoish:
Because `internal/logger.MaybeFromContext()` can sometimes return a
nil logger, it is pluisble to, when constructing loggers outside of the
package, to get a nil logger.

as in: 

- https://github.com/superfly/flyctl/blob/master/flaps/flaps.go#L96

- https://github.com/superfly/flyctl/blob/master/flaps/flaps.go#L72


alichay:
I ran into this myself, recently, trying to work with the API. nice catch, thanks!

PR #2147, opened 2023-04-21
add `apps set-platform-version` for manual migration

alichay:
this allows people to migrate certain apps with volumes with no downtime, if they're willing to do a little bit of live surgery.

sorry dov


matttpt (approved):
Two UX thoughts:

* Most other `apps` commands take the app name as an argument. (OTOH, `migrate-to-v2` doesn't, and maybe it's worth keeping it consistent with that instead?)
* Even though it's hidden, should this still include some kind of "I promise I know what I'm doing" prompt?

(These are just first impressions. Don't let them keep you from merging if they're no good!)

PR #2150, opened 2023-04-21
Add the same validation we have in flaps so we can fail early in the deploy

rugwirobaker:
Companion https://github.com/superfly/nomad-firecracker/pull/1083


alichay (approved):
We should note in comments near these limits in flaps that they're replicated in flyctl, that way they don't end up out of sync

dangra:
hey! IMO this is not the place for validations, we have a method that can validate app configuration before it reaches this point and provides better feedback to users.

 see internal/appconfig/validation.go

it should go there 🙏

rugwirobaker:
@dangra a new validation function? Cause I kept clicking through the function and got me there.

dangra:
+1 to reuse if there is one for checks

PR #2173, opened 2023-04-24
Fix `LoadAppNameIfPresent` by using correct error in `RequireAppName`

matttpt:
`LoadAppNameIfPresent` expects the error returned from `RequireAppName` to be `errRequireAppName` so that it can ignore it. (It looks like we recently improved the error message, but did so by replacing `errRequireAppName` instead of updating its message.)

See the [community forum](https://community.fly.io/t/fly-v0-0-540-fly-cli-proxy-new-a-argument/12469).

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -548,7 +548,7 @@ func appConfigFilePaths(ctx context.Context) (paths []string) {
 	return
 }
 
-var errRequireAppName = fmt.Errorf("we couldn't find a fly.toml nor an app specified by the -a flag. If you want to launch a new app, use '%s launch'", buildinfo.Name())
+var errRequireAppName = fmt.Errorf("the config for your app is missing an app name, add an app_name field to the fly.toml file or specify with the -a flag`")
 
 // RequireAppName is a Preparer which makes sure the user has selected an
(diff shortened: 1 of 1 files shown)


internal/command/command.go
```diff
@@ -548,7 +548,7 @@ func appConfigFilePaths(ctx context.Context) (paths []string) {
 	return
 }
 
-var errRequireAppName = fmt.Errorf("we couldn't find a fly.toml nor an app specified by the -a flag. If you want to launch a new app, use '%s launch'", buildinfo.Name())
+var errRequireAppName = fmt.Errorf("the config for your app is missing an app name, add an app_name field to the fly.toml file or specify with the -a flag`")
```
kzys, line 551: We don't have to change the error message here. Do you have reasons to not mention `fly launch`?

kzys (approved):

internal/command/command.go
matttpt: Oh, that's just because #2009 recently adjusted the wording for this error message. I'm just transferring the new message from a custom error in `RequireAppName` to `errRequireAppName`, which had the old wording and was now unused.

PR #2180, opened 2023-04-25
Fix FLYCTL-BE sentry error

dangra:
image

### DIFF
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -4,6 +4,7 @@ package appconfig
 
 import (
 	"fmt"
+	"net/url"
 	"reflect"
 
 	"github.com/superfly/flyctl/api"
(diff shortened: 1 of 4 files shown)


kzys:
Personally prefer to have typed values (URL) over strings. What do you want by having strings instead?

dangra:
@kzys I was trying to avoid returning an error as part of the method signature, a string looked much simpler but I understand the motivation to return a typed value. Will redo it a bit to avoid any error propagation still return `*url.URL`

kzys:
If appconfig.Config doesn't have any ports, wouldn't it need to return an error still?

dangra:
> If appconfig.Config doesn't have any ports, wouldn't it need to return an error still?

not an error but a nil value

dangra:
> Personally prefer to have typed values (URL) over strings. What do you want by having strings instead?

please see https://github.com/superfly/flyctl/pull/2180/commits/825269d3fb79f80c738042bfef5e57fc7380ff40

PR #2182, opened 2023-04-25
Workaround web bug with app.Deployed being false for machines app

dangra:


matttpt (approved):
LGTM, just two nits:

* `deployForSecrets` doesn't need a named return value any more
* This adds a link to a private issue, though to be fair the same link is already in the `deploy` command code

PR #2183, opened 2023-04-25
Fix "Ignoring Dockerfile due to config" warning

kzys:
If `[build]` is in fly.toml, appConfig.Build.Dockerfile will be "", but it shouldn't trigger "Ignoring ..." warning.

https://community.fly.io/t/deploy-warning/12424

### DIFF
--- a/internal/command/deploy/deploy_build.go
+++ b/internal/command/deploy/deploy_build.go
@@ -20,6 +20,30 @@ import (
 	"github.com/superfly/flyctl/terminal"
 )
 
+func multipleDockerfile(ctx context.Context, appConfig *appconfig.Config) error {
+	if len(appConfig.BuildStrategies()) == 0 {
+		// fly.toml doesn't know anything about building this image.
+		return nil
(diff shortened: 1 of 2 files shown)


internal/command/deploy/deploy_build.go
```diff
@@ -20,6 +20,30 @@ import (
...
+	}
+
+	config, _ := resolveDockerfilePath(ctx, appConfig)
+	if config == "" {
+		// No Dockerfile in fly.toml.
+		return nil
+	}
+
+	if found != config {
+		return fmt.Errorf("Use %s (in fly.toml) instead of %s", config, found)
```
benwaffle, line 42: I'd like to keep the word "ignoring", it's impactful enough to make people pay attention to the warning.

```suggestion
		return fmt.Errorf("Ignoring %s, and using %s (from fly.toml).", found, config)
```

benwaffle (approved):

dangra (approved):

PR #2184, opened 2023-04-26
Move lock files to ~/.fly to fix multi-user flyctl

### LINKED ISSUE #977: Support multi-user usage
**Describe the bug**
When running in a multi-user environment, `flyctl` will fail to acquire the `/tmp/flyctl.config.lock` and `/tmp/flyctl.cache.lock` file.

* Operating system
  * `Linux archlinux 5.17.7-arch1-2 #1 SMP PREEMPT Sun, 15 May 2022 05:00:14 +0000 x86_64 GNU/Linux`
* `fly version`
  * `flyctl v0.0.325 linux/amd64 Commit: da2b638 BuildDate: 2022-04-28T04:26:35Z`

**Reproduction**

- Use `flyctl` as one user. The lock files will be generated
- Log in as another user and use `flyctl`.
  ```
  ➜  ~ fly auth login
  WARN failed loading cache file from /home/[redacted]/.fly/state.yml: open /tmp/flyctl.cache.lock: permission denied
  Error open /tmp/flyctl.config.lock: permission denied
  ```

benwaffle:
Right now, when running flyctl in a multi-user environment, you run into
permission issues with the shared lock files in /tmp:

* `flyctl.cache.lock`
* `flyctl.config.lock`
* `flyctl.agent.lock`
* `flyctl.agent.start.lock`

So, I've moved them all to `~/.fly`. In `~/.fly` we already have
`config.yml` and `fly-agent.sock`, so it makes sense that the lock files
are also not shared.

Closes #977

### DIFF
--- a/agent/start.go
+++ b/agent/start.go
@@ -97,10 +97,18 @@ func (alreadyStartingError) Error() string {
 	return "another process is already starting the agent"
 }
 
-var lockPath = filepath.Join(os.TempDir(), "flyctl.agent.start.lock")
+var internalLockPath string
+
+func lockPath() string {
+	if internalLockPath == "" {
+		internalLockPath = filepath.Join(flyctl.ConfigDir(), "flyctl.agent.start.lock")
+	}
+
+	return internalLockPath
+}
 
 func lock(ctx context.Context) (unlock filemu.UnlockFunc, err error) {
-	switch unlock, err = filemu.Lock(ctx, lockPath); {
+	switch unlock, err = filemu.Lock(ctx, lockPath()); {
 	case err == nil:
 		break // all done
 	case ctx.Err() != nil:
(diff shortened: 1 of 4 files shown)


alichay (approved):
good change!

while it's totally fine and mergable as-is, what are your thoughts about not caching the paths? that way it's a little bit simpler, and we don't have to use globals - the performance hit, if any, should be basically negligible

benwaffle:
@alichay thanks, that's so much cleaner, it was pretty late when I wrote this 🤦🏻‍♂️

kzys (approved):

benwaffle:
@dangra I ran preflight checks, it's behaves the same as `master`. (`TestAppsV2MigrateToV2_Volumes` fails but the rest pass)

PR #2186, opened 2023-04-26
Use builtin go buildinfo for git version information

mjsir911:
This makes things more workable with offbeat build procedures eg package managers that might not fetch the repo from git, golang autoembeds this info when available & fails gracefully when not, or when manually requested with `-buildinfo=false` or `-buildvcs=false` (this is what gentoo does)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,4 @@
 NOW_RFC3339 = $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
-GIT_SHA = $(shell git rev-parse HEAD 2>/dev/null || no_git)
-GIT_BRANCH = $(shell git symbolic-ref --short HEAD 2>/dev/null ||:)
-ifneq ($(GIT_BRANCH),)
-GIT_COMMIT = $(GIT_SHA) ($(GIT_BRANCH))
-else
-GIT_COMMIT = $(GIT_SHA)
-endif
 
 all: build cmddocs
 
@@ -15,7 +8,7 @@ generate:
 
 build: generate
 	@echo Running Build
-	go build -o bin/flyctl -ldflags="-X 'github.com/superfly/flyctl/internal/buildinfo.buildDate=$(NOW_RFC3339)' -X 'github.com/superfly/flyctl/internal/buildinfo.commit=$(GIT_COMMIT)'" .
+	go build -o bin/flyctl -ldflags="-X 'github.com/superfly/flyctl/internal/buildinfo.buildDate=$(NOW_RFC3339)'" .
 
 test: FORCE
 	go test ./... -ldflags="-X 'github.com/superfly/flyctl/internal/buildinfo.buildDate=$(NOW_RFC3339)'" --run=$(T)
--- a/internal/buildinfo/version.go
+++ b/internal/buildinfo/version.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strconv"
 	"time"
+	"runtime/debug"
 
 	"github.com/blang/semver"
 	"github.com/superfly/flyctl/terminal"
@@ -13,7 +14,6 @@ import (
 var (
 	buildDate = "<date>"
 	version   = "<version>"
-	commit    = "<commit>"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


kzys (approved):

dangra (changes requested):
The linter spotted some issues worth checking

mjsir911:
yeah, clued me in on the tests for this. Unfortunately I can't think of an easy way to mock `debug.ReadBuildInfo()` so will just remove the git revision checking for now. Going to research how it's done elsewhere though.

kzys:
I'm fine not testing the commit part. We don't have much code around.

dangra (approved):

tvdfly (commented):
Can we bring back the branch name for dev builds, though? It's helpful to see which branch something was built on when doing a bunch of local dev.

For example, this is what I have right now after running `make` on a branch:

```
% fly version
flyctl v0.0.0-1682385080+dev darwin/arm64 Commit: c62ae16a0a9fa41d3aedc56627632b7872bc4c95 (client-init-for-libs) BuildDate: 2023-04-25T01:11:20Z
```

we don't need (or probably want) the branch names in the builds that get released.

mjsir911:
> Can we bring back the branch name for dev builds? It's helpful to see which branch something was built on when doing a bunch of local dev.
> 
> For example, this is what I have right now after running `make` on a branch:
> 
> ```
> % fly version
> flyctl v0.0.0-1682385080+dev darwin/arm64 Commit: c62ae16a0a9fa41d3aedc56627632b7872bc4c95 (client-init-for-libs) BuildDate: 2023-04-25T01:11:20Z
> ```
> 
> we don't need (or probably want) the branch names in the builds that get released.

Hey tvdfly,

Thanks for the request!, if this is a hard requirement on the version info, this PR is pretty much a non-starter.

The git command used to embed this data is revision (and timestamp) only: [`-c log.showsignature=false show -s --format=%H:%ct`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/cmd/go/internal/vcs/vcs.go;drc=bb3965695d3bc3586fba55c7b8d5e8a56cd4c0c9;l=340).

The go team does seem amenable to adding more detailed data into `buildinfo`, but that would be a pretty drawn out process.

Related issues:
- https://github.com/golang/go/issues/22147
- https://github.com/golang/go/issues/29814
- https://github.com/golang/go/issues/37475

I might end up filing a proposal to add branch name (or just custom information?)

Will let y'all close out this PR if you decide to, but thanks for considering it!

tvdfly:
What if we make our own branch var in the place we used to set the commit sha, and then set it in the makefile? If the var is set we can show it and if not we ignore it.

mjsir911:
> What if we make our own branch var in the place we used to set the commit sha, and then set it in the makefile? If the var is set we can show it and if not we ignore it.

Part of the intent of the PR was moving this tagged-info logic outside of the makefile. Doing this works but brings us back to the same initial place where building without being inside of a git repo isn't supported nicely.

Could also sunset this PR and just make something to better handle git-less builds

tvdfly:
@mjsir911 I made a branch on top of this one with what I'm thinking: https://github.com/superfly/flyctl/commit/e2a6baf6

The branch name just doesn't show up if git isn't installed or doesn't work or whatever.

Is that still going to be a problem in the cases you outline? It doesn't seem like it to me, and I'm not very familiar with the issues the motivated the original change. Appreciate your help and this contribution!

mjsir911:
> @mjsir911 I made a branch on top of this one with what I'm thinking: #2284 (this commit in particular: [e2a6baf6](https://github.com/superfly/flyctl/commit/e2a6baf6) )
> 
> The branch name just doesn't show up if git isn't installed or doesn't work or whatever.
> 
> Is that still going to be a problem in the cases you outline? It doesn't seem like it to me, and I'm not very familiar with the issues that motivated the original change. Appreciate your help and this contribution!

Sorry about the slow response, thanks for the merge! The change you linked makes everything work smoothly, yes.

I appreciate you working with me through this one.

PR #2190, opened 2023-04-26
First pass at adding --attach-volume-mount-point

bzanchet:


dangra (commented):
not sure of adding another flag tbh.

what about using the known "volume:mountpoint" format we currently use in `fly m run --volume` flag 

https://github.com/superfly/flyctl/blob/6132cb09c8f8843fb0ef774b3777066c9ed96453/internal/command/machine/run.go#L183-L187

PR #2192, opened 2023-04-27
Add an alias to `version update`: `upgrade`

urcadox:
Hello,

Every time I want to upgrade flyctl, I write `fly version upgrade` instead of `fly version update`. I guess I'm used to other CLI tools where `upgrade` is used for full version updates/upgrades.

Unless I'm the only one, this would be a nice addition. :D

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -353,14 +353,14 @@ func startQueryingForNewRelease(ctx context.Context) (context.Context, error) {
 }
 
 // shouldIgnore allows a preparer to disable itself for specific commands
-// E.g. `shouldIgnore([][]string{{"version", "update"}, {"machine", "status"}})`
-// would return true for "fly version update" and "fly machine status"
+// E.g. `shouldIgnore([][]string{{"version", "upgrade"}, {"machine", "status"}})`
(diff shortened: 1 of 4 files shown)


dangra:
@urcadox I tend to agree, also think it should be `fly version upgrade` first and `update` the alias for backwards compatibility, but before merging we need a few more maintainers to chime in and share their opinions. 

Please consider fixing the PR according to the linter suggestions. thanks

alichay:
yeah I totally support this. once the linter's happy, I'd be happy to get this in there :)

urcadox:
Great, I will fix the lint issues, and probably add another commit to transform the command into `upgrade`, with `update` as an alias.

urcadox:
I made the updates (feel free to use or not to use the commit that renames the command and makes `update` an alias; btw I did not test this as I'm not on my usual machine right now, and I don't have an easy way to install a recent enough version of Go)

alichay:
I made a couple tweaks, which means I ended up making a new PR from my branch. Going to go ahead and close this in favor of #2259. Thanks for getting the ball rolling and contributing, we really appreciate it!

PR #2195, opened 2023-04-27
Do not add a tcp check on launch

dangra:


kzys (approved):
Because we don't know what we need to check and customers most likely have HTTP apps?

kzys:
Synced-up offline. `fly launch` shouldn't have health checks. It should generate the most development-ready fly.toml. So customers can iterate on it to make it production-ready.

PR #2198, opened 2023-04-28
fix ruby dockerfile to avoid permissions error on Gemfile.lock

vyedin:


rubys:
Over time, we probably should revisit whether it makes sense to create a non-root user at all for these applications.  If we know the framework, we can make informed choices as to what files can be protected and which ones need to be writable.

vyedin:
Makes sense. I just wanted to make it work as easily as possible.

PR #2199, opened 2023-04-28
Remove sleep in redis connect

haileys:
The `fly redis connect` command currently sleeps for 3 seconds after spawning a `proxy.Connect` goroutine to wait for the proxy socket to start up.

This PR introduces a new function `proxy.Start`, which binds a local listener synchronously before spawning a goroutine to run the proxy asynchronously. This means that we can rely on the local listener being ready to accept connections as soon as `proxy.Start` returns and no further synchronisation is required.

I put together a small expect script to measure the total time taken to connect a Redis session and issue a ping command:

```expect
spawn bin/flyctl redis connect

expect "Select a database to connect to"
send -- "\n"

expect "127.0.0.1:16379>"
send -- "ping\n"

expect "127.0.0.1:16379>"
send -- "exit\n"
```

On my machine this change results in an improvement from about 4.7 seconds to 2 seconds. The reason it shaves off 2.7 seconds and not 3 seconds is that both `proxy.Connect` and `proxy.Start` make a synchronous call to the API to update the list of wireguard peers. This is about a 300ms round trip on my connection.

There are two remaining calls to `proxy.Connect` - these are both proxy commands (`machine api-proxy` and `proxy`) and don't exhibit the same requirement for synchronisation as `redis connect`.

### DIFF
--- a/internal/command/redis/connect.go
+++ b/internal/command/redis/connect.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
-	"time"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


dangra (approved):
nice finding 💎

Seems `Server.ProxyServer()`  return an error only when `Listener.SetDeadline()` fails which is unlikely yet possible.

PR #2200, opened 2023-04-28
The --attach-volume flag on machines clone now takes the mount path

bzanchet:
When clonning a machine without volumes and trying to attatch an existing volume to it, we need a way to set the mount path.

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -3,6 +3,7 @@ package machine
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/google/shlex"
@@ -56,7 +57,7 @@ func newClone() *cobra.Command {
 		},
 		flag.String{
 			Name:        "attach-volume",
-			Description: "Existing volume to attach to the new machine",
+			Description: "Existing volume to attach to the new machine in the form of <volume_id>[:/path/inside/machine]",
 		},
 		flag.String{
 			Name:        "process-group",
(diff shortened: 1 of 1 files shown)


internal/command/machine/clone.go
```diff
@@ -159,15 +160,34 @@ func runMachineClone(ctx context.Context) (err error) {
...
-		if len(source.Config.Mounts) != 1 {
-			return fmt.Errorf("Can't attach the volume as the source machine doesn't have any volumes configured")
+	var volID string
+	if volumeInfo := flag.GetString(ctx, "attach-volume"); volumeInfo != "" {
+		splitVolumeInfo := strings.Split(volumeInfo, ":")
+
+		if len(source.Config.Mounts) > 1 {
+			return fmt.Errorf("Can't use --attach-volume for machines with more than 1 volume.")
+		} else if len(source.Config.Mounts) == 1 && len(splitVolumeInfo) > 1 {
+			return fmt.Errorf("Can't set a mount path on a machine with a volume, please use only the volume id on '%s'", volumeInfo)
```
dangra, line 170: In the case the mount exists, It is possible to change both: the volume and the path when launching a new machine. there is nothing wrong with it.

bzanchet:
The intended behavior is:

- Cloning a machine w/o volumes:
`--attach-volume <volume_id>` -> error, needs a mount path
`--attach-volume <volume_id>:/data` -> works, mounts volume at `/data`

- Cloning a machine w/ one volume
`--attach-volume <volume_id>` -> works, mounts volume at existing mount path
`--attach-volume <volume_id>:/data` -> error, can't override mount path

internal/command/machine/clone.go
```diff
@@ -159,15 +160,34 @@ func runMachineClone(ctx context.Context) (err error) {
...
+		} else if len(source.Config.Mounts) == 0 && len(splitVolumeInfo) != 2 {
+			return fmt.Errorf("Couldn't find a mount path on '%s'", volumeInfo)
+		}
+
+		// patch the source config so the loop below attaches the volume on the passed mount path
+		if len(source.Config.Mounts) == 0 && len(splitVolumeInfo) == 2 {
+			volID = splitVolumeInfo[0]
```
dangra, line 177: `volID = splitVolumeInfo[0]` can be factored out of the `if` block and then you can factor even more when `len(source.Config.Mounts) == 0`

internal/command/machine/clone.go
bzanchet: yeap-- maybe in a follow up PR?

I thought this is a very specific edge-case that shouldn't prevent anything meaningful. At the same time, it could potentially prevent some errors if the user mistypes the mount path.

internal/command/machine/clone.go
bzanchet: good catch

internal/command/machine/clone.go
dangra: there is nothing wrong with changing the mount path and `fly machine` are low level surgical commands the user must be aware of. 

Even `fly deploy` allows changing the mount path, it prints a warning message, that is all.

dangra (approved):
@bzanchet there is nothing fundamentally wrong in this PR so I am approving for the sake of moving forward. I see it as a bit shortsighted that we limit cloning into a new mount path only for source machines without existing mounts when it is totally possible and valid option for machines with existent mounts.

internal/command/machine/clone.go
bzanchet: > Even `fly deploy` allows changing the mount path, it prints a warning message, that is all.

fair enough, didn't know about it-- and it was simple enough to change it.

can you please take another look? (I've tested quite a few cases locally, but it's my first PR on `flyctl` and the first time something I wrote in `go` goes live hehe)

bzanchet: also, I didn't quite like changing values on `source.Config.Mounts` but as far as I could tell it's the only way w/o a big-ish refactor

dangra (approved):
🚢

Issue #1006, opened 2022-06-14
Specifying a build-target in fly.toml has no effect


jsierles:
`build-target` in the `build` `fly.toml` section should work identically to the functional `--build-target` flag.

PR #2206, opened 2023-04-28
Fix build_target option in fly.toml

wjordan:
The `build_target` build option in `fly.toml` was broken and didn't have a config test, this PR fixes the option and adds a test.

### DIFF
--- a/flyctl/app_config.go
+++ b/flyctl/app_config.go
@@ -182,7 +182,7 @@ func (ac *AppConfig) unmarshalNativeMap(data map[string]interface{}) error {
 			case "dockerfile":
 				b.Dockerfile = fmt.Sprint(v)
 				insection = true
-			case "build_target":
+			case "build_target", "build-target":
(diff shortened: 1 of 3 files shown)


kzys (approved):

flyctl/app_config.go
```diff
@@ -182,7 +182,7 @@ func (ac *AppConfig) unmarshalNativeMap(data map[string]interface{}) error {
 			case "dockerfile":
 				b.Dockerfile = fmt.Sprint(v)
 				insection = true
-			case "build_target":
+			case "build_target", "build-target":
```
dangra, line 185: is `build-target` needed? that seems to be new but wasn't tested so curious what else changed

wjordan: I'm not the authority on this but I just saw an inconsistency with a previous fix in #1006 and decided to fix this as well. I don't know how those two code paths are related though.

PR #2208, opened 2023-05-01
Perform sanity checks after Apps V2 deploy

pborzenkov:
Now that the healthchecks are disabled by default, it's very easy
to rollout a broken version on every machine in the app.

This PR introduces a "sanity check" step during the rollout. We are
gonna wait for up to 10 seconds to ensure that the newly deployed
machine is up and running and not restarting with non-zero exit code.
This won't catch all errors, but should prevent the most obvious ones
like OOMs right after deploy, missing configuration, etc.

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -258,6 +258,10 @@ func (md *machineDeployment) updateExistingMachines(ctx context.Context, updateE
 			return err
 		}
 
+		if err := md.doSanityChecks(ctx, lm, indexStr); err != nil {
+			return err
+		}
+
 		if !md.skipHealthChecks {
 			if err := lm.WaitForHealthchecksToPass(ctx, md.waitTimeout, indexStr); err != nil {
 				return err
@@ -312,6 +316,10 @@ func (md *machineDeployment) spawnMachineInGroup(ctx context.Context, groupName 
 	// Otherwise wait for the machine to start
 	indexStr := formatIndex(i, total)
 	if err := lm.WaitForState(ctx, api.MachineStateStarted, md.waitTimeout, indexStr); err != nil {
+		return "", err
+	}
+
+	if err := md.doSanityChecks(ctx, lm, indexStr); err != nil {
 		return "", err
 	}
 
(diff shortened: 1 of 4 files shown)


internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -384,3 +392,31 @@ func (md *machineDeployment) warnAboutProcessGroupChanges(ctx context.Context, d
 	}
 	fmt.Fprint(md.io.Out, "\n")
 }
+
+func (md *machineDeployment) doSanityChecks(ctx context.Context, lm machine.LeasableMachine, indexStr string) error {
+	if md.skipSanityChecks {
+		return nil
+	}
+
+	if err := lm.WaitForSanityChecksToPass(ctx, indexStr); err != nil {
```
kzys, line 401: Nit: Early-return here?

https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_return_early_rather_than_nesting_deeply

kzys (approved):

internal/command/deploy/machines_deploymachinesapp.go
pborzenkov: Done!

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -384,3 +392,31 @@ func (md *machineDeployment) warnAboutProcessGroupChanges(ctx context.Context, d
...
+	if md.skipSanityChecks {
+		return nil
+	}
+
+	if err = lm.WaitForSanityChecksToPass(ctx, indexStr); err == nil {
+		md.logClearLinesAbove(1)
+		return nil
+	}
+
+	time.Sleep(2 * time.Second) // Wait 2 secs to be sure logs have reached OpenSearch
```
dangra, line 406: does it really needs 2 extra seconds when it already waited 10 secs for sanity checks?

dangra (approved):
I like the idea but a bit worried about the extra 10s per machine for those that already have to wait N seconds for health checks. 

Either we should do sanity checks only when there are no health checks, or finally do all machine post-launch testing in parallel.

internal/command/deploy/machines_deploymachinesapp.go
pborzenkov: Good catch. It's not required (and was copy-pasted from `release_command` :)).

pborzenkov:
> I like the idea but a bit worried about the extra 10s per machine for those that already have to wait N seconds for health checks.

This happens before health checks. So health check have some additional time to start passing (if all goes well) while sanity check is in progress. While testing, I didn't notice any significant slowdown in deploy times.

> or finally do all machine post-launch testing in parallel.

That's kinda defeats the purpose, as a bad deployment will replace all machines with a non-functional version, and not only the first one.

pborzenkov:
But, yeah, doing this only when no health checks are defined maybe a way to go. I'll merge it as is and see how it goes.

dangra:
> That's kinda defeats the purpose, as a bad deployment will replace all machines with a non-functional version, and not only the first one
I mean, do sanity-checks + health-checks  waiting in parallel **per machine**. So it is not 10s + the time health checks takes to pass.

dangra:
> This happens before health checks. So health check have some additional time to start passing (if all goes well) while sanity check is in progress. While testing, I didn't notice any significant slowdown in deploy times.

This is a good point, HC will happen in the background and only happens that we are checking them 10s later. It is similar to a grace period.

PR #2212, opened 2023-05-02
Add additional detection to Deno scanner

Celeo:
This change adds support for detecting "deno.land" detection in TypeScript files for determining if a directory is written in Deno.

<https://deno.land>, besides previously being the homepage of the Deno project and now redirecting to the new <https://deno.com> site, hosts the project's standard library at <https://deno.land/std> and many thousands of third-party packages at <https://deno.land/x/>.

As of right now, I **have** tested this some local code, but **have not** ran the preflight tests.

Disclaimers: first-time contributor, new to Fly, and haven't touched Go in a long time. :)

image

image

### DIFF
--- a/scanner/deno.go
+++ b/scanner/deno.go
@@ -1,7 +1,7 @@
 package scanner
 
 func configureDeno(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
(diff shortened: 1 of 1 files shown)


Celeo:
Hi,
It wouldn't be a file of that name, but rather that text within TypeScript (or, really, JavaScript) files, like [this](https://github.com/search?q=deno.land&type=Code).

kzys:
Ah, I see. I misunderstood the way dirContains works :facepalm:

PR #2213, opened 2023-05-02
Instrument total run duration for all commands

haileys:


internal/cli/cli.go
```diff
@@ -35,11 +35,12 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
...
 	cs := io.ColorScheme()
 
-	defer func() {
-		metrics.FlushPending()
-	}()
+	defer metrics.FlushPending()
 
-	switch _, err := cmd.ExecuteContextC(ctx); {
+	cmd, err := cmd.ExecuteContextC(ctx)
+	defer metrics.RecordCommandFinish(cmd)
```
alichay, line 41: my only note is that maybe we don't want to track failed commands?

haileys: That's a good call. I've flip flopped back and forth on it, but I think for now I'll restrict this to successful commands only. We don't want timeouts and etc affecting the data.

alichay (approved):
if we end up wanting to track failures, maybe we could make that a separate metric?
no reason not to go more granular then add things up later as we need to, imo!

PR #2215, opened 2023-05-02
Added the --no-public-ips flag to fly deploy

### LINKED ISSUE #2084: Launching internal only services
I deployed an app today that I only wanted available internally, in this case via flycast. It would have been a security issue if it was public.

The app has the standard http services (port 80 http, 443 tls+http), so `fly deploy` helpfully automatically provisions public ips on the first deploy. That concerned me.

I didn't know (and didn't check, I was in a hurry!) if I provisioned a flycast ip with `fly ips allocate-v6 --private` before the first deploy if `fly deploy` would still provision public ips.

I ended up running `fly launch` and skipped the deployment. I commented out all the services config in fly.toml, and ran then ran `fly deploy`. After that I provisioned the flycast ip, uncommented the services config, then deployed. That worked, yay!

I think we could improve this experience for developers. Some ideas:

* Add a `--no-public-ips-on-first-deploy` flag to `fly deploy` (we can probably come up with a better name for the flag if we do this 😄 )
* Add something to the services config that indicates it should only be accessible via flycast, e.g., something like:

```toml
[[services]]
  flycast_only = true
```

For reference, I was using this version of flyctl:
```
% fly version
fly v0.0.521 darwin/arm64 Commit: 4523084f BuildDate: 2023-04-14T15:35:03Z
```

billyb2:
This should fix #2084

TODO:
- [x] Add a test
- [x] Make it work with nomad


billyb2:
This PR can't work with Nomad apps as is, since IP allocations are done in the backend for said apps. I just made flyctl error if you use that flag for nomad, which I think is fine since we recommended apps v2 > apps v1 anyway.

internal/command/deploy/deploy.go
```diff
@@ -81,6 +81,10 @@ var CommonFlags = flag.Set{
 		Description: "Create spare machines that increases app availability",
 		Default:     true,
 	},
+	flag.Bool{
+		Name:        "no-public-ips",
+		Description: "Do allocate any new public IP addresses",
```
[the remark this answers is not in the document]
billyb2: Yikes, fixed that

internal/command/deploy/deploy_first.go
```diff
@@ -24,7 +25,7 @@ func (md *machineDeployment) provisionFirstDeploy(ctx context.Context) error {
 
 func (md *machineDeployment) provisionIpsOnFirstDeploy(ctx context.Context) error {
 	// Provision only if the app hasn't been deployed and have services defined
-	if !md.isFirstDeploy || len(md.appConfig.AllServices()) == 0 {
+	if !md.isFirstDeploy || len(md.appConfig.AllServices()) == 0 || flag.GetBool(ctx, "no-public-ips") {
```
dangra, line 28: Do not do flags fetching here please, follow the pattern and pass them as deployment args at:
https://github.com/superfly/flyctl/blob/9ac4ba2da6c3a64cfd292f19f4d06aa43b255cd8/internal/command/deploy/deploy.go#L195-L205

billyb2: Yep my fault, I'll fix this now

internal/command/deploy/deploy_first.go
```diff
@@ -9,12 +9,14 @@ import (
...
 )
 
-func (md *machineDeployment) provisionFirstDeploy(ctx context.Context) error {
+func (md *machineDeployment) provisionFirstDeploy(ctx context.Context, allocPublicIPs bool) error {
 	if !md.isFirstDeploy || md.restartOnly {
 		return nil
 	}
-	if err := md.provisionIpsOnFirstDeploy(ctx); err != nil {
-		fmt.Fprintf(md.io.ErrOut, "Failed to provision IP addresses, use `fly ips` commands to remmediate it. ERROR: %s", err)
+	if allocPublicIPs {
```
dangra, line 16: this is fine but we already do checks inside `provisionIpsOnFirstDeploy`, better to have them all in one place as it is easy to follow for future ourselves.

billyb2: Done

PR #2217, opened 2023-05-02
Detect python app by setup.py and setup.cfg

igordertigor:
These are pretty standard indicators for python apps. In particular in cases where the authors introduce a separate "build" step.


kzys (approved):
Looks good to me. Want to hear Python folks' thoughts cc @katiayn

katiayn:
@kzys LGTM too! 👍🏽

PR #2218, opened 2023-05-02
[django] Detect python version to be used on the Dockerfile

katiayn:
Detect python version by running:
```
python3 --version
Python 3.11.2
```
or 
```
python --version
Python 3.11.2
```

For [supported versions (3.7 or newer)](https://devguide.python.org/versions/#supported-versions), use the specific version:
- 3.7-slim-buster
- 3.8-slim-buster
- 3.9-slim-buster
- 3.10-slim-buster
- 3.11-slim-buster

For unsupported version, warn the user about unsupported version and set `3.10` as the default. 
Use the detected version to build the image:
```Dockerfile
# Dockerfile
ARG PYTHON_VERSION={{ .pythonVersion }}-slim-buster
...
```


scanner/django.go
```diff
@@ -41,6 +44,69 @@ func configureDjango(sourceDir string, config *ScannerConfig) (*SourceInfo, erro
 
 	vars := make(map[string]interface{})
 
+	var pythonSupported string = "3.7" // https://devguide.python.org/versions/#supported-versions
+	var pythonVersion string = "3.10" // Keep the default version "3.10" (hardcoded on the Dockerfile)
+	var pythonVersionOutput string = "Python 3.10.0"
+	var pythonVersionFound bool = false
```
kzys, line 50: Go has a bit of type inference. So you can do `pythonSupported := "3.7"` instead.

scanner/django.go
```diff
@@ -41,6 +44,69 @@ func configureDjango(sourceDir string, config *ScannerConfig) (*SourceInfo, erro
...
+		cmd := exec.Command("python", "-V")
+		out, err := cmd.CombinedOutput()
+		if err == nil {
+		    pythonVersionOutput = string(out)
+		}
+	} else {
+	    pythonVersionOutput = string(out)
+	}
+
+    re := regexp.MustCompile(`Python ([0-9]+\.[0-9]+)\.[0-9]`)
```
kzys, line 64: How about extracting this part as a separate function? It would make testing the logic easier.

scanner/django.go
```diff
@@ -41,6 +44,69 @@ func configureDjango(sourceDir string, config *ScannerConfig) (*SourceInfo, erro
...
+    match := re.FindStringSubmatch(pythonVersionOutput)
+
+    if len(match) > 1 {
+        // necessary to split and compare the major/minor values because "3.11" is HIGHER than "3.6"
+        // e.g. "3.11" -> "3", "11"
+        version := strings.Split(match[1], ".")
+
+        major, parseMajorErr := strconv.Atoi(version[0])
```
kzys, line 72: Does Python follow SemVer? flyctl uses https://github.com/blang/semver already and maybe you could.

katiayn: @kzys Nice tip! ! I'll check the repo and see if we could use it, it'd make things much easier! 😃 Thanks!

scanner/django.go
katiayn: Sure, I'll do that! Thanks for the suggestion!

kzys (approved):
Looks good to me. The indent looks a bit off and you may be mixing tabs and spaces. But we can fix that throughout the project later.

katiayn:
@kzys I'll check that 👀 and create another PR later then! Thanks for the review and help! 🌟

PR #2220, opened 2023-05-03
Websocket metrics transport

haileys:
Metrics are currently sent to the collector via HTTP requests. That part is just a detail, the important implication is that this forces us to wait for a response to every metric we send. Because there are some metrics that are sent right before flyctl exits (reporting command runtime, success, etc), this has the effect of adding an annoying delay before exit roughly about the round trip time to the metrics collector.

This PR reworks our instrumentation to stream metrics up to the collector over a websocket and without requiring server acknowledgement. We still block on the flush to socket to make sure that the metrics are actually sent, but once they've made it past the `write(2)` boundary flyctl no longer has to care about them.

I benchmarked this change with 30 trials of the command `time flyctl m list`. On my connection, the improvement is significant:

image

The mean runtime for this command is about 290ms faster using the WS transport compared to HTTP - this checks out, my mean RTT to ORD is about 240ms. There is also less variance in the timings when using the WS transport - fewer round trips mean less opportunity for a latency spike.

### DIFF
--- a/internal/metrics/api.go
+++ b/internal/metrics/api.go
@@ -2,69 +2,100 @@ package metrics
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
-	"net/http"
-	"strings"
+	"net/url"
 	"sync"
-	"sync/atomic"
-	"time"
 
 	"github.com/superfly/flyctl/internal/buildinfo"
 	"github.com/superfly/flyctl/internal/config"
 	"github.com/superfly/flyctl/terminal"
+	"golang.org/x/net/websocket"
 )
 
-const (
-	timeout = 6 * time.Second
-)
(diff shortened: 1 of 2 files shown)


rugwirobaker:
Could we run the metrics app in more regions so people don't have to roundtrip to `ord`? This is really nice by the way

haileys:
@rugwirobaker We definitely could, that was my first idea actually! There's less of a pressing need if we merge this, but it's still a good idea

dangra (approved):
I didn't find the server side changes but this looks really good.

alichay (approved):
nice!

haileys:
> I didn't find the server side changes but this looks really good.

https://github.com/superfly/flyctl-metrics/pull/3 :)

PR #2221, opened 2023-05-03
Valid iata codes

senyosimpson:
Ensure we have a valid iata code for a region when migrating from nomad


dangra (approved):
Looks good. I'd add a comment on what case we are stripping the "2" suffix. In fact, we are probably safe by taking the first three letter in all cases.

dangra (approved):

PR #2223, opened 2023-05-03
Rename Sanity Check to Smoke Check

pborzenkov:


dangra:
what was wrong with sanity checks? :)

tvdfly (approved):

dangra:
> what was wrong with sanity checks? :)

I found the answer from @tvdfly in the internal forum, quoting here for visibility 👍 

> "Can we drop sanity? It’s an ableist term."

dangra (approved):

PR #2226, opened 2023-05-03
Deploy once and create a larger file to force sync

dangra:
I was still getting a failure when running the volume migration test, decided to give a try to creating a larger file and syncing in a second call and now it works reliable (for me! please check)


alichay (approved):
it's finally been solved! so glad to finally be done with this cursed test!!!

PR #2231, opened 2023-05-04
Run gofmt from golangci-lint

kzys:
This is borderline refactoring and I'm still under [a 3-month "no refactoring" rule for new hires](https://fly.io/docs/hiring/working/).

However, it has some values;
- flyctl is open-source and have external contributions. Those customers who are passionate enough to contribute flyctl should have prompt response regarding minor style issues. So reviewers can focus more on interesting discusssions.
- Our framework engineers are not familiar about Go and may not configure IDEs/editors and Go's style is different from other major languages (e.g. tabs over spaces). They should also get prompt response regarding minor style issues.

What do folks you think?

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,7 @@ linters:
   enable:
     # - gofumpt
     # - goimports
+    - gofmt
     - gosimple
     - govet
     - ineffassign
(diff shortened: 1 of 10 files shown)


dangra (approved):
+1 We use to have gofumpt when golangci was added but somehow is was too much of an ask. `gofmt` seems like a good restart to me.

there is another .golangci.yml under api/ folder that could use the same change.

kzys:
@dangra Didn't notice api/.golangci.yml. Let me check.

PR #2233, opened 2023-05-05
Instrument calls to GraphQL API and Flaps

haileys:


internal/instrument/call.go
```diff
@@ -0,0 +1,50 @@
+package instrument
```
haileys, line 1: Had to introduce this new package under `internal` to break what would have been a circular package dependency between `api` and `internal/metrics`. I think it's a nice result though.

internal/metrics/command.go
```diff
@@ -2,21 +2,31 @@ package metrics
...
+	Duration        float64 `json:"d"`
+	GraphQLCalls    int     `json:"gc"`
+	GraphQLDuration float64 `json:"gd"`
+	FlapsCalls      int     `json:"fc"`
+	FlapsDuration   float64 `json:"fd"`
 }
 
 func RecordCommandContext(ctx context.Context) {
+	mu.Lock()
+	defer mu.Unlock()
```
haileys, line 28: unrelated to this PR, but found we were missing mutex lock here. Unlikely to cause real problems, but strictly speaking, it is necessary.

haileys:
The lint check is currently failing with:

```
github.com/superfly/flyctl/api imports
	github.com/superfly/flyctl/internal/instrument: module github.com/superfly/flyctl@latest found (v0.0.553), but does not contain package github.com/superfly/flyctl/internal/instrument
```

I don't understand Go modules well enough to know what the problem is here. Anyone have any tips for me?

api/client.go
```diff
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	genq "github.com/Khan/genqlient/graphql"
+	"github.com/superfly/flyctl/internal/instrument"
```
kzys, line 15: In Go, packages (things you declare with `package xxx`) and modules (things on go.mod) are different. Module-wise, `github.com/superfly/flyctl/api` depends `github.com/superfly/flyctl` but `flyctl` also depends `api`.

If flyctl mostly calls RunWithContext() and a few methods, you can define a wrapper struct like https://github.com/containerd/containerd/blob/6020903f2c2f0b6b148b02ca7bdd1cd9c6efac3d/pkg/cri/instrument/instrumented_service.go#L50 to call GraphQL.Begin(). Another option may be defining an interface for instrumentation and pass its implementation, which may be something similar to https://github.com/open-telemetry/opentelemetry-go.

haileys: Oh interesting. It looks like `api` is a special case here being mentioned in `go.mod`?

kzys (approved):

PR #2234, opened 2023-05-05
Revert "Instrument calls to GraphQL API and Flaps"

jsierles:
Reverts superfly/flyctl#2233

This allows us to unblock docs deploys and probably others.


haileys:
@jsierles My bad, taking a look now.

PR #2239, opened 2023-05-05
Add `fly proxy balance` command

jeromegn:
Misfired a merge, here's the original PR: #2229

### DIFF
--- a/flyproxy/balance.go
+++ b/flyproxy/balance.go
@@ -1,0 +1,28 @@
+package flyproxy
+
+type BalanceResponse struct {
+	Destination Destination         `json:"dest"`
+	Total       uint                `json:"total"`
+	Chosen      *BalancedInstance   `json:"chosen"`
+	Rejected    []*BalancedInstance `json:"rejected"`
+}
+
+type BalancedInstance struct {
+	ID          string            `json:"id"`
+	Region      string            `json:"region"`
+	State       string            `json:"state"`
+	Concurrency int               `json:"concurrency"`
+	Healthy     bool              `json:"healthy"`
+	NodeHealthy bool              `json:"node_healthy"`
+	NodeRttMs   float64           `json:"node_rtt_ms"`
+	Rejection   *BalanceRejection `json:"rejection"`
+}
+
+type BalanceRejection struct {
+	ID   string `json:"id"`
+	Desc string `json:"description"`
+}
+
+type BalanceOptions struct {
+	Destination string
+}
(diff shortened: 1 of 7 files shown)


dangra (commented):
@jeromegn is it ready to review or should we convert it to a draft while you work on it?

jeromegn:
@dangra Ready for review

dangra (approved):
So I tried the vanilla `fly proxy PORT` and it shows a warning and forwards the port as before.  `fly proxy forward PORT` does the same an it is pretty much the same old code, looks good.

The interesting part was learning that `fly proxy balance` is like _EXPLAIN_ in SQL lingua, in the sense it shows how requests will be routed from caller's perspective, run once and exits.

I was expecting "balance" to be used as a way to forward a _service port_ to the proxy, not directly to a machine, then show in real time how requests were getting balanced across machines.

jeromegn:
> I was expecting "balance" to be used as a way to forward a _service port_ to the proxy, not directly to a machine, then show in real time how requests were getting balanced across machines.

I'm not sure I understand that feature.

Now if putting it under `fly proxy` is confusing, I think a new top-level subcommand is required. Possibly `fly router`.

This is the start of a range of commands to inspect and modify our proxy's routing behaviour. It won't actually deal with local tunelling / forwarding of any kind.

dangra:
> Now if putting it under `fly proxy` is confusing, I think a new top-level subcommand is required. Possibly `fly router`.
> This is the start of a range of commands to inspect and modify our proxy's routing behaviour. It won't actually deal with local tunelling / forwarding of any kind.

I still feel like an outsider when talking about Fly Proxy, but had read many forum posts the `fly proxy` subcommand seems the natural place for all Fly Proxy related commands.  Something `fly router` would be a new term people won't associate that easily to Fly Proxy.

The outlier is `fly proxy` used as _port forwarder_ which doesn't go through FlyProxy at all but directly to app instances (right?). Now I get why you wanted to move the port forwarding to a new root command like `fly tunnel`. I still think `fly proxy forward` is fine.

For the new command introduced in this PR, I would call it `fly proxy explain` (not "fly proxy balance") but would have to consider what other features you have in mind.
 

> > I was expecting "balance" to be used as a way to forward a _service port_ to the proxy, not directly to a machine, then show in real time how requests were getting balanced across machines.
> 
> I'm not sure I understand that feature.

Current port forwarding (aka `fly proxy PORT`) skips Fly Proxy completely, all traffic goes directly to a selected machine (or alloc). But when I read about `fly proxy balance`, I was expecting traffic to go through Fly Proxy.

The command output would be the current state of proxy routing considering regions and started/stopped instances. All of it updated every few seconds.

PR #2242, opened 2023-05-05
Release flyctl image using ko

andros21:
Hi, a proposal for different approach for `flyctl` image build

No more build from `Dockerfile` but using [`ko`](https://github.com/ko-build/ko) as a `goreleaser` step,
This is certainly an [advisable approach](https://www.chainguard.dev/unchained/move-over-dockerfiles-the-new-way-to-craft-containers)  to place a go application into a container

The size of final image is practically the same:
```console
$ podman inspect -f "{{ .Size }}" flyio/flyctl:latest | numfmt --to=si
56M
$ podman inspect -f "{{ .Size }}" ghcr.io/andros21/flyctl:latest | numfmt --to=si
58M
```
The `ko` default base image is [`cgr.dev/chainguard/static`](https://github.com/chainguard-images/images/tree/main/images/static)

The CVE reported by scanners, e.g. `trivy`, are the same too, only related to the go binary `flyctl`
No additional build time is needed for `ko build` because `flyctl` is cached.

Differences caused by default/available  `kos` config and how `ko build` operates:
* a SBOM is created and pushed in the registry, this default behavior if unwanted can be disabled `sbom: none`
* lack of `skip_push` option, so also pre-releases will be pushed ... I don't know how to fix
* a new entry-point of the image: `/ko-app/flyctl`
 
**References:**
* https://goreleaser.com/customization/ko/
* https://ko.build/configuration/
---
**Note:** I tested the new release process on [my fork](https://github.com/andros21/flyctl) but if it's preferable/safer I can:
* revert `dockers` config and `Dockerfile`
* adding `kos` config pushing to github registry `repository: ghcr.io/superfly/flyctl`
To test it aside, leaving untouched the current build/push to docker registry


dangra:
but what is the point? the Dockerfile is an industry standard and not an issue for flyctl afaik. What are we solving by moving to KO ?

andros21:
It will solve nothing practically, only an alternative approach
IMHO is slightly cleaner and more stylish than `Dockerfile` where scratch image is patched with ca-cert and `/tmp`
If I have to outline two benefits: SBOM and reproducible build

Current `Dockerfile` with multi-stage build remain a valid approach, nothing to say, I open this only to consider new and improved ways to craft containers

Free to close this

dangra:
It's hard to let down changes when time was already invested, but honestly I still don't see the value of merging this PR considering still needs work to make it to feature parity (i.e. pre-releases), final image is larger, reproducible builds are not an issue (in fact, I would say current way meets reproducible builds criteria) and it introduces a dependency (ko + chainguard images) which seems respectable yet new and unnecessary.

kzys:
While I could agree that flyctl matches what [ko suggested](https://github.com/ko-build/ko);

> It's ideal for use cases where your image contains a single Go application without any/many dependencies on the OS base image (e.g., no cgo, no OS package dependencies).

We and the industry have learned Dockerfile already and comfortable editing that. ko hasn't gained that level of adaption and still considered as [an early stage project](https://www.cncf.io/sandbox-projects/) in CNCF.

Sorry, this is certainly the chicken or the egg problem. But I don't think that it is the right time to adopt ko here.

PR #2249, opened 2023-05-08
[django] Auto set django console command

katiayn:
Auto set `console_command` on fly.toml for Django apps on `fly launch`:

```
console_command = "/code/manage.py shell"
```

Reference: https://community.fly.io/t/easily-run-a-console-in-a-new-machine-with-fly-console/12773


dangra (approved):
looks good. Only needs a pass of `gofmt` to fix the linter error. 

btw, when working in flyctl, it is a good idea to install its [pre-commit](https://pre-commit.com/) hooks.

dangra (approved):
:shipit:

PR #2252, opened 2023-05-09
Skip GetAppCompact in machines list

haileys:
The Flaps client actually doesn't need `AppCompact` in the happy-path case. Only in the case that the user has set `FLY_FLAPS_BASE_URL=peer` indicating that we should hit Flaps over Wireguard rather than `https://api.machines.dev/` do we need the app organization slug, and so we can defer calling `GetAppCompact` until then.

### DIFF
--- a/flaps/flaps.go
+++ b/flaps/flaps.go
@@ -57,18 +57,14 @@ type newClientOpts struct {
 
 func newWithOptions(ctx context.Context, opts newClientOpts) (*Client, error) {
 	app := opts.AppCompact
-	appName := opts.AppName
-	if app != nil {
-		appName = app.Name
-	}
 	// FIXME: do this once we setup config for `fly config ...` commands, and then use cfg.FlapsBaseURL below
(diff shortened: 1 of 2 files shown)


haileys:
This improves runtime for `flyctl m list -a hailey-test` from about ~750ms to 340ms

flaps/flaps.go
```diff
@@ -42,35 +42,38 @@ type Client struct {
...
+	Logger api.Logger
 }
 
-func NewWithOptions(ctx context.Context, opts *NewClientOpts) (*Client, error) {
-	app := opts.AppCompact
-	appName := opts.AppName
-	if app != nil {
-		appName = app.Name
-	}
+func newWithOptions(ctx context.Context, opts newClientOpts) (*Client, error) {
```
dangra, line 63: this function was public because of reasons https://github.com/superfly/flyctl/pull/2176

dangra (approved):
Pre approving.

`NewWithOptions` should remain public tho.

PR #2253, opened 2023-05-09
Check that app is listening on the correct address during deployment

pborzenkov:
The PR adds an additional check during Apps V2 deploy. flyctl now verifies that 
an app is listening on the correct address (0.0.0.0) and internal_port for each
service defined in fly.toml. If that's not the case, a warning is shown for
the first machine in each process group:

```
Watch your app at https://fly.io/apps/pbor-test/monitoring

Updating existing machines in 'pbor-test' with rolling strategy
  [1/2] Machine 4d89169ef66698 [app] update finished: success

WARNING The app is listening on the incorrect address and will not be reachable by fly-proxy.
Make sure your app is listening on the following addresses:
  - 0.0.0.0:8080
Found these processes inside the machine with open listening sockets:
  PROCESS        | ADDRESSES                             
-----------------*---------------------------------------
  /app           | 127.0.0.1:8080                        
  /.fly/hallpass | [fdaa:1:d615:a7b:13b:3d2c:8f58:2]:22

  [2/2] Machine 9185793eb11048 [app] update finished: success
  Finished deploying
```

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -581,3 +581,21 @@ type MachineExecResponse struct {
 	StdOut   string `json:"stdout,omitempty"`
 	StdErr   string `json:"stderr,omitempty"`
 }
+
+type MachinePsResponse []ProcessStat
+
+type ProcessStat struct {
+	Pid           int32          `json:"pid"`
+	Stime         uint64         `json:"stime"`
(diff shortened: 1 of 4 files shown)


dangra (approved):
🚀 nice one

PR #2254, opened 2023-05-09
add support for specifying a minimum number of running machines

senyosimpson:
First attempt here. Let me know what else I'd need to change to make this work

### DIFF
--- a/internal/appconfig/service.go
+++ b/internal/appconfig/service.go
@@ -9,15 +9,16 @@ import (
 )
 
 type Service struct {
-	Protocol          string                         `json:"protocol,omitempty" toml:"protocol"`
-	InternalPort      int                            `json:"internal_port,omitempty" toml:"internal_port"`
-	AutoStopMachines  *bool                          `json:"auto_stop_machines,omitempty" toml:"auto_stop_machines,omitempty"`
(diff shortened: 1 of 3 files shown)


internal/appconfig/service.go
```diff
@@ -89,12 +92,13 @@ func (c *Config) AllServices() (services []Service) {
...
-		Concurrency:  svc.Concurrency,
-		Autostop:     svc.AutoStopMachines,
-		Autostart:    svc.AutoStartMachines,
+		Protocol:           svc.Protocol,
+		InternalPort:       svc.InternalPort,
+		Ports:              svc.Ports,
+		Concurrency:        svc.Concurrency,
+		Autostop:           svc.AutoStopMachines,
+		Autostart:          svc.AutoStartMachines,
+		MinRunningMachines: svc.MinRunningMachines,
```
senyosimpson, line 101: Not sure about naming here

dangra (approved):
looks good. I couldn't come up with a better naming either, tried with [Apache Server multi processing module](https://httpd.apache.org/docs/2.4/mod/prefork.html) that suggests `MinSpareMachines` as name but it is a bit different.

senyosimpson:
This way of doing things has some [UX issues](https://flyio.discourse.team/t/keeping-the-minimum-number-of-machines-running-with-autoscaling/2818/11?u=senyo) so we're going to try another route

PR #2257, opened 2023-05-10
poc: support running a minimum number of machines

senyosimpson:
POC to support running a minimum number of machines. This absolutely does not work yet and the code is horrific (noob Gopher here) and there's some hardcoded values while I try test it locally but I'm keen to get some feedback. I think there's a bunch of plumbing I need to actually get it working.

1. A user sets a number of minimum machines to run in `fly.toml`
1. We restrict this to only work on the primary region. The minimum number of machines will all be run in the primary region.
    1. If the number of minimum machines is larger than the number of machines in the primary region, we issue a warning and all machines in the primary region are set to keep running
1. We set `autostop = false` on the number of machines specified by the user for the given process group.

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -441,6 +441,7 @@ type MachineService struct {
 	InternalPort             int                        `json:"internal_port,omitempty" toml:"internal_port,omitempty"`
 	Autostop                 *bool                      `json:"autostop,omitempty"`
 	Autostart                *bool                      `json:"autostart,omitempty"`
+	MinRunningMachines       *int                       `json:"min_running_machines,omitempty"`
 	Ports                    []MachinePort              `json:"ports,omitempty" toml:"ports,omitempty"`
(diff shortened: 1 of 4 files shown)


dangra:
> We set autostop = false on the number of machines specified by the user for the given process group.

Just wanted to express my disenchantment with this solution, it is so half-baked, the approach shows it deficiencies when the no-auto-stoppable machine crashes (or its host goes down).

And not just that, it will require a good amount of tuning to `fly scale` command too to keep min running machines on the expected count.

I'll help with the PR review anyways.

senyosimpson:
> Just wanted to express my disenchantment with this solution, it is so half-baked, the approach shows it deficiencies when the no-auto-stoppable machine crashes (or its host goes down).

Good points. The one way we can solve this in the proxy is if we can pass `primary_region` in the `fly_proxy` settings alongside autostop/autostart. If that's straightforward enough then we can get rid of this option

dangra:
Superseded by https://github.com/superfly/flyctl/pull/2268

Issue #2245, opened 2023-05-08
Dockerfile should be allowed to be .dockerignore'd


gabrielalmeida:
Fly requires Dockerfile to be part of its context when running `fly deploy` and will fail if `Dockerfile` is part of `.dockerignore`:

> Error: failed to fetch an image or build from source: error building: failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /data/docker/tmp/buildkit-mount150277830/apps/dashboard/Dockerfile: no such file or directory

This alone is unexpected and differs from how Docker works.

[.dockerignore reference](https://docs.docker.com/engine/reference/builder/#dockerignore-file): 
> You can even use the .dockerignore file to exclude the Dockerfile and .dockerignore files. **These files are still sent to the daemon because it needs them to do its job**. But the ADD and COPY instructions do not copy them to the image.

Thus, Dockerfile will always be part of the "internal docker build" context and causes a `COPY . .` cache to be invalidated leading to expensive unnecessary rebuilds.

For instance, due to this limitation, a single change on Dockerfile in an unrelated layer may cause all dependencies to be installed again and a new build to run, going from a fully cached build to an expensive 10min+.

As put by @jacobgreenleaf on a [comment](https://github.com/superfly/flyctl/pull/1213/files#diff-7dd2551e6b56af8196ece355e6635fd2d69ded0dc977c10e4276f3a43b6fb9c2R76-R79) on PR https://github.com/superfly/flyctl/pull/1213:

> // When a user writes a dockerignore file, they might include a rule for "Dockerfile", or ".dockerignore".
// Those files must still be sent to the Docker daemon via this archive, however the user's intent of having
// the file excluded from the image that gets built is still preserved because their dockerignore file
// is transmitted in the archive as-is before these exclusions are added.

The PR was closed without comments so I've created this issue.

matttpt:
#2260 should fix this in the next `flyctl` release. Feel free to re-open this if not.

Thanks @gabrielalmeida for the detailed issue and @jacobgreenleaf for the first look at this in #1213 (#2260 turned out to be pretty similar)!

PR #2260, opened 2023-05-10
Add `Dockerfile` to build context regardless of path

matttpt:
The `Dockerfile` must always be in the build context sent to the Docker daemon, even if it's covered by `.dockerignore`. See issue #2245. (For reference, `docker build` ensures that the `Dockerfile` is sent to the daemon [here](https://github.com/docker/cli/blob/bc3f905a2aa9e7ea84591851860b820bb31caa07/cli/command/image/build/dockerignore.go#L28), and the daemon itself removes the `Dockerfile` if it's in `.dockerignore` [here](https://github.com/moby/moby/blob/f70d9933d160b12b8d6247c77e721c94fae9eefe/builder/remotecontext/detect.go#L117), I think.)

Right now we make sure that `/Dockerfile` and `/dockerfile` are added to the build context even if they're excluded by the `.dockerignore`. However, if the `Dockerfile` is located elsewhere, we should add an override for *that* path instead. Otherwise, it isn't sent to the daemon and it results in an error like the one shown in #2245.

Admittedly, this might cause strange behavior if the `Dockerfile` path has any characters with special meaning for `.dockerignore` matching rules (e.g. `*` and `?`). However, `docker build` itself [doesn't](https://github.com/docker/cli/blob/bc3f905a2aa9e7ea84591851860b820bb31caa07/cli/command/image/build/dockerignore.go#L39) seem to worry about this, so I think it'll be fine.

PR #1213 (now closed) was in part a WIP fix for this, and I arrived at more or less the same thing. So thanks to jacobgreenleaf for taking the first look at this!

### DIFF
--- a/internal/build/imgsrc/archive.go
+++ b/internal/build/imgsrc/archive.go
@@ -33,11 +33,7 @@ func CreateArchive(dockerfile, workingDir, ignoreFile string, compressed bool) (
 		compressed: compressed,
 	}
 
-	excludes, err := readDockerignore(workingDir, ignoreFile)
-	if err != nil {
-		return nil, errors.Wrap(err, "error reading .dockerignore")
-	}
-	archiveOpts.exclusions = excludes
+	relativeDockerfilePath := ""
 
 	// copy dockerfile into the archive if it's outside the context dir
 	if !isPathInRoot(dockerfile, workingDir) {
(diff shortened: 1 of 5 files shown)


dangra (approved):
Such a gem 💎

PR #2262, opened 2023-05-10
Disallow `experimental` features in V2 configs

alichay:
People are getting burned by failed migrations because the migrator doesn't warn them that `experimental` config sections are unsupported for Apps V2.

There's room for improvement and making this more user friendly (such as linking to docs and suggesting solutions for specific fields) but this is a good first step.

### DIFF
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -102,12 +102,13 @@ type Build struct {
 }
 
 type Experimental struct {
-	Cmd          []string `toml:"cmd,omitempty" json:"cmd,omitempty"`
-	Entrypoint   []string `toml:"entrypoint,omitempty" json:"entrypoint,omitempty"`
-	Exec         []string `toml:"exec,omitempty" json:"exec,omitempty"`
-	AutoRollback bool     `toml:"auto_rollback,omitempty" json:"auto_rollback,omitempty"`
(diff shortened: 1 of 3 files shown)


matttpt (commented):
Out of curiosity I tried running the migration preflight tests, and for `TestAppsV2MigrateToV2_Volumes` I got the experimental features error. I tried adding a line to print the experimental section:

```
appconfig.Experimental{Cmd:[]string{}, Entrypoint:[]string{}, Exec:[]string{}, AutoRollback:false, EnableConsul:false, EnableEtcd:false}
```

I'm not sure why `auto_rollback` is unset. I'm wondering if it has to do with this test creating the app manually (`fly apps create`) and not through `fly launch`. Maybe it won't cause problems in practice?

Everything else looks good, just wanted to run this by you before it's merged!

matttpt (approved):
LGTM!

dangra (commented):
This PR will make the situation better but it is not complete due to:
1. Config.Experimental type doesn't have fields for all possible experimental features
2. Some experimental features can be easily supported like Cmd, Exec and Entrypoint
3. EnableConsul and EnableEtcd can be ignored because the app secrets are already set and the machines will get them once migrated (needs confirmation).
4. There are some missing experimental features like "allowed_public_ports" that are real blockers

PR #2263, opened 2023-05-11
Exclude fly auth token from metrics

haileys:
This command is so fast that the metrics WebSocket hasn't even had a chance to connect by the time it's ready to exit, leading to an annoying delay after printing the auth token on every invocation.

It does no RPC and no other network I/O, it just spits out a value from config on disk, so metrics are pretty low value for this subcommand anyway. Let's just skip metrics so the command can be instant.


alichay (approved):
graceful solution :)

PR #2264, opened 2023-05-11
Improve migrate-to-v2 error msg when worker cannot fork volume

tvdfly:
This error is confusing:

> failed while migrating: Disk id 432 is not a valid candidate

It means the worker hosting the volume does not have capacity for the second volume that is created during the migration.

This change changes the error message to:

> failed while migrating: unfortunately the worker hosting your volume vol_8675309bob (the_vol_name) does not have capacity for another volume to support the migration; some other options: 1) try again later and there might be more space on the worker, 2) run a manual migration https://community.fly.io/t/manual-migration-to-apps-v2/11870, or 3) wait until we support volume migrations across workers (we're working on it!)

Which will hopefully be less confusing :-)


tvdfly:
Please merge after reviewing (unless there is feedback to address, of course). Thanks!

PR #2265, opened 2023-05-11
Better key value handling

guillaumervls:
Passing commas in values of `KEY=value`-like flags, like env vars or build args, would break. In `--env FOO=bar,baz`, `env` was parsed into `FOO=bar` and `baz`.

This was because we used [`StringSlice`](https://pkg.go.dev/github.com/spf13/pflag#StringSlice) instead of [`StringArray`](https://pkg.go.dev/github.com/spf13/pflag#StringArray).

[`StringSlice`](https://pkg.go.dev/github.com/spf13/pflag#StringSlice) is useful by providing a shorthand for flags where commas shouldn't appear in the values (like IDs). But for arbitrary values like env vars or build args, [`StringArray`](https://pkg.go.dev/github.com/spf13/pflag#StringArray) is a safer choice. Docker CLI `-e` and `--build-arg` work like this and ignore commas.

--- 

Also added tests to check that we can close https://github.com/superfly/flyctl/issues/1710.

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -209,7 +209,7 @@ func deployToMachines(ctx context.Context, appConfig *appconfig.Config, appCompa
 		AppCompact:            appCompact,
 		DeploymentImage:       img.Tag,
 		Strategy:              flag.GetString(ctx, "strategy"),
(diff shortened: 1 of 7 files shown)


jsierles (approved):
🔥

PR #2270, opened 2023-05-11
Add support for experimental cmd, entrypoint and exec features

dangra:


alichay:
I could be mistaken - wouldn't this overwrite every process group?

dangra:
@alichay Yes, but I guess that is what happens for Nomad apps too.

alichay:
huh. that's really weird behavior. oh well, legacy is legacy. let's get this working for people

alichay (approved):

alichay (approved):

PR #2278, opened 2023-05-12
Block migration only for *the app* that relies on allowed_public_ports

dangra:


internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -596,6 +601,17 @@ func (m *v2PlatformMigrator) validate(ctx context.Context) error {
 	return nil
 }
 
+func (m *v2PlatformMigrator) validateKnownUnmigratableApps(ctx context.Context) error {
+	// This is as last meassure for very special cases
+	knownUnmigratableApps := []string{
+		"globalconfig-dht", // https://flyio.discourse.team/t/x/2860/6
+	}
+	if slices.Contains(knownUnmigratableApps, m.appCompact.ID) {
+		return fmt.Errorf("Your app is known to break under V2 platform. Please contact support")
```
alichay, line 610: imo the wording seems almost kind of harsh, their app is already being singled out.
mind if I try rewording it?

```suggestion
		return fmt.Errorf("Your app uses features incompatible with the V2 platform. Please contact support to discuss how to successfully migrate")
```

alichay (approved):
one note, just a suggestion. great work tracking all this down and fixing it at the source :)

PR #2279, opened 2023-05-12
feat: add support for canary deployment strategy for Apps V2

jipperinbham:


internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -98,77 +98,98 @@ func (md *machineDeployment) deployMachinesApp(ctx context.Context) error {
 	processGroupMachineDiff := md.resolveProcessGroupChanges()
 	md.warnAboutProcessGroupChanges(ctx, processGroupMachineDiff)
 
-	if len(processGroupMachineDiff.machinesToRemove) > 0 {
-		// Destroy machines that don't fit the current process groups
-		if err := md.machineSet.RemoveMachines(ctx, processGroupMachineDiff.machinesToRemove); err != nil {
-			return err
+	// TODO: validate _somewhere_ no mounts/volumes are used.
```
alichay, line 101: aside from this, looks good!
nice drive-by cleanup as well, some of the deploy code definitely needed a little love

jipperinbham: @dangra is there a central place where do validate the fly.toml? My thought is if we can rely on the config being valid, I don't need to care about it at this point.

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -286,10 +307,19 @@ func (md *machineDeployment) updateExistingMachines(ctx context.Context, updateE
 	return nil
 }
 
-func (md *machineDeployment) spawnMachineInGroup(ctx context.Context, groupName string, i, total int, standbyFor []string) (string, error) {
+type metadata struct {
+	key   string
+	value string
+}
+
+func (md *machineDeployment) spawnMachineInGroup(ctx context.Context, groupName string, i, total int, standbyFor []string, meta ...metadata) (machine.LeasableMachine, error) {
```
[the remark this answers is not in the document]
jipperinbham: Yea, the leasable machine stuff is a bit hard to follow because there's also a goroutine running in the background to constantly renew them. The main driver for the change to this function signature is I need to destroy the canary machine and that function expects a LeasableMachine.

internal/command/deploy/machines_deploymachinesapp.go
jipperinbham: > I thought we were skipping canary and falling back to Rolling in that case?

Is that what we want to do? Or rather, is that what we do now?

Thanks for the specific code points! I'll work on getting the validation in place.

dangra: > Is that what we want to do? Or rather, is that what we do now?

You are right, failing is what we do now ([source](https://github.com/superfly/web/blob/a1f46ab04f52962e0904e6860167003f7bd86470/app/models/nomad/mutations/update_strategy.rb#L12-L17))

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -98,77 +98,98 @@ func (md *machineDeployment) deployMachinesApp(ctx context.Context) error {
 	processGroupMachineDiff := md.resolveProcessGroupChanges()
 	md.warnAboutProcessGroupChanges(ctx, processGroupMachineDiff)
 
-	if len(processGroupMachineDiff.machinesToRemove) > 0 {
-		// Destroy machines that don't fit the current process groups
-		if err := md.machineSet.RemoveMachines(ctx, processGroupMachineDiff.machinesToRemove); err != nil {
-			return err
+	// TODO: validate _somewhere_ no mounts/volumes are used.
+	if md.strategy == "canary" {
```
dangra, line 102: Our previous chat left me thinking that canaries are a waste of time on initial deploy (see `md.isFirstDeploy` bool)

jipperinbham: Ah, nice! Will incorporate this into the deploy.

internal/command/deploy/machines_deploymachinesapp.go
jipperinbham: Went with https://github.com/superfly/flyctl/pull/2279/commits/e49507c1650c9a3e6d21299c0415a0796bff9c10 but not sure if there's any side effects with setting volumes earlier.

internal/command/deploy/machines_deploymachinesapp.go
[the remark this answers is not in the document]
jipperinbham: I'm having a hard time following the link to the line you're concerned about.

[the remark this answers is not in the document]
dangra: This line https://github.com/superfly/flyctl/blob/39195cfab0af1343c1b4724f8ace531656805748/internal/command/deploy/machines_deploymachinesapp.go#L339 in current version of this PR. 

It releases the lease just before returning the leasable machine.

internal/command/deploy/machines_deploymachinesapp.go
dangra: I think we have to revert the setVolumes() change and look for len(appconfig.Mounts) instead.

dangra: btw, the appconfig validation is the right thing to do IMO. Checking it here is a bit late.

jipperinbham: Yea... I initially started adding code to `cfg.validateDeploySection` but `Validate` doesn't take the `MachineDeploymentArgs` deployment as a parameter and it's invoked from several different places which wouldn't have a variable to send there. So my compromise was to do the validation later after creating the `machineDeployment` object which seems to have its own set of validations in the various `setXXX` calls.

internal/command/deploy/machines_deploymachinesapp.go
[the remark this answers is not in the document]
jipperinbham: What would you suggest? Everything seems to work off of a LeasableMachine object.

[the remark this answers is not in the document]
dangra: I mean, the machine is not leased by the time it is returned, that was what I wanted to point out. I guess we can ignore it because the only use case for the returned unleased machine is to be destroyed so it is good anyways.

internal/command/deploy/machines_deploymachinesapp.go
dangra: I see. I can redo it later.  

To me, "[determineAppConfigForMachines](https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines.go#L83)" is where md.appConfig.Deploy.Strategy should be set if passed as argument. Pretty much the same we do for other arguments like primaryRegion and envvars. And below that, within same function NewMachineDeployment, it runs [validations](https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines.go#L87).

jipperinbham (commented):
I can't approve but LGTM!

dangra (approved):
let's cheat then!

PR #2280, opened 2023-05-12
`fly sftp get /path/to/base.txt` should create base.txt locally

kzys:
It was creating /path/to/base.txt locally and it is counter-intuitive.

In fact, sftp shell's get command used the remote file's basename even before this change.

### DIFF
--- a/internal/command/ssh/sftp.go
+++ b/internal/command/ssh/sftp.go
@@ -195,7 +195,7 @@ func runGet(ctx context.Context) error {
 	}
 
 	fmt.Printf("%d bytes written to %s\n", bytes, local)
-	return nil
+	return f.Sync()
 }
 
 var completer = readline.NewPrefixCompleter(
@@ -364,6 +364,11 @@ func (sc *sftpContext) getDir(rpath string, args []string) {
 	}
 
 	z.Close()
+
+	err = f.Sync()
+	if err != nil {
+		sc.out("failed to sync %s: %s", lpath, err)
+	}
 }
 
 func (sc *sftpContext) chmod(args ...string) error {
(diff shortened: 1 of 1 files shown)


internal/command/ssh/sftp.go
```diff
@@ -509,6 +516,10 @@ func (sc *sftpContext) get(args ...string) error {
 		} else {
 			sc.out("wrote %d bytes", bytes)
 		}
+		err = f.Sync()
+		if err != nil {
+			sc.out("failed to sync %s: %s", localFile, err)
```
catflydotio, line 521: Can we give the user more information about what happened and what to do about it?

catflydotio:
I would guess not in this case, but is this change going to break anything for existing users?

internal/command/ssh/sftp.go
kzys: The `err` part should have what happened. E.g. "no space left on device" if this downloads 2GB file and the local host doesn't have the space.

"What to do about" is difficult. This error is not coming from Fly's platform. So we don't have much concrete suggestions.

kzys:
> I would guess not in this case, but is this change going to break anything for existing users?

Yes and [some might expect](https://www.hyrumslaw.com/). That said, at least three people are confused about (see https://community.fly.io/t/access-sqlite-file/9429) and `fly ssh sftp shell`'s `get` command is doing this "take its basename" approach. So this behavior mismatch is too wild to keep

internal/command/ssh/sftp.go
catflydotio: All good. 

I'd say these error messages are very normal. We tell the user "wrote %d bytes" which is technically true even if it can seem weird if the sync to disk then goes wrong. I was just wondering if we could very easily do better than "normal," while you've got a PR going already. I wonder this about most of our error messages. ;)

catflydotio:
I agree with your previous answer for various reasons; this was a belt-and-braces check.

PR #2282, opened 2023-05-15
Enable autostart for pg services when scale to zero

DAlperin:


flypg/launcher.go
```diff
@@ -155,6 +162,7 @@ func (l *Launcher) LaunchMachinesPostgres(ctx context.Context, config *CreateClu
 						},
 					},
 					Concurrency: concurrency,
+					Autostart:   &autoStart,
```
pborzenkov, line 165: AFAIK, you also need to remove `machineConf.DisableMachineAutostart`, otherwise `flaps` always overwrites service meta tags and will clear `autostart=true`.

dangra (approved):
@DAlperin just to be clear, AutoStart is enabled by default when not specified here. The only change here is that it will be disabled for non-scale-to-zero clusters.  I guess that is the intention.

flypg/launcher.go
DAlperin: `DisableMachineAutostart` is marked deprecated in favor of the service.Autostart field. I think this was changed last week

dangra: Good point. Do not use `DisableMachineAutostart`, it is deprecated.

pborzenkov: Yeah, but it's being set in this file below (line 212) and AFAIK `flaps` uses it if it's set ignoring the rest of the `autostart`/`autostop` parameters.

PR #2283, opened 2023-05-15
`metrics`: add opt-out, `fly metrics` command

alichay:
```php
flyctl on  feat/metrics/opt-out [$] via 🐹 v1.20.2 
❯ ./bin/flyctl settings analytics
Anonymized analytics: enabled

This can be controlled with fly settings analytics <enable/disable>

flyctl on  feat/metrics/opt-out [$] via 🐹 v1.20.2 
❯ ./bin/flyctl settings analytics disable
Anonymized analytics: disabled

flyctl on  feat/metrics/opt-out [$] via 🐹 v1.20.2 
❯ ./bin/flyctl settings analytics
Anonymized analytics: disabled

This can be controlled with fly settings analytics <enable/disable>

flyctl on  feat/metrics/opt-out [$] via 🐹 v1.20.2 
❯ ./bin/flyctl settings analytics enable
Anonymized analytics: enabled

flyctl on  feat/metrics/opt-out [$] via 🐹 v1.20.2 
❯ ./bin/flyctl settings analytics
Anonymized analytics: enabled

This can be controlled with fly settings analytics <enable/disable>
```

### DIFF
--- a/internal/command/help/help.go
+++ b/internal/command/help/help.go
@@ -114,7 +114,7 @@ Monitoring and managing things:
 		fmt.Printf(`
 Access control:
 `)
-		listCommands([]string{"orgs", "auth", "move"})
+		listCommands([]string{"orgs", "auth", "move", "metrics"})
 
 		fmt.Printf(`
 More help:
--- a/internal/command/help/help.go
+++ b/internal/command/help/help.go
@@ -114,7 +114,7 @@ Monitoring and managing things:
 		fmt.Printf(`
 Access control:
 `)
-		listCommands([]string{"orgs", "auth", "move", "metrics"})
+		listCommands([]string{"orgs", "auth", "move"})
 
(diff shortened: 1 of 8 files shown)


dangra (approved):
Somehow I think a top level command for an internal CLI setting is too much. I would like to group all flyctl related knobs into a single subcommand

The usual for tools like this is `config` but it is already taken, so other options are `fly options`, `fly settings`, `fly local-config` ... 

not a blocker tho.

dangra:
btw, if we go for the top level command, brew.sh uses `brew analytics enable/disable`  which could be better name for an internal option. "metrics" already has some meaning on fly apps world.

alichay:
great notes. I think I'm going to rework this with that in mind. ty!

haileys:
love this!

PR #2287, opened 2023-05-16
Disable provisioning of new nomad apps

alichay:
note: turboku is untested. it used to force nomad, and now it doesn't.

1. removes the `--machines` flag on `fly apps create`, hides the `--nomad` flag for `apps create` (and if it's specified, throws an error).
2. hides the `--force-nomad` and `--force-machines` flags for deploy/launch. they are now unused, but I didn't want to potentially break pipelines that use these flags. (and --force-nomad being specified can throw more helpful errors than just unknown flag)
3. deploy pulls which platform to use solely from the existing app (assumes machines if the platform version != "nomad")
4. fly launch always uses `machines` and throws an error if `--force-nomad` is specified.

### DIFF
--- a/internal/command/apps/create.go
+++ b/internal/command/apps/create.go
@@ -10,6 +10,7 @@ import (
 	"github.com/superfly/flyctl/iostreams"
 
 	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/internal/buildinfo"
 	"github.com/superfly/flyctl/internal/command"
 	"github.com/superfly/flyctl/internal/config"
 	"github.com/superfly/flyctl/internal/flag"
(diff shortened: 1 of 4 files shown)


internal/command/apps/create.go
```diff
@@ -47,14 +47,11 @@ may be fetched with 'fly config save -a <app_name>'`
 			Name:        "network",
 			Description: "Specify custom network id",
 		},
-		flag.Bool{
-			Name:        "machines",
-			Description: "Use the machines platform",
-		},
```
dangra, line 53: Can we make it a no-op? I am afraid it could break some pipelines

alichay: What do you mean?

alichay: Yeah, good call!

internal/command/apps/create.go
```diff
@@ -74,6 +71,10 @@ func RunCreate(ctx context.Context) (err error) {
 		apiClient     = client.FromContext(ctx).API()
 	)
 
+	if flag.GetBool(ctx, "nomad") {
+		return fmt.Errorf("nomad platform is no longer supported")
```
kzys, line 80: "... is deprecated" or "creation of an app on Nomad platform is ...". We still support the platform itself.

alichay:
should have addressed everything. squashed the commits though because it was getting a bit unruly for such a small change haha

kzys (approved):

PR #2294, opened 2023-05-17
Add helpers to encourage migration

ndarilek:
I extracted the messaging from `fly deploy` to a separate function so we can a) call it from more places and b) change how assertive/aggressive it gets over time (E.g. maybe we eventually offer to run the migration automatically on interactive terminals prior to forcing it.)

I also called this function from `fly status`. I don't have any Nomad apps to test with but I inverted the check to make sure it worked, and it seemed to. I think I uninverted the invert but definitely check since I'm unfortunately out of coffee today so who knows what my brain is doing to get back at me for that.

### DIFF
--- a/internal/command/apps/restart.go
+++ b/internal/command/apps/restart.go
@@ -72,8 +72,13 @@ func runRestart(ctx context.Context) error {
 }
 
 func runNomadRestart(ctx context.Context, app *api.AppCompact) error {
-	command.PromptToMigrate(ctx)
 	client := client.FromContext(ctx).API()
+
+	appName := flag.FirstArg(ctx)
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		command.PromptToMigrate(ctx, app)
+	}
 
 	if _, err := client.RestartApp(ctx, app.Name); err != nil {
 		return fmt.Errorf("failed restarting app: %w", err)
--- a/internal/command/apps/restart.go
+++ b/internal/command/apps/restart.go
@@ -3,6 +3,7 @@ package apps
 import (
 	"context"
 	"fmt"
+
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/flaps"
 
@@ -71,6 +72,7 @@ func runRestart(ctx context.Context) error {
 }
 
 func runNomadRestart(ctx context.Context, app *api.AppCompact) error {
+	command.PromptToMigrate(ctx)
 	client := client.FromContext(ctx).API()
 
 	if _, err := client.RestartApp(ctx, app.Name); err != nil {
--- a/internal/command/apps/restart.go
+++ b/internal/command/apps/restart.go
@@ -74,11 +74,7 @@ func runRestart(ctx context.Context) error {
 func runNomadRestart(ctx context.Context, app *api.AppCompact) error {
 	client := client.FromContext(ctx).API()
 
-	appName := flag.FirstArg(ctx)
-	app, err := client.GetAppCompact(ctx, appName)
-	if err != nil {
-		command.PromptToMigrate(ctx, app)
-	}
+	command.PromptToMigrate(ctx, app)
 
 	if _, err := client.RestartApp(ctx, app.Name); err != nil {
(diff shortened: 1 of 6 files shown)


alichay (approved):
nice!

alichay (approved):
nice!

ndarilek:
I've been told that `AppCompact` is the only reliable source of which platform an app is on.

Is it correct that this can only be retrieved from an API call? Or does this get put in the context at some point? I'm still not entirely clear what ends up in the context, what has to be retrieved from the API, or whether retrieving something from the context performs background magic like API calls or filesystem reads.

Thanks.

alichay:
> I've been told that `AppCompact` is the only reliable source of which platform an app is on.
> 

yeah I just saw that! that's totally right. the places where you call the new function should have this info _anyway_, so if you just make the helper take an appcompact, that is probably the path of least resistance?

ndarilek:
Awesome, thanks for the feedback. I think this should do it, and also answers my questions about whether or not this makes sense to add to non-app commands.

internal/command/apps/restart.go
```diff
@@ -72,9 +72,14 @@ func runRestart(ctx context.Context) error {
...
 
 func runNomadRestart(ctx context.Context, app *api.AppCompact) error {
-	command.PromptToMigrate(ctx)
 	client := client.FromContext(ctx).API()
 
+	appName := flag.FirstArg(ctx)
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		command.PromptToMigrate(ctx, app)
+	}
```
alichay, line 81: Getting the app is unnecessary here, we already have one passed as an argument to the function :)
```suggestion
	command.PromptToMigrate(ctx, app)
```

ndarilek:
Ugh, and once again with the tire fire that is GitHub Actions accessibility.

I see there are lint failures, but I can't find where they are with my screen reader. Further, when I use my usual trick of checking the raw logs, I get this:

```
2023-05-18T17:57:09.1055385Z Running [/home/runner/golangci-lint-1.52.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
2023-05-18T17:57:56.5540969Z ##[error]SA4009: argument app is overwritten before first use (staticcheck)
2023-05-18T17:57:56.5551838Z ##[error]SA4009(related information): assignment to app (staticcheck)
2023-05-18T17:57:56.5552434Z 
2023-05-18T17:57:56.5552977Z ##[error]issues found
2023-05-18T17:57:56.5553790Z Ran golangci-lint in 47447ms
```

Am I just missing the file/line number of the error? I imagine it's somewhere in that giant log but I don't have the rest of eternity to review it line-by-line. :)

Thanks.

alichay:
> Ugh, and once again with the tire fire that is GitHub Actions accessibility.
> 
> I see there are lint failures, but I can't find where they are with my screen reader. Further, when I use my usual trick of checking the raw logs, I get this:
> 
> ```
> 2023-05-18T17:57:09.1055385Z Running [/home/runner/golangci-lint-1.52.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
> 2023-05-18T17:57:56.5540969Z ##[error]SA4009: argument app is overwritten before first use (staticcheck)
> 2023-05-18T17:57:56.5551838Z ##[error]SA4009(related information): assignment to app (staticcheck)
> 2023-05-18T17:57:56.5552434Z 
> 2023-05-18T17:57:56.5552977Z ##[error]issues found
> 2023-05-18T17:57:56.5553790Z Ran golangci-lint in 47447ms
> ```
> 
> Am I just missing the file/line number of the error? I imagine it's somewhere in that giant log but I don't have the rest of eternity to review it line-by-line. :)
> 
> Thanks.

The issue it's complaining about is the thing I mentioned in the comment above. it just doesn't like that app is re-assigned

ndarilek:
Ah, I didn't know where that particular action vomited its report. Thanks, I'll check there in the future in case that is any more accessible.

ndarilek:
Also, is there any way to run these lints locally? I wish VS Code's problems view surfaced them somehow, since I generally check that before pushing code.

alichay (approved):

PR #2296, opened 2023-05-17
nit: r/costs/cost

DazWilkin:
"Looks like you're accessing a paid feature. Dedicated IPv4 addresses now **costs** $2/mo. Are you ok with this?"


dangra:
thanks!

PR #2297, opened 2023-05-17
Allow settings specific memory size and number of cpus in deploys

billyb2:


internal/command/deploy/deploy.go
```diff
@@ -206,6 +218,8 @@ func deployToMachines(ctx context.Context, appConfig *appconfig.Config, appCompa
 		WaitTimeout:           time.Duration(flag.GetInt(ctx, "wait-timeout")) * time.Second,
 		LeaseTimeout:          time.Duration(flag.GetInt(ctx, "lease-timeout")) * time.Second,
 		VMSize:                flag.GetString(ctx, "vm-size"),
+		VMCPUs:                flag.GetInt(ctx, "vm-cpus"),
+		VMMemory:              flag.GetInt(ctx, "vm-memory"),
```
dangra, line 222: missing CPUKind here?

internal/command/deploy/machines.go
```diff
@@ -383,12 +386,31 @@ func (md *machineDeployment) latestImage(ctx context.Context) (string, error) {
...
+
+	if vmCPUKind == "" {
+		md.machineGuest.CPUKind = "shared"
+	} else {
+		if md.machineGuest.CPUKind == "shared" || md.machineGuest.CPUKind == "performance" {
+			md.machineGuest.CPUKind = vmCPUKind
+		} else {
+			return fmt.Errorf("invalid machine CPU kind requested, '%s', expected to start with 'shared' or 'performance'", vmCPUKind)
+		}
+	}
```
dangra, line 405: this whole block is awkward 🤔 

I was expecting it to:
1. Override `md.machineGuest.CPUKind` only if vmCPUKind is not an empty string
2. Default to `shared` only `md.machineGuest.CPUKind` is not set
3. (Optionally) fail if vmCPUKind is not one of `[]string{"shared", "performance"}` 

as it currently reads, it does a few unexpected things:
1. Sets `md.machineGuest.CPUKind` to "shared" if --vm-cpu-kind is not passed but overrides what `--vm-size` could have set
2. Checks `md.machineGuest.CPUKind` is one of "shared" or "performance", but doesn't check the value of `--vm-cpu-kind` is valid.

billyb2: Another good catch! This should've been caught by my tests, but those also had a bug :/.

internal/command/deploy/machines.go
```diff
@@ -383,12 +386,31 @@ func (md *machineDeployment) latestImage(ctx context.Context) (string, error) {
 	return resp.App.CurrentReleaseUnprocessed.ImageRef, nil
 }
 
-func (md *machineDeployment) setMachineGuest(vmSize string) error {
-	if vmSize == "" {
-		return nil
-	}
+func (md *machineDeployment) setMachineGuest(vmSize string, vmCPUKind string, vmCPUs int, vmMem int) error {
 	md.machineGuest = &api.MachineGuest{}
```
dangra, line 390: I think we have to start from a known baseline default like `shared-cpu-1x`  (that is what `nil` does)

Add just after this line the following code:
```go
md.machineGuest.SetSize("shared-cpu-1x")
```
 
and then only try to override size, cpus, kind and memory if the flags are not the empty values.

billyb2:
Closing, this is superseded by the excellent PR at #2300

PR #2300, opened 2023-05-17
`deploy`: allow setting cpu and memory

alichay:
Adds `--vm-memory` and `--vm-cpus` to `fly deploy`.

This introduces quite a bit of complexity, so let's talk about the side effects:

- `--memory` and `--cpus` flags have been canonicalized to `--vm-[memory/cpus]` throughout flyctl. Aliases exist for the old parameters, so that existing scripts and docs still work.
- `flag` got support for grabbing from multiple named parameters. This allows aliasing parameter names for graceful deprecation.

The way deploy determines an appropriate guest for a machine got a little bit more robust:
1. if the machine already exists (deploying over an existing app), we reuse the existing guest.
2. if the user passes `--vm-size` to deploy, that takes precedence.
3. if there's still no guest, but there *are* existing machines (so, new machine in existing process group),
        we take the median machine size and use that. (find the median CPU size for the process group, out of all machines that size in the group, take the median memory, and use that guest)
4. if `--vm-cpus` or `--vm-memory` is specified, and there is still no guest (so, not an existing machine, no `--vm-size`, and the process group doesn't exist, we use `shared-cpu-1x`.
        * If we didn't specify either of those parameters, and there's still no guest, we provide a nil guest.
5. finally, with the resulting guest, apply `--vm-cpus` or `--vm-memory`.

Currently, most preflight tests pass, and the ones that don't pass also don't pass on master, so maybe that's not an issue?

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -311,12 +311,9 @@ type metadata struct {
 	value string
 }
 
-func (md *machineDeployment) guestForMachine(mach *api.Machine, group string) *api.MachineGuest {
+func (md *machineDeployment) guestForGroup(group string) *api.MachineGuest {
 
 	var guest *api.MachineGuest
-	if mach != nil && mach.Config != nil && mach.Config.Guest != nil {
-		guest = helpers.Clone(mach.Config.Guest)
-	}
 	if md.machineGuest.RealizedGuest != nil {
 		guest = helpers.Clone(md.machineGuest.RealizedGuest)
 	}
@@ -349,7 +346,7 @@ func (md *machineDeployment) guestForMachine(mach *api.Machine, group string) *a
 }
 
 func (md *machineDeployment) spawnMachineInGroup(ctx context.Context, groupName string, i, total int, standbyFor []string, meta ...metadata) (machine.LeasableMachine, error) {
-	launchInput, err := md.launchInputForLaunch(groupName, md.guestForMachine(nil, groupName), standbyFor)
+	launchInput, err := md.launchInputForLaunch(groupName, md.guestForGroup(groupName), standbyFor)
 	if err != nil {
 		return nil, fmt.Errorf("error creating machine configuration: %w", err)
 	}
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -311,8 +311,45 @@ type metadata struct {
 	value string
 }
 
+func (md *machineDeployment) guestForMachine(mach *api.Machine, group string) *api.MachineGuest {
+
+	var guest *api.MachineGuest
+	if mach != nil && mach.Config != nil && mach.Config.Guest != nil {
+		guest = helpers.Clone(mach.Config.Guest)
+	}
+	if md.machineGuest.RealizedGuest != nil {
+		guest = helpers.Clone(md.machineGuest.RealizedGuest)
+	}
+	if guest == nil {
+		siblings := lo.FilterMap(md.machineSet.GetMachines(), func(m machine.LeasableMachine, _ int) (*api.Machine, bool) {
+			if m.Machine().ProcessGroup() == group {
+				return m.Machine(), true
+			}
+			return nil, false
+		})
+		median := machine.GetMedianGuest(siblings)
+		if median != nil {
+			guest = median
+		}
+	}
+	if guest == nil {
+		if md.machineGuest.CPUs == 0 && md.machineGuest.MemoryMB == 0 {
+			return nil
+		}
+		guest = helpers.Clone(api.MachinePresets["shared-cpu-1x"])
+	}
+
+	if md.machineGuest.MemoryMB != 0 {
+		guest.MemoryMB = md.machineGuest.MemoryMB
+	}
+	if md.machineGuest.CPUs != 0 {
+		guest.CPUs = md.machineGuest.CPUs
+	}
+	return guest
+}
+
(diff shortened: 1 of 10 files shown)


dangra:
is it a dupe of https://github.com/superfly/flyctl/pull/2297 ?

alichay:
> is it a dupe of #2297 ? It is a bit more complete tho. :)

ugh did the thing again where stuff was worked on simultaneously, oops!
this is more full-featured but also quite a bit bigger, they both have their own merits :)
definitely going to steal his test, though, adding that's a good call

internal/command/machine/run.go
```diff
@@ -47,16 +47,31 @@ var sharedFlags = flag.Set{
...
+		Name:        "vm-cpus",
+		Description: "Number of CPUs",
 	},
 	flag.Int{
 		Name:        "memory",
 		Description: "Memory (in megabytes) to attribute to the machine",
+		Hidden:      true,
+	},
+	flag.Int{
+		Name:        "vm-memory",
```
dangra, line 73: Would be cool to have flag aliases instead of duplicating them here

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -311,8 +311,45 @@ type metadata struct {
 	value string
 }
 
+func (md *machineDeployment) guestForMachine(mach *api.Machine, group string) *api.MachineGuest {
```
dangra, line 314: Hey, I don't think this is a good idea. If I read this correctly, it will update existing machines but the original intention of `--vm-*` flags was to serve as backup for `fly scale vm` command when it is not possible to set vm size before first deploy.

dangra: yeap, I think we have to keep its original meaning and add a better text description to the flags.

dangra:
Hey, I am reviving https://github.com/superfly/flyctl/pull/2297 while we work out the details in this PR to unblock a customer

internal/command/machine/run.go
alichay: I looked into this but this PR is already *big*, and that required an entire rework of the argument system. I got part-way through and realized I wasn't actually doing anything useful

alichay:
> Hey, I am reviving #2297 while we work out the details in this PR to unblock a customer

yeah this one kind of ballooned in complexity haha

anyway, I think this one's finally ready.

dangra:
> One thing of note: this PR doesn't have the flag vm-cpukind. nowhere else iirc do we ask people to just say "performance" vs "shared", as opposed to the templates like "performance-2x", and I also think the duplicated flags vm-size vs vm-cpukind can get a little confusing

I had the same doubt but it was there and I didn't want to overthink it. The vm-size should suffice, happy to see --vm-cpukind go away.

alichay:
fwiw github is warning about a merge conflict that doesn't exist?? the only conflict is two import lines being added, they don't conflict with anything

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -311,8 +311,45 @@ type metadata struct {
 	value string
 }
 
+func (md *machineDeployment) guestForGroup(group string) *api.MachineGuest {
```
dangra, line 314: hey, I still don't see the value of this function :( 

Starting with that it accepts a group name, but it is only called to spawn new machines (launchInputForLaunch) when there are no machines for that group. If there are no "siblings", which never are, it computes a median across all machines.

This is basically the same as computing such median in setMachineGuest(), and it is useful only if we agree the median is better than just defaulting to the minimum and deterministic shared-cpu-1x. 

Once the siblings and median idea is gone, the only left is basically what was before, start with `shared-cpu-1x` then apply vm-size, vm-cpus and vm-memory on top when set.  No need for `realizedGuest` either, just let `md.machineGuest` be a `*api.MachineGuest`

internal/command/machine/run.go
dangra: honestly, with #2297 merged, the only value I see in this PR is the global support for `--vm-*` flags family and its aliases. I'd also remove `--vm-cpukind` as it adds more noise than clarity and `--vm-size` serves as baseline for it.

internal/command/deploy/machines_deploymachinesapp.go
alichay: > Starting with that it accepts a group name, but it is only called to spawn new machines (launchInputForLaunch) when there are no machines for that group. If there are no "siblings", which never are, it computes a median across all machines.

i don't think this is necessarily true! when service config changes, deploy might end up creating new machines in a group with existing machines, right?

internal/command/machine/run.go
alichay: > honestly, with #2297 merged, the only value I see in this PR is the global support for `--vm-*` flags family and its aliases.
Yeah, I definitely see where you're coming from. The only reason I didn't close it after that got merged is for the global flags, and because the median guest calculation might be useful for other things (like scale count, maybe). that said, this _does_ also handle the edge case of service changes affecting machine creation better (imo)

> I'd also remove `--vm-cpukind` as it adds more noise than clarity and `--vm-size` serves as baseline for it.
This has already been announced to people. it'd be pretty lame to announce something then remove it the _next day_.

internal/command/deploy/machines_deploymachinesapp.go
dangra: > i don't think this is necessarily true! when service config changes, deploy might end up creating new machines in a group with existing machines, right?

"deploy" only  creates machines when there are new groups, it doesn't if there are machines for exiting groups. 

The most common case is to update a machine, that won't change the machine guest.

And the last case is "replacing a machine" instead of updating it, this happens when volume config changes, but that reuses the original machine's guest.

hmmm.. it could be for canaries, need to check.

alichay:
we're not getting anywhere here, so I'm going to break this up into child PRs

PR #2301, opened 2023-05-18
Fix inaccurate process group changes displayed during deployment

matttpt:
As Andrea pointed out today, we often create more machines than we say we will!


dangra (approved):
:shipit:

PR #2306, opened 2023-05-18
Experiment: add tab completion for `--app`, `--org`, and `--region`

alichay:
I had to do some ugly reshuffling to get this to build without circular dependencies (hello, `flagnames` package!), but the end result is _really_ slick.

https://github.com/superfly/flyctl/assets/13626644/6a188fdc-50d1-442f-8481-261c403f62eb

### DIFF
--- a/internal/cmdutil/preparers/preparers.go
+++ b/internal/cmdutil/preparers/preparers.go
@@ -1,0 +1,81 @@
+package preparers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/internal/config"
+	"github.com/superfly/flyctl/internal/flag/flagctx"
+	"github.com/superfly/flyctl/internal/httptracing"
+	"github.com/superfly/flyctl/internal/instrument"
+	"github.com/superfly/flyctl/internal/logger"
+	"github.com/superfly/flyctl/internal/state"
+)
+
+type Preparer func(context.Context) (context.Context, error)
+
+func LoadConfig(ctx context.Context) (context.Context, error) {
+	logger := logger.FromContext(ctx)
+
+	cfg := config.New()
+
+	// Apply config from the config file, if it exists
+	path := filepath.Join(state.ConfigDirectory(ctx), config.FileName)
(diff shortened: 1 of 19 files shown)


matttpt (approved):
Overall this looks great to me. I've enjoyed test-driving this branch locally!

I agree that the reorganization is ugly, but I likewise can't think of a better way to get around the circular dependencies—short of trying to merge the flags and preparers into one "command line processing" package, which could easily be uglier. (Though perhaps *all* the preparers could be moved into the new module at some point? Having them split might be a little confusing if you don't already know about it.)

I also wonder if it makes sense to add a (relatively short) timeout to the context for the completion functions. That way it might not freeze for too long if it can't connect or if the API is being very slow. Either way is probably fine, though, since ^C is enough to get out of it.

That's all I've got. This is really cool. Have a ✅!

alichay:
> Though perhaps _all_ the preparers could be moved into the new module at some point?

Hilariously, this *also* causes a circular dependency. It's a nightmare!!!!!!

great idea with the cancelable context, I'll do that then merge. ty for looking this monster over!

smorimoto:
Very nicely laid-out code and work! [I was just thinking about working on it myself](https://community.fly.io/t/auto-complete-the-app-name-on-flyctl/12933), so this is really great. But if I try this, at least it doesn't work with fish shell... Does anyone have any idea?

alichay:
I use fish too. I will say that the way you have to configure completions is...... unintuitive haha.

Do you have a `~/.config/fish/completions/flyctl.fish`?
* If not, you'll have to run `flyctl completion fish > ~/.config/fish/completions/flyctl.fish`
* If you do, check and see if completions work for `flyctl` rather than just `fly`.
   * If they work for flyctl, you can do what I did (it's a hack), and add `alias fly=flyctl` to your `~/.config/fish/config.fish`

smorimoto:
Oh, weirdly it seems to work with `flyctl` 🤯

alichay:
Fair enough! Yeah, if you add the alias to your shell config it should make it work with `fly` as well.

It's a shame shell completion doesn't follow symlinks, but I guess I get why it wouldn't

smorimoto:
Yeah, well anyway it's amazing you let me literally solve this in 10 minutes instead of spending 1 hour debugging. Thank you! 🙂

P.S. Promote this by opening a thread on the community forum! Many people may miss this, but this feature should not be missed!

smorimoto:
OH NO. I just read this 😅 https://community.fly.io/t/quality-of-life-shell-completion-for-flyctl/13651

PR #2310, opened 2023-05-19
Do not override machine image on fly m update

dangra:
Prevent `fly m update` to change the machine image when it is not meant to change like in:

image

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -263,6 +263,11 @@ func runMachineRun(ctx context.Context) error {
 		return fmt.Errorf("the app %s uses an earlier version of the platform that does not support machines", app.Name)
 	}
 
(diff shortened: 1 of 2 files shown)


matttpt (approved):
Probably not a big deal, but just FYI, it does seem to change the error message in this one edge case:

```
% fly m run ""                 
Searching for image '' remotely...
Error: not a valid image: 

% ~/src/flyctl/bin/fly m run ""
Error: could not launch machine: failed to launch VM: config.image required
```

dangra:
@matttpt 👍  : https://github.com/superfly/flyctl/pull/2310/commits/7e18fb1ec6e8b10d7d05b6cbab81516bb24e50ff added a checked for empty image path

```
$ fly m run ""
Error: image argument can't be an empty string
```

PR #2312, opened 2023-05-19
Monitor migrated machine

senyosimpson:
WIP on ensuring an app has actually been successfully migrated. The first approach here only works for HTTP services and will attempt to make a request to the app via its `fly.dev` domain. The same strategy won't work for TCP and UDP. There, we might need to look at fly-proxy logs or something. Or potentially have a custom endpoint in the proxy that can tell us if there's connectivity (for TCP). For UDP, I have no idea.

There's some caveats here too. If an app requires downtime and nomad allocs are not scaled to zero, then this doesn't really meaningfully test anything.

### DIFF
--- a/internal/command/migrate_to_v2/migrate_to_v2.go
+++ b/internal/command/migrate_to_v2/migrate_to_v2.go
@@ -581,7 +581,7 @@ func (m *v2PlatformMigrator) Migrate(ctx context.Context) (err error) {
 		}
 	}
 
-	tb.Detail("Checking if app is responsive")
+	tb.Detail("Ensuring app is responsive")
 	err = m.checkAppResponsive()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


rugwirobaker:
> The same strategy won't work for TCP

Could we run the flyctl proxy and ping through its `6pn`? https://github.com/superfly/flyctl/blob/0826bcaf38839144329e42668b4fbd20b3ffc2ec/internal/command/redis/connect.go#L81-L95

senyosimpson:
> Could we run the flyctl proxy and ping through its 6pn?

Spent the day getting that to work. It's possible _but_ the problem is, the service in the VM needs to be bound to accept ipv6. I'm going to guess that most applications won't do that. Heck it took me forever to realise why my application wasn't responding on my 6pn address, turns out i was only listening on ipv4.

For TCP, the only other alternative I've thought about is having a custom "endpoint" that connects to a TCP service and reports back that it has successfully connected or not. That feels non-trivial to me so before I do that, I'm going to go looking at the splits of how many TCP, UDP and HTTP services we have.

PR #2316, opened 2023-05-21
feat: Make `put` command handle wildcards

klevente:


internal/command/ssh/sftp.go
```diff
@@ -396,6 +396,41 @@ func (sc *sftpContext) chmod(args ...string) error {
 	return nil
 }
 
+func (sc *sftpContext) uploadFile(lpath string, rpath string, permbits int64) {
```
klevente, line 399: extracted already existing logic for uploading a file to a separate function

internal/command/ssh/sftp.go
```diff
@@ -396,6 +396,41 @@ func (sc *sftpContext) chmod(args ...string) error {
...
+	if _, err := sc.ftp.Stat(rpath); err == nil {
+		sc.out("put %s -> %s: file exists on VM", lpath, rpath)
```
klevente, line 401: for now, ignore files which are already present on the container

internal/command/ssh/sftp.go
```diff
@@ -418,46 +453,64 @@ func (sc *sftpContext) put(args ...string) error {
...
+		if !inf.IsDir() {
+			sc.out("put %s -> %s: remote path %s is not a directory", lpath, rdir, rdir)
+			return nil
+		}
+
+		n := len(matches)
+		sc.out("put %s -> %s: matched %d files, uploading...", lpath, rdir, n)
+
+		for i, file := range matches {
+			file = filepath.ToSlash(file)
```
klevente, line 508: had to force `/`s because on Windows, resolved glob patterns are separated by `\`

internal/command/ssh/sftp.go
```diff
@@ -418,46 +453,64 @@ func (sc *sftpContext) put(args ...string) error {
...
+			filename := path.Base(file)
+			rpath := path.Join(rdir, filename)
```
klevente, line 510: handles both the case when user supplied `dir` and `dir/` as a target directory for bulk upload

kzys (commented):
Personally I'd let shells handle wildcards. Powerful shells like zsh can do more than filepath.Glob() does.

There are `flyctl ssh sftp get` already. I could add `flyctl ssh sftp put` instead. Would it work for you?

klevente:
@kzys Yep that makes way more sense actually, as that would also enable devs to write scripts for uploading files programmatically. It'd also satisfy my use-case fully.

Just a question: I checked how `fly sftp get` works, and it operates on a single file as well. I'm not sure how we can let the shell expand the provided glob before passing it into `fly`, but I am not well-versed with Unix shell stuff to be honest.

Also, if we decide to go this route, it'd be cool to have a flag like `--overwrite` that overwrites any existing files in the container. WDYT?

mrluc:
I came here because I would *looove* `fly sftp put` 😍  @kzys

Stumbled across this thread because `fly sftp shell` doesn't allow completions, or the star syntax, which taken together make using it a schlep -- adding `put` would probably solve

PR #2317, opened 2023-05-22
Respsect '--vm-*' deploy flags when sizing release machine

pborzenkov:


dangra (commented):
Any motivation to merge this now? I think it is in collision path with #2300 and we can work on getting that one merged instead.

pborzenkov:
> Any motivation to merge this now? I think it is in collision path with https://github.com/superfly/flyctl/pull/2300 and we can work on getting that one merged instead.

There is a support ticket in Helpscout from a user that can't run `release_command` because it doesn't allow to choose VM size during initial deploy. AFAIK, this affects only the first deployment, as after the first successful one the release machine will be sized accordingly.

But I'd rather integrate this one into @alichay's PR.

dangra (approved):
let get it in If this is blocking an actual customer. We can absorb this PR into #2300 which can take longer.

PR #2318, opened 2023-05-22
fly proxy: fix FD leak

pborzenkov:
`defer` is not called until the function returns, so `fly proxy`
has been leaking one socket per connection.


dangra (approved):

pborzenkov:
Yeah, it's about `Close`. We don't really need to check the error there, I'll make a PR to tell the linter to ignore this one.

PR #2321, opened 2023-05-22
Set `console_command` for Laravel apps

matttpt:
This is the command that @fideloper suggested for the `fly console` [Fresh Produce post](https://community.fly.io/t/easily-run-a-console-in-a-new-machine-with-fly-console/12773).


dangra (approved):
this looks good to me. Follows fresh produce post to the letter :)

PR #2325, opened 2023-05-22
WIP: migrate-to-v2 Nomad Autoscaling

billyb2:
Currently, some machines are made that are set to autostop as a replacement.


internal/command/migrate_to_v2/machines.go
```diff
@@ -12,8 +12,16 @@ import (
...
-func (m *v2PlatformMigrator) resolveMachineFromAlloc(alloc *api.AllocationStatus) (*api.LaunchMachineInput, error) {
-	mConfig, err := m.appConfig.ToMachineConfig(alloc.TaskName, nil)
+func (m *v2PlatformMigrator) createLaunchMachineInput(oldAllocID string, skipLaunch bool) (*api.LaunchMachineInput, error) {
+	taskName := ""
+
+	if len(m.oldAllocs) > 0 {
+		taskName = m.oldAllocs[0].TaskName
+	} else {
+		taskName = "app"
+	}
```
alichay, line 22: Shouldn't this be looking for the taskName of the alloc specified by `oldAllocID`?
```suggestion
	alloc, hasAlloc := lo.Find(m.oldAllocs, func(alloc *api.AllocationStatus) (*api.AllocationStatus, bool) {
		return alloc, alloc.ID == oldAllocID
	})
	if hasAlloc {
		taskName = alloc.TaskName
	} else {
		taskName = "app"
	}
```

internal/command/migrate_to_v2/machines.go
```diff
@@ -32,26 +42,76 @@ func (m *v2PlatformMigrator) resolveMachineFromAlloc(alloc *api.AllocationStatus
 
 	// We have manual overrides for some regions with the names <region>2 e.g ams2, iad2.
 	// These cause migrations to fail. Here we handle that specific case.
-	region := alloc.Region
+	if m.appConfig == nil {
+		// FIXME better error message here
+		return nil, fmt.Errorf("Could not find app config")
+	}
+
+	region := m.appConfig.PrimaryRegion
```
alichay, line 50: ditto above
```suggestion
	region := ""
	if hasAlloc {
		region = alloc.Region
	} else {
		region = m.appConfig.PrimaryRegion
	}
```

alichay (commented):
I could be wrong but it looks like this'll end up making new machines all copies of oldAllocs[0] in the app's primary region, instead of respecting the old allocs.
(ik this is WIP, just wanted to point this out just in case!)

internal/command/migrate_to_v2/machines.go
billyb2: yep, you're right!

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -688,10 +690,6 @@ func (m *v2PlatformMigrator) validateKnownUnmigratableApps(ctx context.Context)
 }
 
 func (m *v2PlatformMigrator) validateScaling(ctx context.Context) error {
-	// FIXME: for now we fail if there is autoscaling.. remove this once we create any extra machines based on autoscaling config
-	if m.autoscaleConfig.Enabled {
-		return fmt.Errorf("cannot migrate app %s with autoscaling config, yet; watch https://community.fly.io for announcements about autoscale support with migrations", m.appCompact.Name)
-	}
 	return nil
```
benwaffle, line 693: should we just delete this method?

billyb2: Probably should, yea.

benwaffle (approved):
Works for me. I'm not convinced autoscaling on nomad even works at all. But I see it working fine after migration.

benwaffle:
> actually deploying the autostop machines as stopped.

I think it's ok to ship without this feature, as they will autostop on their own within ~30min it seems.

internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -692,22 +690,7 @@ func (m *v2PlatformMigrator) validate(ctx context.Context) error {
...
-	// This is as last meassure for very special cases
-	knownUnmigratableApps := []string{
-		"globalconfig-dht", // https://flyio.discourse.team/t/x/2860/6
-	}
-	if slices.Contains(knownUnmigratableApps, m.appCompact.ID) {
-		return fmt.Errorf("Your app uses features incompatible with the V2 platform. Please contact support to discuss how to successfully migrate")
-	}
-	return nil
-}
-
```
alichay, line 705: Why'd we get rid of this?

PR #2327, opened 2023-05-23
Respect '--token' flag when talking to flaps/nats

pborzenkov:
This is a second attempt that also fixes BuildKit auth.


dangra:
How do we prevent a regression? does it need a preflight test case?

pborzenkov:
I've also tested this with `testflight` locally this time, all green.

dangra (approved):

PR #2333, opened 2023-05-24
agent: switch to nhooyr.io/websocket for websockets

pborzenkov:
The main motivation is much better error reporting. We have some weird
agent issues even when Wireguard connection is tunneled over Websockets,
some of them where even Websocket connection can't be established.

Unfortunately, golang.org/x/net/websocket doesn't return detailed errors
and makes it very hard to debug such issues.

As a bonus point, Websocket connection can now be established over
proxy, so it should work in more restrictive network environments.


matttpt (commented):
Overall LGTM! Two questions:

* Do you expect any issues having removed the read/write deadlines?
* It seems weird to me that this uses text frames, but I'm guessing that that was the way it was before?

kzys:
You should also mention that golang.org/x/net/websocket is deprecated-ish! I wasn't aware about.

wg/ws.go
```diff
@@ -136,33 +137,42 @@ func (wswg *WsWgProxy) Port() (int, error) {
...
-		InsecureSkipVerify: true,
-	}
-
-	// oh well, if it'll end horror
-	ws, err := websocket.DialConfig(conf)
+	ws, _, err := websocket.Dial(ctx, rurl, &websocket.DialOptions{
+		HTTPClient: &http.Client{
+			Transport: &http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				// It's fine. The traffic inside the tunnel is already encrypted by WG
```
kzys, line 149: Would it have Wireguard over WebSocket over HTTP?

pborzenkov:
> Overall LGTM! Two questions:
> 
> * Do you expect any issues having removed the read/write deadlines?

Nope. This package uses context internally, so reads and writes cancel just fine. I've also tested with putting my laptop to sleep for a longer period and resuming, the connection re-establishes just fine.

> * It seems weird to me that this uses text frames, but I'm guessing that that was the way it was before?

Yep. We can provide something like `/binary` endpoint on the backend that talks binary protocol, but I don't want to touch it for now. I also don't see any problems with this except when something is trying to interpret the data: I've tried running the connection via `mitmproxy` while debugging and it fails because `mitmproxy` tries to convert the data to string and it's not a valid UTF-8 string.

wg/ws.go
pborzenkov: > Would it have Wireguard over WebSocket over HTTP?

I think it's fine to have HTTP, but right now backend is listening on HTTPS only and I don't want to touch it at the moment.

dangra (approved):

PR #2338, opened 2023-05-24
`migrate-to-v2`: filter out failed nomad allocs

alichay:
Preflight passes, but I'd really appreciate if someone who knows nomad better that I do could double-check that filtering by `!alloc.Failed && alloc.LatestVersion` shouldn't introduce weird side effects

### DIFF
--- a/internal/command/migrate_to_v2/migrate_to_v2.go
+++ b/internal/command/migrate_to_v2/migrate_to_v2.go
@@ -236,6 +236,9 @@ func NewV2PlatformMigrator(ctx context.Context, appName string) (V2PlatformMigra
 	if err != nil {
 		return nil, err
 	}
+	allocs = lo.Filter(allocs, func(alloc *api.AllocationStatus, _ int) bool {
+		return !alloc.Failed && alloc.LatestVersion
+	})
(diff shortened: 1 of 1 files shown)


internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -236,6 +236,9 @@ func NewV2PlatformMigrator(ctx context.Context, appName string) (V2PlatformMigra
 	if err != nil {
 		return nil, err
 	}
+	allocs = lo.Filter(allocs, func(alloc *api.AllocationStatus, _ int) bool {
+		return !alloc.Failed && alloc.LatestVersion
```
dangra, line 240: I've no idea what "latest_version" means :) ... digging ⌛

alichay: at a glance: it looks like web returns all running allocations, even those that should be destroyed but haven't yet.
LatestVersion is true if alloc.version == last release version on the app 
it's an edge case, but this should filter out weird allocations that just don't die.

dangra (approved):
🚢

PR #2340, opened 2023-05-24
Add `--vm-size` to `fly console`

matttpt:
A number of other commands have `--vm-size`, so it makes sense for `fly console` to have it too.


dangra (approved):
It looks good

nit: the logic in this function keeps repeating across the codebase, at some point we will have to factor it out into a reusable function

PR #2341, opened 2023-05-24
Fixes for autostart/autostop-related config fields

matttpt:
1. When generating an app config from machines, copy the `autostart`, `autostop`, and `min_machines_running` fields.
2. Don't use `omitempty` for TOML serialization of `auto_start_machines` and `auto_stop_machines`. The encoder always omits `nil` values since they can't be represented in TOML. `omitempty` actually omits non-`nil` `false` values as well.

The motivation for this is to make sure that using `fly config save` and then `fly deploy` doesn't inadvertently change the configuration (especially for [Postgres](https://community.fly.io/t/how-to-turn-off-postgres-scaling-to-zero-after-1-hour/12522/2)).

### DIFF
--- a/internal/appconfig/service.go
+++ b/internal/appconfig/service.go
@@ -11,8 +11,8 @@ import (
 type Service struct {
 	Protocol           string                         `json:"protocol,omitempty" toml:"protocol"`
 	InternalPort       int                            `json:"internal_port,omitempty" toml:"internal_port"`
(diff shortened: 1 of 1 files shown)


dangra (approved):
LGTM and Makes sense! 

> Don't use omitempty for TOML serialization of auto_start_machines and auto_stop_machines. For these fields the encoder always omits nil values since they can't be represented in TOML. omitempty actually omits non-nil false values as well.

I am sure my future self will add the "omitempty" back if this comment is not above those same lines 😄

PR #2342, opened 2023-05-24
Disable cgo in Makefile

kzys:
flyctl could be copied into Linux containers and it must not assume any libc versions.

Note that cgo has been disabled in Dockerfile.dev, .goreleaser.yml and .goreleaser.dev.yml.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ generate:
 
 build: generate
 	@echo Running Build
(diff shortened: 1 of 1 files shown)


dangra:
Is it disabled for Window builds too?

kzys:
Seems like so.

https://github.com/superfly/flyctl/blob/b7b9de2b4c1739254e7e4c75bfd7a215a7ca0fb8/.goreleaser.yml#L26

dangra (approved):

PR #2349, opened 2023-05-26
different deploy token scopes

DAlperin:


internal/command/tokens/create.go
```diff
@@ -59,45 +61,82 @@ func newDeploy() *cobra.Command {
 			Description: "The duration that the token will be valid",
 			Default:     time.Hour * 24 * 365 * 20,
 		},
+		flag.String{
+			Name:        "scope",
+			Shorthand:   "s",
+			Description: "either 'app' or 'org'",
+			Default:     "app",
+		},
```
btoews, line 69: I don't have strong feelings about flyctl UX, but are we sure we want `fly tokens create deploy -s org` and not `fly tokens create org`?

internal/command/tokens/create.go
```diff
@@ -59,45 +61,82 @@ func newDeploy() *cobra.Command {
...
+			"deploy_organization",
+			&gql.LimitedAccessTokenOptions{},
+			expiry,
+		)
+		if err != nil {
+			return fmt.Errorf("failed creating deploy token: %w", err)
+		}
+
+		token = resp.CreateLimitedAccessToken.LimitedAccessToken.TokenHeader
 	}
```
btoews, line 133: Maybe a default case here?

internal/command/tokens/create.go
```diff
@@ -59,45 +61,82 @@ func newDeploy() *cobra.Command {
...
-	)
-	if err != nil {
-		return fmt.Errorf("failed creating deploy token: %w", err)
+	scope := flag.GetString(ctx, "scope")
+	switch scope {
+	case "app":
+		appName := appconfig.NameFromContext(ctx)
+		//if appName == "" {
+		//	return command.ErrRequireAppName
+		//}
```
btoews, line 90: Debugging?

internal/command/tokens/create.go
```diff
@@ -59,45 +61,82 @@ func newDeploy() *cobra.Command {
...
+			org.ID,
+			"deploy_organization",
+			&gql.LimitedAccessTokenOptions{},
+			expiry,
+		)
+		if err != nil {
+			return fmt.Errorf("failed creating deploy token: %w", err)
+		}
+
+		token = resp.CreateLimitedAccessToken.LimitedAccessToken.TokenHeader
```
btoews, line 132: Wouldn't hurt to DRY up these parts between the two sections.

internal/command/tokens/list.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/command/orgs"
+	"github.com/superfly/flyctl/internal/flag"
+	"github.com/superfly/flyctl/internal/render"
+	"github.com/superfly/flyctl/iostreams"
+)
+
+func newList() *cobra.Command {
+	const (
+		short = "List deploy tokens"
```
btoews, line 19: We might have non-deploy tokens at some point

```suggestion
		short = "List tokens"
```

internal/command/tokens/revoke.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/flag"
+)
+
+func newRevoke() *cobra.Command {
+	const (
+		short = "Revoke deploy tokens"
```
btoews, line 14:
```suggestion
		short = "Revoke tokens"
```

btoews (approved):
There's an undocumented alias allowing `fly tokens deploy` instead of having to type `fly tokens create deploy`. Up to you if we want to keep adding those aliases for other token profiles. That's implemented here https://github.com/superfly/flyctl/blob/ff8a399ac3d67a89468fd54b382b473bdfa37d2e/internal/command/tokens/tokens.go#L17-L18

PR #2350, opened 2023-05-26
add --image and --fly-credentials to fly console

guillaumervls:
This enables passing a custom Docker image and arguments for the console machine. With it users can share & reuse third-party scripts/tools to interact with their apps. Use cases include setting up OSS third-party apps, like DBs, like MySQL ⬇️ 

You can test it with https://github.com/guillaumervls/flymysql: instead of typing the chonky shell code in the README, now you can just run _(assuming you've built this branch's version of `flyctl` and aliased it as `dfly`)_
```
fly apps create <app-name>
dfly console --app <app-name> --fly-credentials --image guillaumervls/flymysql:latest setup
```
And ✨ you've got a MySQL DB on Fly!


kzys (commented):
This is *wild*.

internal/command/console/console.go
```diff
@@ -159,7 +224,37 @@ func runConsole(ctx context.Context) error {
...
+			var cfg config
+			err = mapstructure.Decode(resp.App.Image.Config, &cfg)
+			if err != nil {
+				return fmt.Errorf("failed to decode config: %w", err)
+			}
+			consoleCommand = fmt.Sprintf(
+				"/bin/sh -c 'cd %s && %s %s'",
+				cfg.Config.WorkingDir,
+				strings.Join(cfg.Config.Entrypoint, " "),
+				strings.Join(flag.Args(ctx), " "),
```
kzys, line 251: It seems unnecessary fragile because of https://owasp.org/www-community/attacks/Command_Injection.

Sadly https://pkg.go.dev/golang.org/x/crypto/ssh doesn't support shell-less execution like os/exec. Would https://github.com/kballard/go-shellquote work?

PR #2361, opened 2023-05-29
`deploy`: add `--release-command-timeout`, prompt change timeout on err

alichay:
This adds a flag `--release-command-timeout`, so that the release command can have a separate timeout from the general purpose `wait-timeout` used on app machines. This timeout starts at five minutes, instead of two, and supports a `none` option to allow the release command to go on forever.

Release command failures (and normal wait failures) fail with this message when timing out now:
```
  Waiting for 4d891163a52087 to have state: destroyed
Error: release command failed - aborting deployment. error waiting for release_command machine 4d891163a52087 to finish running: timeout reached waiting for machine to destroyed failed to wait for VM 4d891163a52087 in destroyed state: Get "https://api.machines.dev/v1/apps/ali-process-dns-example/machines/4d891163a52087/wait?instance_id=01H1JV6NGZGGJK2N611XQXM8PE&state=destroyed&timeout=2": net/http: request canceled
note: you can change this timeout with the --release-command-timeout flag
```

(I'm not even sure if we *want* to post this on every wait failure, but it's a better status quo than now!)

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -3,6 +3,7 @@ package deploy
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


tvdfly (approved):
🚢🚢🚢  thx!

PR #2363, opened 2023-05-29
`deploy`: canary default for volume-less apps

alichay:
Switch to the canary strategy for apps without volumes.

Canary now throws an error if deploying on an app with volumes - do we want that?

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -407,10 +407,31 @@ func (md *machineDeployment) setMachineGuest(vmSize string, vmCPUKind string, vm
 	return nil
 }
 
+func (md *machineDeployment) deployMethodIsCompatible(method string) error {
+	switch strings.ToLower(method) {
+	case "canary", "bluegreen":
+		if len(md.appConfig.Mounts) > 0 {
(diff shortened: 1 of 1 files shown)


internal/command/deploy/machines.go
```diff
@@ -407,10 +407,31 @@ func (md *machineDeployment) setMachineGuest(vmSize string, vmCPUKind string, vm
 	return nil
 }
 
+func (md *machineDeployment) deployMethodIsCompatible(method string) error {
+	switch strings.ToLower(method) {
```
kzys, line 411: Was it case-insensitive?

alichay: when we deploy, we use ToUpper on the strategy, so we accept values of any case.
https://github.com/superfly/flyctl/blob/d614f1ad9bd5ca729dc2de186cbf0f326d206819/internal/command/deploy/machines.go#L436

SCREAMING_CASE string constants are pretty ugly so I opted to check against lowercase ones, but it shouldn't make a difference :)

kzys: I see. I wanted to save future Fly customers from having UPPER_CASE vs. lower_case debate, but seems the ship has sailed :)

kzys (approved):

PR #2366, opened 2023-05-29
CI for preflight test

tvdfly:
Add new tool: make ci-preflight. It runs preflight tests with json output and sends to a file, then a new processing script reads the file and shows only the pass/fail/skip results (i.e. no output).

Add a new github action workflow for running the preflight tests once per day and after pushes to the master branch.


.github/workflows/preflight.yml
```diff
@@ -0,0 +1,42 @@
...
+    - cron: '21 17 * * *'
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: Brief reason for running this workflow manually
+        required: false
+        default: User initiated run
+        type: string
+  push:
+    branches: [master, ci-preflight]
```
tvdfly, line 14: reminder to self: remove the ci-preflight branch from here before merging the pr

scripts/delete_preflight_apps.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -eo pipefail
+if [ "${FLY_PREFLIGHT_TEST_FLY_ORG}x" = "x" ] ; then
```
kzys, line 3: `if [ "${FLY_PREFLIGHT_TEST_FLY_ORG}" == "" ] ; then` should work here.

https://www.shellcheck.net/wiki/SC2268

scripts/report-ci-preflight-results/main.go
```diff
@@ -0,0 +1,100 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+)
+
+func main() {
+	if err := run(); err != nil {
+		time.Sleep(50 * time.Millisecond) // let stdout flush
```
kzys, line 14: How does it work?

tvdfly: I want the error that's printed on the next line to at the very end of the output, because it will have the command to rerun things. stdout/stderr being unbuffered and the way that github actions capture those two output streams means it's not guaranteed to be at the end of the github action output. adding this little sleep was my i'm a lazy engineer and this will mostly work fix :-) 

this is what happened before I put in this short sleep: https://github.com/superfly/flyctl/actions/runs/5115901837/jobs/9197567462

kzys: Because we have a lot in stdout? I could take a look that after master merge.

kzys (approved):

scripts/report-ci-preflight-results/main.go
tvdfly: I deleted all this after finding a somewhat simpler way to avoid logging the history output. So it should all be moot now :-)

scripts/delete_preflight_apps.sh
tvdfly: np, will change. please use ["insert a suggestion"](https://haacked.com/archive/2019/06/03/suggested-changes/) for nits like this in the future, so the pr author can do the 1 click accept to incorporate the change. thx!

PR #2375, opened 2023-05-31
wireguard: re-read the config before prunning the peers

pborzenkov:
So we always working with the latest on-disk config and don't overwrite
it with the stale data.

### DIFF
--- a/internal/wireguard/wg.go
+++ b/internal/wireguard/wg.go
@@ -143,6 +143,13 @@ type WireGuardStates map[string]*wg.WireGuardState
 func GetWireGuardState() (WireGuardStates, error) {
 	states := WireGuardStates{}
 
+	// Re-read the config to make sure we are working with the latest on-disk version.
+	// Otherwise we might write stale data that was there when the agent was started
+	// (like deleted token).
+	if err := viper.ReadInConfig(); err != nil && !os.IsNotExist(err) {
+		return states, err
+	}
+
 	if err := viper.UnmarshalKey(flyctl.ConfigWireGuardState, &states); err != nil {
 		return nil, errors.Wrap(err, "invalid wireguard state")
 	}


internal/wireguard/wg.go
```diff
@@ -141,6 +141,11 @@ func C25519pair() (string, string) {
 type WireGuardStates map[string]*wg.WireGuardState
 
 func GetWireGuardState() (WireGuardStates, error) {
+	// Re-read the config to make sure we are working with the latest on-disk version.
+	// Otherwise we might write stale data that was there when the agent was started
+	// (like deleted token).
+	flyctl.InitConfig()
```
kzys, line 147: flyctl.InitConfig does a lot and it would even saves the config again.

https://github.com/superfly/flyctl/blob/d614f1ad9bd5ca729dc2de186cbf0f326d206819/flyctl/flyctl.go#L90-L114

What is the specific bug you'd like to fix? Is it possible to do this reload before pruning, instead of all Getter calls?

pborzenkov: > flyctl.InitConfig does a lot
I think I can do just `viper.ReadInConfig()`

>  and it would even save the config again.
Only if we are migrating from an old config format, no?

> What is the specific bug you'd like to fix?
Agent is long-running and right now operates on the cached copy of the config. Any changes made to the config get overwritten by it (like token, for example).

> Is it possible to do this reload before pruning, instead of all getter calls?
`GetWireGuardState() is also called from `validateTunnelsUnlocked` which terminates tunnels for invalid peers. So it needs up-to-date peer info as well.

kzys (approved):

kzys:
Looks good to me!

> Agent is long-running and right now operates on the cached copy of the config. Any changes made to the config get overwritten by it (like token, for example).

You don't have to do that in this PR, but it may make sense to do that in agent/server/server.go in that case. GetWireGuardState doesn't have to know about the agent mode and the agent might call different functions later.

PR #2376, opened 2023-05-31
`scale count`: fix `process-group` flag ignored

alichay:
Now, even if `app` is present, if `--process-group xyz` is specified, `scale count` will try to change `xyz` instead of `app`.

### DIFF
--- a/internal/command/scale/count.go
+++ b/internal/command/scale/count.go
@@ -55,20 +55,18 @@ func runScaleCount(ctx context.Context) error {
 
 	args := flag.Args(ctx)
 
-	groupName := ""
-
 	processNames := appConfig.ProcessNames()
-	if !slices.Contains(processNames, api.MachineProcessGroupApp) {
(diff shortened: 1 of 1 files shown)


internal/command/scale/count.go
```diff
@@ -55,20 +55,18 @@ func runScaleCount(ctx context.Context) error {
...
+		if !slices.Contains(processNames, groupName) {
 			return fmt.Errorf("--process-group flag is required when no group named 'app' is defined")
 		}
 	}
 
-	if groupName == "" {
-		groupName = appConfig.DefaultProcessName()
+	if !slices.Contains(processNames, groupName) {
+		return fmt.Errorf("process group '%s' not found", groupName)
 	}
```
matttpt, line 70: Whoa, I didn't realize it previously let you operate on arbitrary/nonexistent process groups!

matttpt (approved):

PR #2377, opened 2023-05-31
fix: attempt to infer machine guest for canary from process group

jipperinbham:
Currently, the canary machine is always the default guest `shared-cpu-1x` or based on any flags provided in `fly deploy`. This is causing issues for many apps where the canary fails because it needs to have the same guest configuration as the other machines in its process group.

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -79,6 +79,26 @@ func (md *machineDeployment) restartMachinesApp(ctx context.Context) error {
 	return md.updateExistingMachines(ctx, machineUpdateEntries)
 }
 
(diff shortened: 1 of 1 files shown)


internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -79,6 +79,26 @@ func (md *machineDeployment) restartMachinesApp(ctx context.Context) error {
 	return md.updateExistingMachines(ctx, machineUpdateEntries)
 }
 
+func (md *machineDeployment) inferCanaryGuest(name string) *api.MachineGuest {
```
senyosimpson, line 82: Just a personal one but this would read easier to me if it was something like

```go
mInGroup := lo.Filter(machines, machine.ProcessGroup() == name)
mInGroup[0].Config.Guest
```

I think this achieves the same thing (at the expense of having to iterate over everything)

PR #2378, opened 2023-05-31
Remove fly console's 15 seconds hard timeout

kzys:
Some machines take more to boot.


matttpt (approved):
LGTM!

IIRC the flaps client can accept `Wait` timeouts of up to 60 seconds, so you could bump that for less frequent API calls if you'd like.

PR #2379, opened 2023-05-31
Fix machine check warning status

benwaffle:
Per corrosion:
```
./state.db> select distinct status from consul_checks;
+----------+
| status   |
+----------+
| passing  |
| critical |
| warning  |
+----------+
```

### DIFF
--- a/internal/format/format.go
+++ b/internal/format/format.go
@@ -124,7 +124,7 @@ func passingChecks(checks []api.CheckState) (n int) {
 
 func warnChecks(checks []api.CheckState) (n int) {
 	for _, check := range checks {
-		if check.Status == "warn" {
+		if check.Status == "warning" {
 			n++
 		}
 	}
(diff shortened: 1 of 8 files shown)


kzys (commented):
The time to defile constants?

benwaffle:
Maybe, but I can't see a clean way to make the JSON unmarshaller validate the string enum

benwaffle:
@kzys something like that?

internal/format/format.go
```diff
@@ -124,7 +124,7 @@ func passingChecks(checks []api.CheckState) (n int) {
 
 func warnChecks(checks []api.CheckState) (n int) {
 	for _, check := range checks {
-		if check.Status == "warn" {
+		if check.Status == "warning" {
```
kzys, line 127: Here too?

kzys (approved):

internal/format/format.go
benwaffle: that's a different field in types.go. I only changed machine_types.go

PR #2383, opened 2023-06-01
`migrate-to-v2 troubleshoot`: tool to diagnose & repair failed migrations

alichay:
This should make supporting future blocks of automated migrations much easier :)

It does two things, right now

1. If it's run on a `machines` app, it'll look for any running allocs. If there are any, it'll switch the app to `detached`, scale to zero, then switch back to `machines`.
2. If it's run on a `detached` app, it'll prompt the user to look at all running allocs, all running machines, and pick which side to keep. They can also just exit here, if they'd prefer to talk it out with support.


matttpt (commented):
Since I'm not working on migrations I'm not super qualified to comment on this, but it looks good (and really cool!) to me. Added one question about the code below.

PR #2394, opened 2023-06-01
`migrate-to-v2`: mark replaced volumes as "replaced"

alichay:
Marks replaced volumes with a status flag to make internal bookkeeping easier during the transition


dangra:
@alichay this is outdated and not sure we are going to use it at this point. Would you be ok to close the PR?

PR #2400, opened 2023-06-04
Fix `Dockerfile` typo in error message.

dsomel21:
## Description

I was trying to deploy a Rails app and noticed this typo. 

## Screenshot

Screen Shot 2023-06-04 at 7 03 32 AM


dsomel21:
Also, I am not familiar with Docker... since it failed to generate one (a `Dockerfile`), should I just create one manually?

rubys:
Try running `bundle install` then launch again.

PR #2411, opened 2023-06-06
Update copy in flyctl to be more consistent with Fly terminology, capitalization, and other copy conventions

bradgessler:
Making improvements to CLI copy as I try launching Rails apps, all in the name of improving DevUX.

### DIFF
--- a/gql/schema.graphql
+++ b/gql/schema.graphql
@@ -9189,7 +9189,7 @@ enum ServiceHandlerType {
   HTTP
 
   """
-  Handle TLS for PostgreSQL connections
+  Handle TLS for Postgres connections
   """
   PG_TLS
 
--- a/internal/command/apps/open.go
+++ b/internal/command/apps/open.go
@@ -48,11 +48,11 @@ func runOpen(ctx context.Context) error {
 
 	app, err := client.FromContext(ctx).API().GetAppCompact(ctx, appName)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


gql/schema.graphql
```diff
@@ -9189,7 +9189,7 @@ enum ServiceHandlerType {
   HTTP
 
   """
-  Handle TLS for PostgreSQL connections
+  Handle TLS for Postgres connections
```
matttpt, line 9192: I think this file is generated from `web`, so you may need to change this text over there too to prevent it from getting overwritten in the future.

internal/command/apps/open.go
```diff
@@ -48,11 +48,11 @@ func runOpen(ctx context.Context) error {
 
 	app, err := client.FromContext(ctx).API().GetAppCompact(ctx, appName)
 	if err != nil {
-		return fmt.Errorf("failed retrieving app %s: %w", appName, err)
+		return fmt.Errorf("Failed retrieving app %s: %w", appName, err)
```
matttpt, line 51: Just a nitpick, but IIRC typical Go style is to leave the messages for `error`s [uncapitalized](https://google.github.io/styleguide/go/decisions.html#error-strings). I think `flyctl` *generally* adheres to this—but not always, and these are top-level errors anyway, so if you think it looks better for the user, then that's fine by me :smile:

matttpt (approved):
Two quick notes below, but still LGTM as-is!

PR #2421, opened 2023-06-08
migrate-to-v2: create machines in bom region if allocs are in maa region

lillianberryfly:
https://status.flyio.net/incidents/f2chl7gggnx3


jipperinbham (approved):
LGTM, I'm indifferent on whether adding a log out it necessary so will leave that decision up to you.

PR #2425, opened 2023-06-08
migrate-to-v2: troubleshoot shouldn't assume automated migration

benwaffle:
Migrations done by users themselves (locally) don't have admin-bot@fly.io releases. This was probably affecting https://community.fly.io/t/error-creating-a-new-machine-in-group-app-requires-an-unattached-data-volume-create-it-with-fly-volume-create-data/13403/10

This also applies to apps that have always been on v2 from the start

### DIFF
--- a/internal/command/migrate_to_v2/troubleshoot.go
+++ b/internal/command/migrate_to_v2/troubleshoot.go
@@ -163,8 +163,6 @@ func (t *troubleshooter) refreshAllocs(ctx context.Context) error {
 }
 
 func (t *troubleshooter) hasMigrated(ctx context.Context) bool {
-	client := client.FromContext(ctx).API()
-
 	if t.app.PlatformVersion == appconfig.DetachedPlatform {
 		return true
 	}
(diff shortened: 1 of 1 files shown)


internal/command/migrate_to_v2/troubleshoot.go
```diff
@@ -183,20 +181,7 @@ func (t *troubleshooter) hasMigrated(ctx context.Context) bool {
...
-	for _, release := range releases {
-		// Technically, I don't think this is the only time we could use admin-bot@fly.io,
-		// but we use it infrequently and soon we'll be done dealing with this,
-		// so it's probably an acceptable way to determine this for now.
-		if release.User.Email == "admin-bot@fly.io" {
-			return true
-		}
-	}
-	return false
+	return true
```
benwaffle, line 184: This makes me think the entire function should just be
```
return t.app.PlatformVersion != appconfig.NomadPlatform
```

alichay: Yeah, that's what I was about to say. It used to be more important that we got this perfect, but it's a lot more graceful now so we can just say "if not nomad, keep going"

PR #2427, opened 2023-06-09
make wait timeout configurable for migrations

jipperinbham:


internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -37,6 +37,8 @@ import (
 	"golang.org/x/exp/slices"
 )
 
+var defaultWaitTimeout = 5 * time.Minute
```
alichay, line 40: nit: should probably be const

alichay (approved):

PR #2435, opened 2023-06-12
Add back in making wg websockets enabled by default

Lucais11:


dangra (approved):
The failing preflight test seems like an unrelated timeout caused by the backend.

PR #2437, opened 2023-06-13
use full paths when calling exec command

rubys:
See:
  https://tip.golang.org/doc/go1.19#os-exec-path
  https://community.fly.io/t/unable-to-run-flyctl-launch/13504

### DIFF
--- a/scanner/rails.go
+++ b/scanner/rails.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -13,6 +14,7 @@ import (
 )
 
 var healthcheck_channel = make(chan string)
+var bundle, ruby string
 
 func configureRails(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
 	// `bundle init` will create a file with a commented out rails gem,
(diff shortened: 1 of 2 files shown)


kzys:
Putting relative paths in PATH is bad. Go changed the behavior as you noted. Rust did it too in https://github.com/rust-lang/rust/pull/87704.

Is to coming from customers' own configuration (like ~/.bashrc) or tools (like Bundler)?

rubys:
Putting relative paths in PATH is commonplace for Ruby.  Example: https://github.com/rbenv/rbenv/wiki/Understanding-binstubs#adding-project-specific-binstubs-to-path ; note the security concern mentioned on that page.

scanner/rails.go
```diff
@@ -26,14 +28,39 @@ func configureRails(sourceDir string, config *ScannerConfig) (*SourceInfo, error
...
+
+		if err != nil {
+			return nil, errors.Wrap(err, "failure finding bundle executable")
+		}
+	}
+
+	ruby, err = exec.LookPath("ruby")
+	if err != nil {
+		if errors.Is(err, exec.ErrDot) {
+			bundle, err = filepath.Abs(ruby)
```
kzys, line 48: bundle -> ruby

scanner/rails.go
```diff
@@ -71,9 +98,21 @@ func configureRails(sourceDir string, config *ScannerConfig) (*SourceInfo, error
 			},
 		}
 	} else {
+		// find absolute path to rake executable
+		rake, err := exec.LookPath("rake")
+		if err != nil {
+			if errors.Is(err, exec.ErrDot) {
+				bundle, err = filepath.Abs(ruby)
```
kzys, line 105: bundle -> rake?

kzys (changes requested):

kzys:
Oh rbenv...

kzys (approved):

PR #2438, opened 2023-06-13
Show roles in `fly machines` selection menu

matttpt:
I found this useful when working with an HA Fly Postgres app.

### DIFF
--- a/internal/command/machine/select.go
+++ b/internal/command/machine/select.go
@@ -201,9 +201,29 @@ func sortAndBuildOptions(machines []*api.Machine) []string {
 		if group := machine.ProcessGroup(); group != "" {
 			details += fmt.Sprintf(", process group '%s'", group)
 		}
+		role := getMachineRole(machine)
+		if role != "" {
+			details += fmt.Sprintf(", role '%s'", role)
+		}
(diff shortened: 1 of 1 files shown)


dangra (approved):

internal/command/machine/select.go
```diff
@@ -201,11 +201,31 @@ func sortAndBuildOptions(machines []*api.Machine) []string {
...
 	return options
 }
 
+func getMachineRole(machine *api.Machine) string {
+	if machine.State != "started" {
+		return ""
+	}
+	for _, check := range machine.Checks {
+		if check.Name == "role" {
+			if check.Status == "passing" {
```
dangra, line 219: use `api.Passing` instead

dangra (commented):
Looks good

I think we have to limit it to postgres machines tho by checking presence of `api.MachineConfigMetadataKeyFlyManagedPostgres` in machine's metadata

matttpt:
Thanks, I forgot about those constants!

I wondered about limiting it to Postgres machines, but it turns out that for [`ssh console -s`](https://community.fly.io/t/fly-ssh-console-s-now-shows-roles-on-the-machine-selector/13278) and the [monitoring page](https://community.fly.io/t/monitoring-page-role-column/13214) Lubien intentionally left it open to all apps with `role` health checks.

PR #2443, opened 2023-06-13
`migrate-to-v2`: scale based on raw number of allocs, not filtered

alichay:
scale up/down based on the raw number of allocs, rather than the filtered data used for machine spawning that gets deduplicated.


dangra (approved):
looks good, but consider I tried to search for edge cases and red flags but got lost :sorry:

alichay:
> looks good, but consider I tried to search for edge cases and red flags but got lost :sorry:

np, this code is getting a little messy :sob:  ty!

PR #2446, opened 2023-06-14
add processes=app in by default

senyosimpson:
Adds the default process (`app`) to the fresh config that's created  by `fly launch`. This is in an effort to make it more transparent to the user.

### DIFF
--- a/test/preflight/fly_launch_test.go
+++ b/test/preflight/fly_launch_test.go
@@ -43,6 +43,7 @@ func TestFlyLaunch_case01(t *testing.T) {
 			"auto_stop_machines":   true,
 			"auto_start_machines":  true,
 			"min_machines_running": int64(0),
+			"processes":            []string{"app"},
 		},
 	}
 	require.EqualValues(f, want, toml)
--- a/test/preflight/fly_launch_test.go
+++ b/test/preflight/fly_launch_test.go
@@ -43,7 +43,7 @@ func TestFlyLaunch_case01(t *testing.T) {
 			"auto_stop_machines":   true,
 			"auto_start_machines":  true,
 			"min_machines_running": int64(0),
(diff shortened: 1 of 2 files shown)


matttpt (commented):
Looks like the `TestFlyLaunch_case01` preflight test needs a quick update to account for the change:

```diff
Diff:
--- Expected
+++ Actual
@@ -5,3 +5,3 @@
  },
- (string) (len=12) "http_service": (map[string]interface {}) (len=5) {
+ (string) (len=12) "http_service": (map[string]interface {}) (len=6) {
   (string) (len=19) "auto_start_machines": (bool) true,
@@ -10,3 +10,6 @@
   (string) (len=13) "internal_port": (int64) 8080,
-  (string) (len=20) "min_machines_running": (int64) 0
+  (string) (len=20) "min_machines_running": (int64) 0,
+  (string) (len=9) "processes": ([]interface {}) (len=1) {
+   (string) (len=3) "app"
+  }
  },
```

test/preflight/fly_launch_test.go
```diff
@@ -43,6 +43,7 @@ func TestFlyLaunch_case01(t *testing.T) {
 			"auto_stop_machines":   true,
 			"auto_start_machines":  true,
 			"min_machines_running": int64(0),
+			"processes":            append(make([]interface{}, 0), "app"),
```
dangra, line 46: that reads so weird :))   use `[]any{"app"}` instead 🙏

dangra (approved):

test/preflight/fly_launch_test.go
senyosimpson: Ah, I'd assumed this didn't work but just realised I tried `[]interface{"app"}` instead of `[]interface{}{"app"}`. Great suggestion, added it.

PR #2447, opened 2023-06-15
Mention individual tests in Slack

kzys:
Our preflight tests are currently not stable. To avoid the broken window
situation, this commit changes the tests' Slack message to mention
individual failing tests.


dangra (approved):
:shipit:

PR #2452, opened 2023-06-17
Deploy took too long and the oldest leases have expired

rugwirobaker:
```sh
....
  [34/34] Machine e784e22df29d38 [app] update finished: success
  Finished deploying
WARN failed to release lease for machine 91857706c09083: lease not found
WARN failed to release lease: lease not found
WARN failed to release lease for machine 4d891169c09987: lease not found
WARN WARN failed to release lease for machine 5683dd97cd618e: lease not found
WARN failed to release lease: lease not found
WARN failed to release lease: lease not found
failed to release lease for machine 3d8d91e6ce6528: lease not found
WARN failed to release lease for machine 3287004c636185: lease not found
WARN failed to release lease: lease not found
WARN failed to release lease: lease not found
```

### DIFF
--- a/internal/machine/machine_set.go
+++ b/internal/machine/machine_set.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/machine/machine_set.go
```diff
@@ -127,7 +128,7 @@ func (ms *machineSet) ReleaseLeases(ctx context.Context) error {
 	hadError := false
 	for err := range results {
 		contextTimedOutOrCanceled := errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled)
-		if err != nil && (!contextWasAlreadyCanceled || !contextTimedOutOrCanceled) {
+		if err != nil && (!contextWasAlreadyCanceled || !contextTimedOutOrCanceled || !strings.Contains(err.Error(), "lease not found")) {
```
dangra, line 131: Is the goal to ignore the error if it the lease already expired? if so, I think you have to change the condition because it is making it more narrow than it should.

If you apply De Morgan's law it reads:
```go
if err != nil && !(contextWasAlreadyCanceled && contextTimedOutOrCanceled && strings.Contains(err.Error(), "lease not found")) {
```

PR #2453, opened 2023-06-17
`platform vm-sizes`: sort 16x under 8x

alichay:
really solving the important issues here :^)

screenshot of terminal showing 16x sorted under 8x, but no other changes to output


dangra (approved):
🤣

Issue #2250, opened 2023-05-08
Sort output of `fly platform regions` more helpfully


catflydotio:
If we sorted by continent, then country, then actual city name, rather than airport code, it would be easier to choose regions to deploy to or expand into.

redjonzaci:
@catflydotio I wonder if that merged PR solved this issue?

catflydotio:
@redjonzaci If you run `fly platform regions`, you can see that the list is now in alphabetical order by city name. This issue is about making it easier to find a region in the right part of the world.

It's really up to the flyctl team whether they want to close this issue. I don't want to close it out from under them.

redjonzaci:
I understand. I thought you were part of the flyctl team, but sorry for the tags!

PR #2454, opened 2023-06-19
Update sorting on platform regions (#2250)

dobecad:
Updated the `flyctl platform regions` command to sort by city name rather than airport code. 

I believe it's easier to find the region I am looking for when the left-most column is sorted by the city name, rather than the airport code.


kzys (approved):
Thanks! Sounds reasonable to me.

dangra:
I have crossed feelings because I usually use this command output to map airport codes to city names, but I can understand the sequence of thoughts where sorting by name is preferable.

PR #2455, opened 2023-06-19
`deploy`: batch deploys >= 6 into batches of five machines at a time

alichay:
For rolling and canary, this batches machine updates where machines >= 6 into groups of 5. This speeds up deploys a _lot_.
The first update is always solo, so that the most common deploy issues are easy to troubleshoot.

There's also a flag `--strategy-batch=false` that can be used to disable batching.

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -117,8 +117,9 @@ var CommonFlags = flag.Set{
 	flag.Bool{
 		Name: "strategy-batch",
 		Description: fmt.Sprintf(
-			"For \"rolling\" and \"canary\": batch Machine updates into groups of %d when updating more than %d.",
-			BATCHING_GROUP_SIZE, BATCHING_CUTOFF,
+			"For \"rolling\" and \"canary\": batch Machine updates into %d groups when updating more than %d.",
+			batchingGroupCount,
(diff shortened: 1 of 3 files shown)


dangra (approved):
Awesome how simply you solved it! 

When I looked at it, was worried about interleaving console output between different updates from different go routines but doing it in batches makes it very easy to follow. Not a perfect incremental rolling update but much better than current for sure.

I think Nomad is configured to perform a 20% concurrent rolling update, that is why the migration limit was set to 5 allocs. If we can adjust the batch size to 20% of updateEntries the effect will be the same.

alichay:
> Awesome how simply you solved it! 
> 
> When I looked at it, was worried about interleaving console output between different updates from different go routines but doing it in batches makes it very easy to follow. Not a perfect incremental rolling update but much better than current for sure.
> 

I got like three days into trying to get that to work but I eventually just gave up when I realized I was deep in a pit going nowhere. I couldn't get something that didn't clobber over error messages 😔

> I think Nomad is configured to perform a 20% concurrent rolling update, that is why the migration limit was set to 5 allocs. If we can adjust the batch size to 20% of updateEntries the effect will be the same.

ooh that's a great idea. I'll swap to doing that then merge :)

PR #2458, opened 2023-06-20
Feat/appsv2 blue green deployments

gwuah:
This pull request adds blue green deployments for appsV2. 
There's a lot going on, so I broke the feature down into commits to make code review easy.

How it works
- generates top-level checks from service-level checks. 
- create new(green) machines running the new deployment. (these machines are cordoned from the proxy)
- concurrently waits for the green machines to be started
- concurrently waits for green machines to pass health-checks
- marks the green machine as ready for traffic using the un-cordon api
- destroys the blue machines
- profit

Improvements
Currently, we use custom top-level checks  to determine if the new machines are healthy for traffic. This isn't ideal and it requires changes in flaps/flyd to clean up. I've already started working on a solution in https://github.com/superfly/nomad-firecracker/pull/1297

### DIFF
--- a/internal/command/deploy/strategy_bluegreen.go
+++ b/internal/command/deploy/strategy_bluegreen.go
@@ -364,6 +364,9 @@ func (bg *blueGreen) attachCustomTopLevelChecks() {
 					cc.Type = &serviceProtocol
 				}
 
+				if entry.launchInput.Config.Checks == nil {
+					entry.launchInput.Config.Checks = make(map[string]api.MachineCheck)
+				}
 				entry.launchInput.Config.Checks[fmt.Sprintf("bg_deployments_%s", *check.Type)] = cc
 			}
 		}


internal/command/deploy/strategy_bluegreen.go
```diff
@@ -0,0 +1,437 @@
...
+}
+
+func (bg *blueGreen) Rollback(ctx context.Context, err error) error {
+
+	if strings.Contains(err.Error(), ErrDestroyBlueMachines.Error()) {
+		// todo(@gwuah)
+		// if we fail to destroy previous deployment, there's no need to cancel the deployment
+		// we can just show the user how to cleanup (flydev machines destroy --force <machine-id>)
+		// or we can retry?
+		return nil
```
alichay, line 426: Probably want to do this before shipping?

internal/command/deploy/strategy_bluegreen.go
```diff
@@ -0,0 +1,437 @@
...
+		io:              md.io,
+		colorize:        md.colorize,
+		clearLinesAbove: md.logClearLinesAbove,
+		aborted:         atomic.Bool{},
+		healthLock:      sync.RWMutex{},
+		stateLock:       sync.RWMutex{},
+	}
+
+	// Hook into Ctrl+C so that aborting the migration
+	// leaves the app in a stable, unlocked, non-detached state
```
alichay, line 62: nit: copy-paste documentation lol

flaps/flaps.go
```diff
@@ -577,6 +611,8 @@ func (f *Client) NewRequest(ctx context.Context, method, path string, in interfa
 		return nil, err
 	}
 
+	// fmt.Println("[targetEndpoint]]", targetEndpoint)
+
```
alichay, line 615:
```suggestion
```

alichay (commented):
looks good! I didn't totally understand the whole deploy process but I saw most of the highlights of what I was expecting :)

One thing I'm unsure about but is probably just me not understanding the process is that it seems the green machines are uncordoned when they're healthy, but never get cordoned in the first place. I haven't followed the implementation of cordon, is this something that these machines will be by default, by nature of how they're configured?

alichay (approved):
cool! I thought that might be related but didn't want to assume :)

billyb2 (approved):
very nice

PR #2462, opened 2023-06-20
migrate-to-v2: bugfix validate volumes after resolving them

tvdfly:
previously, (at least some) migrations with volumes would not try to migrate the volumes, because the migration setup was not setting usesForkedVolumes. Now, we run the volume validation after resolving volumes to ensure it gets set correctly.


alichay (changes requested):
Good catch! We should probably no longer call `validateVolumes` in `validate` then, right?

tvdfly:
> Good catch! We should probably no longer call `validateVolumes` in `validate` then, right?

I'm not really sure; this change made my migration with volumes work so I didn't investigate further. :-)

alichay:
shouldn't cause any issue to remove that original call then. if this function works when there's no volume state defined, that's a coincidence and shouldn't be depended on imo

alichay (approved):

Issue #2388, opened 2023-06-01
`flyctl launch` misdetects Remix apps as NodeFramework apps


m5r:
**Describe the bug**
Running `npx create-remix@latest` followed by `flyctl launch` in the newly created app's folder does not detect it as a Remix app anymore:

```text
$ flyctl launch
Creating app in /home/mokhtar/app/ddd
Scanning source code
Detected a NodeJS app
? Choose an app name (leave blank to generate one):
```

If I remove `configureNodeFramework` from the [list of scanners on line 93](https://github.com/superfly/flyctl/blob/master/scanner/scanner.go#L93), I get the correct behavior:

```text
$ ../flyctl/out launch
Creating app in /home/mokhtar/app/ddd
Scanning source code
Detected a Remix app
```

I'm running:
* Manjaro 22
* flyctl v0.1.23 linux/amd64 Commit: 4aa1bbf64fc3d6696450dec51ccd0bfd5782454c BuildDate: 2023-05-30T15:32:31Z

jaaneh:
Also seeing this when trying to deploy a Next.js app. Happened so far on both flyctl `v0.1.23` and `v0.1.24`.
```
❯ fly launch    
Creating app in /home/jan/dev/myapp
Scanning source code
Detected a NodeJS app
? Choose an app name (leave blank to generate one): 
```

Adbib:
You can copy Dockerfile, fly.toml, start.sh files from an old project and try to deploy it again.
 It worked for me

hawkcookie:
I have the same problem. cli version is `v0.1.38`.

redjonzaci:
`flyctl` correctly detected a Remix app after running `fly launch` inside the folder created by `npx create-remix@latest`.
@m5r can you confirm it works for you too?

m5r:
Kind of late response but yes, it works for me too. Thanks @dobecad for fixing it!
I'm closing the issue

PR #2472, opened 2023-06-24
Move nodeFramework scanner after node (#2388)

dobecad:
### Change Summary

What and Why:
According to #2388 and local testing, `flyctl` was detecting Remix applications as generic NodeJs applications. This change aims to ensure the scanner package correctly distinguishes a Remix app from a NodeJs app.

How:
Looking at the [scanner function](https://github.com/superfly/flyctl/blob/master/scanner/scanner.go#L86), the `configureNodeFramework` scanner is being run before the `configureNode` scanner. However, the check for Remix occurs in the [configureNode](https://github.com/superfly/flyctl/blob/master/scanner/node.go#L19) scanner. As such, the `configureNodeFramework` scanner should run after the `configureNode` scanner as a fallback in case the `configureNode` scanner does not identify a framework.

Related to:
#2388

Testing:
Can be tested by first creating two separate apps (`npx create-remix@latest` for Remix and `npm init` for NodeJs) and then testing `flyctl launch` on each app.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/scanner/scanner.go
+++ b/scanner/scanner.go
@@ -90,10 +90,9 @@ func Scan(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
 		configurePhoenix,
 		configureRails,
 		configureRedwood,
-		configureNodeFramework,
 		/* frameworks scanners are placed before generic scanners,
-		   since they might mix languages or have a Dockerfile that
-			 doesn't work with Fly */
+		since they might mix languages or have a Dockerfile that
+		doesn't work with Fly */
 		configureDockerfile,
 		configureLucky,
 		configureRuby,
@@ -104,6 +103,7 @@ func Scan(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
 		configureNuxt,
 		configureNextJs,
 		configureNode,
+		configureNodeFramework,
 		configureStatic,
 	}
 


jaaneh:
This would be great. I still cannot deploy Next apps without copying Dockerfile's manually, since it detects as NodeJS.

rubys:
The hope and intent is that the node framework scanner does a better job at producing Dockerfiles for Remix and Next.js apps than the previous templates.  Can I ask what specific problem you are seeing with the framework scanner?

More detail here: https://fly.io/blog/flydotio-heart-js/, but the TL;DR version is that there are so many options (example: pnpm vs yarn vs npm, swapfiles, Microsoft Windows) that apply to all frameworks that is is better to have a single scanner.  See also https://github.com/fly-apps/dockerfile-node#overview

rubys (changes requested):
See comment: https://github.com/superfly/flyctl/pull/2472#issuecomment-161878861

dobecad:
The specific problem I and some of the users in #2388 were experiencing was that Remix and Next.js apps would be identified as NodeJs apps. Since the NodeJs scanner is more generic than the other JS framework scanners, I believe the `configureNodeFramework` scanner should be run last with respect to the other JS framework scanners, such that any of the specific JS framework scanners have a chance at identifying the application, before identifying the application as a NodeJs app.

If the NodeJs scanner is meant to meant to produce better Dockerfiles than the specific JS framework scanners, then is there any point in still using the other scanners? Would it make sense to just optionally prompt the user to enter their JS framework when a generic NodeJs app is detected?

rubys:
> Since the NodeJs scanner is more generic than the other JS framework scanners

Can I stop you right there?  The NodeJS scanner and the NodeJSFramework scanners are two different things.  The NodeJSFramework scanner is more advanced than all of the other NodeJS scanners, and produces a superior Dockerfile, including framework specific tailoring when necessary.

> If the NodeJs scanner is meant to meant to produce better Dockerfiles than the specific JS framework scanners, then is there any point in still using the other scanners?

I haven't tested the NodeJSFramework scanner with Node < 16.

> Would it make sense to just optionally prompt the user to enter their JS framework when a generic NodeJs app is detected?

There is no need to prompt.  Detection is done in two phases.  First the flyctl scanner verifies that minimal requirements are in place, specifically: https://github.com/superfly/flyctl/blob/ca53dd216e7fe025912c58b4cef91a35cf8f071c/scanner/nodeFramework.go#L21-L25

Assuming these needs are met, the [`@flydotio/dockerfile`](https://www.npmjs.com/package/@flydotio/dockerfile) module is installed as a development dependency and then executed.  Having this module be written in javascript to be run under nodejs means that it can:
  * do more advanced feature detection
  * do more advanced tailoring of the generated Docerfiles (ejs is much more capable than go templates)
  * attract [contributions from the community](https://github.com/fly-apps/dockerfile-node/graphs/contributors) who, after all, are more familiar with JS/Node than they are with Go.

Click through some of the [tests](https://github.com/fly-apps/dockerfile-node/tree/main/test) to see examples of the Dockerfiles that are produced given different inputs.

Give it a try!  And if the Dockerfile that is produced for your application doesn't meet your needs, either [open an issue](https://github.com/fly-apps/dockerfile-node/issues) or [start a discussion](https://community.fly.io/).

dobecad:
The part that's confusing me is that if I create a Remix app, I expect `flyctl launch` to say, "Detected a Remix app". Is this expectation wrong? Is it expected behavior for the [srcInfo.Family](https://github.com/superfly/flyctl/blob/master/internal/command/launch/launch.go#L114) to always be `NodeJs` if the application has a `package.json` with a `start` script?

My understanding of the [Scan](https://github.com/superfly/flyctl/blob/master/scanner/scanner.go#L88) function is that it will return the first result found in the `configure*` functions. If the `configureNodeFramework` scanner comes before the `configureNode` or `configureNextJs` scanners, then the app is identified as a `NodeJs` application, rather than `Remix`/`NextJs`/etc. Is this expected behavior of `flyctl`? It seems misleading to me that there are checks in [configureNode](https://github.com/superfly/flyctl/blob/master/scanner/node.go#L42) that set the `srcInfo.Family` to `Remix`, if the `configureNodeFramework` will always set the `srcInfo.Family` to `NodeJs`

rubys:
I would be glad to merge a pull request that changed the text of the output where it says "Detected a ... app", but didn't otherwise change the behavior in terms of the Dockerfile or fly.toml produced. 

I would also be glad to merge a pull request that fixes a specific problem identified with the Dockerfile or fly.toml produced, but other than fixing that specific problem leaves the outputs unchanged.

PR #2473, opened 2023-06-24
add check section to http_service block

gwuah:
### Change Summary
This change adds a check section to the http_service block in fly.toml

What and Why:
Until now, users didn't have a way to define checks for their [http_services](https://fly.io/docs/reference/configuration/#the-http_service-section)

- I've seen people do very weird things (http_service block + check in an almost empty services block. Not even sure if that works.)
- Others also assume it works out of the box. https://community.fly.io/t/healthcheck-based-routing/12997/9?u=kwaw

### Documentation

- [ ] Fresh Produce - I don't think this is fresh produce worthy.
- [ ] In superfly/docs, or asked for help from docs team - I'll try and add this to the docs
- [ ] n/a

### DIFF
--- a/internal/appconfig/definition_test.go
+++ b/internal/appconfig/definition_test.go
@@ -213,6 +213,21 @@ func TestToDefinition(t *testing.T) {
 			"proxy_proto_options": map[string]any{
 				"version": "v2",
 			},
+			"checks": []map[string]any{
+				{
+					"interval":        "1m21s",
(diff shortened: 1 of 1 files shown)


alichay:
good chance you already know this but just in case: I don't think these checks get added to the flattened service definition yet, meaning they're unused

alichay (approved):

Issue #1437, opened 2022-11-03
Can't use quotes in build secrets


mca10:
`fly deploy` command fails when including a build secret containing quotation marks. No amount of escaping seems to get around the issue.

* Operating system: macOS Monterey Version 12.6
* `fly version`: `fly v0.0.420 darwin/amd64 Commit: 977028ba BuildDate: 2022-10-25T14:30:02Z`

The issue happens with or without an existing `fly.toml` file

Command:
`fly deploy --build-secret my_secret='"'`

Output:
```text
Error: invalid argument "my_secret=\"" for "--build-secret" flag: parse error on line 1, column 11: bare " in non-quoted-field
Usage:
  flyctl deploy [WORKING_DIRECTORY] [flags]

Flags:
  -a, --app string             Application name
      --auto-confirm           Will automatically confirm changes without an interactive prompt.
      --build-arg strings      Set of build time variables in the form of NAME=VALUE pairs. Can be specified multiple times.
      --build-only             Build but do not deploy
      --build-secret strings   Set of build secrets of NAME=VALUE pairs. Can be specified multiple times. See https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information
      --build-target string    Set the target build stage to build if the Dockerfile has more than one stage
  -c, --config string          Path to application configuration file
      --detach                 Return immediately instead of monitoring deployment progress
      --dockerfile string      Path to a Dockerfile. Defaults to the Dockerfile in the working directory.
  -e, --env strings            Set of environment variables in the form of NAME=VALUE pairs. Can be specified multiple times.
  -h, --help                   help for deploy
  -i, --image string           The image tag or ID to deploy
      --image-label string     Image label to use when tagging and pushing to the fly registry. Defaults to "deployment-{timestamp}".
      --local-only             Only perform builds locally using the local docker daemon
      --nixpacks               Deploy using nixpacks to generate the image
      --no-cache               Do not use the build cache when building the image
      --now                    Deploy now without confirmation
      --push                   Push image to registry after build is complete
  -r, --region string          The region to operate on
      --remote-only            Perform builds on a remote builder instance instead of using the local docker daemon
      --strategy string        The strategy for replacing running instances. Options are canary, rolling, bluegreen, or immediate. Default is canary, or rolling when max-per-region is set.

Global Flags:
  -t, --access-token string   Fly API Access Token
  -j, --json                  json output
      --verbose               verbose output

[...]

PR #2476, opened 2023-06-26
Remove surrounding double quotes from secrets

benrogmans:
### Change Summary

What and Why:
If secrets have surrounding double quotes, remove them before setting the secret. Environment variables may contain prefixed and suffixed double quotes to escape special characters. When including the double quotes in secrets, they lead to problems when using them.

How:
If the string starts with " or ends with ", the double quotes are removed. This should respect escaped double quotes as the first character then no longer is a double quote but a backslash.

With regards to line 40: The backtick syntax in Go is a raw string literal and it can contain arbitrary characters, including newlines, backslashes, and even double quotes. There is no need to escape them.

Related to:
This PR addresses issues #589 and #1437.

Please note that this is my first PR in this codebase and that I have not been able to test the below code. I just noticed the issue is open for a long time already and I am personally in need of a fix 😄 Any feedback is appreciated, thanks for considering my PR.

### DIFF
--- a/internal/command/secrets/parser.go
+++ b/internal/command/secrets/parser.go
@@ -37,10 +37,15 @@ func parseSecrets(reader io.Reader) (map[string]string, error) {
 				// Switch to multiline
 				parserState = parserStateMultiline
 				parsedKey = parts[0]
-				parsedVal.WriteString(strings.TrimPrefix(parts[1], "\"\"\""))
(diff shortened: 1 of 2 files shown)


dangra:
hi @benrogmans, thanks for working on a fix. 

Could you also add a test case to https://github.com/superfly/flyctl/blob/master/internal/command/secrets/parser_test.go ?

dangra:
thanks!

PR #2481, opened 2023-06-27
infer region if attach-volume specified in clone

senyosimpson:
### Change Summary

What and Why:
When running the command `fly m clone <machine> --attach-volume=<vol id>:<mount path>` and the region flag is not specified, the region is not inferred from the volume. This causes failures if you clone a machine in a different region from the volume you are trying to attach. This remedies that issue.

How:

The logic is changed to follow these rules:

* If a volume is specified and the region is specified, check if they are in the same region. If not, fail
* If a volume is specified and the region is not specified, infer the region from the volume
* If only the region is specified, use it
* If none of the above, infer from the region of the machine

Related to:
n/a
---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -103,8 +103,25 @@ func runMachineClone(ctx context.Context) (err error) {
 	}
 	flapsClient := flaps.FromContext(ctx)
 
+	var vol *api.Volume
+	if volumeInfo := flag.GetString(ctx, "attach-volume"); volumeInfo != "" {
+		splitVolumeInfo := strings.Split(volumeInfo, ":")
+		volID := splitVolumeInfo[0]
+
+		vol, err = client.GetVolume(ctx, volID)
+		if err != nil {
+			return fmt.Errorf("could not get existing volume: %w", err)
+		}
+	}
+
 	region := flag.GetString(ctx, "region")
-	if region == "" {
+	if vol != nil && region != "" {
+		if vol.Region != region {
+			return fmt.Errorf("specified region %s but volume is in region %s, use the same region as the volume", region, vol.Region)
+		}
+	} else if vol != nil && region == "" {
+		region = vol.Region
+	} else {
 		region = source.Region
 	}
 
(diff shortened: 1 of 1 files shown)


internal/command/machine/clone.go
```diff
@@ -103,8 +103,25 @@ func runMachineClone(ctx context.Context) (err error) {
 	}
 	flapsClient := flaps.FromContext(ctx)
 
+	var vol *api.Volume
+	if volumeInfo := flag.GetString(ctx, "attach-volume"); volumeInfo != "" {
```
senyosimpson, line 107: A downside of this implementation is that the logic is duplicated further down (see below). I thought about storing the results in variables and using them further below but the logic started to get kind of messy. Still trying to see if I can get it right but if this is fine as is, happy to leave it.

https://github.com/superfly/flyctl/blob/7c5e9480d437495e37a3f1c572ec2b84614cdae9/internal/command/machine/clone.go#L177-L214

internal/command/machine/clone.go
```diff
@@ -103,8 +103,25 @@ func runMachineClone(ctx context.Context) (err error) {
...
-	if region == "" {
+	if vol != nil && region != "" {
+		if vol.Region != region {
+			return fmt.Errorf("specified region %s but volume is in region %s, use the same region as the volume", region, vol.Region)
+		}
+	} else if vol != nil && region == "" {
+		region = vol.Region
+	} else {
 		region = source.Region
 	}
```
alichay, line 126: Looks like it might use the source machine's region when there's no volume and a region flag *was* specified?
```suggestion
	} else if region == "" {
		region = source.Region
	}
```

internal/command/machine/clone.go
alichay: duplicated code is definitely gross but I think it's probably fine here. plumbing everything through would probably end up reading less clearly.

alichay (changes requested):
I think there's a subtle bug in region selection. Aside from that, looks good!

internal/command/machine/clone.go
senyosimpson: oh yep, great catch

alichay (approved):
awesome!

PR #2484, opened 2023-06-27
Refactor version checks to work with both Semver and Calver-compatible versions

ndarilek:
### Change Summary

What and Why:

In an attempt to encourage users to upgrade, we thought it might make sense to switch versions from Semver to Calver.

* Semver makes explicit guarantees about compatibility and breakage that we're not technically honoring.
* Even if we make it explicit that we're not using Semver in that way, as a developer-oriented product, end users seem likely to assume that we are based on the version format alone.
* As a rapidly-evolving platform, it makes sense that our versions should carry some meaning. In the absence of any better ideas, Calver seems like a safe bet.

How:

* Create a generic `Version` interface encapsulating the ways we interact with flyctl versions, implemented for both Semver and calver.
* Implement methods for parsing each format, including a generic parser that attempts a Calver parse but falls back to Semver on failure.
* Implement version comparison functions that sort and compare normally across versions of the same type, but which declare all Semver versions to be less than Calver versions.

With this in place, the hope is that we can transition over to Calver and drop Semver support once enough users have upgraded.

---

### Documentation

- [ ] Fresh Produce
- [ ] Eventually, when closer to release


internal/buildinfo/version.go
```diff
@@ -8,17 +8,126 @@ import (
...
 
 	"github.com/blang/semver"
+	"github.com/loadsmart/calver-go/calver"
 	"github.com/superfly/flyctl/terminal"
 )
 
+type Version interface {
+	String() string
+	EQ(other Version) bool
+	Outdated(string) (bool, error)
```
kzys, line 18: It is a bit odd to see that Outdated takes two versions. I was expecting something like below.

```
v, _ := buildinfo.ParseSemver("0.0.1")
outdated, _ := v.Outdated() // returns true
```

How about calling this method LT (like semver) or Before (like time.Time)?

Also how about taking Version instead of string, to be more like EQ()?

internal/buildinfo/version.go
```diff
@@ -8,17 +8,126 @@ import (
...
+		return false
+	}
+}
+
+func (v *SemverVersion) Outdated(other string) (bool, error) {
+	otherParsed, err := ParseSemver(other)
+	if err != nil {
+		return false, err
+	}
+	return v.LT(otherParsed.Version), nil
```
alichay, line 47: Perhaps in SemverVersion's `Outdated` we should try to parse `other` as calver first, and on success unconditionally return true?

internal/buildinfo/version.go
```diff
@@ -8,17 +8,126 @@ import (
...
+		} else {
+			v2 := s[j].(*SemverVersion)
+			return v1.Version.LT(v2.Version)
+		}
+	}
+}
+
+func (s Versions) Swap(i, j int) {
+    s[i], s[j] = s[j], s[i]
+}
```
alichay, line 121: CI is complaining about formatting on `version.go`. It suggests running `gofmt -s`?

internal/buildinfo/version.go
```diff
@@ -105,15 +212,3 @@ func parseVesion(v string) semver.Version {
 	}
 	return parsedV
 }
```
alichay, line 214: CI's complaining about `parseVesion` being unused. Seems like everything that uses it got yanked, so it should be safe to remove this too.

alichay (commented):
aside from what kaz said, I have one other question. I also copied over the lint issues from CI because those are a pain to get to

this is awesome, great work!!

internal/buildinfo/version_test.go
```diff
@@ -25,5 +25,5 @@ func TestDevMeta(t *testing.T) {
 
 	loadMeta()
 
-	assert.Equal(t, fmt.Sprintf("0.0.0-%d+dev", BuildDate().Unix()), Version().String())
+	assert.Equal(t, fmt.Sprintf("0.0.0-%d+dev", BuildDate().Unix()), ParsedVersion().String())
```
alichay, line 28: This is going to start expecting a calver, based on the time of the test. not sure the best way to make this easily testable...

agent/start.go
```diff
@@ -45,10 +45,8 @@ func StartDaemon(ctx context.Context) (*Client, error) {
 	env := os.Environ()
 	env = append(env, "FLY_NO_UPDATE_CHECK=1")
 
-	versionPre := buildinfo.Version().Pre
-
-	if len(versionPre) > 0 {
-		versionNum := versionPre[0].VersionNum
+	if buildinfo.IsDev() {
+		versionNum := buildinfo.ParsedVersion()
 		env = append(env, fmt.Sprintf("FLY_DEV_VERSION_NUM=%d", versionNum))
```
alichay, line 50: I could be misreading this but I think `versionNum` here would be of type `Version`, right?
It's being formatted as `%d`, though - do we want a function that extracts some sort of "build number"-adjacent integer for this case?

internal/buildinfo/version.go
alichay: IMO it makes sense to operate on two versions, but I agree that it could be called LT or Before just to match convention

agent/start.go
ndarilek: I wasn't really clear on what problem this code was solving. Clearly it's injecting something in the environment for pre-release builds but I wasn't sure why it did that.

Unfortunately calver-go doesn't expose the micro/build number. We can get a time associated with the components of the year/month/day but that won't take the micro component into account.

Exposing this value upstream would of course be very easy, but I wanted to be clear on what this code was doing before doing that. For this purpose it may be fine to just use a unix timestamp, though of course that'd change on every build so may not be desirable.

internal/buildinfo/version_test.go
ndarilek: Going to mark this as resolved since (I think) I modified the code to prefer the version var if it parses, and that seems to be set in the testcase. Could be that gremlin rules apply and we shouldn't run this test straddling midnight, but we can cross that bridge if we get there and it actually exists. :)

ndarilek:
Marking this as ready for review since it's further along than a draft, though I'm still unclear about what we're doing with versions re: the Wireguard agent as per my previous comment. Think that might be the final hangup.

agent/start.go
ndarilek: Maybe. I'm still unclear what purpose putting the version in the environment is supposed to serve. I'm guessing it has something to do with the agent checking whether there's a mismatch, but the variable only appears in 2 places--here and `buildinfo`. It also seems like it may not be intended to contain the entire version, but I'm not sure.

Maybe I'm over-complicating things, but with calver we can fairly easily generate a fake but sensible version if one doesn't exist (I.e. for dev builds.) We can also fairly easily detect that a given generated version is newer/older than whatever binary may have started here. I'm just not clear on what the problem we're solving here is. Maybe the linked PR gives enough context, but not quite for me as a n00b. :)

CC @billyb2

alichay: yeah I don't really follow this bit of code either. my instinct is to either yank most of this out and simplify it, or just ship what's here and hope for the best. either of those are honestly fine choices.

only thing that gives me pause is the thing I mentioned above: `fmt.Sprintf("FLY_DEV_VERSION_NUM=%d", versionNum)` is using the `%d` formatter to print a Version, which seems wrong, right?

ndarilek: You're almost certainly right--I don't know what this specific format string does and I likely copy-pasted something wrong. Fixing that is on my list of things to do, but I too am inclined to yank this code given that I can't find any reference to the environment variable that explains what's going on. And if we yank this code then we're yanking the format string too.

The initial code just dumped a timestamp in that variable. I wanted to be clear on what exactly was going there before I dove into figuring out which data and format string would be best.

ndarilek:
I haven't tried making a dev build. I haven't changed anything in the release process yet.

Haven't really messed with the makefile until today when I discovered `make test` worked, so I'll dig into that a bit more tomorrow. Thanks for pointing me to it.

agent/start.go
billyb2: This code is pretty horrendous looking back :/.

From what I remember, and from reading [an old Discourse post](https://flyio.discourse.team/t/launching-proxy-fails-on-dev-version/1802), this fixes a bug where the fly agent would be different then the version that flyctl used. The fix for this was to set the version number as an environment variable, that fly agent uses. 

The micro component, in this case, is used since fly version numbers look like v0.0.0-1398431841. I think as long as the FLY_DEV_VERSION_NUMBER is a uint, it should be fine. This code might also no longer be relevant, but I'm not sure what changes have been made to flyctl since it was introduced.

(Also wow, I'm getting nostalgia for code I wrote 4 months ago)

billyb2: > yeah I don't really follow this bit of code either. my instinct is to either yank most of this out and simplify it, or just ship what's here and hope for the best. either of those are honestly fine choices, but the first is a bit more work for little reward.
> 
> only thing that gives me pause is the thing I mentioned above: `fmt.Sprintf("FLY_DEV_VERSION_NUM=%d", versionNum)` is using the `%d` formatter to print a Version, which seems wrong, right? Should it be grabbing a unix timestamp from the version, or something?

As for this specifically, this code is correct, it's just a terrible variable name. I'm getting the Version's .Pre member, and then just calling that version number, since that was value that mattered for dev versions at the time

ndarilek: OK, thanks for the context. I went ahead and removed the environment variable code for now, particularly since I probably wasn't doing the right thing with it in version.go anyway. If we need it back we can re-add it.

Right now my expectation is that releases will have versions like 2023.07.01.1 while dev versions will have versions like 2023.07.01.1234567890 (E.g. followed by the build/run timestamp.) So they'll always be different and we shouldn't need to worry about outdated agents.

Thanks again.

ndarilek:
Thoughts on merging this as is and working out the release process later? There are a few related problems we need to solve and I don't know the best way forward:

1. We need a way to calculate micro version components in the Makefile. I'm not sure where `make build` is used, so it may be sufficient to just hard-code this to 0 or 1 and assume the Makefile is only being used for local test builds.
2. We may or may not need a way to calculate the next released version number. Not sure if we're manually tagging things right now, but it may be sufficient to do that, and should be less error-prone. It may be fine to just have a script try adding tags like v2023.07.01.1, v2023.07.01.2, etc, bumping the micro version until the tag applies.
3. GoRleaser only works with semver versions. We can probably work around this by using v2023.07.01-1 as the tag format, though I'm not sure if it'd also strip out the 0-padding in the version normalization. But I'm anticipating a bit of struggle making that tooling work, and we may also need to patch calver-go upstream to allow for different separators between the date and micro version components.

If we'd like to solve those as part of this PR then I'm fine with that too--just thought it might be worth splitting the problem in case someone else has ideas on how to solve the remaining issues separately.

alichay:
> Thoughts on merging this as is and working out the release process later?

making a prerelease off of this branch to see how it behaves with the release process, and to make sure it won't impact updating or how versions are *currently* parsed. There's also one issue, below:

> * We need a way to calculate micro version components in the Makefile. I'm not sure where `make build` is used, so it may be sufficient to just hard-code this to 0 or 1 and assume the Makefile is only being used for local test builds.

yeah - makefile is not used for production builds, its versions have always been nonsense semver with a commit hash appended to the end. Not too worried about this, it's not used in prod, we can figure out something better later if it crops up.

> * We may or may not need a way to calculate the next released version number.

I believe all the versioning logic comes from `bump_version.sh` and its sibling script `semver`. We're totally OK to skip this for now. Might be nice to just implement this as a tiny go executable so that we can implement version parsing in a real programming language instead of a shell script :skull:

> 3\. GoRleaser only works with semver versions. We can probably work around this by using v2023.07.01-1 as the tag format, though I'm not sure if it'd also strip out the 0-padding in the version normalization. But I'm anticipating a bit of struggle making that tooling work, and we may also need to patch calver-go upstream to allow for different separators between the date and micro version components.

Not to go all Not Invented Here, but if we forked `calver-go` I imagine we could drastically simplify their code as well if we only support one calver scheme.

This, unfortunately, *is* a blocker. Not too concerned with 0 padding, we should be able to cross that bridge when we get to it, but we need to support the dash as part of the pattern.
Alternatively, if we wanted to be hackish, we could have the parse function replace dashes with `.` before sending it to `calver-go` and have the String implementation for the wrapper type substitute the fourth dot to a dash. This is a bit ugly, but I wanted to provide another option.

kzys:
> This, unfortunately, _is_ a blocker. Not too concerned with 0 padding, we should be able to cross that bridge when we get to it, but we need to support the dash as part of the pattern.

Is the semver incompatibility coming from having four-dotted numbers (`YYYY.0M.0D.MICRO`)? If we use `YYYY.0M.0D-MICRO` or `YYYY.0M.MICRO`, would it work?

ndarilek:
I tried integrating this with the new auto-update code on master. Think I got it, but it's a Monday after time off so please re-review since I'm almost sure I missed something.

Right now the calver `SeverelyOutdated` function is a stub until we have enough calver versions and can decide what that looks like. I also had to fork calver-go to add getters for the various version components. Not sure if we have any policies for working off of forks, but this may need an update if my PR there gets merged.

Thanks.

ndarilek:
Is there a way to run these lints locally? Not sure why the CI output of this particular action is so hard to get at accessibly.

alichay:
I think CI is running `golangci-lint run`.

ndarilek:
Just got that working. Thanks, that helps a lot!

ndarilek:
I want to get semver compatibility done today so I can put this to rest for a bit. In the interest of not having to maintain a full-on fork or gut the current library, I'm trying to figure out the minimal changes that need to be made.

As a refresher, the blocker is that goreleaser only accepts semver-compatible versions, and I don't want to redo our release tooling for this. We've decided that the format we want is 0YYYY.0M.0D-MICRO.

The hackiest solution is probably modifying `*Semver) ParseVersion(string)` to just replace `-` with `.` and pass it to the existing parsing code. The disadvantage, of course, is that we'd be generating releases and artifacts of the form 2023.07.05-1 but reporting the version everywhere else as 2023.07.05.1. This doesn't seem like a huge inconsistency to me but I wanted to check that assumption before charging ahead with it.

There are of course more invasive solutions to this problem, but if the easiest one works fine then why not use that?

Thanks a bunch.

ndarilek:
When an initial calver parse attempt fails, I replace a single `-` with a `.` and try again.

I've also added in a couple tests to cover what my sleep-addled brain is calling the dedashing hack--making sure it works, and making sure it won't accidentally parse semver versions as calver.

I think this should be ready and backwards-compatible with our current scheme. If anyone could glance over my resolution to the conflicts introduced by auto-updating and make sure that logic is till sound, I'd appreciate it. Then, if anyone can point me to how to cut a pre-release, we can try that and make sure our current process doesn't break.

Thanks.

internal/buildinfo/version.go
```diff
@@ -2,23 +2,164 @@ package buildinfo
...
+		}
+		if v.Minor < otherVer.Minor {
+			return true
+		}
+		if v.Patch+5 < otherVer.Patch {
+			return true
+		}
+		return false
+	} else {
+		return false
```
alichay, line 65: If we return true when V is older than otherVer, shouldn't we return true if other is not semver?

That would imply V is semver but other is calver

internal/buildinfo/version.go
```diff
@@ -2,23 +2,164 @@ package buildinfo
...
+	_, ok := parsedVersion.(*CalverVersion)
+	if ok {
+		return false
+	} else {
+		other := parsedVersion.(*SemverVersion)
+		return v.Version.GT(other.Version)
+	}
+}
+
+func (v *SemverVersion) SeverelyOutdated(other Version) bool {
```
alichay, line 51: nit: for clarity, we could call `other` `latest` and `otherVer` `latestVer`?

internal/buildinfo/version.go
```diff
@@ -2,23 +2,164 @@ package buildinfo
...
+		// compatibility for goreleaser.
+		dedashedVersion := strings.Replace(version, "-", ".", 1)
+		v, err = calver.Parse(calverFormat, dedashedVersion)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return &CalverVersion{*v}, nil
+}
+
```
alichay, line 88: We could add a String method here that adds the dashes back, so we can keep our formatting consistent.

This *should* work?
```suggestion

func (v CalverVersion) String() string {
	verStr := []rune(v.Version.String())
	count := 0
	for i, v := range verStr {
		if v == '.' {
			count++
		}
		if count == 3 {
			verStr[i] = '-'
			break
		}
	}
	return string(verStr)
}

```

internal/buildinfo/version.go
```diff
@@ -2,23 +2,164 @@ package buildinfo
...
+		return v.Version.CompareTo(&other.Version) == 1
+	} else {
+		return true
+	}
+}
+
+func (v *CalverVersion) SeverelyOutdated(other Version) bool {
+	// todo: Figure out how to do this when we actually have outdated calver
+	// versions.
+	return false
```
alichay, line 110: I really wish calver-go had an API for getting a time.Time from a version. That's really what this needs...

ndarilek: [My fork](https://github.com/ndarilek/calver-go) does. I've PR'd the changes--if they aren't merged upstream, I'm happy to move that over to superfly.

internal/buildinfo/version.go
ndarilek: Almost certainly, my brain was coffee-deprived when I wrote that. I'll check it over in a bit.

internal/buildinfo/version.go
alichay: In that case, why don't we just start with checking if the version is a month or two old? We can always adjust if needed later :)

alichay:
Once SeverelyOutdated gets implemented and the tests are fixed (the string formatting change broke the tests lol), I think that'd be a great time to test everything out. I had no idea this was such a rabbithole

internal/buildinfo/version.go
ndarilek: Sure thing. Is there a more idiomatic way of doing date comparisons in golang other than checking if the timestamps are > something like 86400 * 30 seconds apart? That was my initial inclination but figured I'd save a round trip if there's a better way.

ndarilek:
Heh, no kidding, figured this would be a lot easier than it turned out to be.

I've implemented `SeverelyOutdated` for calver. Going to cut a pre-release after lunch. Can I do that on a branch or does it need to be merged?

ndarilek:
Ah, has to be on the prerelease branch. Can I just merge into that, or do I need approval?

alichay:
Just do it on this branch. It'll yell at you about it but that branch is kind of broken right now - it's fine.

internal/buildinfo/version.go
alichay: I think that's a perfectly fine way to implement that :)

alichay:
> v0.1.57-pre-1 was cut off this branch and correctly reports its version. My assumption is that the code correctly handles backwards-compatibility

🎉 horray!

> Since I think we're waiting a bit on this, I'd like to merge this just to close the branch, then maybe pick up work on B and collaborate on someone else with A in maybe a month or so, or whenever we're ready to finish the transition. Thoughts?

Yeah, sounds great! you definitely deserve a break from working on version code haha

ndarilek:
ugh, this just doesn't end. :(

This is the code where we removed the environment variable stuff a while back. My impression was that folks weren't entirely clear what was happening here, though maybe I misread.

Do we know what's getting a bad version? If I run `fly agent run` in one terminal and `fly agent ping` in another, everything seems to be fine. I wish I understood what's causing this specific error and why, since I'm fairly sure we're handling the case of an outdated agent.

Is it time to just declare this a failed experiment and move on? Seems like it's close, but there always seems like a new wrinkle, and now that wrinkle is in a bit of code there previously doesn't seem to have been a whole lot of clarity on.

Thanks.

alichay:
Ok. The agent bug is fixed, and the migrate-to-v2 issue is a breaking change to web. I'm going to try to sort that out.

I think this is almost done! That said, if you want to hand this off to someone else that's totally understandable haha

ndarilek:
I don't mind continuing with it. It just seems like way more than I expected it to be, and given the emphasis here on quick and short tasks, I'm a bit worried that spending so much time on this in my first month or so might look bad for me. I seriously thought this would just be a few days of work at most.

That said, what needs to be done to merge this onto master? I think it probably makes sense to carve off integration into the release process for a bit later so we can make sure flaps and everything else isn't going to choke on calver for some reason. Do we need another pre-release? Happy to cut one if so.

alichay:
> given the emphasis here on quick and short tasks, I'm a bit worried that spending so much time on this in my first month or so might look bad for me.
Totally understandable! It's fine, this stuff happens, but I totally get where you're coming from
> I seriously thought this would just be a few days of work at most.
Ugh me too.

> I think it probably makes sense to carve off integration into the release process for a bit later so we can make sure flaps and everything else isn't going to choke on calver for some reason.
Definitely. I'd like for flyctl to be able to parse calver versions for a while before we start worrying about using it, that way we can display pretty version numbers when it's time to update to a calver version :)

> Do we need another pre-release? Happy to cut one if so.
I think you should rebase off master first, just so that CI reruns and we can make sure that all the problems we were seeing were related to `web`. Once that's done, if it's all green, I'd just say it's time to approve and merge

ndarilek:
Are these the web-related problems? Haven't checked over any of the usual preflight failures we've been hitting so I'm not sure if these are those. :)

alichay:
> Are these the web-related problems? Haven't checked over any of the usual preflight failures we've been hitting so I'm not sure if these are those. :)

Yep! I thought i'd already fixed it, but the problem was more tricky than I thought it'd be (can you please rebase again)

ndarilek:
All checks passed, I assume we're good to merge this?

alichay (approved):
you definitely earned this approval haha

PR #2492, opened 2023-06-27
Automatic updating

alichay:
### Change Summary

What and Why: Adds autoupdating to flyctl. Staying up-to-date is now much more important than it used to be, so this not only provides a gentle nudge towards keeping things up-to-date, but it makes it potentially more convenient.

There is a config setting to disable autoupdating, and it's automatically disabled in CI environments.

How:

* If the version cache is out-of-date (more than a week old), the latest version info will be synchronously downloaded. If it's not more than a week old, we'll retain previous behavior and download async.
  * The benefit here is that an installation of flyctl that's lain dormant for months should be able to be run, and in that single run determine that it's out of date, update, _then_ run the user's command.

* If the CLI is "severely" out-of-date (defined currently as `0.0.5` versions behind), it'll download the new version, then re-run the command against the newer version.

* If the CLI is out of date, but not severely, it will either:
   * For linux/macOS, update flyctl in the background while running the original command.
      * This is seamless and provides wonderful UX.
   * For Windows users, prompt to update. I couldn't get background updates seamless on Windows for various reasons.
      * Between prompting and autoupdating when severely out of date, this is _fine_. Still an improvement over the status quo!

Related to:
 1. Project board (internal): https://github.com/orgs/superfly/projects/59/views/1
 2. Community announcement: [Flyctl Versions, Autoupdating, and the CLI Apocalypse](https://community.fly.io/t/flyctl-versions-autoupdating-and-the-cli-apocalypse/13794)

---

### Documentation

- [X] [Fresh Produce](https://community.fly.io/t/flyctl-versions-autoupdating-and-the-cli-apocalypse/13794)
- [ ] n/a

---

### TODO 
 - [x] Test that this works. I'm going to be using prereleases to test this.
 - [x] Tone-down the updates so that it's every couple versions or something - non-stop updating would get frustrating real fast.
 - [x] Check for updates synchronously if state.yml doesn't exist or if its modification date is > one week. This ensures that people spinning up dusty copies of flyctl will always get updates.

### DIFF
--- a/internal/command/settings/autoupdate.go
+++ b/internal/command/settings/autoupdate.go
@@ -13,7 +13,7 @@ import (
 )
 
 func newAutoUpdate() *cobra.Command {
-	metricsRoot := command.New("autoupdate", "Control automatic updates", "", runAutoupdateStatus)
+	autoupdateRoot := command.New("autoupdate", "Control automatic updates", "", runAutoupdateStatus)
 
 	optIn := command.New("enable", "Enable automatic updates", "", func(ctx context.Context) error {
 		return setAutoupdateEnabled(ctx, true)
@@ -22,10 +22,10 @@ func newAutoUpdate() *cobra.Command {
 		return setAutoupdateEnabled(ctx, false)
 	})
 
-	metricsRoot.AddCommand(optIn)
-	metricsRoot.AddCommand(optOut)
+	autoupdateRoot.AddCommand(optIn)
+	autoupdateRoot.AddCommand(optOut)
 
-	return metricsRoot
+	return autoupdateRoot
 }
 
 func printAutoupdateEnabled(ctx context.Context, enabled bool) {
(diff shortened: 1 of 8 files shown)


internal/command/settings/autoupdate.go
```diff
@@ -0,0 +1,63 @@
...
+	fmt.Fprintf(io.Out, "Automatic updating: %s\n", enabledStr)
+}
+
+func runAutoupdateStatus(ctx context.Context) error {
+	var (
+		cfg = config.FromContext(ctx)
+		io  = iostreams.FromContext(ctx)
+	)
+
+	printAutoupdateEnabled(ctx, cfg.SendMetrics)
```
dangra, line 45: looks like a bug, should be `AutoUpdate`, isn't it?

internal/command/settings/autoupdate.go
```diff
@@ -0,0 +1,63 @@
...
+
+	fmt.Fprintf(io.Out, "\nThis can be controlled with 'fly settings autoupdate <enable/disable>'\n")
+
+	return nil
+}
+
+func setAutoupdateEnabled(ctx context.Context, enabled bool) error {
+	path := state.ConfigFile(ctx)
+
+	if err := config.SetSendMetrics(path, enabled); err != nil {
```
dangra, line 55: same issue here

dangra (approved):
Looks good! only a small bug that should be fixed before merging.

Issue #2497, opened 2023-06-28
`fly redis connect` should receive `redis-cli` path


chiptus:
currently, when running `fly redis connect` without having redis-cli installed in $PATH results in `Could not find redis-cli in your $PATH. Install it or point your redis-cli at: someurlError: exec: "redis-cli": executable file not found in $PATH`

It would be nice if I could do something like `fly redis connect --cli-path /path/to/redis-cli` and fly would use that instance of redis-cli. for example when I use the npm package `redis-cli` it installs a binary called `rdcli` which should have the same features like redis-cli without installing redis.

redjonzaci:
@dangra can we close this since the PR was merged? Whenever you find the time

PR #2498, opened 2023-06-28
feat(redis): add proxy command

chiptus:
### Change Summary

What and Why:
Adds a command to redis `fly redis proxy`
Sometimes `redis connect` doesn't work because redis-cli doesn't exist, but we want to proxy the redis db to test it locally.

How:
adds a new file in the redis package that manages creating the proxy.
also moved the logic of creating the proxy params to `getProxyParams` that is shared between `fly redis connect` and the new command

Related to: #2497 

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/redis/connect.go
+++ b/internal/command/redis/connect.go
@@ -7,15 +7,11 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/superfly/flyctl/agent"
-	"github.com/superfly/flyctl/gql"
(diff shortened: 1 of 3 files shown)


dangra:
@jsierles this seems like a sensible thing but as you are close to extensions, would like to get your opinion on it.

jsierles (approved):
LGTM

dangra:
thanks @chiptus !

PR #2510, opened 2023-06-30
Use the right strategy when setting secrets

gwuah:


internal/command/secrets/secrets.go
```diff
@@ -113,6 +113,7 @@ func v2deploySecrets(ctx context.Context, app *api.AppCompact, release *api.Rele
 		AppCompact:       app,
 		RestartOnly:      true,
 		SkipHealthChecks: detach,
+		Strategy:         cfg.Deploy.Strategy,
```
rugwirobaker, line 116: The strategy in cfg is probably not set so we're getting a panic

PR #2515, opened 2023-06-30
Prompt users about provisioning extensions in high latency regions

jsierles:
Extensions like PlanetScale are deployed on AWS, and only some Fly regions have low latency connections to them. This PR ensures that if a latency-sensitive extension is deployed in a region with high latency to the nearest supported external region, the user gets prompted on whether they'd like to continue with a potentially high latency instance.

### DIFF
--- a/internal/command/extensions/core/core.go
+++ b/internal/command/extensions/core/core.go
@@ -13,6 +13,7 @@ import (
 	"github.com/superfly/flyctl/internal/flag"
 	"github.com/superfly/flyctl/internal/prompt"
 	"github.com/superfly/flyctl/iostreams"
+	"golang.org/x/exp/slices"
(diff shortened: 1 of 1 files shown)


fideloper:
LGTM. 

Sanity check: It makes sense to go by the primary region rather than find all the regions the app is deployed into, right?  The primary region is likely the one that'll accept write requests in a multi-region deployment.

Is it possible on older versions of fly.toml that no primary region is set?

fideloper (approved):

jsierles:
Primary region needs to be set on newer apps. If someone tries this on a nomad app, they'll just have to pick a region, which seems fine to me.

fideloper:
Right on!

PR #2519, opened 2023-07-02
Inject flyctl version into machine metadata

gwuah:
The endgoal is to allow us quickly determine when machines in an app have been created with different flyctl versions -> different configuration settings -> different platform behaviour. 

I don't think they happen frequently, but when they do, they're a PAIN to resolve. ask @senyosimpson


billyb2 (approved):
👍🏾

dangra:
We could test the metadata key by comparing its value to the buildinfo version instead o deleting it. Just a late nitpick

PR #2523, opened 2023-07-03
Extract ephemeral machine code from `fly console`; use in `fly pg import`

matttpt:
### Change Summary

This moves `fly console`'s ephemeral machine creation code into the `internal/machine` package to make it reusable, and switches `fly pg import` over to using it.

The immediate benefit is that `fly pg import` gets the same cleanup features as `fly console`. The ephemeral machine will auto-destroy when stopped, and `flyctl` will attempt to clean up the ephemeral machine using a new context (with a short timeout) even when an error occurs.

Long-term, this will make it easier to add other commands that need ephemeral machines. E.g. a command for running third-party images to set up databases could make use of this (as in #2350).

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/console/console.go
+++ b/internal/command/console/console.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/http"
-	"time"
 
 	"github.com/samber/lo"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


dangra:
@matttpt I don't see any preflight for `fly pg import` nor `fly console`. Do you think it is worth adding one before merging?

dangra (approved):

PR #2526, opened 2023-07-04
Handle SIGUSR2 and dump goroutines on the signal

kzys:
When flyctl hangs, dumping all goroutines may be helpful to understand its internal state.

Docker uses SIGUSR1 for that, but it is already taken by wg/signals_unix.go.

https://community.fly.io/t/flyctl-v0-1-47-cannot-deploy-clone-and-list-machines/13940 https://github.com/moby/moby/blob/v24.0.2/daemon/debugtrap_unix.go#L17

### Change Summary

What and Why:

flyctl v0.1.47 was hanging occasionally and we had no clues. This change may help us understand what's going on.

How:

Handling SIGUSR2 and dumping all goroutines' stacktraces.

Related to:

https://community.fly.io/t/flyctl-v0-1-47-cannot-deploy-clone-and-list-machines/13940

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/debug_unix.go
+++ b/debug_unix.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"runtime/pprof"
+
+	"golang.org/x/sys/unix"
+)
+
+// handleDebugSignal handles SIGUSR2 and dumps debug information.
+func handleDebugSignal(ctx context.Context) {
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, unix.SIGUSR2)
+
+	for {
+		select {
+		case <-sigCh:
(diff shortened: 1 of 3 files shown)


pborzenkov:
There is a nice package called `gops` that integrates a lot of diagnostics - https://github.com/google/gops
Not only goroutines dump, but also profiler, memory stats, etc.
Maybe worth to consider it?

kzys:
Having gops on our own hosts (e.g. backend servers) may be nice. For customer hosts (e.g. their laptops), I don't want to take a TCP port.

https://github.com/google/gops/blob/v0.3.27/agent/agent.go#L74

dangra (approved):

PR #2536, opened 2023-07-05
`logging`: log equiv. of `LOG_LEVEL=debug` to files in `~/.fly/logs`

alichay:
### Change Summary

What and Why:

To make troubleshooting easier and issues more reproducible, we'll keep logs for the last 10 (this is a totally arbitrary number) runs of flyctl. These logs are the equivalent of `LOG_LEVEL=debug`, although they won't contain the actual output - only the debug logs.

How:

A lot of cleanup around logging and the `terminal` package. We had like four reimplementations of this logging system scattered around flyctl lmao, this consolidates most of it.

Related to: [internal discussion](https://flyio.discourse.team/t/flyctl-diagnostics-pipeline/3403)

---

### Documentation

- [ ] Fresh Produce
- [X] n/a (this is more a tool for support rather than something user-facing)

----

Sorry about the mega-commit! I kind of dug myself into a hole, and had to rewrite things to get out of it - in the process I just totally lost any hope of a reasonable commit history.

### DIFF
--- a/internal/logger/global_logfile.go
+++ b/internal/logger/global_logfile.go
@@ -1,0 +1,37 @@
+package logger
+
+import (
+	"bufio"
+
+	"github.com/superfly/flyctl/internal/logger/logfile"
+)
+
+var globalLogFile = logFile{
+	destroyed: true,
+}
+
+func InitLogFile() error {
+	if !globalLogFile.destroyed {
+		return logfileAlreadyInitializedError
+	}
+	rawFile, err := logfile.CreateLogFile()
+	globalLogFile = logFile{
+		file:      rawFile,
+		writer:    bufio.NewWriter(rawFile),
+		destroyed: false,
+	}
+	return err
+}
+
+func CloseLogFile() error {
+	if globalLogFile.destroyed {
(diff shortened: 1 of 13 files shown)


internal/logger/global_logfile.go
```diff
@@ -0,0 +1,37 @@
+package logger
+
+import (
+	"bufio"
+
+	"github.com/superfly/flyctl/internal/logger/logfile"
+)
+
+var globalLogFile = logFile{
+	destroyed: true,
```
dangra, line 10: does it need a mutex ?

dangra (approved):

internal/logger/global_logfile.go
alichay: I don't think go mutexes require initialization, right?

dangra: sorry, I blindly asked.

The question was towards protecting the global variable from changing when InitLogFile or CloseLogFile are called.

alichay: ahhh ok :)
I think it _should_ be fine?

PR #2537, opened 2023-07-06
Status:display machine metadata.role if not empty

rugwirobaker:
### Change Summary

What and Why:

`fly status` should display a role column if one is specified in the machine metadata

LiteFS for example now sets a role metadata key that can be replica or primary. This makes it easy to identify different machines in a cluster.

How:

Related to:

https://community.fly.io/t/dynamic-machine-metadata/13115
---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/status/machines.go
+++ b/internal/command/status/machines.go
@@ -169,10 +169,10 @@ func RenderMachineStatus(ctx context.Context, app *api.AppCompact, out io.Writer
 			if len(machine.Config.Standbys) > 0 {
 				hasStandbys = true
 			}
-			var role = "N/A"
+			var role string
 
-			if machine.Config.Metadata["role"] != "" {
-				role = machine.Config.Metadata["role"]
+			if v := machine.Config.Metadata["role"]; v != "" {
+				role = v
 			}
 			rows = append(rows, []string{
 				getProcessgroup(machine),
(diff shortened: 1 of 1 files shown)


internal/command/status/machines.go
```diff
@@ -169,12 +169,18 @@ func RenderMachineStatus(ctx context.Context, app *api.AppCompact, out io.Writer
 			if len(machine.Config.Standbys) > 0 {
 				hasStandbys = true
 			}
+			var role = "N/A"
```
dangra, line 172: can it be an empty string? It looks too heavy to show N/A when generally not used.

internal/command/status/machines.go
```diff
@@ -169,12 +169,18 @@ func RenderMachineStatus(ctx context.Context, app *api.AppCompact, out io.Writer
...
+
+			if machine.Config.Metadata["role"] != "" {
```
dangra, line 174: nit: 

```go
if v := machine.Config.Metadata["role"]; v != "" {
  role = v
}
```

dangra (approved):

internal/command/status/machines.go
```diff
@@ -169,12 +169,18 @@ func RenderMachineStatus(ctx context.Context, app *api.AppCompact, out io.Writer
 			if len(machine.Config.Standbys) > 0 {
 				hasStandbys = true
 			}
+			var role string
+
+			if v := machine.Config.Metadata["role"]; v != "" {
+				role = v
+			}
```
dangra, line 176: the interesting bit is that latest changes made all this logic redundant:

```go
role := machine.Config.Metadata["role"]
```

PR #2538, opened 2023-07-06
Fix background task runner, use for metrics

michaeldwan:
This PR makes the following changes to the background task runner so it's safer and actually useful for more things:
- Initialize the task manager before the command invocation starts so it's available for command preparers and finalizers immediately
- Add a Start function that accepts a context which will get passed to tasks rather than using a context provided when the task manager was initialized. This lets command preparers enqueue and use context values later on in tasks
- Add a ShutdownWithTimeout function that waits up to the timeout for tasks to finish
- Add a RunFinalizer function that accepts a task to run _after_ Shutdown is called. This provides a safe way to run stuff before exiting with the safety of a global timeout

With those changes in place, metrics can now be safely flushed before exit using a `RunFinalizer` task.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -7,10 +7,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/internal/metrics"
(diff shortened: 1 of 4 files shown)


alichay (approved):
I just discovered the task runner a couple of days ago, love to see it getting more features (especially finalizers!)

dangra (approved):

billyb2 (approved):
very cool

PR #2539, opened 2023-07-07
Send flaps_call metrics

billyb2:
### Change Summary
This is a second attempt to get [this PR](https://github.com/superfly/flyctl/pull/2503) working

What and Why:
Now that michael's changes on [the task runner](https://github.com/superfly/flyctl/pull/2538) are merged, flaps calls shouldn't hang.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/flaps/flaps.go
+++ b/flaps/flaps.go
@@ -619,12 +619,16 @@ func sendFlapsCallMetric(ctx context.Context, endpoint string, timing instrument
 
 	call := matches[index]
 
-	cmd_ctx := command_context.FromContext(ctx)
-	commandName := cmd_ctx.Parent().Name() + "-"
-	if commandName == "flyctl-" {
-		commandName = ""
+	cmdCtx := command_context.FromContext(ctx)
+	var nameParts []string
+	for cmdCtx != nil {
+		if cmdCtx.Name() == "flyctl" {
+			break
+		}
+		nameParts = append([]string{cmdCtx.Name()}, nameParts...)
+		cmdCtx = cmdCtx.Parent()
 	}
-	commandName += cmd_ctx.Name()
+	commandName := strings.Join(nameParts, "-")
 
 	metrics.Send(ctx, "flaps_call", flapsCall{
 		Call:       call,
--- a/flaps/flaps.go
+++ b/flaps/flaps.go
@@ -618,9 +618,17 @@ func sendFlapsCallMetric(ctx context.Context, endpoint string, timing instrument
 	}
 
 	call := matches[index]
+
+	cmd_ctx := command_context.FromContext(ctx)
+	commandName := cmd_ctx.Parent().Name() + "-"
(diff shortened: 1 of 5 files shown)


alichay:
question: what does a flaps request URL look like now?

Like, for the request `/v1/apps/<app>/machines/<machine_id>/start`, would the `flapsCall` ID be sent as like a query parameter or something?

billyb2:
We've made changes to flyctl-metrics, so it should be able to handle the load now.

flaps/flaps.go
```diff
@@ -590,6 +596,36 @@ func (f *Client) sendRequest(ctx context.Context, method, endpoint string, in, o
...
+	if index >= len(matches) {
+		err := fmt.Errorf("flaps endpoint %s did not match regex %s", endpoint, flapsCallRegex.String())
+		sentry.CaptureException(err)
+		return
+	}
+
+	call := matches[index]
+	metrics.Send(ctx, "flaps_call", flapsCall{
+		Call:       call,
+		Command:    command_context.FromContext(ctx).Name(),
```
alichay, line 623: two questions:
what would a value for call look like, and what values does `command_context.FromContext(ctx).Name()` give?

for call, I'm worried about cardinality, and for Command, I'm worried that both `fly apps restart` and `fly machines restart` would both return `restart`.

billyb2: `call` should always be the basic flaps call that's being done, like `lease` or `wait`. Any extra data like `?instance_id` should get filtered out by the regex.

As for both `apps restart` and `machines restart` returning restart, you're right, that's a bug I missed, and I'll fix that now.

flaps/flaps.go
```diff
@@ -590,6 +596,44 @@ func (f *Client) sendRequest(ctx context.Context, method, endpoint string, in, o
...
+	}
+
+	call := matches[index]
+
+	cmd_ctx := command_context.FromContext(ctx)
+	commandName := cmd_ctx.Parent().Name() + "-"
+	if commandName == "flyctl-" {
+		commandName = ""
+	}
+	commandName += cmd_ctx.Name()
```
alichay, line 627: nit: for future-proofing, this could use a loop to dig up to the true root node, although I don't think we have any commands that use flaps that go that deep so it shouldn't matter in practice.

it's a little uglier, although maybe more future proof(?). definitely your call!

```suggestion
	cmdCtx := command_context.FromContext(ctx)
	var nameParts []string
	for cmdCtx != nil {
		if cmdCtx.Name() == "flyctl" {
			break
		}
		nameParts = append([]string{cmdCtx.Name()}, nameParts...)
		cmdCtx = cmdCtx.Parent()
	}
	commandName := strings.Join(nameParts, "-")
```

alichay (approved):
good stuff! made one suggestion, but feel free to merge it without that :D

PR #2545, opened 2023-07-09
Add Metrics Regarding Flaps Calls

alichay:


internal/machine/ephemeral.go
```diff
@@ -123,7 +123,9 @@ func makeCleanupFunc(ctx context.Context, machine *api.Machine) func() {
 	return func() {
 		const stopTimeout = 15 * time.Second
 
-		stopCtx, cancel := context.WithTimeout(context.Background(), stopTimeout)
+		// FIXME: is there a reason we *need* to use context.Background here, instead of the normal context
+		// As far as I can tell, this is the only place in the codebase that does this
+		stopCtx, cancel := context.WithTimeout(ctx, stopTimeout)
```
billyb2, line 128: I need to verify that this is fine (though I haven't experienced any issues due to this change)

flaps/flaps_machines.go
```diff
@@ -1,10 +1,13 @@
+// FIXME: how do we handle flaps calls timing out. We still need to send metrics for it
```
billyb2, line 1: Kind of out of scope for this PR, plus it's a miniscule number of metrics that we don't need to worry about for right now.

billyb2:
This change would allow us to have more information regarding which flaps calls are failing. Metrics like these would help us to see why, for example, why deployments are failing.

flaps/flaps.go
```diff
@@ -184,6 +188,12 @@ func (f *Client) _sendRequest(ctx context.Context, method, endpoint string, in,
 		}
 	}()
 
+	defer func() {
+		if statusCode != nil {
+			*statusCode = resp.StatusCode
+		}
+	}()
```
dangra, line 195: These lines triggered some eye twitching :))

Why not returning the statuscode (along the error) if we are already changing the function signature? 

I mean, this approach of returning values is from languages that doesn't support returning multiple values and that is not the Go case.

flaps/flaps_machines.go
```diff
@@ -17,9 +19,165 @@ import (
...
+}
+
+// FIXME: rename this, bad name.
+type flapsActionInfo struct {
+	action          flapsAction
+	machineID       string
+	queryParameters map[string]string
+}
+
+func queryParamsToString(params map[string]string) string {
```
dangra, line 117: better to reuse https://pkg.go.dev/net/url#Values.Encode

flaps/flaps.go
```diff
@@ -204,6 +214,38 @@ func (f *Client) _sendRequest(ctx context.Context, method, endpoint string, in,
...
+func sendFlapsCallMetric(ctx context.Context, endpoint flapsAction, statusCode int, duration time.Duration) {
+	cmdCtx := command_context.FromContext(ctx)
+
+	// Iterate backwards through the command name to figure out the command being run.
+	// For example, `fly m run` becomes `machine-run`, `deploy --flags` becomes `deploy`
+	// TODO(billy): Unit test this
+	var nameParts []string
+	for cmdCtx != nil {
+		// Don't include the binary name in the nameParts
+		if cmdCtx.Name() == "flyctl" {
```
dangra, line 233: haven't tested it, does it work when the user calls flyctl through the `fly` alias?

flaps/flaps_machines.go
```diff
@@ -17,9 +19,165 @@ import (
...
+	cordon:       "cordon",
+	uncordon:     "uncordon",
+}
+
+// FIXME: rename this, bad name.
+type flapsActionInfo struct {
+	action          flapsAction
+	machineID       string
+	queryParameters map[string]string
+}
```
dangra, line 115: this type is half way to become the param's struct for sendRequestMachines function

It could absorb Method and Headers too, in fact, Method can be derived from Action as it does for endpoint.

flaps/flaps.go
billyb2: Yea that's understandable, I just didn't know how annoying it would be to always return a status code when it's only needed in one function.

flaps/flaps.go
billyb2: This is a good point, I didn't consider other binary names.

flaps/flaps.go
```diff
@@ -198,10 +202,42 @@ func (f *Client) _sendRequest(ctx context.Context, method, endpoint string, in,
...
+	cmdCtx := command_context.FromContext(ctx)
+
+	// Iterate backwards through the command name to figure out the command being run.
+	// For example, `fly m run` becomes `machine-run`, `deploy --flags` becomes `deploy`
+	// TODO(billy): Unit test this
+	var nameParts []string
+	for cmdCtx != nil {
+		// Don't include the binary name in the nameParts
+		if cmdCtx.Parent() == nil {
+			break
```
dangra, line 228: nit: it can be part of the `for` condition like in:
```go
for cmdCtx != nil && cmdCtx.Parent() != nil {
```

flaps/flaps_actions.go
```diff
@@ -0,0 +1,113 @@
+package flaps
+
+import "net/http"
+
+type flapsAction int
```
dangra, line 5: + 💯  for moving this into its own file.

dangra (approved):
well done. thanks for considering all the feedback!

flaps/flaps.go
billyb2: go is too advanced for my small brain

flaps/flaps.go
```diff
@@ -220,10 +220,11 @@ func sendFlapsCallMetric(ctx context.Context, endpoint flapsAction, statusCode i
...
+	//
+	// We check if the parent is nil so we don't include the binary name in the nameParts
 	// TODO(billy): Unit test this
 	var nameParts []string
-	for cmdCtx != nil {
-		// Don't include the binary name in the nameParts
+	for cmdCtx != nil && cmdCtx.Parent() != nil {
 		if cmdCtx.Parent() == nil {
 			break
 		}
```
dangra, line 230: oops, you left the "if" block there 🍂

PR #2547, opened 2023-07-10
Scale command can scale apps with volumes

senyosimpson:
#### What and Why:

Currently, `fly scale count`does not support scaling apps with volumes (instead, you have to run `fly machine clone`). This work adds that support. This creates a consistent experience as devs can reach for the same command to scale their app count up/down.

How:

There's different behaviours depending on whether the machine count is being scaled up or down.

For up:
1. We check if there are existing, unattached volumes in the region the machine is being created
2. If there are, then those are attached to the newly created machines
3. If there are no available volumes, we create a new volume for every machine that is created

For down, we don't do anything to the volume. They just become unattached. The user can manually delete them later.

#### Thoughts

We should probably support a few options to tweak the behaviour:

1. To create machines with new volumes, even if there are existing volumes
2. ~~To delete the volumes attached to machines when the machines are deleted~~
3. To specify a snapshot ID so a volume can be cloned from a snapshot of an existing volume

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -66,7 +66,7 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 	availableVols := lo.Map(volumes, func(v api.Volume, _ int) *availableVolume {
 		return &availableVolume{
 			Volume:    &v,
-			Available: true,
+			Available: !v.IsAttached(),
 		}
 	})
 
@@ -303,12 +303,12 @@ func computeActions(machines []*api.Machine, expectedGroupCounts map[string]int,
 		mConfig.Standbys = nil
 
 		for regionName, delta := range regionDiffs {
-			var availableVols []*availableVolume
+			var availableVolumes []*availableVolume
 			if len(mConfig.Mounts) > 0 {
 				volumeName := mConfig.Mounts[0].Name
 				groupVolumes := volumeGroups[volumeName]
-				availableVols = lo.Filter(groupVolumes, func(av *availableVolume, _ int) bool {
-					return !av.Volume.IsAttached() && av.Volume.Region == regionName
+				availableVolumes = lo.Filter(groupVolumes, func(av *availableVolume, _ int) bool {
+					return av.Available && av.Volume.Region == regionName
 				})
 			}
 
@@ -318,7 +318,7 @@ func computeActions(machines []*api.Machine, expectedGroupCounts map[string]int,
 				Delta:         delta,
 				Machines:      perRegionMachines[regionName],
 				MachineConfig: mConfig,
-				Volumes:       availableVols,
+				Volumes:       availableVolumes,
 			})
 		}
 	}
@@ -343,7 +343,7 @@ func computeActions(machines []*api.Machine, expectedGroupCounts map[string]int,
 			var availableVolumes []*availableVolume
 			if len(mConfig.Mounts) > 0 {
 				availableVolumes = lo.Filter(volumes, func(av *availableVolume, _ int) bool {
-					return !av.Volume.IsAttached() && av.Volume.Region == regionName
+					return av.Available && av.Volume.Region == regionName
 				})
 			}
 
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -172,7 +172,7 @@ func launchMachine(ctx context.Context, action *planItem, volume *api.Volume, ap
 			{
 				Volume: volume.ID,
 				Path:   mnt.Path,
-				Name:   mnt.Name,
+				Name:   volume.Name,
 			},
 		}
 	}
@@ -272,6 +272,9 @@ func computeActions(machines []*api.Machine, expectedGroupCounts map[string]int,
 	machineGroups := lo.GroupBy(machines, func(m *api.Machine) string {
 		return m.ProcessGroup()
 	})
+	volumeGroups := lo.GroupBy(volumes, func(v *availableVolume) string {
+		return v.Volume.Name
+	})
 
 	for groupName, groupMachines := range machineGroups {
 		expected, ok := expectedGroupCounts[groupName]
@@ -301,8 +304,10 @@ func computeActions(machines []*api.Machine, expectedGroupCounts map[string]int,
 		for regionName, delta := range regionDiffs {
 			var availableVols []*availableVolume
 			if len(mConfig.Mounts) > 0 {
-				availableVols = lo.Filter(volumes, func(av *availableVolume, _ int) bool {
-					return !av.Volume.IsAttached() && av.Volume.Region == regionName
+				volumeName := mConfig.Mounts[0].Name
+				groupVolumes := volumeGroups[volumeName]
+				availableVols = lo.Filter(groupVolumes, func(av *availableVolume, _ int) bool {
+					return !av.Volume.IsAttached() && av.Volume.Region == regionName && av.Volume.Name == volumeName
 				})
 			}
 
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -63,6 +63,11 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 		return err
 	}
 
+	appCompact, err := apiClient.GetAppCompact(ctx, appConfig.AppName)
+	if err != nil {
+		return err
+	}
+
 	defaults := newDefaults(appConfig, latestCompleteRelease, machines)
 
 	actions, err := computeActions(machines, expectedGroupCounts, volumes, regions, maxPerRegion, defaults)
@@ -119,7 +124,7 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 					volume = &action.Volumes[i]
 				}
 
-				m, v, err := launchMachine(ctx, action, volume, appConfig)
+				m, v, err := launchMachine(ctx, action, volume, appCompact)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


senyosimpson:
Code still needs work but just opening the PR in the meantime

internal/command/scale/count_machines.go
```diff
@@ -107,11 +96,28 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
...
+	trackedVolumes := lo.FilterMap(volumes, func(v api.Volume, _ int) (*trackedVolume, bool) {
+		if !v.IsAttached() {
+			tv := trackedVolume{
+				volume: v,
+				used:   false,
+			}
+			return &tv, true
+		}
+
+		return &trackedVolume{}, false
```
dangra, line 113: you can return `nil, false` here as the volume will be ignored.

internal/command/scale/count_machines.go
```diff
@@ -132,8 +138,56 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
...
+	colorize := io.ColorScheme()
 	flapsClient := flaps.FromContext(ctx)
+	apiClient := client.FromContext(ctx).API()
+
+	app, err := apiClient.GetAppCompact(ctx, appConfig.AppName)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(action.MachineConfig.Mounts) > 0 {
```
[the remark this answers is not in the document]
senyosimpson: Yep, agreed!

senyosimpson:
Example output:

```
❯ fly scale count 7 --region jnb
App 'helix' is going to be scaled according to this plan:
  +6 machines for group 'app' on region 'jnb' with size 'shared-cpu-1x'
  +1 new volumes and using 5 existing volumes in region 'jnb'
? Scale app helix? Yes
Executing scale plan
  Using unattached volume vol_xxxxxx
  Created xxxxxx group:app region:jnb size:shared-cpu-1x volume:vol_xxxxxx
  Using unattached volume vol_xxxxxx
  Created xxxxxx group:app region:jnb size:shared-cpu-1x volume:vol_xxxxxx
  Using unattached volume vol_xxxxxx
  Created xxxxxx group:app region:jnb size:shared-cpu-1x volume:vol_xxxxxx
  Using unattached volume vol_xxxxxx
  Created xxxxxx group:app region:jnb size:shared-cpu-1x volume:vol_xxxxxx
  Using unattached volume vol_xxxxxx
  Created xxxxxx group:app region:jnb size:shared-cpu-1x volume:vol_xxxxxx
  Volume volhelix will start empty
  Created xxxxxx group:app region:jnb size:shared-cpu-1x volume:vol_xxxxxx
```

internal/command/scale/count_machines.go
```diff
@@ -132,14 +150,89 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
...
+func createVolume(ctx context.Context, mount api.MachineMount, volume *api.Volume, appID string, region string) (*api.Volume, error) {
+	io := iostreams.FromContext(ctx)
+	colorize := io.ColorScheme()
+	apiClient := client.FromContext(ctx).API()
+
+	var err error
+
+	withNewVolumes := flag.GetBool(ctx, "with-new-volumes")
+	if volume != nil && !withNewVolumes {
+		fmt.Fprintf(io.Out, "  Using unattached volume %s\n", colorize.Bold(volume.ID))
```
dangra, line 195: wdyt about returning earlier here to avoid the nested "else" block completely?

internal/command/scale/count_machines.go
```diff
@@ -192,12 +286,20 @@ func computeActions(machines []*api.Machine, expectedGroupCounts map[string]int,
 		mConfig.Standbys = nil
 
 		for regionName, delta := range regionDiffs {
+			var availableVolumes []api.Volume
+			if len(mConfig.Mounts) > 0 {
+				availableVolumes = lo.Filter(volumes, func(v api.Volume, _ int) bool {
+					return !v.IsAttached() && v.Region == regionName
+				})
```
dangra, line 293: there is an edge case, not so common but still possible, the mounts can be shared with multiple process groups.
If that were the case, scaling up will reuse the same volume list for different process groups

internal/command/scale/count_machines.go
```diff
@@ -132,14 +150,89 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
...
+		if err != nil {
+			return nil, nil, err
+		}
+		action.MachineConfig.Mounts = []api.MachineMount{
+			{
+				Volume:    volume.ID,
+				Path:      mnt.Path,
+				Name:      mnt.Name,
+				Encrypted: mnt.Encrypted,
+				SizeGb:    mnt.SizeGb,
```
dangra, line 168: no need to set these, they are ignored by flyd.

internal/command/scale/count_machines.go
```diff
@@ -111,11 +125,24 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 		switch {
 		case action.Delta > 0:
 			for i := 0; i < action.Delta; i++ {
-				m, err := launchMachine(ctx, action)
+				var volume *api.Volume
+				if i < len(action.Volumes) {
+					if av := action.Volumes[i]; av.Available {
+						av.Available = false
```
senyosimpson, line 131: @dangra so that we don't reuse same volumes across process groups, I went with the idea of mapping `[]api.Volume => []*availableVolume` which has the definition (`availableVolume { *api.Volume, available: bool }`).

So we can check the availability flag and update it after using them so we don't reuse it later with another process group

dangra: I see it now but didn’t find where the volume allocation consider the "mount's name" when reusing a existent volume. It is like any volume will do as long as it is in the same region.

dangra: consider the fly.toml from https://community.fly.io/t/mounts-per-process-group-for-apps-v2/12304

it defines different mounts with different volume names per process group

senyosimpson: Ahhh sure, missed that. I'll get that done

internal/command/scale/count_machines.go
```diff
@@ -257,7 +257,7 @@ type planItem struct {
 	Region        string
 	Delta         int
 	Machines      []*api.Machine
-	MachineConfig *api.MachineConfig
+	MachineConfig api.MachineConfig
```
dangra, line 260: why? it will make a shallow copy of the machine config, it doesn't prevent the copy from sharing other sub fields. It doesn't look like a sensible change IMO. If you need to clone the config before modifying it, use `helpers.Clone`

internal/command/scale/count_machines.go
```diff
@@ -159,7 +159,7 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 	return nil
 }
 
-func launchMachine(ctx context.Context, action *planItem, volume *api.Volume, appCompact *api.AppCompact) (*api.Machine, *api.Volume, error) {
+func launchMachine(ctx context.Context, action planItem, volume *api.Volume, appCompact *api.AppCompact) (*api.Machine, *api.Volume, error) {
```
dangra, line 162: hmm, what is the point of passing the action item as a value?

internal/command/scale/count_machines.go
senyosimpson: Ah, nice, didn't know about `helper.Clone`, makes sense. The reason for the copy is due to this line:

https://github.com/superfly/flyctl/blob/f8080d10a19f095d6a2b08f7495cec20d3209378/internal/command/scale/count_machines.go#L166-L177

Since the same `action` is used if the delta is bigger than 1, it gets modified each iteration when we assign the new mount values. On the next iteration, `machineConfig.Mounts` has the values of the new volume. That's a problem mostly because we support restoring from a snapshot. The `GetVolumeSnapshots` takes the volume id. So if that changes to the newly created volume, it will have no snapshots and break the logic.

internal/command/scale/count_machines.go
```diff
@@ -192,12 +302,22 @@ func computeActions(machines []*api.Machine, expectedGroupCounts map[string]int,
 		mConfig.Standbys = nil
 
 		for regionName, delta := range regionDiffs {
+			var availableVols []*availableVolume
+			if len(mConfig.Mounts) > 0 {
+				volumeName := mConfig.Mounts[0].Name
+				groupVolumes := volumeGroups[volumeName]
+				availableVols = lo.Filter(groupVolumes, func(av *availableVolume, _ int) bool {
+					return !av.Volume.IsAttached() && av.Volume.Region == regionName && av.Volume.Name == volumeName
```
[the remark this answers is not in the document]
senyosimpson: Meant to change (had old code before that used filtering first)

senyosimpson:
Updated to support scaling up when volumes belong to different processes

```
❯ fly scale count app=2 web=2
App 'helix' is going to be scaled according to this plan:
  +1 machines for group 'web' on region 'jnb' with size 'shared-cpu-1x'
  +1 new volumes for group 'web' in region 'jnb'
  +1 machines for group 'app' on region 'jnb' with size 'shared-cpu-1x'
  +1 new volumes for group 'app' in region 'jnb'
? Scale app helix? Yes
Executing scale plan
  Volume volhelix2 will start empty
  Created xxxxxx group:web region:jnb size:shared-cpu-1x volume:vol_xxxxxx
  Volume volhelix will start empty
  Created xxxxxx group:app region:jnb size:shared-cpu-1x volume:vol_xxxxxx
```

dangra (approved):

PR #2554, opened 2023-07-12
Distributed Elixir support for fly launch

chrismccord:
### Change Summary

What and Why:
Today `fly launch` Just Works™ for a vanilla Phoenix app, but it lacks support for distributed elixir out-of-the-box, which is one of the best features of Phoenix + Fly. This change will make distributed elixir Just Work™ for fly apps on the next release of Phoenix.

How:
The phoenix team released https://github.com/phoenixframework/dns_cluster , which provides simple DNS poll based clustering. The next Phoenix release will include this dependency by default when generating a `phx.new` project.

In addition to the automatic clustering, I am also using the unique Docker image id from `FLY_IMAGE_REF` as the node basename. This ensures only matching nodes across deployments will connect to eachother. This change can be merged ahead of Phoenix release and will still work for older phoenix applications.

---

### Documentation

- [ ] Fresh Produce

### DIFF
--- a/scanner/phoenix.go
+++ b/scanner/phoenix.go
@@ -2,7 +2,6 @@ package scanner
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


chrismccord:
@jsierles I'm not sure because I need to conditionally generate it, or append the contents to an existing file, but only if I don't find a particular match. I don't know if the templates plumbing can accommodate that?

jsierles:
Good point. Go templates can't append, so if this works, all good!

jsierles (approved):

PR #2556, opened 2023-07-12
Define Go version in one place and use GoReleaser for preflight tests

kzys:
### Change Summary

What and Why:

During #2553, I hard-coded Go versions in N different places. This PR cleans up the mess and makes preflight tests closer to what customers would have.

How:

- The first commit cleans up #2553. I don't want to fix N different places for upgrade/downgrade Go versions.
- The second commit is not really used by this PR. I used dev_release job and realized that the file is broken. Then realized test_build was enough for the third commit.
- The third commit was connecting GoReleaser to preflight. So we can test what customers would use.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,11 @@ on:
   push:
   workflow_dispatch:
 
+# We right now cannot use Go 1.20.6 due to
+# https://community.fly.io/t/git-hub-actions-deployment-error/14126
+env:
+  go_version: 1.20.5
+
 jobs:
   test_build:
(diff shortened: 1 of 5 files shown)


dangra (approved):
If I understand this correctly, it does:
1. Build a snapshot release and upload it as artifact
2. trigger preflight workflow
3. download the artifact with flyctl binary
4. run preflight tests using previous step binary

LGTM!

kzys:
Yes. Previously we were building flyctl again before starting preflight tests.

PR #2568, opened 2023-07-14
Send all metrics in a temporary separate process

billyb2:
### Change Summary

What and Why :
This should help us to avoid sending to the metrics server blocking flyctl from completing, as well as preventing any latency with flyctl-metrics from preventing deployments (and other commands) to flyctl.

How:
When flyctl exits, or receives a SIGINT, we fork a separate `fly metrics send`  process who's sole task is sending all the metrics we collected over the course of running flyctl to the metrics server. Once it finishes sending (or fails 3 times in sending it), the process exits. No zombies!

Related to:
https://github.com/superfly/flyctl-metrics/pull/20

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


internal/command/command.go
```diff
@@ -107,6 +107,13 @@ func newRunE(fn Runner, preparers ...preparers.Preparer) func(*cobra.Command, []
 		task.FromContext(ctx).Start(ctx)
 
 		sendOsMetric(ctx, "started")
+		task.FromContext(ctx).RunFinalizer(func(ctx context.Context) {
+			err := metrics.FlushMetrics()
+			if err != nil {
+				fmt.Fprintln(os.Stderr, "Error spawning metrics process: ", err)
```
rugwirobaker, line 113: You should probably use the `iostreams.StdErr` that already contained in the context.

internal/command/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+
+	cmd = command.New("send", short, long, run)
+	cmd.Hidden = true
+	cmd.Args = cobra.NoArgs
+
+	return
+}
+
+func run(ctx context.Context) error {
+	stdin_bytes, err := io.ReadAll(os.Stdin)
```
rugwirobaker, line 50: Same here, use the iostreams package.

rugwirobaker (changes requested):

internal/command/root/root.go
```diff
@@ -139,6 +140,7 @@ func New() *cobra.Command {
 		group(console.New(), "upkeep"),
 		settings.New(),
 		group(mysql.New(), "dbs_and_extensions"),
+		metrics.New(),
```
rugwirobaker, line 143: We should probably make this a hidden command

rugwirobaker (approved):
LGTM

PR #2569, opened 2023-07-14
Upgrade dependencies

michaeldwan:
This upgrades a number of deps that we've fallen behind on, and in the process resolved a circular dependency I encountered while upgrading cobra. 

I tested it quite a bit locally, but let's hold off merging until the test suite is fixed.


userbradley:
If I can be cheeky and ask, how come you dont use Dependabot? From my understanding, Dependabot supports Gomod

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates

dangra:
hey @userbradley we only use dependabot for security updates but no doubt it is worth a shot for keeping all dependencies up to date. I'll take a look.

dangra:
there you go https://github.com/superfly/flyctl/pull/2575

userbradley:
Glad I could be of service @dangra. Enjoy a PR from Dependabot every time someone sneezes!

PR #2571, opened 2023-07-14
Fix 5-second hang when preparers fail

matttpt:
### Change Summary

Command preparers may submit tasks to the task manager, but those tasks are not run until the task manager is started. If a preparer fails, the manager is never started, and calls to `Shutdown` or `ShutdownWithTimeout` wait for tasks that will never run.

To reproduce, run `fly m list` with no `fly.toml`, `-a` flag, or `FLY_APP`, so that the `RequireAppName` preparer fails. Observe that there's a 5-second delay while the task manager waits for a metrics task to run before timing out. This isn't a serious issue, but it's quite annoying: the error message is printed only *after* the hang, and `^C` has no effect.

So, during shutdown, don't wait for tasks unless the task manager was started.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/task/task.go
+++ b/internal/task/task.go
@@ -4,6 +4,7 @@ package task
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/superfly/flyctl/internal/logger"
@@ -61,7 +62,8 @@ type Manager interface {
 }
 
 type manager struct {
-	queue chan Task
+	queue   chan Task
+	started atomic.Bool
 	sync.WaitGroup
 }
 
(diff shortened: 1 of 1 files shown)


michaeldwan:
This is a good fix! I changed the way the task manager is kicked off while looking into a related issue last week. It now starts running before preparers so tasks can be queued immediately without waiting for the manager to be initialized. That fixed one source of hangs, but I hadn't considered when a preparer fails after a previous one adds something to the queue. Nice find.

FWIW this code is really gross, buggy, and old (my fault...) I'm testing out a change that runs preparers within the context of a command to simplify logic and have more control over the lifecycle of the cli process. Should have something to show once #2569 is merged.

michaeldwan (approved):

PR #2587, opened 2023-07-17
support defining files with fly deploy

jipperinbham:
### Change Summary

Currently, only `fly machine run|update` support the flags for adding files to a machine and these changes brings that feature to `fly deploy`.

What and Why:

Additional flags added for specifying files to be added to a machine config and created in the VM.

How:

```
--file-local=/path/inside/machine=<local/path>
--file-literal=/path/inside/machine=VALUE
--file-secret=/path/inside/machine=SECRET
```

Related to: https://github.com/superfly/flyctl/pull/2506

---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/machines_launchinput.go
+++ b/internal/command/deploy/machines_launchinput.go
@@ -7,6 +7,7 @@ import (
 	"github.com/samber/lo"
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/internal/buildinfo"
+	machinecmd "github.com/superfly/flyctl/internal/command/machine"
 	"github.com/superfly/flyctl/internal/machine"
 	"github.com/superfly/flyctl/terminal"
 )
@@ -46,6 +47,8 @@ func (md *machineDeployment) launchInputForLaunch(processGroup string, guest *ap
 	if len(standbyFor) > 0 {
 		mConfig.Standbys = standbyFor
 	}
+
+	mConfig.Files = md.files
 
 	return &api.LaunchMachineInput{
 		Region:     region,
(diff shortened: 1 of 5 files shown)


internal/command/deploy/machines_launchinput.go
```diff
@@ -47,6 +48,8 @@ func (md *machineDeployment) launchInputForLaunch(processGroup string, guest *ap
 		mConfig.Standbys = standbyFor
 	}
 
+	mConfig.Files = md.files
```
dangra, line 51: Shouldn't files come from fly.toml?  I buy it could be useful to override them using flags on `fly deploy` but generally speaking you want to declare them in the app configuration file. do you think?

jipperinbham: I wasn't sure TBH so figured I'd start with flags and see what others thought.

internal/command/deploy/machines_launchinput.go
```diff
@@ -47,6 +47,9 @@ func (md *machineDeployment) launchInputForLaunch(processGroup string, guest *ap
 		mConfig.Standbys = standbyFor
 	}
 
+	// Merge files from the app config and the parsed flags.
+	machine.MergeFiles(mConfig, md.files)
```
jipperinbham, line 51: I kept going in circles trying to figure out how to merge files from flags and files from appconfig.Config before here and eventually gave up and went with this.

dangra (commented):
this is looking great ! 

My only complain are the missing test cases for things that have a reasonable test suite like `internal/appconfig/serde_test.go` and `internal/command/deploy/machines_launchinput_test.go`

As a minimum please add the file section to `internal/appconfig/testdata/full-reference.toml`

internal/command/deploy/machines_launchinput.go
dangra: I'd go with merging it at `determineAppConfigForMachines` because even if not the best, it is consistent with what we are doing now for other flags.

dangra: Right, but that is what we do for env, strategy, region flags...  having a different treatment for files makes it worse.

dangra (approved):

PR #2603, opened 2023-07-21
install.sh: rename into place

haileys:
There's actually two changes in this PR, but they would conflict if I split them out into separate PRs so I've combined them into one PR. [Diff best viewed ignoring whitespace changes](https://github.com/superfly/flyctl/pull/2603/files?w=1).

The first commit wraps the entire install script in a `main` function which is immediately invoked at the end of the script. This is necessary because we suggest users to pipe curl to shell, and if the curl transfer fails, `sh` will happily execute the partial script. This might seem pedantic, but the hairs on my neck started standing up as I was writing `rm -rf`, so I thought it best to be extra careful.

The second commit, and the main change that prompted this PR, changes the behaviour of the install script to extract the flyctl tarball to a temporary directory first, chmod +x's the binary there, and then renames into place.

This is necessary to avoid this race condition:

image

What I think is happening here is the auto-updater happens to be running in the background at the same time I ran that first `fly m stop` command in the screenshot. `tar` had `~/.fly/bin/flyctl` open for writing, and as a result, we get an error trying to execute it. This race condition could also manifest as a permission denied error in the case that the user tries to invoke `flyctl` in between `tar` exiting and `chmod` doing its thing.

The usual way to work around this kind of problem on unix is to exploit the atomicity of the `rename(2)` syscall by doing all of our work in a temporary directory before finally renaming into place. Because `rename(2)` is guaranteed to be atomic, it's not possible for a `flyctl` invocation to race with it - it'll cleanly execute either the old or the new version.

### DIFF
--- a/installers/install.sh
+++ b/installers/install.sh
@@ -18,18 +18,20 @@ main() {
 	flyctl_install="${FLYCTL_INSTALL:-$HOME/.fly}"
 
 	bin_dir="$flyctl_install/bin"
+	tmp_dir="$flyctl_install/tmp"
 	exe="$bin_dir/flyctl"
 	simexe="$bin_dir/fly"
 
-	if [ ! -d "$bin_dir" ]; then
-		mkdir -p "$bin_dir"
-	fi
+	mkdir -p "$bin_dir"
+	mkdir -p "$tmp_dir"
 
-	curl -q --fail --location --progress-bar --output "$exe.tar.gz" "$flyctl_uri"
-	cd "$bin_dir"
-	tar xzf "$exe.tar.gz"
-	chmod +x "$exe"
-	rm "$exe.tar.gz"
+	curl -q --fail --location --progress-bar --output "$tmp_dir/flyctl.tar.gz" "$flyctl_uri"
+	# extract to tmp dir so we don't open existing executable file for writing:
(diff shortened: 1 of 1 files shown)


installers/install.sh
```diff
@@ -4,50 +4,56 @@
...
+	mkdir -p "$bin_dir"
+	mkdir -p "$tmp_dir"
+
+	curl -q --fail --location --progress-bar --output "$tmp_dir/flyctl.tar.gz" "$flyctl_uri"
+	# extract to tmp dir so we don't open existing executable file for writing:
+	tar -C "$tmp_dir" -xzf "$tmp_dir/flyctl.tar.gz"
+	chmod +x "$tmp_dir/flyctl"
+	# atomically rename into place:
+	mv "$tmp_dir/flyctl" "$exe"
+	rm -rf "$tmp_dir"
```
dangra, line 34: if $FLYCTL_INSTALL is set to `/` this is going to be problematic. 
What if instead it only attempts to remove `$tmp_dir/flyctl.tar.gz` ?

haileys: Good catch. Pushed a commit up fixing this!

dangra (approved):
:shipit:

PR #2605, opened 2023-07-21
Wait 60m to delete preflight apps

billyb2:
Otherwise, we sometimes destroy apps mid-test

### Change Summary

What and Why:
Increase the amount of time needed before we destroy an app.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/scripts/clean-up-preflight-apps/main.go
+++ b/scripts/clean-up-preflight-apps/main.go
@@ -52,7 +52,7 @@ func run() error {
 		return err
 	}
 	for _, app := range resp.Organization.Apps.Nodes {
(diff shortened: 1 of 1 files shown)


dangra:
Apps are created on each test case and purged after 30min. If test cases fails because of this, it means that test cases are taking longer than 30min which is a long long time for any app operation we are doing in preflight.

To me, the problem must be somewhere else, some API not responding, hanging or failing too often that retries pile up.

billyb2:
Hmm, true.  Any action taking an hour is odd. You're right, I shoould approach this from a different perspective.

PR #2606, opened 2023-07-21
Set the application platform when provisioning Sentry projects

jsierles:
This PR sends the detected application platform when provisioning a Sentry project. Sentry projects require a platform setting.


internal/command/extensions/sentry/sentry.go
```diff
@@ -17,3 +17,116 @@ func New() (cmd *cobra.Command) {
...
+	"ruby",
+	"ruby-rack",
+	"ruby-rails",
+	"rust",
+	"unity",
+	"unreal",
+}
+
+var PlatformMap = map[string]string{
+	"Bun":           "javascript",
```
rubys, line 101: This is just a question.  Deno does not have as a goal to be node-compatible, though recently it has done some effort along these lines.  Bun does have a goal to be node-compatible... it isn't perfect, but is steadily improving along those lines.  With this context in mind, why does Deno map to node and Bun map to javascript?

rubys (approved):

internal/command/extensions/sentry/sentry.go
jsierles: That was a bit arbitrary. I don't know that there's a big difference in how Sentry handles node versus javascript projects. We can switch it to node.

jsierles: I added a comment with a link to the list of supported platforms.

PR #2607, opened 2023-07-21
Faster Preflight Tests

billyb2:
### Change Summary
Preflight tests should now run faster, making our CI as a whole go faster.

One issue with this PR is that running at higher parallelism levels locally (`go test --parallel=16`), tests will fail for reasons I can't quite decipher. (They also fail running `make preflight` on master, so I think this is an issue with my setup).

What and Why:
Eventually, preflight tests passing should be a requirement for new fly releases, to help avoid regressions.

How:
This PR adds more Github workers, as well as making some changes to tests so they can be run in parallel.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/scripts/preflight.sh
+++ b/scripts/preflight.sh
@@ -47,7 +47,7 @@ gotesplit \
     -total "$total" \
     -index "$index" \
     github.com/superfly/flyctl/test/preflight/... \
-    -- --tags=integration -v -timeout=60m $test_opts | tee "$test_log"
+    -- -parallel=8 -p8 --tags=integration -v -timeout=60m $test_opts | tee "$test_log"
 test_status=$?
 
 set -e
--- a/scripts/preflight.sh
+++ b/scripts/preflight.sh
@@ -47,7 +47,7 @@ gotesplit \
     -total "$total" \
     -index "$index" \
     github.com/superfly/flyctl/test/preflight/... \
-    -- -parallel=8 -p8 --tags=integration -v -timeout=60m $test_opts | tee "$test_log"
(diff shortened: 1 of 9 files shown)


scripts/preflight.sh
```diff
@@ -47,7 +47,7 @@ gotesplit \
     -total "$total" \
     -index "$index" \
     github.com/superfly/flyctl/test/preflight/... \
-    -- --tags=integration -v -timeout=60m $test_opts | tee "$test_log"
+    -- --parallel=4 --tags=integration -v -timeout=60m $test_opts | tee "$test_log"
```
dangra, line 50: not sure what is happening with [this build](https://github.com/superfly/flyctl/actions/runs/5660882717/job/15337837456?pr=2607) but all other jobs took less than 5min while that is 10m+ and still running.

I'd lower the test timeout to 10min if we confirm it is stuck due to server side issues.

billyb2: I'm gonna wait for it to timeout, so I can view which specific test failed. I'm pretty sure it's a server side issue though.

billyb2: the test has failed on two consecutive commits, I'm gonna look into it further.

billyb2: i'm thinking this is most likely a backend issue. The [same test](https://github.com/superfly/flyctl/actions/runs/5661445183/job/15339542250) was failing for me on a completely separate commit

dangra (approved):
LGTM! 🚢

PR #2609, opened 2023-07-21
New `fly launch` command

alichay:
### Change Summary

What and Why: Revamps the Fly Launch command to make it easier to deploy new apps and more clear how to work with them after the fact

---

### Documentation

- [X] Fresh Produce (post WIP)
- [ ] n/a

### TODO

- [ ] Test `fly auth` flow still works


alichay:
This is WIP, but everything's behind hidden flags so it's not a huge concern. I want to get this merged so that we can start to build on this as a foundation :)

dangra (approved):
My only worry is the duplication of codepaths for configs coming from scanners. We will have to be careful to not loose added functionality when fully switching over.

alichay:
> My only worry is the duplication of codepaths for configs coming from scanners. We will have to be careful to not loose added functionality when fully switching over.

not too worried about scanner stuff falling out of sync - they're a first class citizen of the proposal, so before it ships I imagine it'll end up getting worked on by framework folks

PR #2613, opened 2023-07-23
Make first non-"Dev" pg choice the default

rubys:
Closing due to lack of interest....


kzys (commented):
Sounds good to me. I'm inclined to make our default safe, rather than inexpensive.

PR #2618, opened 2023-07-24
Add `app delete/remove/rm` as an alias for `app destroy`.

ndarilek:
### Change Summary

Add `fly apps delete/remove/rm` as aliases for `fly apps destroy`.

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/apps/destroy.go
+++ b/internal/command/apps/destroy.go
@@ -35,6 +35,7 @@ from the Fly platform.
 
 	destroy.ValidArgsFunction = completion.Adapt(completion.CompleteApps)
 
+	destroy.Aliases = []string{"rm"}
 	return destroy
 }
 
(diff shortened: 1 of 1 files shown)


internal/command/apps/destroy.go
```diff
@@ -35,6 +35,7 @@ from the Fly platform.
 
 	destroy.ValidArgsFunction = completion.Adapt(completion.CompleteApps)
 
+	destroy.Aliases = []string{"rm"}
```
dangra, line 38: Why not but we probably want to standarize a bit and accept all variations:

`fly machines` can do `destroy`, `remove` and `rm`. 
`fly volumes` accepts `destroy`, `delete` and `rm`

ndarilek: Happy to add more, I went conservative for a first pass but I've also tried `fly app delete` before too.

dangra (approved):
:shipit:

PR #2619, opened 2023-07-25
autoupdate: be proactive against update loops

alichay:
### Change Summary

What and Why: Three things:
1. blocks autoupdate checking for `fly settings autoupdate` for obvious reasons.
2. Sets the `FLY_NO_UPDATE_CHECK` environment variable when autoupdate relaunches flyctl, to make it impossible for an update loop to occur
3. Attempts to use the binary `$(brew prefix)/bin/flyctl` if we were running in homebrew before, because homebrew does strange things with symlinks that *might* be making updates inconsistent. When not under homebrew, we do what we did before: `realpath(argv[0])`
4. wait a little longer after updates before re-running. let fs operations settle down and everything before expecting flyctl to be fully updated.

Related to: this person's unfortunate error https://community.fly.io/t/flyctl-versions-autoupdating-and-the-cli-apocalypse/13794/3?u=allison

---

### Documentation

- [ ] Fresh Produce
- [X] n/a

### DIFF
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -356,7 +356,7 @@ func Relaunch(ctx context.Context, silent bool) error {
 
 	// Wait a bit for the update to take effect.
 	// Windows seemed to need this for whatever reason.
-	time.Sleep(250 * time.Millisecond)
+	time.Sleep(400 * time.Millisecond)
 
 	binPath, err := GetCurrentBinaryPath()
 	if err != nil {
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -3,6 +3,7 @@ package update
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


internal/update/update.go
```diff
@@ -182,37 +183,62 @@ func latestHomebrewRelease(ctx context.Context, channel string) (*Release, error
...
+
+	brewPrefixBytes, err := exec.Command(brewExe, "--prefix").Output()
+	if err != nil {
+		return "", err
+	}
+
+	brewBinPrefix := filepath.Join(strings.TrimSpace(string(brewPrefixBytes)), "bin") + string(filepath.Separator)
+
+	_brewBinDir = api.StringPointer(brewBinPrefix)
+	return *_brewBinDir, nil
```
dangra, line 210: you lost me there :) 

why returning the dereference of the pointer to BrewBinPrefix and not brewBinPrefix directly?

dangra: ahhh.. because of the global variable to memoize the result 🤔

dangra (approved):
most of this code runs as a single thread, the global variables shouldn't be an issue... until they are, could be good to wrap their write access with a sync.Mutex. Or even use `sync.Once` to initialize them.

PR #2623, opened 2023-07-25
Fix various issues in Django scanner

ndarilek:
### Change Summary

Our Django scanner had the following issues:
* Failure to determine an existing image if the locally-installed Python was a beta.
* Failure to install necessary build dependencies if psycopg was required.

These changes do the following:
* If the version number of the locally-installed Python contains any non-numeric components (E.g. "3.12.0b4), then that version is used explicitly and pinned, otherwise the existing version logic that uses a minor version continues to work as normal.
* If we pin a version, we explicitly call that out in the terminal output and in a comment in the Dockerfile advising the user to select an official release before deploying to production. IMO this strikes a good balance between getting users up and running while advising them about the compromise we made along the way.
* Images based on Debian Buster are no longer being built, so I bump to the Bullseye-based images instead.
* If psycopg is detected, then gcc and libpq-dev are installed so the image builds.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/scanner/django.go
+++ b/scanner/django.go
@@ -2,14 +2,15 @@ package scanner
 
 import (
 	"fmt"
-	"github.com/blang/semver"
-	"github.com/logrusorgru/aurora"
-	"github.com/mattn/go-zglob"
-	"github.com/superfly/flyctl/helpers"
 	"os/exec"
 	"path"
 	"regexp"
 	"strings"
+
+	"github.com/blang/semver"
+	"github.com/logrusorgru/aurora"
+	"github.com/mattn/go-zglob"
+	"github.com/superfly/flyctl/helpers"
 )
 
 // setup django with a postgres database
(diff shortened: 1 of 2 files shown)


dangra:
@ndarilek the linter is complaining with:

```
trim trailing whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing scanner/templates/django/Dockerfile
pre-commit hook(s) made changes.
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.
All changes made by hooks:
diff --git a/scanner/templates/django/Dockerfile b/scanner/templates/django/Dockerfile
index de6a625..2e1ca56 100644
--- a/scanner/templates/django/Dockerfile
+++ b/scanner/templates/django/Dockerfile
@@ -10,10 +10,10 @@ ENV PYTHONUNBUFFERED 1
 
 {{ if .hasPostgres -}}
 # install psycopg2 dependencies.
-RUN apt-get update && apt-get install -y \                                                                                                                                    
-    libpq-dev \                                                                                                                                                               
-    gcc \                                                                                                                                                                     
-    && rm -rf /var/lib/apt/lists/*                                                                                                                                            
+RUN apt-get update && apt-get install -y \
+    libpq-dev \
+    gcc \
+    && rm -rf /var/lib/apt/lists/*
 
 {{ end -}}
 
Error: Process completed with exit code 1.
```

ndarilek:
Thanks, think I fixed them.

dangra:
LGTM but will wait a bit for @katiayn to sign-off ✅

katiayn (approved):
LGTM!

PR #2624, opened 2023-07-25
Fix the sensitive environment variable warning

billyb2:
### Change Summary

What and Why:
We're currently printing the substring that was matched, instead of the actual environment variable.

How:

Related to:
https://community.fly.io/t/remove-superfluous-warning-x-may-be-a-potentially-sensitive-environment-variable/14388

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


billyb2:
> Are you printing the _value_ here as well?

It should just print the key
```
WARN PRIVATE_KEY may be a potentially sensitive environment variable. Consider setting it as a secret, and removing it from the [env] section: https://fly.io/docs/reference/secrets/
```

PR #2627, opened 2023-07-26
update: only update if our installation is updateable

alichay:
### Change Summary

What and Why: Disable updating when flyctl binary is not in `FLYCTL_INSTALL`, `~/.fly`, or `$(brew --prefix)/bin`. This won't work, and just leads to update loops.

Related to: https://community.fly.io/t/flyctl-versions-autoupdating-and-the-cli-apocalypse/13794/7?u=allison

---

### Documentation

- [ ] Fresh Produce
- [X] n/a


dangra (approved):
Autoupdate is hard!

PR #2631, opened 2023-07-27
use flaps api for volumes

lillianberryfly:
### Change Summary

What and Why:

Moves volumes operations to use flaps instead of web.

How:

Related to:

---

### Documentation

- [x] Fresh Produce
- [x] n/a


internal/command/migrate_to_v2/migrate_to_v2.go
```diff
@@ -411,7 +411,7 @@ func (m *v2PlatformMigrator) rollback(ctx context.Context, tb *render.TextBlock)
 		for _, vol := range m.createdVolumes {
 			if err := retry.Do(
 				func() error {
-					_, err := m.apiClient.DeleteVolume(ctx, vol.vol.ID, m.appLock)
+					_, err := m.flapsClient.DeleteVolume(ctx, vol.vol.ID)
```
dangra, line 414: Does it fail if the app is locked?

internal/command/migrate_to_v2/pg.go
```diff
@@ -92,20 +92,19 @@ func (m *v2PlatformMigrator) migratePgVolumes(ctx context.Context) error {
...
 			// TODO: make use of https://github.com/superfly/nomad-firecracker/pull/1013
-			input := api.CreateVolumeInput{
-				AppID:     app.ID,
+			input := api.CreateVolumeRequest{
 				Name:      fmt.Sprintf("%s_machines", vol.Name),
 				Region:    region,
-				SizeGb:    vol.SizeGb,
-				Encrypted: vol.Encrypted,
+				SizeGb:    &vol.SizeGb,
+				Encrypted: api.Pointer(vol.Encrypted),
```
dangra, line 99: Can we address this comment here: 
> // TODO: make use of https://github.com/superfly/nomad-firecracker/pull/1013

dangra: or ignore it. I introduced a bug when removing `_machines` suffix and can fix it in another PR.

internal/command/migrate_to_v2/volumes.go
```diff
@@ -61,13 +61,10 @@ func (m *v2PlatformMigrator) migrateAppVolumes(ctx context.Context) error {
...
-		newVol, err := m.apiClient.ForkVolume(ctx, api.ForkVolumeInput{
-			AppID:          m.appFull.ID,
-			SourceVolumeID: vol.ID,
-			MachinesOnly:   true,
-			Remote:         true,
+		newVol, err := m.flapsClient.CreateVolume(ctx, api.CreateVolumeRequest{
+			SourceVolumeID: &vol.ID,
+			MachinesOnly:   api.Pointer(true),
 			Name:           vol.Name,
-			LockID:         m.appLock,
```
dangra, line 70: Is flaps api ignoring the app lock if exists?

PR #2633, opened 2023-07-28
Postgres fork-from fixes

davissp14:
### Change Summary

The postgres `--fork-from` feature will now attempt to resolve the source image.  If we are unable to do so, the latest image will be used by default.

Users can override this by manually specifying the `--image-ref` flag.

### DIFF
--- a/internal/command/postgres/create.go
+++ b/internal/command/postgres/create.go
@@ -230,11 +230,13 @@ func run(ctx context.Context) (err error) {
 			return fmt.Errorf("The target volume size %dGB must be greater than or equal to the volume fork target: %dGB", pgConfig.DiskGb, vol.SizeGb)
 		}
 
+		// Attempt to resolve the image ref from the machine tied to the fork volume.
+		if pgConfig.ImageRef == "" {
+			pgConfig.ImageRef = resolveImageFromForkVolume(vol, machines)
+		}
+
 		// Resolve the fork-from app manager
 		params.Manager = resolveForkFromManager(ctx, machines)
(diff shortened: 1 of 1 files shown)


internal/command/postgres/create.go
```diff
@@ -403,6 +406,7 @@ type ClusterParams struct {
 	Manager    string
 	ForkFrom   string
 	Autostart  bool
+	ImageRef   string
```
matttpt, line 409: Just curious—if I understand right, this shadows the `ImageRef` field in the embedded `PostgresConfiguration` (that gets set from the `--image-ref` option). Is this intended?

internal/command/postgres/create.go
```diff
@@ -590,3 +594,20 @@ func resolveForkFromManager(ctx context.Context, machines []*api.Machine) string
 
 	return flypg.StolonManager
 }
+
+func resolveImageFromForkVolume(vol *api.Volume, machines []*api.Machine) string {
+	// If the volume is not associated with a machine, we can't resolve the image and wil
+	// fallback to the default image.
+	if len(machines) == 0 {
+		return ""
+	}
```
dangra, line 603: not needed ?  the iteration over an empty `machines` array will fallback to last `return ""`

dangra (approved):

lubien (approved):

internal/command/postgres/create.go
davissp14: Yeah, this is wrong.  Let me fix.

PR #2635, opened 2023-07-28
add --require-unique-zone to machine clone

lubien:
### Change Summary

What and Why: You can now do `--volume-require-unique-zone` to create a machine with volume on another zone instead of having first to do `vol create --require-unique-zone` then `machine clone --atach-volume-id`. This will benefit folks trying to get out of capacity issues in builders too.

How: Added a flag `--volume-require-unique-zone` like we do in volumes

Related to: internal discussions about Volumes and Capacity

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -78,7 +78,7 @@ func newClone() *cobra.Command {
 			Description: "Comma separated list of machine ids to watch for. You can use '--standby-for=source' to create a standby for the cloned machine",
 		},
 		flag.Bool{
-			Name:        "require-unique-zone",
+			Name:        "volume-requires-unique-zone",
 			Description: "Require volume to be placed in separate hardware zone from existing volumes. Default false",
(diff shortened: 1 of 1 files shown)


internal/command/machine/clone.go
```diff
@@ -77,6 +77,11 @@ func newClone() *cobra.Command {
 			Name:        "standby-for",
 			Description: "Comma separated list of machine ids to watch for. You can use '--standby-for=source' to create a standby for the cloned machine",
 		},
+		flag.Bool{
+			Name:        "require-unique-zone",
```
dangra, line 81: can we name it `--volume-requires-unique-zone` ?  the amount of flags tend to grow on the `fly machines clone|run` commands

lubien: Ah good call

dangra (approved):

PR #2637, opened 2023-07-31
Remove local test parallelism

billyb2:
### Change Summary

Tests will no longer run parallel using `go test`, but it still will for Github Actions

What and Why:'

So initially, I thought this was a [web issue](https://flyio.discourse.team/t/preflight-tests-failing-with-500-errors/3605), but after spending quite a bit of time with this, I can't be sure. It looks like *something* in fly can't handle mutliple commands runnning at the same time, possibly because some state is overwritten. For now, I think the best way of handling this is just disabling go test concurrency, whhile keeping it in Github Actions.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ test: FORCE
 # to run one test, use: make preflight-test T=TestAppsV2ConfigSave
 preflight-test: build
 	if [ -r .direnv/preflight ]; then . .direnv/preflight; fi; \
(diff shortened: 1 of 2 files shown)


dangra:
Time to completion still looks good 💪  
image

dangra (approved):

PR #2616, opened 2023-07-24
add sentry option when generating a Rails dockerfile

rubys:

PR #2615, opened 2023-07-24
Provision a Sentry project automatically for newly deployed apps

jsierles:
### Change Summary

What and Why:

With this change, anew apps launched on Fly get a free Sentry project associated with them. This is the goal of a Fly.io partnership with Sentry to provide a free Sentry account to Fly.io customers for error tracking.

How:

This PR ensures the provisioning happens, and also sets the `SENTRY_DSN` so VMs will pick them up. 

### Documentation

- [ ] Fresh Produce
- [ ] In superfly/docs, or asked for help from docs team
- [ ] n/a

PR #2639, opened 2023-07-31
Updates to Sentry and PlanetScale extension provisioning

jsierles:
This PR supersedes #2615 and #2616 due to both of those PRs needing to be merged together. Also, some refactoring was required to prevent cyclical package imports and to allow choosing whether to just stage, or stage _and_ deploy secrets.
 
The original changes where:

* Provision Sentry projects automatically for new apps
* Add Sentry instrumentation automatically to Rails apps

Other notable changes:

We also place automatic Sentry instrumentation behind a flag, until it's ready to be released, so it may be tested by engineering teams.

 Sentry instrumentation is added to Rails apps, regardless, but it isn't activated unless `SENTRY_DSN` is set.


internal/command/deploy/deploy_first.go
```diff
@@ -14,14 +16,28 @@ func (md *machineDeployment) provisionFirstDeploy(ctx context.Context, allocPubl
...
 	if err := md.provisionIpsOnFirstDeploy(ctx, allocPublicIPs); err != nil {
-		fmt.Fprintf(md.io.ErrOut, "Failed to provision IP addresses, use `fly ips` commands to remmediate it. ERROR: %s", err)
+		fmt.Fprintf(md.io.ErrOut, "Failed to provision IP addresses. Use `fly ips` commands to remediate it. ERROR: %s", err)
 	}
 	if err := md.provisionVolumesOnFirstDeploy(ctx); err != nil {
 		return fmt.Errorf("failed to provision seed volumes: %w", err)
 	}
+
+	// Provision Sentry on first deployment, unless we're using CI, where it's more likely to see ephemeral applications
+	if !env.IsCI() && md.provisionExtensions {
```
dangra, line 26: what if it sets the default value of `--provision-extensions` to `!env.IsCI()` instead of this? it will behave as we want and still allow testing on CI if needed.

internal/command/errors/errors.go
```diff
@@ -0,0 +1,10 @@
+package errors
+
+import (
+	"github.com/spf13/cobra"
+	sentry_ext "github.com/superfly/flyctl/internal/command/extensions/sentry"
+)
+
+func New() (cmd *cobra.Command) {
+	return sentry_ext.Dashboard()
```
dangra, line 9: wdyt about extending `fly open` to `fly open sentry` or `fly open errors` instead of creating another toplevel command for what is an already polluted command root.

Another option is `fly app errors`

internal/command/deploy/deploy.go
```diff
@@ -47,6 +47,7 @@ var CommonFlags = flag.Set{
 	flag.NoCache(),
 	flag.Nixpacks(),
 	flag.BuildOnly(),
+	flag.ProvisionExtensions(),
```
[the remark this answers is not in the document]
jsierles: This is true. I did this because I thought it needed to be shared with launch, but launch pulls the shared flags from here.

internal/command/errors/errors.go
jsierles: I don't have strong feelings here. I'll move to `apps errors` for now.

jsierles: For now I just removed this until the aliasing can be done in a cleaner fashion. Now aliases have unfortunate help output.

Issue #2646, opened 2023-08-01
Unable to add read regions to upstash redis databases non-interactively via flyctl


MarkIannucci:
**Describe the bug**
We are unable to add read replicas to upstash redis databases non-interactively.

Operating system:
`flyctl v0.1.68 linux/amd64 Commit: 8a1757c2e652cc3c2a4d90280ab6ac021e7b1919-dirty BuildDate: 2023-07-31T19:22:13Z`

**Paste your command**
`flyctl redis create --name "read-replicas-for-me" --enable-eviction --org bifrost --plan 200M --region iad,ord,den`

**Command output:**

`Error: region iad,ord,den not found`

**Additional context:**

It looks like [there's some go code](https://github.com/superfly/flyctl/blob/d6bd4610b8fd06a3643b0da27fadb6958d4cb5f1/internal/command/redis/create.go#L196) which can support this functionality, but there isn't a flag in the `redis create` command to support it.  If I were to submit a PR for this, it would be my first time writing Go.

Our deployment is fully automated except for adding additional replica regions.  It would be great to get to that happy place where we can test against read replicas in our CI pipeline.

PR #2649, opened 2023-08-03
Allow passing replica regions non-interactively for Upstash Redis

jsierles:
This PR adds a `replica-regions` flag to `flyctl redis create/update` to allow fully non-interactive setup of Redis clusters.


MarkIannucci:
closes #2646.

PR #2662, opened 2023-08-07
Add preflight tests for `fly console` and `fly pg import`

matttpt:
This adds basic preflight tests to make sure that `fly console` and `fly pg import` work as expected. The specific impetus for this is to be sure that #2523 doesn't break anything when merged.


dangra (approved):
❤️

PR #2666, opened 2023-08-08
Prevent bluegreen deployments from spamming CI logs

magJ:
### Change Summary

What and Why:

Currently, bluegreen deployments spam CI logs with thousands of lines of redundant status output.
This change only prints deployment progress when state/health changes for non-interactive use.

How:

Keep track of last reported state/health, and only print new state when it changes.

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


magJ:
@gwuah Could you please take a look at this?

magJ:
> But, interactive or not, it doesn't make sense to log+clear the same thing over & over again. So, kindly amend your PR with the requested changes and I'll give my ✅

The problem with your proposed changes, is that `clearLinesAbove` will clear too many lines in interactive mode, and begin eating away at the past terminal output.  

I could add some additional logic to keep track of how many lines should be cleared, but it didn't seem worthwhile, as the interactive mode works fine as-is.

magJ:
@gwuah I added some feedback to your PR. 

Happy to have this PR closed and that one merged, the issue I raised isn't a big deal.
Although I noticed on your PR you mentioned:

> #2666 works fine for 1 machine deployment. With multiple machine deployments it doesn't display the right state.

I figure it's better to ask you about it here rather than in the other PR.

I tested this PR with multiple machine deployments, and it seemed to work fine for me, what state was being shown incorrectly?

PR #2669, opened 2023-08-08
Added the ability to disable autoupdates

### LINKED ISSUE #2614: Allow disabling autoupdate at build-time
**Describe the bug**

In some deployment systems, flyctl will not be able to update itself, yet it tries to. Nixpkgs is one of those that I know of.
Currently this is worked around by patching the default in the source (https://github.com/NixOS/nixpkgs/pull/244494/commits/3dcfe643700283f591f61c57ffb14e0a96fa9c07) but that's not a great solution long-term.

A compile option, so this could be overwritten from go's flags would be very welcome.

billyb2:
### Change Summary

Added a variable to disable autoupdates at build time

What and Why:

Fixes #2614 

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -32,8 +32,15 @@ type Release struct {
 	Timestamp   time.Time `yaml:"timestamp"`
 }
 
+// For disabling auto updates at build time
+var disableAutoUpdate = "false"
+
 // Check reports whether update checks should take place.
 func Check() bool {
+	if disableAutoUpdate == "true" {
+		return false
+	}
+
 	switch {
 	case env.IsTruthy("FLY_UPDATE_CHECK"):
 		return true
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -37,11 +37,9 @@ var disableAutoUpdate = "false"
 
 // Check reports whether update checks should take place.
 func Check() bool {
-	if disableAutoUpdate == "true" {
-		return false
-	}
-
 	switch {
+	case disableAutoUpdate == "true":
+		return false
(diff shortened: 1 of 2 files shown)


internal/update/update.go
```diff
@@ -32,8 +32,16 @@ type Release struct {
 	Timestamp   time.Time `yaml:"timestamp"`
 }
 
+// For disabling auto updates at build time
+var disableAutoUpdate = "false"
+
 // Check reports whether update checks should take place.
 func Check() bool {
+	fmt.Println(disableAutoUpdate)
```
dangra, line 40: extra debug line

internal/update/update.go
```diff
@@ -32,8 +32,15 @@ type Release struct {
...
 }
 
+// For disabling auto updates at build time
+var disableAutoUpdate = "false"
+
 // Check reports whether update checks should take place.
 func Check() bool {
+	if disableAutoUpdate == "true" {
+		return false
+	}
```
dangra, line 42: this test can be blended into the `switch` statement below and would be more idiomatic

billyb2: 👍🏾

dangra (approved):

michaeldwan:
Hey before we merge this, what's the use case? We're not going to do custom builds for different package managers. Is this for the nix people?

billyb2:
Exactly that. Some environments don't do well with binaries updating themselves, and having a general way for people packaging flyctl to disable it is good IMO, especially considering it's just a few lines of code. I can remove the addition to the Makefile, since just having the variable is enough.

PR #2673, opened 2023-08-08
Return a better "could not reserve resource" error

billyb2:
-------

### Change Summary

What and Why:

The error being returned when deploying to an existing machine, "could not reserve resource", can be confusing and doesn't give clear indication as to how to continue.

How:

Intercept "could not reserve resource" when attempting a second deploy to the region, and suggest deploying to a new region.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -408,7 +408,11 @@ func (md *machineDeployment) updateMachineByReplace(ctx context.Context, e *mach
 
 	newMachineRaw, err := md.flapsClient.Launch(ctx, *e.launchInput)
 	if err != nil {
-		return err
+		if strings.Contains(err.Error(), "could not reserve resource for machine") {
+			return errors.New("The region you're trying to deploy is likely at capacity. Consider deploying to a new region with fly deploy --region <region> or trying again in a few hours\n")
+		} else {
+			return err
+		}
 	}
 
 	lm = machine.NewLeasableMachine(md.flapsClient, md.io, newMachineRaw)


rugwirobaker (approved):
LGTM

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -408,7 +408,11 @@ func (md *machineDeployment) updateMachineByReplace(ctx context.Context, e *mach
 
 	newMachineRaw, err := md.flapsClient.Launch(ctx, *e.launchInput)
 	if err != nil {
-		return err
+		if strings.Contains(err.Error(), "could not reserve resource for machine") {
```
michaeldwan, line 411: Is the error coming back from flaps a generic string, or does it have a type? If it's just a generic error, switch it to a type so we can do proper error inspection with `errors.Is` or `errors.As`. Otherwise if the error message changes at all this code won't get hit anymore.

shortdiv (approved):

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -408,7 +408,11 @@ func (md *machineDeployment) updateMachineByReplace(ctx context.Context, e *mach
...
+			return errors.New("The region you're trying to deploy is likely at capacity. Consider deploying to a new region with fly deploy --region <region> or trying again in a few hours\n")
```
michaeldwan, line 412: This is a good opportunity to add improved messaging and maybe a link to docs like we talked about in Slack.

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -408,7 +408,11 @@ func (md *machineDeployment) updateMachineByReplace(ctx context.Context, e *mach
...
+		} else {
+			return err
+		}
```
dangra, line 415: oh wait, the place for this is https://github.com/superfly/flyctl/blob/965aba6d8e011302f48a039e632a5575e20653f6/internal/command/deploy/machines_deploymachinesapp.go#L355-L360

michaeldwan (commented):
So this is merged, but let's consider some changes. We need to make sure this works:
- from any other place in the stack that might throw a capacity error, like resuming a machine, scaling resources, custom operations from postgres or litefs, etc. You can't predict these things now, but you can create a standard error type that future code uses. And if you work with error types, this improvement will keep working as intended as long as future code uses the right errors
- if the error message changes this help text should still be shown to the user

internal/command/deploy/machines_deploymachinesapp.go
dangra: good point +1

billyb2:
i'm gonna revert the PR, to address the comments

internal/command/deploy/machines_deploymachinesapp.go
michaeldwan: For posterity, here's some interfaces we have that lets us tag a custom error type so the downstream error handlers can render them, taking into account a lot more than we can do inside random functions: https://github.com/superfly/flyctl/blob/master/internal/flyerr/flyerr.go#L14-L40

If an error has any meaning to flyctl at all, it needs a custom error type. Even if it's `var CustomError = errors.New("boom")` in a package. That'll let us refer to an error as a type, which makes everything a lot easier.

PR #2675, opened 2023-08-09
Handle docker images >2GB

benwaffle:
Use a new compressedSizeFull BigInt field that can handle numbers > 2147483647, to support docker images >2GB


dangra (approved):
there is a schema.graphql file on the root that could use an update too

PR #2678, opened 2023-08-09
`fly deploy`: don't wait for standby machines

matttpt:
### Change Summary

What and Why:

The standby check appears to have been misplaced relative to the `WaitForState` call.

Related to:

* Should fix an apparent bug introduced in #2660.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


dangra (approved):
My bad ✋  thanks

PR #2679, opened 2023-08-09
Concurrently update and wait per updated machine on rolling deploys

dangra:
### Change Summary

What and Why: Rolling deploys uses concurrent machine updates but waits sequentially for machines to reach its started state, pass smoke and health checks. This is faster than previous all sequential approach but still inconvenient for apps with a large number of machines. It updates 33% of machines at the same time but doesn't consider process group, making them vulnerable to complete group downtime.

How: Switch to a pool based solution that will update and wait for 33% of machines per process group in parallel. 

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


internal/command/deploy/strategy_rolling.go
```diff
@@ -0,0 +1,171 @@
...
+			indexStr,
+			md.colorize.Bold(lm.FormattedMachineId()),
+			md.colorize.Green("success"),
+		)
+		if inBatch {
+			fmt.Fprint(md.io.ErrOut, "\n")
+		}
+	}
+
+	md.warnAboutIncorrectListenAddress(ctx, lm)
```
pborzenkov, line 169: This function may modify `md.listenAddressChecked` map. Should the map now be protected with a mutex?

PR #2683, opened 2023-08-10
Extension provisioning updates

jsierles:
This PR addresses:

* Display of links to extension provider ToS
* Displaying of SSO links/commands at provisioning time
* Copy and clarity improvements for the overall provisioning experience
* Special treatment for auto-provisioned extensions like Sentry


rubys:
While I don't understand all of this (example: coupons?), apparently sentry requires an agreement to ToS and Rails will automatically default to including sentry.  Based on this understanding, my one point of feedback is that if the person running fly launch declines the ToS, we shouldn't configure sentry.

Scanner callbacks are passed a list of options selected:

https://github.com/superfly/flyctl/blob/eb18df0ba905e98d881b2ca1f968c19285e650bc/scanner/rails.go#L160

And these options are used to determine what flags to pass to dockerfile-rails:

https://github.com/superfly/flyctl/blob/eb18df0ba905e98d881b2ca1f968c19285e650bc/scanner/rails.go#L216-L227

jsierles:
Coupons was just a leftover from not having merged changes to the schema from`main`.

Regarding ToS, we've decided to make agreement implicit in our own ToS. We'll display copy to this effect once per org, to avoid interrupting the deployment flow.

PR #2684, opened 2023-08-11
`bluegreen`: return errors properly

alichay:
### Change Summary

What and Why: Bluegreen would print errors but not return an error code unless rollback _also_ failed. Now we return the error. (this probably double-prints the error message, but we can refine that later)

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -323,7 +323,10 @@ func (md *machineDeployment) updateUsingBlueGreenStrategy(ctx context.Context, u
 	bg := BlueGreenStrategy(md, updateEntries)
 	if err := bg.Deploy(ctx); err != nil {
 		fmt.Fprintf(md.io.ErrOut, "Deployment failed after error: %s\n", err)
-		return bg.Rollback(ctx, err)
+		if rollbackErr := bg.Rollback(ctx, err); rollbackErr != nil {
+			return rollbackErr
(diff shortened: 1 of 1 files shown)


internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -323,7 +323,10 @@ func (md *machineDeployment) updateUsingBlueGreenStrategy(ctx context.Context, u
 	bg := BlueGreenStrategy(md, updateEntries)
 	if err := bg.Deploy(ctx); err != nil {
 		fmt.Fprintf(md.io.ErrOut, "Deployment failed after error: %s\n", err)
-		return bg.Rollback(ctx, err)
+		if rollbackErr := bg.Rollback(ctx, err); rollbackErr != nil {
+			return rollbackErr
+		}
+		return err
```
michaeldwan, line 329: Do we still need the original error in the event that rollback failed?

alichay: What do you mean?

michaeldwan: Since `err` is returned if deployment fails, but `rollbackErr` is returned if deployment failed _and_ rollback failed, I was curious if anything upstream depended on `err` being returned. But I see that the old code didn't return it at all, so I guess not!

billyb2 (approved):

Issue #2543, opened 2023-07-09
Flyctl y/N prompt are not visible in light theme


tshu-w:
**Describe the bug**
In the light theme, the y/N prompt in Flyctl is not visible, making it difficult to interact with the prompt. This issue can cause confusion and hinder the user experience, especially for users who prefer the light theme. It would be helpful to address this and ensure that the y/N prompt is properly visible in both light and dark themes.

* Operating system: macOS 13.4.1
* `fly version`: fly v0.1.52 darwin/arm64 Commit: 9928dce858ee7d597f7af9068fca6e72893f3c56-dirty BuildDate: 2023-07-07T23:13:50Z

** Paste your `fly.toml`

```toml
Default
```

** Command output: **
<img width="396" alt="Screenshot 2023-07-09 at 15 38 36" src="https://github.com/superfly/flyctl/assets/13161779/6d3d2e6b-d77c-4188-af21-a083ca6c7de6">

usrme:
First, I'm not affiliated with Fly in any way, but this mimics the behavior of the dependency that they use, `go-survey/survey`:

![image](https://github.com/superfly/flyctl/assets/5902545/557de86c-bb52-4115-9b6b-52751a2448ec)

Which itself uses the ANSI color code `white` for confirmation prompts: https://github.com/go-survey/survey/blob/fa37277e6394c29db7bcc94062cb30cd7785a126/confirm.go#L33. This obviously breaks down in light themes. 

Unfortunately `go-survey/survey` only provides a way to [change the format and color of the icons](https://github.com/go-survey/survey#changing-the-icons) preceding the messages and it seems that the project isn't _that_ actively maintained. Just changing the aforementioned `white` to `default`, to match whatever the terminal emulator theme's default color is, would fix this though:

![image](https://github.com/superfly/flyctl/assets/5902545/33a80f37-ddd4-4a4b-a070-b7dbb2f6b303)

[...]

Issue #1831, opened 2023-03-04
Usability difficulties in light terminals due to inappropriate use of fixed greys


chris-morgan:
* Operating system: Arch Linux
* `fly version`: flyctl v0.0.475 linux/amd64 Commit: a5acc104 BuildDate: 2023-03-03T18:38:39Z

I use a light terminal, and there are a couple of places where colour is used inappropriately in flyctl, making text difficult to read.

1.  During `fly launch` on a Rails app, in the “? Would you like to set up a Postgresql database now? (y/N) ” prompt, the text “(y/N)” can be *completely invisible*.
    I’ve diagnosed the root cause and reported this (well, commented on a bug that was unreasonably closed in a needinfo kind of way three years ago) in https://github.com/go-survey/survey/issues/258#issuecomment-1454725104.
    Figured it was still worth mentioning since I’m mentioning the second issue, which *does* belong in flyctl.

2.  Various messages are printed using hard-coded shades of grey, or potentially “bright black” which is unreliable.
    This is bad because you don’t know what the default background colour is, so the colour chosen could be high-contrast, low-contrast or even zero-contrast.
    For example, in this output from `fly deploy`:

    ```
    ==> Verifying app config
    --> Verified app config
    ==> Building image
    Remote builder ‹redacted› ready
    ==> Creating build context
    --> Creating build context done
    ==> Building image with Docker
    --> docker host: 20.10.12 linux x86_64
    ```

    The `==>` lines are green (colour 2, which is by custom safe to use),
    and the `-->` lines a very light grey (colour 252, which will almost always be #d0d0d0).

    This case comes from [internal/cmdfmt/messages.go:16](https://github.com/superfly/flyctl/blob/5881257a7e0d8fc9593d3381c3a87f07fda6615b/internal/cmdfmt/messages.go#L16).
    On dark colour schemes, #d0d0d0 is likely to be fairly close to to the default foreground colour (e.g. VS Code and xterm’s default schemes use #e5e5e5, Windows 10 Console’s uses #cccccc, Alacritty’s #c5c8c6).
    The difference may or may not be noticeable, and may be higher or lower contrast.
    This is clearly bad.
    On light schemes, it’ll regularly be almost completely invisible.
    This is clearly very bad.

[...]

PR #2691, opened 2023-08-14
fix(#2543): override color 'white' to avoid usability issues

usrme:
### Change Summary

What and Why:

The color `white` from `survey` is overridden to avoid usability issues in lighter console emulator themes.

How:

I saw how the GitHub CLI approached this issue [here](https://github.com/cli/cli/blob/d9ea221b1a8b1af7856ed4896c92f67acbf295da/cmd/gh/main.go#L74-L81) and used the same approach. While the issue in question was #2543 and related to confirmation prompts, I believe this change should be done outside of the `Confirm` function under `internal/prompt` to avoid any other similar issues where `white` is used instead of the more sane `default`.

Related to:

- https://github.com/superfly/flyctl/issues/1831
- https://github.com/superfly/flyctl/issues/2543

### DIFF
--- a/internal/prompt/prompt.go
+++ b/internal/prompt/prompt.go
@@ -10,9 +10,10 @@ import (
 	"sync"
 
 	"github.com/AlecAivazis/survey/v2"
+	surveyCore "github.com/AlecAivazis/survey/v2/core"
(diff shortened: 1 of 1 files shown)


dangra:
Thanks!

PR #2692, opened 2023-08-14
Better Error Messages

billyb2:
This PR is directly related to [this slab post](https://flyio.slab.com/posts/flyctl-errors-are-for-humans-b1b4houk)

-------

### Change Summary

TL;DR: Errors like this would help users (and us) with debugging, as well as helping to stabliize error messages for the CLI. [Trying to follow the CLI dev guide](https://clig.dev/) as well as possible

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -103,8 +103,10 @@ func isUnchangedError(err error) bool {
 }
 
 func printError(io *iostreams.IOStreams, cs *iostreams.ColorScheme, cmd *cobra.Command, err error) {
-	fmt.Fprintf(io.ErrOut, cs.Red("Error: "))
-	fmt.Fprintln(io.Out, err.Error())
+	fmt.Fprint(io.ErrOut, cs.Red("Error: "))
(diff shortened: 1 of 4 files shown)


michaeldwan:
At a high level, this looks good. I don't think the "Fault" concept is necessary though. That's more of a thought process for us to run through when deciding how to communicate issues (and allocate energy to avoiding them) rather than a runtime decision. 

Somewhat related... [this PR](https://github.com/superfly/flyctl/pull/2695/files) is another example of when a suggestion and help url would improve the error ux. Perhaps a good next step is picking a common error that uses `errors.New` or `fmt.Errorf` today and replacing it with an error type that takes advantage of these interfaces. Thoughts?

billyb2:
Sounds like a plan. I'll remove the Fault code and put the PR up for review.

alichay (approved):
this is wonderful! If you've tested this, it looks good to me :)

PR #2694, opened 2023-08-14
Ignore destroyed volumes

matttpt:
#### What and Why:

Now that we use the Machines API to work on volumes, the volume states that we receive correspond directly to the volume states in `flyd`. Several of these states are for volumes that have been effectively "destroyed" but aren't truly destroyed yet. We should filter out volumes in these states so that we don't try to use them.

#### How:

`GetVolumes` now filters out destroyed volumes, so the list of returned volumes is like what we previously got from GraphQL. `GetAllVolumes` provides an unfiltered version.

#### Related:

* https://community.fly.io/t/trying-to-bring-up-a-machine-after-deleting-its-volume/14806

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/flaps/flaps_volumes.go
+++ b/flaps/flaps_volumes.go
@@ -5,15 +5,19 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/samber/lo"
 	"github.com/superfly/flyctl/api"
+	"golang.org/x/exp/slices"
 )
+
+var destroyedVolumeStates = []string{"scheduling_destroy", "fork_cleanup", "waiting_for_detach", "pending_destroy", "destroying"}
 
 func (f *Client) sendRequestVolumes(ctx context.Context, method, endpoint string, in, out interface{}, headers map[string][]string) error {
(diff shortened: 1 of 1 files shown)


flaps/flaps_volumes.go
```diff
@@ -25,6 +26,21 @@ func (f *Client) GetVolumes(ctx context.Context) ([]api.Volume, error) {
...
 }
 
+func (f *Client) GetVolumes(ctx context.Context) ([]api.Volume, error) {
+	volumes, err := f.GetAllVolumes(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return lo.Filter(volumes, func(v api.Volume, _ int) bool {
+		switch v.State {
+		case "scheduling_destroy", "fork_cleanup", "waiting_for_detach", "pending_destroy", "destroying":
```
jipperinbham, line 36: Could also use the slices.Contains function and make the slices of invalid states a const/var.

jipperinbham (approved):

PR #2700, opened 2023-08-16
Various extension provisioning updates

jsierles:
In this PR we:

Introduce the `fly apps errors` as a more memorable alias of `fly ext sentry create`. Only "default" extensions like Sentry or PlanetScale get special command aliases like this.

We allow CI environments to auto-provision Sentry projects. The thought is that provisioning in CI is low risk, and could be useful for some customers. To prevent this behavior in CI, one may now pass `--no-extensions` to disable auto-provisioning.

We prevent beta extensions from being auto-provisioned if an org is not allow-listed.

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -360,7 +360,7 @@ func deployToMachines(
 		AllocPublicIP:         !flag.GetBool(ctx, "no-public-ips"),
 		UpdateOnly:            flag.GetBool(ctx, "update-only"),
 		Files:                 files,
-		ProvisionExtensions:   flag.GetBool(ctx, "provision-extensions"),
+		ExcludeRegions:        excludeRegions,
 		NoExtensions:          flag.GetBool(ctx, "no-extensions"),
 		OnlyRegions:           onlyRegions,
 	})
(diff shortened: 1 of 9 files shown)


internal/command/deploy/deploy.go
```diff
@@ -357,7 +361,7 @@ func deployToMachines(
 		UpdateOnly:            flag.GetBool(ctx, "update-only"),
 		Files:                 files,
 		ProvisionExtensions:   flag.GetBool(ctx, "provision-extensions"),
-		ExcludeRegions:        excludeRegions,
```
dangra, line 360: bad merge ?

internal/command/deploy/machines.go
```diff
@@ -53,6 +53,7 @@ type MachineDeploymentArgs struct {
 	UpdateOnly            bool
 	Files                 []*api.File
 	ProvisionExtensions   bool
+	NoExtensions          bool
```
dangra, line 56: it needs a corresponding entry in: https://github.com/superfly/flyctl/blob/870516e939379bf565e51fb2338f1eee96b44cff/internal/command/deploy/machines.go#L147-L170

otherwise the value of --no-extensions is ignored

dangra (changes requested):

PR #2704, opened 2023-08-16
Add a --debug flag that prints stacktraces

billyb2:
### Change Summary

What and Why:

Being able to print stack traces upon encountering an error is *very* useful for debugging.

How:

This adds a new `--debug` flag that prints a stack trace if an error is encountered

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/flag/flagnames/constants.go
+++ b/internal/flag/flagnames/constants.go
@@ -13,6 +13,9 @@ const (
 	// LocalOnly denotes the name of the local-only flag.
 	LocalOnly = "local-only"
 
+	// Deub denotes the name of the debug flag.
+	Debug = "debug"
+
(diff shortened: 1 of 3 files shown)


internal/flag/flagnames/constants.go
```diff
@@ -13,6 +13,9 @@ const (
 	// LocalOnly denotes the name of the local-only flag.
 	LocalOnly = "local-only"
 
+	// Deub denotes the name of the debug flag.
```
dangra, line 16: typo

billyb2: Whoops

dangra (approved):

PR #2703, opened 2023-08-16
`launch` v2: crossing the finish line

alichay:
Finishing touches for Launch V2.

 - [x] Split db providers out of Plan proper, into db-specific plans to make room for other providers
 - [x] Provision databases based on the plan
 - [ ] Finish upstash redis support (blocked)
 - [x] Toggleable high-availability

PR #2711, opened 2023-08-17
scan for db, redis in rails

rubys:
In support of the "launch v2" effort

### DIFF
--- a/scanner/rails.go
+++ b/scanner/rails.go
@@ -76,9 +76,32 @@ func configureRails(sourceDir string, config *ScannerConfig) (*SourceInfo, error
 		AutoInstrumentErrors: true,
 	}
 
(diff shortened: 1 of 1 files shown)


alichay (approved):
The other paths are pretty self-explanatory, but does litestack imply LiteFS?

rubys:
> does litestack imply LiteFS?

No, and there is no way to scan the source and determine if litefs is wanted.  But if litefs is selected (presumably by ui-ex) that could be passed as an option (or a part of the plan) and the dockerfile generator can make that happen.

litestack is something that will configure rails to use sqlite for everything including web sockets and background jobs.

alichay:
merged into my branch @ #2703, ty!

PR #2714, opened 2023-08-18
Rename the auto-confirm flag to yes

billyb2:
### Change Summary

What and Why:

This makes auto-confirming much more consistent across the CLI. The tests were purposely not changed to test backwards compatibility.

How:

Use `flag.Yes()` everywhere, and make `--auto-confirm` and alias for `--yes` so it shouldn't break anything

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -60,7 +60,7 @@ var CommonFlags = flag.Set{
 		Shorthand:   "e",
 		Description: "Set of environment variables in the form of NAME=VALUE pairs. Can be specified multiple times.",
 	},
-	flag.Yes().SetDescription("Will automatically confirm changes when running non-interactively."),
+	flag.Yes(),
 	flag.Int{
 		Name:        "wait-timeout",
 		Description: "Seconds to wait for individual machines to transition states and become healthy.",
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -60,10 +60,7 @@ var CommonFlags = flag.Set{
 		Shorthand:   "e",
 		Description: "Set of environment variables in the form of NAME=VALUE pairs. Can be specified multiple times.",
 	},
-	flag.Bool{
-		Name:        "auto-confirm",
-		Description: "Will automatically confirm changes when running non-interactively.",
-	},
+	flag.Yes().SetDescription("Will automatically confirm changes when running non-interactively."),
 	flag.Int{
 		Name:        "wait-timeout",
 		Description: "Seconds to wait for individual machines to transition states and become healthy.",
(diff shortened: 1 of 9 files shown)


michaeldwan:
The change looks good, but I'd suggest using `--force` rather than `--yes` since it's usually the flag to do something potentially destructive without confirmation. `--yes` can be an alias. What do you think?

internal/flag/flag.go
```diff
@@ -327,9 +332,15 @@ func Yes() Bool {
 		Name:        flagnames.Yes,
 		Shorthand:   "y",
 		Description: "Accept all confirmations",
+		Aliases:     []string{"auto-confirm"},
 	}
 }
 
+// Checks whether the --yes flag or the --auto-confirm flag is present
+func HasYes(ctx context.Context) bool {
```
dangra, line 340: The convention we have been following so far is to call the function `flag.GetFlagName()` for `flag.FlagName()`, in this case it should be `flag.GetYes()`

billyb2:
> The change looks good, but I'd suggest using `--force` rather than `--yes` since it's usually the flag to do something potentially destructive without confirmation. `--yes` can be an alias. What do you think?

Hm, I personally prefer `--yes` since it's already the convention in flyctl (as well as many other CLIs), though I see the appeal of `--force`. If you think force would be better though, I'll change it to that.

michaeldwan:
> Hm, I personally prefer `--yes` since it's already the convention in flyctl (as well as many other CLIs), though I see the appeal of `--force`. If you think force would be better though, I'll change it to that.

Okay I've changed my mind... `--yes` makes sense it's only for skipping confirmation prompts, which appears to be how we're using `--yes` and `--force` right now. So sticking with yes makes sense. Carry on!

internal/command/deploy/deploy.go
```diff
@@ -60,10 +60,7 @@ var CommonFlags = flag.Set{
 		Shorthand:   "e",
 		Description: "Set of environment variables in the form of NAME=VALUE pairs. Can be specified multiple times.",
 	},
-	flag.Bool{
-		Name:        "auto-confirm",
-		Description: "Will automatically confirm changes when running non-interactively.",
-	},
+	flag.Yes().SetDescription("Will automatically confirm changes when running non-interactively."),
```
michaeldwan, line 63: Is there a benefit to changing the flag description for specific commands, or can we use the same flag description for everything? eg "Skip confirmation prompt"

billyb2: No not necessarily, I mostly just changed it so this didn't change much, but honestly it doesn't matter much.

michaeldwan (approved):
This is excellent. Do you think you can come back and update the commands that currently have a `--force` flag (if it makes sense to do so)?

PR #2712, opened 2023-08-17
Improve errors for machine wait timeouts

billyb2:
-------

### Change Summary

What and Why:

Improved the messaging when receiving a timeout error after deploying a machine.

How:

Implement some FlyErr things

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] In superfly/docs, or asked for help from docs team
- [x] n/a

PR #2718, opened 2023-08-21
'fly secrets deploy' command for deploying staged secrets

nbw:
### What and Why:

A `fly secrets deploy` command to deploy staged secrets. It would be useful to have a way to deploy secrets using the current deployed app config (i.e: without having to cut a new release which requires the code, etc.). There are some follow up features that I'm exploring (in the UI) that would use this command.

To explain a bit more, it's possible to stage secrets (for machine apps) using `fly secrets set --staging FOO=BAR` but the only way to deploy them is using `fly deploy` which cuts a new release. It would be useful to have a way to deploy secrets _without having to create a new release_, which is exactly what `fly secrets set FOO=BAR` does already when it's not staged:

https://github.com/superfly/flyctl/blob/b61fe0fa4ef70f2a97aac8abf0d23cb5a3d4c50e/internal/command/secrets/secrets.go#L105-L111

### How:

This PR doesn't reinvent the wheel. It simply uses the existing [`DeploySecrets`](https://github.com/superfly/flyctl/blob/b61fe0fa4ef70f2a97aac8abf0d23cb5a3d4c50e/internal/command/secrets/secrets.go#L81) function, which is normally called _after_ unsetting/setting secrets. 

### Caveats:

- only works with machine apps
- requires that at least one deploy has been run prior

## Testing

I've omitted my app name, machine id's, etc. 

```bash
> fly secrets set --stage -a lonely-secret TEST=FOO

Secrets have been staged, but not set on VMs. Deploy or update machines in this app for the secrets to take effect.

> go run . secrets deploy -a [app name]
INFO Using wait timeout: 2m0s lease timeout: 13s delay between lease refreshes: 4s

Updating existing machines in '[app name]' with rolling strategy
  [1/2] Machine ******************* [app] update finished: success
  [2/2] Machine ******************* [app] update finished: success
  Finished deploying

> fly ssh console -a [app name]
Connecting to fdaa:1:***:***:***:****:****:*... complete
root@******************:/usr/src/app# env | grep TEST
TEST=FOO
```

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/secrets/deploy.go
+++ b/internal/command/secrets/deploy.go
@@ -1,0 +1,59 @@
+package secrets
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/flaps"
+	"github.com/superfly/flyctl/internal/appconfig"
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/flag"
+)
+
+func newDeploy() (cmd *cobra.Command) {
+	const (
+		long  = `Deploy staged secrets for an application`
+		short = long
+		usage = "deploy [flags]"
+	)
+
+	cmd = command.New(usage, short, long, runDeploy, command.RequireSession, command.RequireAppName)
+
+	flag.Add(cmd,
+		sharedFlags,
+	)
+
+	return cmd
+}
+
+func runDeploy(ctx context.Context) (err error) {
+	client := client.FromContext(ctx).API()
+	appName := appconfig.NameFromContext(ctx)
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		return err
+	}
+
+	flapsClient, err := flaps.New(ctx, app)
+	if err != nil {
+		return fmt.Errorf("could not create flaps client: %w", err)
+	}
+
+	machines, _, err := flapsClient.ListFlyAppsMachines(ctx)
+	if err != nil {
+		return err
+	}
+
+	if !app.Deployed && len(machines) == 0 {
+		return errors.New("before using 'fly secrets deploy', you must first deploy your app at least once using 'fly deploy'")
+	}
+
+	if app.PlatformVersion != appconfig.MachinesPlatform {
+		return errors.New("secrets deploy is only supported for machine apps")
+	}
+
+	return DeploySecrets(ctx, app, flag.GetBool(ctx, "stage"), flag.GetBool(ctx, "detach"))
+}
--- a/internal/command/secrets/deploy.go
+++ b/internal/command/secrets/deploy.go
@@ -2,7 +2,6 @@ package secrets
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/spf13/cobra"
@@ -11,6 +10,7 @@ import (
 	"github.com/superfly/flyctl/internal/appconfig"
 	"github.com/superfly/flyctl/internal/command"
 	"github.com/superfly/flyctl/internal/flag"
+	"github.com/superfly/flyctl/internal/flyerr"
 )
 
 func newDeploy() (cmd *cobra.Command) {
@@ -50,12 +50,19 @@ func runDeploy(ctx context.Context) (err error) {
 	}
 
 	if app.PlatformVersion != appconfig.MachinesPlatform {
-		return errors.New("secrets deploy is only supported for machine apps")
+		return flyerr.GenericErr{
+				Err:      "secrets deploy is only supported for machine apps",
+		}
 	}
 
-	if !app.Deployed && len(machines) == 0 {
-		return errors.New("before using 'fly secrets deploy', you must first deploy your app at least once using 'fly deploy'")
+	if !(app.Deployed && len(machines) > 0) {
+		return flyerr.GenericErr{
+			Err:      "no machines available to deploy",
+			Descript: "'fly secrets deploy' will only work if the app has been deployed and there are machines available",
+			Suggest:  "Try 'fly deploy' first",
+		}
 	}
 
 	return DeploySecrets(ctx, app, false, flag.GetBool(ctx, "detach"))
 }
+
--- a/internal/command/secrets/deploy.go
+++ b/internal/command/secrets/deploy.go
@@ -23,7 +23,9 @@ func newDeploy() (cmd *cobra.Command) {
 	cmd = command.New(usage, short, long, runDeploy, command.RequireSession, command.RequireAppName)
 
 	flag.Add(cmd,
-		sharedFlags,
+		flag.App(),
+		flag.AppConfig(),
+		flag.Detach(),
 	)
 
 	return cmd
@@ -55,5 +57,5 @@ func runDeploy(ctx context.Context) (err error) {
 		return errors.New("secrets deploy is only supported for machine apps")
 	}
 
-	return DeploySecrets(ctx, app, flag.GetBool(ctx, "stage"), flag.GetBool(ctx, "detach"))
+	return DeploySecrets(ctx, app, false, flag.GetBool(ctx, "detach"))
 }
--- a/internal/command/secrets/deploy.go
+++ b/internal/command/secrets/deploy.go
@@ -41,7 +41,9 @@ func runDeploy(ctx context.Context) (err error) {
 
 	flapsClient, err := flaps.New(ctx, app)
 	if err != nil {
-		return fmt.Errorf("could not create flaps client: %w", err)
+		return flyerr.GenericErr{
+			Err:      fmt.Sprintf("could not create flaps client: %w", err),
+		}
 	}
 
 	machines, _, err := flapsClient.ListFlyAppsMachines(ctx)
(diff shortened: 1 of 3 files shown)


billyb2 (changes requested):
This is a great overall change, just two quick notes:
1. Adding a test for `fly secrets deploy` to make sure there aren't any regressions
2. using a error type that implements some of the flyerr interfaces. [here's an example](https://github.com/superfly/flyctl/pull/2712/files). This is more optional, but I'm trying to make it a standard around flyctl. How I want it to work is pretty poorly documented (which I'll fix), but essentially you have a error type that returns a static error message, a suggestion for said error (if applicable), a documentation URL (if applicable), and a description of the error that's meant more for debugging. If you'd like, I could implement the error and push to this branch.
3. A subsequent PR to our [docs on runtime secrets](https://fly.io/docs/reference/secrets/) would be really helpful to users, alongside a fresh produce :)

This is a great PR, it just needs a test to prevent regressions and you've got a ✅ from me

billyb2 (approved):

michaeldwan:
What's the reason for not wanting to create a new release? As I understand it, any change to the config should create a new release. Could this create problems where the same release version is running but the runtime behavior is actually different?

dangra:
> What's the reason for not wanting to create a new release? As I understand it, any change to the config should create a new release. Could this create problems where the same release version is running but the runtime behavior is actually different?

I agree, but I think there is a confusion. 

Updating secrets without `--stage` flag triggers a _restart only_ deployment which *creates a new release*, although it doesn't rebuild the image from source code. 

This change  is exposing `DeploySecrets` function as a command, IMO that is a good thing for users.

internal/command/secrets/deploy.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return fmt.Errorf("could not create flaps client: %w", err)
+	}
+
+	machines, _, err := flapsClient.ListFlyAppsMachines(ctx)
+	if err != nil {
+		return err
+	}
+
+	if !app.Deployed && len(machines) == 0 {
```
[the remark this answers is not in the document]
dangra: do we need this check here? In fact, if we do it should be after the `app.PlatformVersion == "machines"` check

internal/command/secrets/deploy.go
```diff
@@ -0,0 +1,59 @@
...
+		return errors.New("before using 'fly secrets deploy', you must first deploy your app at least once using 'fly deploy'")
+	}
+
+	if app.PlatformVersion != appconfig.MachinesPlatform {
+		return errors.New("secrets deploy is only supported for machine apps")
+	}
+
+	return DeploySecrets(ctx, app, flag.GetBool(ctx, "stage"), flag.GetBool(ctx, "detach"))
```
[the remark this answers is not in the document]
dangra: +1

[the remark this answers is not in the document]
nbw: yeah in hindsight that makes total sense. good catch

nbw:
@michaeldwan @dangra 

> What's the reason for not wanting to create a new release? 

Creating a new release requires having the code available. I would like to be able to trigger a redeploy without having the code available. This would enable us to allow setting/unsetting secrets from the UI (which we have now, but the secrets are only staged). Effectively this is a "redeploy" rather than "new deploy"

internal/command/secrets/deploy.go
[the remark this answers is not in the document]
nbw: > Should this maybe be an or?

So that logic taken from DeploySecrets, but the error messaging there is not helpful there.

```
# DeploySecrets
	if !app.Deployed && len(machines) == 0 {
		fmt.Fprintln(out, "Secrets are staged for the first deployment")
		return nil
	}
``` 

>  it should be after the app.PlatformVersion == "machines" check

Yeah, that would cover our bases more and make the && make more sense.

nbw:
Thanks for the run-through Daniel. I might have misspoke by saying "Creating a new release requires". I  meant using `fly deploy`.

internal/flyerr/flyerr.go
```diff
@@ -90,6 +90,7 @@ func PrintCLIOutput(err error) {
 	suggestion := GetErrorSuggestion(err)
 
 	if description != "" {
+		fmt.Println()
```
nbw, line 93: @billyb2 I added a new line here.

I don't have my GOPATHs configured correctly (or I just don't understand how Golang works anymore) so I couldn't test this locally because my code is importing `"github.com/superfly/flyctl/internal/flyerr"` from some other path (or from source), but 

I'm hoping:
```
> go run . secrets deploy -a secrets-deploy
Error: no machines available to deploy
'fly secrets deploy' will only work if the app has been deployed and there are machines available

Try 'fly deploy' first
```

will be:

```
> go run . secrets deploy -a secrets-deploy
Error: no machines available to deploy

'fly secrets deploy' will only work if the app has been deployed and there are machines available

Try 'fly deploy' first
```

nbw:
@billyb2 I used `flyerr` (left a comment about a new line change), but I didn't get around to writing tests. I'm not familiar enough to write tests for this module in a short amount of time. If there was already an example in secrets then I'd start there, but since there isn't I'm hoping we can skip tests this time.

👋  could I get a final review pass. thanks.  

I'll word on generating/adding docs after this is merged.

dangra:
> 👋 could I get a final review pass. thanks.

Please get the errors reported by CI addressed.

I think the change is good for users, so +1 to merge. 
Will approve once CI is passing.

internal/flyerr/flyerr.go
billyb2:
```
Error: no machines available to deploy
'fly secrets deploy' will only work if the app has been deployed and there are machines available

Try 'fly deploy' first
```

Very nice!

internal/command/secrets/deploy.go
```diff
@@ -0,0 +1,69 @@
...
+	machines, _, err := flapsClient.ListFlyAppsMachines(ctx)
+	if err != nil {
+		return err
+	}
+
+	if app.PlatformVersion != appconfig.MachinesPlatform {
+		return flyerr.GenericErr{
+			Err: "secrets deploy is only supported for machine apps",
+		}
+	}
```
dangra, line 58: This block of code should happen right after getting the AppCompact at line 40.

internal/flyerr/flyerr.go
```diff
@@ -90,6 +90,7 @@ func PrintCLIOutput(err error) {
...
 		fmt.Printf("\n%s", description)
```
dangra, line 94: not sure what is happening here 🤔 
If you want an extra line, just prefix the string format with another `\n`, isn't it?

nbw: yeah, that makes way more sense. I was following the pattern already existing in this file, but yes. you're correct. good catch.

dangra (approved):
LG. My only caveat is that the use of flyerr.GenericErr in line 43 and 50 seems excesive. A simple `fmt.Errorf()` would do.

internal/command/secrets/deploy.go
```diff
@@ -0,0 +1,69 @@
...
+
+func runDeploy(ctx context.Context) (err error) {
+	client := client.FromContext(ctx).API()
+	appName := appconfig.NameFromContext(ctx)
+	app, err := client.GetAppCompact(ctx, appName)
+	if err != nil {
+		return err
+	}
+
+	if app.PlatformVersion != appconfig.MachinesPlatform {
```
dangra, line 42: btw, you can replace all this with `command.IsMachinesPlatform(ctx, appName)` from `github.com/superfly/flyctl/internal/command` package.

nbw: looked into this and you could do it that way and clean up the error, but then I'd had to retrieve the app twice.

nbw:
> LG. My only caveat is that the use of flyerr.GenericErr in line 43 and 50 seems excesive. A simple `fmt.Errorf()` would do.

@dangra could do. I'm having a little bit of decision paralysis here because I'm getting [recommendations](https://github.com/superfly/flyctl/pull/2718#pullrequestreview-1587604312) that contradict each other. If it's all good with you, I'd like to merge this one as is and we can pull back on that later if we need to. I don't want to waste everyone's time reviewing this.

billyb2:
> > LG. My only caveat is that the use of flyerr.GenericErr in line 43 and 50 seems excesive. A simple `fmt.Errorf()` would do.
> 
> @dangra could do. I'm having a little bit of decision paralysis here because I'm getting [recommendations](https://github.com/superfly/flyctl/pull/2718#pullrequestreview-1587604312) that contradict each other. If it's all good with you, I'd like to merge this one as is and we can pull back on that later if we need to. I don't want to waste everyone's time reviewing this.

That's my fault. Using a flyerr thing is best when there's a suggestion that can be used when the user encounters an error. For a simple error message, `fmt.Errorf` is fine. Also, pretty much everything I noted is optional, what matters most is the PR itself :)

internal/command/secrets/deploy.go
dangra: Oh right, but the thing is creating a flaps instance doesn't require the app, just the appName:

https://github.com/superfly/flyctl/blob/8ff263df70d3755bd1cc1fa10a9fa2ab65997c1c/flaps/flaps.go#L41-L43

PR #2721, opened 2023-08-21
Fix scaling up from zero when group has mounted volumes

dangra:
### Change Summary

What and Why: Scaling up from zero fails due to lack of volume size information

How: Infer size from available volumes or fallback to 1GB+encrypted if none found

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


alichay (approved):
lgtm!

PR #2725, opened 2023-08-22
`migrate-to-v2`: Query app volumes from Flaps

alichay:
### Change Summary

What and Why: Query volumes with the Flaps API for volumes instead of including them in the GraphQL "App" query. Also removes the field from that query, to prevent accidental usage of the old API.

Related to: https://community.fly.io/t/fly-migrate-to-v2-apps-with-volumes-support/12316/44?u=allison (shouldn't fix this problem, but helps narrow down the causes)

---

### Documentation

- [ ] Fresh Produce
- [X] n/a

### DIFF
--- a/internal/command/migrate_to_v2/pg.go
+++ b/internal/command/migrate_to_v2/pg.go
@@ -77,10 +77,9 @@ func (m *v2PlatformMigrator) updateNomadPostgresImage(ctx context.Context) error
 }
 
 func (m *v2PlatformMigrator) migratePgVolumes(ctx context.Context) error {
-	app := m.appFull
 	regionsToVols := map[string][]api.Volume{}
 	// Find all volumes
-	for _, vol := range app.Volumes.Nodes {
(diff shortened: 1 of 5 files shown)


internal/command/migrate_to_v2/pg.go
```diff
@@ -77,10 +77,9 @@ func (m *v2PlatformMigrator) updateNomadPostgresImage(ctx context.Context) error
 }
 
 func (m *v2PlatformMigrator) migratePgVolumes(ctx context.Context) error {
-	app := m.appFull
 	regionsToVols := map[string][]api.Volume{}
 	// Find all volumes
-	for _, vol := range app.Volumes.Nodes {
+	for _, vol := range m.oldAttachedVolumes {
```
alichay, line 82: note that this *looks* like it changes behavior, but it filters by whether there's an attachedAllocation below anyway, so this shouldn't change anything

dangra (approved):
lgtm

Issue #1458, opened 2022-11-15
flyctl ssh console - missing terminal functionality on Windows; arrow keys, forward-delete, home/end etc.


kingIZZZY:
`flyctl ssh console`
Cannot scroll up/down through command history with arrow keys ⬆⬇
Cannot move cursor with ⬅➡ keys or skip words Ctrl+⬅ or Ctrl+➡ to edit line 
Cannot skip to beginning/end of line with Home / End
Seems there are fundamental Terminal functionalities missing - such as would be available with `docker exec -it`

* Windows 10
* `flyctl.exe v0.0.431 windows/amd64 Commit: eeef66af BuildDate: 2022-11-13T17:15:29Z`
* Dockerfile-based PHP project https://github.com/hyperf/hyperf-skeleton
  * Alpine linux `FROM hyperf/hyperf:8.1-alpine-v3.16-swoole`
  * by default it logs into `sh` - even then it should have these basic terminal line editing features
  * tried switching to `bash` - does not help, still can't line edit on the prompt line

ndarilek:
> tried switching to bash - does not help, still can't line edit on the prompt line

Do you mean switched to running bash in the SSH console? I'm wondering if this might be related to other possibly shell-related issues with `flyctl ssh console`. If you did run bash in the console and it still didn't help, then it may not be.

Also, what environment are you running flyctl in under Windows? Terminal, cmd, powershell, etc.? That unfortunately makes a huge difference, though it sounds like you're used to having that functionality and don't here so that may not be the issue.

Thanks.

kingIZZZY:
> * Windows 10

ndarilek:
Sorry, I meant were you running flyctl in the Windows Terminal app, the powershell CLI, cmd, or something else? Each of those handles the cursor a bit differently.

I'll probably be working on this at some point and I'd like to be sure I can replicate it first.

Thanks!

kingIZZZY:
[community.fly.io forum post](https://community.fly.io/t/flyctl-ssh-console-no-up-arrow-no-back-arrow-no-forward-delete-home-end-etc/8684)

I tried in all of the above, CMD, powershell, windows terminal, same behavior. When I try from Docker Desktop terminal for example it does seem to behave normally (i ssh into the local docker which runs on WSL2, then from there I install flyctl and run fly commands - then the terminal does behave correctly)

ndarilek:
Awesome, thanks for the additional context.

hrumhurum:
[...]

Issue #435, opened 2021-06-16
Terminal Resizing in Windows


tqbf:
**Describe the bug**
Until recently, `flyctl ssh console` was broken on Windows, because terminal file descriptor handling on Windows differs from that of Unix. 

The fix disabled window size change handling (which didn't work to begin with) on Windows.

But that leaves us with a flyctl that doesn't pass window size changes to the SSH server on Windows, which is not a dealbreaker, but is a wart. 

We should figure out how to handle window size changes on Windows (there's already a `terminal_windows.go` with a no-op terminal size change handler), so that when Windows people resize their `cmd.exe`, the remote end of the SSH connection is told about it.

tomr-a5:
Thanks for reporting this. #2887 completed Windows SSH console resize support, and current flyctl still sends window changes to the SSH session, so im going to close this as fixed. Please reopen if resizing still fails on the latest flyctl.

PR #2726, opened 2023-08-22
`win32`: Set console codepage to UTF-8

alichay:
### Change Summary

What and Why: Sets the Windows console codepage to UTF-8, hopefully making SSH and interactive sessions more reliable on Windows builds.

Related to: https://github.com/superfly/flyctl/issues/1458

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/cli/cli_windows.go
+++ b/internal/cli/cli_windows.go
@@ -1,0 +1,12 @@
+package cli
+
+import "golang.org/x/sys/windows"
+
+func init() {
+
+	kernel32 := windows.NewLazySystemDLL("kernel32.dll")
+	setConsoleCP := kernel32.NewProc("SetConsoleCP")
+	// Set codepage to UTF-8
+	// https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers#:~:text=Unicode%20(UTF%2D7)-,65001,-utf%2D8
+	setConsoleCP.Call(uintptr(65001))
+}


alichay:
This fixes *some* of the problem but we still have garbled output. It seems like maybe a buffering issue? There's a desync somewhere between what programs *expect* to have written to the screen and what ends up there.

hrumhurum:
Please note that [`SetConsoleCP`](https://learn.microsoft.com/en-us/windows/console/setconsolecp) function only sets the input console encoding. To set the output encoding, [`SetConsoleOutputCP`](https://learn.microsoft.com/en-us/windows/console/setconsoleoutputcp) should also be called. After calling these two functions with CP_UTF8 argument (65001), the console should be in a full Unicode mode with UTF-8 encoding set for both input and output streams.

hrumhurum:
Another nitpick: both `SetConsoleCP` and `SetConsoleOutputCP` functions take `UINT` argument, not `uintptr`. These types may have different sizes depending on a CPU architecture, e.g. 32 bit vs 64 bit. UINT is always 4 bytes long (32 bits) regardless of the architecture.

alichay:
> Please note that [`SetConsoleCP`](https://learn.microsoft.com/en-us/windows/console/setconsolecp) function only sets the input console encoding. To set the output encoding, [`SetConsoleOutputCP`](https://learn.microsoft.com/en-us/windows/console/setconsoleoutputcp) should also be called

Great catch, thank you! Unfortunately, even after setting this (and even enabling system-wide UTF8 support!), mc and nano still produce incorrect results.

image


> both `SetConsoleCP` and `SetConsoleOutputCP` functions take `UINT` argument, not `uintptr`. These types may have different sizes depending on a CPU architecture, e.g. 32 bit vs 64 bit. UINT is always 4 bytes long (32 bits) regardless of the architecture.

I know, I hate it too. For some reason, [golang requires all arguments to be cast to uintptr](https://pkg.go.dev/golang.org/x/sys/windows#LazyProc.Call) (regardless of whether they're pointers or not).

hrumhurum:
> Unfortunately, even after setting this (and even enabling system-wide UTF8 support!), mc and nano still produce incorrect results.

Those are promising results. I think that the sizing problems you see may be caused by a too small terminal window. Please try to maximize the window first before running interactive console apps like `mc` or `nano`. I believe that the sizing defect is caused by the [absence of window resizing support in `flyctl`](https://github.com/superfly/flyctl/issues/435).

hrumhurum:
I fixed the UTF-8 encoding setup for the Windows console: not only the input console encoding should be set to UTF-8, but also the output encoding. The tested and validated changes are in https://github.com/superfly/flyctl/pull/2889 pull request.

PR #2732, opened 2023-08-24
audit tool

michaeldwan:
### Change Summary

What and Why:

Tool to help us wrangle sprawling flyctl commands.

How:

`go run ./cmd/audit <lint|stats|list>`

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/cmd/audit/audit.go
+++ b/cmd/audit/audit.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "audit",
+		Short: "Tool for auditing flyctl commands",
+	}
+
+	rootCmd.AddCommand(newListCmd())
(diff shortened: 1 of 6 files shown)


cmd/audit/audit.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "audit",
+		Short: "Tool for auditing flyctl commands",
+	}
+
+	rootCmd.AddCommand(newListCmd())
```
billyb2, line 15: what are your thoughts on adding this as a fly command, but making it only build for dev builds? Either that, or having some kind of documentation in the README? This is one of those really interesting features of flyctl that feels like it'll get lost

michaeldwan: Yeah it'll get added to dev/contribution docs (when we write them 😉)

PR #2745, opened 2023-08-28
`ssh`: poll console size on Windows to get around lack of SIGWINCH

### LINKED ISSUE #1458: flyctl ssh console - missing terminal functionality on Windows; arrow keys, forward-delete, home/end etc.

alichay:
This paired with #2726 would fix #1458

What and Why: Adds polling for console size on Windows so that SSH gets the right window size. On unix-like operating systems, we respond to SIGWINCH, but the Windows equivalent requires funneling all input through a rube goldberg machine of Windows-specific system calls. It's not realistic to do that, so we just poll for changes.

Related to: #1458, #2726

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


michaeldwan:
Also fixes #435

ndarilek:
Would it be worth merging this, #2726, and cutting a pre-release? Seems like they may be hard to test independently.

I'm happy to test either or both of these if you can tell me how to put each through its paces. In general TUI programs are harder to use with a screen reader since, for instance, a dialog box is just a handful of box-drawing characters and text vs. a node in the accessibility tree announcing itself as a dialog. But something like nano should be easier to test since I generally know what it should behave like.

Thanks for taking these on.

alichay:
> Would it be worth merging this, #2726, and cutting a pre-release? Seems like they may be hard to test independently.

Unfortunately there still seems to be a missing piece. In my testing, this doesn't seem to do anything???

michaeldwan:
I'm wary of merging in order to test. I can give this a whirl on my windows machine when I get home. To test things like this I just go build on my mac then copy the binary to windows like it's 2007.

hrumhurum:
How is it going? Can we have a release with that and with UTF-8 encoding, maybe under an opt-in feature flag? The reason I'm asking is because the problem still persists and it's always a pain to jump over the moon for such a basic functionality. Even UTF-8 encoding alone would enough to give a significant relief.

alichay:
> Even UTF-8 encoding alone would enough to give a significant relief.
Merged!

> How is it going? Can we have a release with that and with UTF-8 encoding, maybe under an opt-in feature flag? The reason I'm asking is because the problem still persists and it's always a pain to jump over the moon for such a basic functionality. 

This PR never worked in my testing, and I sadly don't have the bandwidth to finish this up right now. If you want to and are able, you could try to finish what I started? Otherwise, I think it'll be a while until I'm able to get back to this and give it another look.

We want to fix this issue, but Windows makes it very difficult, and there's quite a few time-sensitive projects going on internally here that I have to give my attention to.

hrumhurum:
> Merged!

Thank you!

> This PR never worked in my testing, and I sadly don't have the bandwidth to finish this up right now. If you want to and are able, you could try to finish what I started?

I'll give it a shot. I will post updates here.

hrumhurum:
I've completed the feature implementation in https://github.com/superfly/flyctl/pull/2887 pull request. Both Windows console resizing and UTF-8 encoding do work now.

dangra (approved):

PR #2750, opened 2023-08-29
Catch Flaps Errors And Return Extra Information 

billyb2:
### Change Summary

What and Why:

Flaps now returns some extra information such as status code a struct with additional info, so we should take advantage of that.

How:

Related to:

https://github.com/superfly/nomad-firecracker/pull/1535

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


internal/command/machine/clone.go
```diff
@@ -288,6 +290,15 @@ func runMachineClone(ctx context.Context) (err error) {
...
 	if err != nil {
+		if errors.As(err, &flaps.LaunchCapacityErr{}) {
+			return flyerr.GenericErr{
+				Err:      err.Error(),
+				Descript: flyerr.GetErrorDescription(err),
+				Suggest:  "Try cloning to a different nearby region, or try again in a few hours",
+				DocUrl:   flyerr.GetErrorDocUrl(err),
+			}
+
+		}
```
michaeldwan, line 301: Do you need to catch and wrap this? If `err` is a `flaps.LaunchCapacityErr` it should get picked up downstream and render the help text. You'd need to add a suggestion function though.

flaps/flaps_machines.go
```diff
@@ -22,6 +23,38 @@ func (f *Client) sendRequestMachines(ctx context.Context, method, endpoint strin
...
 
+const (
+	LaunchCapacityErrStr = "no capacity available in"
+)
+
+type LaunchCapacityErr struct {
+	region string
+}
+
+var launchCapacityErrRe *regexp.Regexp = regexp.MustCompile(fmt.Sprintf(`%s (?P<region>[a-zA-Z0-9]*)`, LaunchCapacityErrStr))
```
michaeldwan, line 34: Curious if we can update flyd/flaps to return errors as JSON with a CODE field and some extra text or data and just pass that straight through. Thoughts?

internal/command/machine/clone.go
billyb2: nope, i'm just thinking about how i could add a suggestion per command

flaps/flaps_machines.go
billyb2: done!

internal/cli/cli.go
```diff
@@ -137,6 +137,11 @@ func printError(io *iostreams.IOStreams, cs *iostreams.ColorScheme, cmd *cobra.C
 		fmt.Fprintln(io.ErrOut)
 	}
 
+	if requestId := flyerr.GetErrorRequestID(err); requestId != "" {
+		fmt.Fprintln(io.ErrOut, "Fly-Request-Id:", requestId)
```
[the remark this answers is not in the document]
billyb2: My thinking was that the request-id is a separate concern to the user than the actual error itself, though I'm down to change it. Would you want to give an example of how you think we should print it?

rugwirobaker:
We're going to define more know errors thought right?

internal/cli/cli.go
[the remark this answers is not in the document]
benwaffle:
```
Error: failed to update VM 148ed510a01458: request returned non-2xx status, 504 (request ID a9bf7c15f6b987cd-lga)
```

So that users can just share the one line with us when debugging. I don't think we need it for errors we handle already, such as "no capacity in ord", as it'll just make the output messier, and our error messages should have enough info for the users to debug.

billyb2:
> We're going to define more know errors thought right?

For sure! I want to get some initial errors I'm using for different flyctl improvements out of the way first, but the goal is to have it everywhere.

internal/cli/cli.go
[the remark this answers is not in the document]
billyb2: Looks good, will do!

billyb2:
Drafting this again since there's been a lot of internal discussion regarding the flaps error API

billyb2:
Superseded by #2862

PR #2751, opened 2023-08-29
fly ssh console: select/filter by process group

wjordan:
### Change Summary

- Appends a process group field to the end of `fly ssh console --select`, if there is more than one process group set on the machines:

```
$ fly ssh console --select
? Select VM:  [Use arrows to move, type to filter]
> lax: aaaaaaaaaaaaaa fdaa:0:aaaa:aaa:aaaa:aaaa:aaaa:a crimson-shadow-1234 (process1)
  lax: bbbbbbbbbbbbbb fdaa:0:bbbb:bbb:bbbb:bbbb:bbbb:b lingering-morning-5678 (process2)
```

- Add `--group` flag to `fly ssh console` to filter target machines by the specified process group.
- Make `--region` flag actually filter target machines by the specified region.

### DIFF
--- a/internal/command/ssh/console.go
+++ b/internal/command/ssh/console.go
@@ -66,6 +66,7 @@ func stdArgsSSH(cmd *cobra.Command) {
 			Description: "Unix username to connect as",
 			Default:     DefaultSshUsername,
 		},
+		flag.ProcessGroup(),
 	)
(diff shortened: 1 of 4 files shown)


dangra (approved):
nice!

PR #2752, opened 2023-08-29
add selection prompt to volumes commands

wjordan:
### Change Summary

The `destroy`, `extend`, `fork`, and `show` subcommands of `fly volumes` will display an interactive prompt to choose from a list of existing volumes on an app, when no volume ID is specified.

The prompt reuses the table formatting from `fly volumes list` for the selection lines (which I've extracted to `renderTable` for reuse).

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/volumes/volumes.go
+++ b/internal/command/volumes/volumes.go
@@ -1,19 +1,25 @@
 package volumes
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
 	"io"
+	"strconv"
 	"time"
 
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/dustin/go-humanize"
 	"github.com/spf13/cobra"
 
 	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/client"
 	"github.com/superfly/flyctl/flaps"
 
 	"github.com/superfly/flyctl/internal/command"
 	"github.com/superfly/flyctl/internal/command/volumes/snapshots"
+	"github.com/superfly/flyctl/internal/render"
 )
 
 func New() *cobra.Command {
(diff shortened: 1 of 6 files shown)


internal/command/volumes/volumes.go
```diff
@@ -80,3 +86,82 @@ func countVolumesMatchingName(ctx context.Context, volumeName string) (int32, er
...
+	selected := 0
+	fmt.Println(title)
+	prompt := &survey.Select{
+		Message:  "Select volume:",
+		Options:  volumeLines,
+		PageSize: 15,
+	}
+	if err := survey.AskOne(prompt, &selected); err != nil {
+		return nil, fmt.Errorf("selecting volume: %w", err)
+	}
```
matttpt, line 165: Perhaps use the `Select` function from `github.com/superfly/flyctl/internal/prompt`? It's ultimately just a wrapper over `github.com/AlecAivazis/survey/v2`, but it provides some tweaks (like overriding the color scheme).

matttpt (commented):
Overall looks good to me! I left one comment on the code below.

Would it make sense to put this behind a `--select` flag for consistency with other commands like `fly ssh console`/`fly proxy`/`fly machines`?

internal/command/volumes/volumes.go
michaeldwan: This is a good idea. I don't remember if that helper function correctly handles non-interactive prompts, but it'd be nice to make sure corner cases like that are handled in one place so we're not repeating tedious code everywhere.

wjordan:
> Would it make sense to put this behind a --select flag for consistency with other commands like fly ssh console/fly proxy/fly machines?

I agree it makes sense to be consistent, though I'd actually prefer if we went the other way and removed the `--select` flag from commands with no default selection that otherwise throw an error when no ID is provided.

I think an error like `Error: a machine ID must be provided unless --select is used` is tedious, automatically using the selection prompt is more intuitive and saves keystrokes for the common (interactive) case.

internal/command/volumes/volumes.go
wjordan: Yes, it also looks like this wrapper helps return an error when running non-interactively as well which is important.

michaeldwan:
> I think an error like `Error: a machine ID must be provided unless --select is used` is tedious, automatically using the selection prompt is more intuitive and saves keystrokes for the common (interactive) case.

@wjordan yes! I'd love to move in this direction.

matttpt:
Just noticed one more thing: the usage for each command can be changed to indicate that the volume ID is optional (i.e. `<id>` → `[id]`).

matttpt (approved):
Aside from updating the usage, LGTM!

PR #2753, opened 2023-08-29
More test fixes

michaeldwan:
- handles http errors during TestAppsV2Example instead of an unhelpful panic
- fixes an issue with the wrong go version being used for precommit and golang-ci 
- couple of other tweaks to the GH action config


billyb2 (approved):
👍

Issue #2461, opened 2023-06-20
`deploy`: rolling deploys should be async


alichay:
future improvement to #2455 

We batch waiting for machines, which saves a lot of time, but we *could* spawn multiple machines async as well. This would make deploys a *lot* faster, but it would also require more robust logging to support the asynchronous status updates of multiple machines.

alimurtaza026:
@dangra Hi, it should be closed if it is done. Why it is open yet?

tomr-a5:
Thanks for reporting this. #2936 replaced batched rolling updates with concurrent per-process-group updates, so I’m going to close this out. Please reopen if rolling deploys are still updating Machines serially on the latest flyctl.

PR #2755, opened 2023-08-29
async statuslogger

alichay:
### Change Summary

What and Why:
Adds an async-ready job status system. You specify a number of jobs, and it gives each job a line at the bottom of the terminal. This is the piece that makes truly batched deploys possible, because without this it's difficult to coordinate logging between threads while preventing output from being overwritten.

Downsides:
  1. It's heavy-handed. Standardizing this means that we have to use this everywhere.
     (although a lot of that legwork is already done)
  3. The way this is implemented means that if you output many lines to stdout at a time, there's a chance you could get slightly garbled output. Essentially, this package leaves two blank lines above the status log section. Every (iirc) 100ms, we try to redraw the status panel, meaning that slow crawls of text output work just fine. If you output a lot of short lines of text, though, you could overlap with the status - then, when it goes to redraw, that happens fine, but the lines above end up with a bit of the status text in them.

     For example, if you ran the following code:
     ```go
     sl := statuslogger.Create(ctx, 2, true)
     sl.Line(0).Log("aaaaaaaaaa")
     sl.Line(1).Log("bbbbbbbbbb")
     fmt.Println("00000")
     fmt.Println("11111")
     fmt.Println("22222")
     sl.Destroy(false)
     ```

     It would produce the following output:
     ```
     00000
     11111--
     22222/2] aaaaaaaaaa
     
     -------
        [1/2] aaaaaaaaaa
        [2/2] bbbbbbbbbb
     -------
     ```

     ~~I have plans to address this (namely, an API to clear the status board, let you output your block of text, then redraw the status), but I wanted to get feedback on what's here before I put more effort into it :)~~

     Edit: that's fixed!
     ```go
     sl := statuslogger.Create(ctx, 2, true)
     sl.Line(0).Log("aaaaaaaaaa")
     sl.Line(1).Log("bbbbbbbbbb")
     // contrived usage of NewContext here because it's an example
     resume := statuslogger.Pause(statuslogger.NewContext(ctx, sl.Line(0))
     fmt.Println("00000")
     fmt.Println("11111")
     fmt.Println("22222")
     resume()
     sl.Destroy(false)
     ```
     gives
     ```
     00000
     11111
     22222
     
     -------
        [1/2] aaaaaaaaaa
        [2/2] bbbbbbbbbb
     -------
     ```

Related to: #2455, #2461

---

### Documentation

- [ ] Fresh Produce
- [X] n/a


michaeldwan:
This is excellent.

dangra:
This is awesome! Once merge I will rebase my work for concurrent rolling deploys at https://github.com/superfly/flyctl/pull/2679

benwaffle:
If you have more status lines than terminal rows (e.g. `fly scale count 100`), then it renders all of the lines, but scrolling behaves weirdly. Can we render only as many rows as are visible, and only rows that are being updated?

benwaffle (approved):

PR #2756, opened 2023-08-29
Add host_dedication_id to flyctl (deploy)

gwuah:
### Change Summary
Add a `host_dedication_id` that fly deploy logic should pass to the machine API when creating a new machine and delete/recreate machines if they don't match on subsequent deployments.

### DIFF
--- a/internal/command/volumes/create.go
+++ b/internal/command/volumes/create.go
@@ -102,11 +102,11 @@ func runCreate(ctx context.Context) error {
 		return client.GetAppBasic(ctx, appName)
 	})
 
-	// if confirm, err := confirmVolumeCreate(ctx, appName); err != nil {
-	// 	return err
-	// } else if !confirm {
-	// 	return nil
-	// }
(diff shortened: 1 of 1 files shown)


internal/command/volumes/create.go
```diff
@@ -99,11 +102,11 @@ func runCreate(ctx context.Context) error {
...
-	if confirm, err := confirmVolumeCreate(ctx, appName); err != nil {
-		return err
-	} else if !confirm {
-		return nil
-	}
+	// if confirm, err := confirmVolumeCreate(ctx, appName); err != nil {
+	// 	return err
+	// } else if !confirm {
+	// 	return nil
+	// }
```
michaeldwan, line 109: Does this happen somewhere else now, or are we not wanting to prompt for confirmation?

michaeldwan (approved):
Changes look good. Question about removing confirmation.

My only gripe is that "Host Dedication ID" is an odd set of words. Is "Dedicated Host ID" incorrect?

michaeldwan:
Would `Host Group ID` make sense?

PR #2736, opened 2023-08-25
Warn if "fly scale count 0" resets the VM size

kzys:
If a VM is not shared-cpu-1x, "fly scale count 0" shows a warning to avoid surprises.

### Change Summary

What and Why:

`fly scale count 0` is having a bit of surprising behavior. While we might need to change the behavior someday, this change at least tells customers about potential surprises.

How:

```
% ../flyctl/bin/flyctl scale count 0
App 'withered-wildflower-5588' is going to be scaled according to this plan:
  -2 machines for group 'app' on region 'sea' with size 'shared-cpu-2x'
  Scaling to zero resets the size of the machines to 'shared-cpu-1x'
? Scale app withered-wildflower-5588? No
```

No warnings if the machine is shared-cpu-1x.

```
% ../flyctl/bin/flyctl scale count 0
App 'withered-wildflower-5588' is going to be scaled according to this plan:
  -2 machines for group 'app' on region 'sea' with size 'shared-cpu-1x'
? Scale app withered-wildflower-5588? No
```

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] In superfly/docs, or asked for help from docs team
- [x] n/a

PR #2766, opened 2023-08-30
Add vm-size option to fly scale count

billyb2:
### Change Summary

What and Why:

This will allow users to set a vm-size when scaling up or down, which also allows users to set a vm size when scaling from 0

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


michaeldwan (approved):

dangra (changes requested):
Please, move the default guess size to `machine_defaults.go`, there is a `guess` struct field for this case.

internal/command/scale/machine_defaults.go
```diff
@@ -42,6 +43,10 @@ func newDefaults(appConfig *appconfig.Config, latest api.Release, machines []*ap
 				break
 			}
 		}
+
+		// If we still don't have a guest size, just set it to the default one
+		guest = new(api.MachineGuest)
+		guest.SetSize(deploy.DefaultVMSize)
```
dangra, line 49: Can we do this outside and pass last-mile default `guest` as an argument to `newDefaults`?

internal/command/scale/count_machines.go
```diff
@@ -61,8 +62,12 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 		return err
 	}
 
+	defaultGuest := new(api.MachineGuest)
+	defaultGuest.SetSize(deploy.DefaultVMSize)
 	defaults := newDefaults(appConfig, latestCompleteRelease, machines, volumes,
-		flag.GetString(ctx, "from-snapshot"), flag.GetBool(ctx, "with-new-volumes"))
+		flag.GetString(ctx, "from-snapshot"), flag.GetBool(ctx, "with-new-volumes"), defaultGuest)
+
+	deploy.ApplyFlagsToGuest(ctx, defaults.guest)
```
dangra, line 70: no need to access the private field of defaults when you can apply it to `defaultGuest` right after creating it.

dangra, line 70: as last nitpick, can we move the `ApplyFlagsToGuest` function somewhere that doesn't require a cross import of `deploy` or any other command?

internal/command/scale/count.go
```diff
@@ -38,6 +38,25 @@ For pricing, see https://fly.io/docs/about/pricing/`
...
+		},
+		flag.String{
+			Name:        "vm-cpukind",
+			Description: "The kind of CPU to use ('shared' or 'performance')",
+		},
+		flag.Int{
+			Name:        "vm-memory",
+			Description: "Memory (in megabytes) to attribute to the VM",
+			Aliases:     []string{"memory"},
+		},
```
dangra, line 59: this set of flags keeps repeating, would be good to merge them into a `Flag.Set{}` and reuse it from all commands that require VM size related flags.

internal/command/scale/count_machines.go
billyb2: ah that's my mistake, i should've seen that

andie787 (commented):
Just wanted to note that there's [an open PR to output a warning ](https://github.com/superfly/flyctl/pull/2736)when you use `fly scale count 0`. Not sure if the warning should now say to use these new flags? Or is this too complicated?
Will update docs when it's all decided 😄

internal/command/scale/machine_defaults.go
```diff
@@ -42,6 +42,8 @@ func newDefaults(appConfig *appconfig.Config, latest api.Release, machines []*ap
 				break
 			}
 		}
+
+		guest = fallbackGuest
```
dangra, line 46: I think you want this before the `for` line.

billyb2:
> Just wanted to note that there's [an open PR to output a warning ](https://github.com/superfly/flyctl/pull/2736)when you use `fly scale count 0`. Not sure if the warning should now say to use these new flags? Or is this too complicated? Will update docs when it's all decided 😄

Oh nice! I think the warning alongside a suggestion to use these new flags would be great. I'll also leave a comment in the PR.

internal/command/scale/count_machines.go
```diff
@@ -61,8 +62,12 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 		return err
 	}
 
+	defaultGuest := new(api.MachineGuest)
+	defaultGuest.SetSize(deploy.DefaultVMSize)
 	defaults := newDefaults(appConfig, latestCompleteRelease, machines, volumes,
-		flag.GetString(ctx, "from-snapshot"), flag.GetBool(ctx, "with-new-volumes"))
+		flag.GetString(ctx, "from-snapshot"), flag.GetBool(ctx, "with-new-volumes"), defaultGuest)
+
+	mach.ApplyFlagsToGuest(ctx, defaultGuest)
```
dangra, line 70: this works but relies on a side effect, it would be more correct if placed after the `.SetSize()` line above, right before handing over _defaultGuest_  to `newDefaults()`

dangra (approved):
Honestly, I can keep going :)) sorry for that. The PR looks mergeable but if you want to polish it a bit more here is an idea:
* Move `internal/machine/util.go` to `internal/flags/machines.go`
* Rename `VMSizeFlags` to `MachineGuestFlags`
* Rename `mach.ApplyFlagsToGuest(ctx, guest)` to `flag.GetMachineGuest(ctx)`
* Move `deploy.DefaultVMSize` to `api.DefaultVMSize`
* Make `flag.GetMachineGuest(ctx)` return a `api.MachineGuest` instance whose base is `api.DefaultVMSize` + `--vm-*` values if set

internal/flag/machines.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	"github.com/superfly/flyctl/api"
+)
+
+// Returns a MachineGuest based on the flags provided overwriting a default VM
+func GetMachineGuest(ctx context.Context) *api.MachineGuest {
+	var guest *api.MachineGuest
```
dangra, line 11: does it work? this is a `nil` value

internal/command/deploy/deploy.go
```diff
@@ -321,8 +280,8 @@ func deployToMachines(
 
 	if guest == nil {
 		guest = &api.MachineGuest{}
-		guest.SetSize(DefaultVMSize)
-		_ = ApplyFlagsToGuest(ctx, guest)
+		guest.SetSize(api.DefaultVMSize)
+		_ = flag.GetMachineGuest(ctx)
```
dangra, line 284: nonsense

billyb2: horrible thing to overlook

remorses:
Why not add the machine size to fly.toml instead? This way every deployment becomes idempotent

Issue #2769, opened 2023-08-30
Accept human-friendly memory & storage sizes


michaeldwan:
`fly scale memory` and friends accept input as a number in megabytes. The number needs to be divisible by 1024 if greater than 1GB, and 256 if less. This is only told to users through error messages. 

`fly volume create` and friends accept input as a number in gigabytes. I don't think it has similar restrictions. 

Instead of forcing users to do conversions, flyctl should accept human friendly sizes as strings. For example:

```
fly scale memory 512mb
fly scale memory 2gb
fly volume extend --size 20gb (or
```

It's out of scope, but that last one would be rad if it accepted `--size +10gb` as well ;)

redjonzaci:
In the code block, did you finish your thought? "(or"

redjonzaci:
I started looking into this, but if you are working on this yourself, I will skip it when I wake up.

redjonzaci:
I added an initial implementation for the examples you mention.
Should I check for other commands that should behave this way?

redjonzaci:
@michaeldwan we just merged my PR for this issue!

PR #2770, opened 2023-08-30
Accept human-friendly memory & storage sizes

redjonzaci:
### Change Summary

What and Why:
`fly scale memory` now accepts 1gb, 1GB, 1.25GB.
`fly volume extend <id> --size` now accepts +1 to add to the current size.

How:

Related to: #2769

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


redjonzaci:
Will take a better look and test when I wake up, but an initial implementation.

redjonzaci:
To accept +1 size, I had to get the volume to determine the current volume size and that results in more time.
But if no + sign as first character, I assume the value should replace the current size.

internal/command/volumes/extend.go
```diff
@@ -72,7 +74,18 @@ func runExtend(ctx context.Context) error {
 		return err
 	}
 
-	sizeGB := flag.GetInt(ctx, "size")
+	sizeFlag := flag.GetString(ctx, "size")
+	sizeGB := 0
+	if sizeFlag[0] == '+' {
+		volume, err := flapsClient.GetVolume(ctx, volID)
```
redjonzaci, line 80: To make +1 work, I had to get the volume to determine its current size, which makes it slower.
But I do this only if there's a + as the first character. Otherwise, I keep treating it as a replacement value for current one.

redjonzaci:
Nope, I didn't consider it, but I can get to it in an hour.

redjonzaci:
On it now.

redjonzaci:
> have you considered https://pkg.go.dev/github.com/docker/go-units#FromHumanSize ?

The issue with this is that we only get the number back, while in the case of `fly scale memory` we want to allow users to pass both MB and GB sizes.
It can be used for `fly volume extend <id> --size` though.

redjonzaci:
@dangra `FromHumanSize` function returns size in bytes and I would need to do the conversion myself
so I am deciding not to use the function at all. Let me know what you think!

dangra:
> while in the case of `fly scale memory` we want to allow users to pass both MB and GB sizes.

not sure what you mean

> . And it doesn't support float numbers like in "1.5gb".

It does: https://go.dev/play/p/ZvUJMKLHggL

dangra:
> `FromHumanSize` function returns size in bytes and I would need to do the conversion myself

converting it to Megabytes or Gigabytes isn't a big deal right?



> so I am deciding not to use the function at all. Let me know what you think

the downside of your handcrafted parsing functions are the lack tests and proven use.
on the other side, go-units is a well tested library backed by a reasonable healthy organization, already battle tested and reused by many others.

redjonzaci:
Also, we appear to be using IEC 60027-2 while `FromHumanSize` uses SI.
https://physics.nist.gov/cuu/Units/binary.html

I get this error from `fly scale memory`:
```
Memory size must be in 256 a MiB increment (256MiB would work)
```

dangra:
> Also, we appear to be using IEC 60027-2 while `FromHumanSize` uses SI.

then switch to `units.RAMInBytes`

redjonzaci:
Sorry I didn't go through all offerings of `units`!
Will keep notes about how to approach the work here as I go.

redjonzaci:
I really appreciate you taking the time to give feedback!

internal/command/volumes/extend.go
```diff
@@ -72,7 +73,22 @@ func runExtend(ctx context.Context) error {
...
+
+	if sizeFlag[0] == '+' {
+		volume, err := flapsClient.GetVolume(ctx, volID)
+		if err != nil {
+			return err
+		}
+		sizeGB += volume.SizeGb
+	}
+
 	if sizeGB == 0 {
```
dangra, line 92: this check has to be right after line 82 and before checking for `+` char presence in `sizeFlag[0]`

internal/command/volumes/extend.go
```diff
@@ -72,7 +73,22 @@ func runExtend(ctx context.Context) error {
 		return err
 	}
 
-	sizeGB := flag.GetInt(ctx, "size")
+	sizeFlag := flag.GetString(ctx, "size")
+	sizeBytes, err := units.FromHumanSize(sizeFlag)
```
dangra, line 77: do we have to change it to `RAMInBytes()` to match expectations?

internal/command/volumes/extend.go
```diff
@@ -72,7 +73,22 @@ func runExtend(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("invalid size: %w", err)
+	}
+
+	sizeGB := int(sizeBytes) / units.GB
```
dangra, line 82: what happens if `--size=3` (no units) is passed as value?

internal/command/volumes/extend.go
redjonzaci: Since this only accepts sizeGB, not necessary I think.

dangra: hmm, if we ever accept 1024GB as volume size that is going to expose the same issue you found with `fly scale memory` command and 1024MB.

internal/command/volumes/extend.go
redjonzaci: I thought of this algorithm:

If it's not 0 and it has a `+` sign (or not), check if it's just an `int` (after `+` sign).
If just an int, set volume size.
If not just an `int`, try parsing with `FromHumanSize` and then set volume size.

internal/command/volumes/extend.go
redjonzaci: Even if we accept 1024GB, it will be divided by a billion and we still get 1024GB.

internal/command/volumes/extend.go
redjonzaci: Implemented that

redjonzaci:
@dangra I don't know if I can add new tests to the suite, but wrote some. 😌

redjonzaci:
Am I supposed to run the workflows on my fork of this repo because they get cancelled?

internal/command/volumes/extend.go
dangra: Fair point. I guess the same applies to `fly scale memory`, no need to use `RAMInBytes` as long as the parsed value is divided by `units.MB` instead of `units.MiB`

internal/command/volumes/extend.go
```diff
@@ -72,11 +74,29 @@ func runExtend(ctx context.Context) error {
...
+	sizeFlag := flag.GetString(ctx, "size")
+	sizeGB, err := strconv.Atoi(sizeFlag)
+	if err != nil {
+		sizeBytes, err := units.FromHumanSize(sizeFlag)
+		if err != nil {
+			return fmt.Errorf("invalid size: %w", err)
+		}
+
+		sizeGB = int(sizeBytes) / units.GB
+	}
```
dangra, line 86: what about moving this parsing logic to `helpers/units.go` and reuse it for `fly vol extend` and `fly scale memory` ? 

There is another commonly used flag name `--vm-size` usually parsed by `internal/flag/machines.go`'s `GetMachineGuest()` that could reuse the helper

dangra:
> @dangra I don't know if I can add new tests to the suite, but wrote some. 😌

thanks for that ! 

I'm afraid CI won't run tests for non-committers due to security restrictions.

internal/command/volumes/extend.go
redjonzaci: And pass in the unit of division

dangra:
btw, run `go mod tidy` so it updates `go.mod / go.sum` files

internal/command/volumes/extend.go
redjonzaci: If I pass `1gb`, `FromHumanSize` returns `10^9` and whichever way I divide it, I can't get to 1024MB.

redjonzaci: I could pass a parser func to `helpers.ParseSize` though if we still want to reuse the logic:

`helpers.ParseSize(sizeFlag, units.RAMInBytes, units.MiB)`
`helpers.ParseSize(sizeFlag, units.FromHumanSize, units.GB)`

redjonzaci:
> > @dangra I don't know if I can add new tests to the suite, but wrote some. 😌
> 
> thanks for that !
> 
> I'm afraid CI won't run tests for non-committers due to security restrictions.

Does that mean you will accept my test? Because I removed it 😂

internal/command/volumes/extend.go
redjonzaci: Done

internal/command/volumes/extend.go
redjonzaci: `GetMachineGuest` can't reuse the helper as the size is set as `string`.

dangra:
Your tests were good! Restore them please

dangra:
thanks @redjonzaci !

PR #2774, opened 2023-08-31
Smoke checks: use uptime to more quickly pass test

tvdfly:
Clusters over a handful of machines will see faster `fly deploy` times with this change.

Previously, the smoke checks started a 10 second timer and ran a series of checks over those 10 seconds. The intent was to ensure that the machine stayed up and didn't repeatedly exit over that time.

Since we added smoke checks we started batching updates for machines, and then run the smoke checks sequentially. This means the smoke test may run multiple minutes after a machine was updated. For example, with an app with 36 machine it will get updated in batches of 12. Then the smoke check will be run on each machine sequentially. It takes ~110 seconds to get to the 12th machine. At that point, it doesn't make sense to wait another 10 seconds observing the machine.

This change updates the smoke check to look for the most recent start event, and then use that to calculate the uptime. We pass the smoke check when the machine has an uptime greater than 10 seconds. For the batch deployment example above, this means the first machine will wait ~10 seconds for the smoke to complete, then the 2nd through 12th machines will pass the smoke checks nearly immediately—assuming they are up and not constantly restarting.

`fly deploy` dropped from 6.5 minutes down to a little under 2 minutes on a 36 machine app I tested.🏎️

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: "Place wintun.dll"
         run: cp deps/wintun/bin/amd64/wintun.dll ./
       - name: Run tests
-        run: make test
+        run: make test test-api
 
   release:
     needs: test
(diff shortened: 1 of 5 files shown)


dangra (approved):
Looks good! 

nit: MostRecentStartTimeAfterLaunch could certainly benefit from a few unit tests

Issue #2746, opened 2023-08-28
--org parameter is ignored


yanc0:
**Describe the bug**

--org argument is set but the command is asking for the organization anyway.

* Ubuntu 22.04.3 LTS
* `flyctl v0.1.84 linux/amd64 Commit: 5735a8b979ebc938c74832c3edd52c509abaa7eb BuildDate: 2023-08-23T16:56:07Z`

**Paste your `fly.toml`**

```toml
app = "radicale"
primary_region = "cdg"

[http_service]
  internal_port = 5232
  force_https = true
  auto_stop_machines = true
  auto_start_machines = true
  min_machines_running = 0
  processes = ["app"]

[mounts]
  source = "radicale_data"
  destination = "/data
```

**Command output:**

```text
$ flyctl ssh issue --agent --org personal
? Select Organization:  [Use arrows to move, type to filter]
> Yann (personal)
  my.org.com (my-org-com)
```

redjonzaci:
I believe `--org` needs to be the first arg as per [flyctl/internal/command/ssh/issue.go:83](https://github.com/superfly/flyctl/blob/master/internal/command/ssh/issue.go#L83).

yanc0:
Thanks you for your help @redjonzaci ! Unfortunately, it does not work either.

redjonzaci:
Did you try `flyctl ssh issue personal`? 😉
Maybe the docs need to be updated as they describe both an [org] arg and --org flag.

<img width="608" alt="image" src="https://github.com/superfly/flyctl/assets/73707194/8baf583d-afce-4726-a77f-fadd125fcf7f">

redjonzaci:
@jsierles sorry to tag you here, but I see you added [issue.go](https://github.com/superfly/flyctl/blob/master/internal/command/ssh/issue.go).
Do we want to proceed with an `org` arg or `--org` flag?

redjonzaci:
@dangra do you think we could do something for this?

dangra:
@redjonzaci I'd drop the "org as first arg" syntax and rely on your recent work on `flag.GetOrg()` only. 

Convert `path` to a flag too.

tomr-a5:
Thanks for reporting this. #2759 made `ssh issue` honor `--org` before prompting, so I’m going to close this out. Please reopen if you can still reproduce it on the latest flyctl.

PR #1572, opened 2023-01-12
issueCertificate updates

btoews:

PR #2777, opened 2023-08-31
Switch path from arg to flag

redjonzaci:
### Change Summary

Related to: #2746


redjonzaci:
@dangra I am not sure how `fly ssh issue` should work with both `--org` and `--path`.
When I specify both, I get the following error:
```
Invalid email address: mail: missing '@' or angle-addr
```

I think it may not be working properly in the current released version.

internal/command/ssh/issue.go
```diff
@@ -42,6 +41,11 @@ validity.`
```
dangra, line 41: positional arguments are not valid now

redjonzaci: Both args and flags should be valid?

dangra: I'd opt not to be backwards compatible with this command.

But I am referring to line 40, we have to change it `cobra.NoArgs`:
```
 	cmd.Args = cobra.MaximumNArgs(3)
```

internal/command/ssh/issue.go
```diff
@@ -97,24 +101,21 @@ func runSSHIssue(ctx context.Context) (err error) {
...
-	case 1:
-		// org only
-	case 2:
+
+	orgFlag := flag.GetOrg(ctx)
+	path := flag.GetString(ctx, "path")
+
+	if orgFlag != "" && path != "" {
+		// org+email+path
+		emails = orgFlag
```
dangra, line 110: it doesn't make sense.

seems the original command accepted 3 args (org, email and path), although usage docs says it only does 2 `issue [org] [path]`.

We need another `--email` flag, a StringSlice or whatever matches the `mail.ParseAddress()` output.

redjonzaci: I will try looking into the history of the command first to get some more context.

redjonzaci: Found the commit where it worked with 3 args, applying fix.

redjonzaci:
@dangra wonder if you could review this when you find some time?

internal/command/ssh/issue.go
```diff
@@ -97,24 +106,22 @@ func runSSHIssue(ctx context.Context) (err error) {
...
-			rootname = args[1]
+			rootname = orgFlag
 		}
-	case 3:
-		// org+email+path
-		emails = args[1]
-		rootname = args[2]
-	default:
-		return errors.New("Too many positional arguments\n")
 	}
```
dangra, line 125: I am afraid I don't have time at the moment but this whole thing looks shady to me. Needs 🧠 power and settle down on how this positional values that now are flags were used to convey the different command combinations.

redjonzaci: I think the argument changes started with https://github.com/superfly/flyctl/pull/1572.
I can do a more in-depth analysis of this tomorrow.

redjonzaci: First, the issue command supported `[org] [email] [path]` args.
And `--org`, `--username`, `--hours`, `--agent`, `--dotssh`, `--overwrite` flags.

PR #1572 made the following changes related to the args of `fly issue ssh` command:
- `--username` flag to accept a string slice instead of string.

- Adds comment saying API args username and email are deprecated. Removes `[email]` arg,
but still uses its value in the switch statement?

- Gets just the name part of the email address and adds it to the principals.

I think the only issue in my code is changing the addition in line 123 to be `email`, not `orgFlag`
as `[email]` arg was `args[1]` previously.

redjonzaci: Actually, after removing the `[email]` arg,
the code seems to expect the value of the second arg to be used for `[email]` or `[path]`.

redjonzaci:
@dangra I think I finally got it right with this last change. 🤞🏼

PR #2733, opened 2023-08-25
Better help

michaeldwan:
This PR revamps help text for the `fly` command and its subcommands. The goal is to make our help text more helpful, easier to read, and easier to maintain going forward. This only addresses the framework for how help text is rendered. Updating the oodles of individual commands and flags will come in subsequent PRs.

Read on for an overview of what's new.

# Improved Formatting

- All command descriptions have proper indentation, and dynamic line wrapping to fit the terminal width. Proper indentation was fixed recently in #2722, this PR adds it to more things. 
- Flag descriptions also have dynamic line wrapping to fit the terminal width.

<details>
<summary>Old Output, wrapped at 120 characters</summary>

```text
$ fly deploy --help | fold -w 120
Deploy Fly applications from source or an image using a local or remote builder.

To disable colorized output and show full Docker build output, set the environment variable NO_COLOR=1.

Usage:
  flyctl deploy [WORKING_DIRECTORY] [flags]

Flags:
  -a, --app string                       Application name
      --auto-confirm                     Will automatically confirm changes when running non-interactively.
      --build-arg stringArray            Set of build time variables in the form of NAME=VALUE pairs. Can be specified m
ultiple times.
      --build-only                       Build but do not deploy
      --build-secret stringArray         Set of build secrets of NAME=VALUE pairs. Can be specified multiple times. See 
https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information
      --build-target string              Set the target build stage to build if the Dockerfile has more than one stage
  -c, --config string                    Path to application configuration file
      --detach                           Return immediately instead of monitoring deployment progress
      --dockerfile string                Path to a Dockerfile. Defaults to the Dockerfile in the working directory.
  -e, --env stringArray                  Set of environment variables in the form of NAME=VALUE pairs. Can be specified 
multiple times.
      --exclude-regions strings          Deploy to all machines except machines in these regions. Multiple regions can b
e specified with comma separated values or by providing the flag multiple times. --exclude-regions iad,sea --exclude-reg
ions syd will exclude all three iad, sea, and syd regions. Applied after --only-regions. V2 machines platform only.
      --file-literal stringArray         Set of literals in the form of /path/inside/machine=VALUE pairs where VALUE is 
the content. Can be specified multiple times.
      --file-local stringArray           Set of files in the form of /path/inside/machine=<local/path> pairs. Can be spe
cified multiple times.
[...]

PR #2780, opened 2023-09-01
add trailing newline to usage template

wjordan:
Followup fix for #2733 for proper `fly help` output.

Note the original the default cobra `UsageTemplate` we used as a source includes a trailing newline:
https://github.com/spf13/cobra/blob/fd865a44e3c48afeb6a6dbddadb8a5519173e029/command.go#L567-L568


michaeldwan (approved):
Good catch!

Issue #966, opened 2022-05-10
`fly destroy` should accept a config file


vjanssens:
**Describe the bug**
`fly destroy` should accept a config file. This would make it easier to run multi-environment setups where you have for example `fly.production.toml` and `fly.staging.toml` with both apps having their own names. I'd rather not remember their specific names but simply point to a config file I know will hold the name.

So instead of:

```
fly destroy my-complex-app-name-production
```

You can run:

```
fly destroy --config ./fly.production.toml
```

* Operating system
fly v0.0.325 darwin/arm64 Commit: da2b638 BuildDate: 2022-04-28T04:00:48Z

jsierles:
Thanks for all the submissions for feature requests. We have all these good ideas in mind, but don't have an ETA for them. For now we'd like to reserve Github issues for bug reports. It would be great to open [a thread on our community forum](https://community.fly.io/), however, where more people will see the requests and chime in. That's one way we gauge interest.

redjonzaci:
@jsierles I could work on this if @vjanssens still needs it.
Depending on the issues we still need to work on, I would need to set them in a priority order.

jsierles:
Not my call, but I'd say this is low priority. Rather than working on this kind of thing one-by-one, I would categorize this under "command consistency" and try to knock them out in groups.

In other words, if some `apps` commands accept `-c` and others do not, we should decide on the correct behavior and update all commands.

redjonzaci:
I can add the `--config` flag, but I am not sure how to read the app name from the config file
so I need to do a deep dive into the `flyctl` codebase before I bother anyone.

Issue #964, opened 2022-05-10
`fly launch` should accept an existing config file for multi-environment setups


vjanssens:
**Describe the bug**
Fly launch should be able to accept an existing config file and use this configuration to 'launch' the application.
I've been working on a fly configuration but needed a separate staging environment. So I renamed my `fly.toml` file to `fly.production.toml` and copied it to `fly.staging.toml`. [This link got me on the right track](https://fly.io/docs/reference/monorepo/#examples). 

Because `fly launch --config ./fly.staging.toml` is not possible, `fly launch` will create a separate/new `fly.toml` file which I don't need because I already created one.

So, instead of running:

   fly launch --org my-organisation-name --name staging-my-app-name

You could do:

   fly launch --config ./staging.fly.toml

And the name will automatically be fetched from the config.

* Operating system

    fly v0.0.325 darwin/arm64 Commit: da2b638 BuildDate: 2022-04-28T04:00:48Z

vjanssens:
A follow-up to this could be to include a field for `organization` in the `fly.toml` specification (https://fly.io/docs/reference/configuration/). This way, you can omit  `--org my-organisation-name` from `fly launch` as well.

suchy:
I'd love to be able to put fly.toml in some config folder :(

redjonzaci:
@vjanssens @suchy do you still need this?

suchy:
Not at the moment but this is generally nice to have.

redjonzaci:
I will start implementing this and let you know how it goes in at most 7 days.

suchy:
Amazing ;-)

redjonzaci:
@suchy is your use case covered by what @vjanssens has described?

suchy:
Yes, 100%

redjonzaci:
What logic I am following: 
1. Try to launch an app from the provided path to a `fly.toml` file.
2. If not found, fallback to current way.

suchy:
Sounds correct 👍🏼 thanks

redjonzaci:
I am continuing my work on the `--config` flag, but I cannot make the decision to add the `organization` field to `fly.toml`.
Hopefully, a Fly.io employee decides in favor of adding that.

redjonzaci:
Initial implementation done,
but there are things to be discussed with the Fly.io team in the coming days.

redjonzaci:
@suchy as promised, following up 7 days later.
The Fly.io team seems to be working on higher priority issues for the moment
[...]

PR #2781, opened 2023-09-02
Add `--config` flag

redjonzaci:
### Change Summary

Related to: #964 #966


redjonzaci:
Leaving at that for now because I need to catch an early flight.

PR #2785, opened 2023-09-04
Allow changing CLI configuration directory

alexblackie:
### Change Summary

We should introduce an environment variable that, if set, will be used instead of hardcoding `~/.fly` everywhere. This allows for the user to use the CLI in non-standard environments, or in more advanced ways (for example, by swapping the config dir to isolate separate accounts).

This is pretty common for infrastructure tools; eg., the Azure CLI allows setting `AZURE_CONFIG_DIR` to do this.

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/helpers/config.go
+++ b/helpers/config.go
@@ -1,0 +1,25 @@
+package helpers
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// GetConfigDirectory will return where config and state files should be
+// stored, either respecting `FLY_CONFIG_DIR` or defaulting to the user's home
+// directory at `~/.fly`.
+func GetConfigDirectory() (string, error) {
+	var dir string
+
+	if value, isSet := os.LookupEnv("FLY_CONFIG_DIR"); isSet {
+		dir = value
+	} else {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+		dir = filepath.Join(homeDir, ".fly")
+	}
+
+	return dir, nil
+}
--- a/helpers/config.go
+++ b/helpers/config.go
@@ -9,17 +9,12 @@ import (
 // stored, either respecting `FLY_CONFIG_DIR` or defaulting to the user's home
 // directory at `~/.fly`.
 func GetConfigDirectory() (string, error) {
-	var dir string
-
 	if value, isSet := os.LookupEnv("FLY_CONFIG_DIR"); isSet {
-		dir = value
-	} else {
-		homeDir, err := os.UserHomeDir()
-		if err != nil {
-			return "", err
-		}
-		dir = filepath.Join(homeDir, ".fly")
+		return value, nil
 	}
-
-	return dir, nil
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(homeDir, ".fly"), nil
 }
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1,18 +1,19 @@
 package agent
 
 import (
-	"os"
 	"path/filepath"
+
+	"github.com/superfly/flyctl/helpers"
 )
 
 // TODO: deprecate
 func PathToSocket() string {
-	dir, err := os.UserHomeDir()
+	dir, err := helpers.GetConfigDirectory()
 	if err != nil {
 		panic(err)
 	}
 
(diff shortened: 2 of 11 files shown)


helpers/config.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// GetConfigDirectory will return where config and state files should be
+// stored, either respecting `FLY_CONFIG_DIR` or defaulting to the user's home
+// directory at `~/.fly`.
+func GetConfigDirectory() (string, error) {
+	var dir string
+
+	if value, isSet := os.LookupEnv("FLY_CONFIG_DIR"); isSet {
+		dir = value
```
dangra, line 15: Go style nitpick: return _value, nil_  here and do not use the `else` block

helpers/config.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"os"
+	"path/filepath"
+)
+
+// GetConfigDirectory will return where config and state files should be
+// stored, either respecting `FLY_CONFIG_DIR` or defaulting to the user's home
+// directory at `~/.fly`.
+func GetConfigDirectory() (string, error) {
+	var dir string
```
dangra, line 12: no need to declare `dir` at the top

dangra (commented):
do you think `internal/update/update.go` need an update too? it is user UserHomeDir + ".fly"

helpers/config.go
```diff
@@ -0,0 +1,25 @@
...
+
+	if value, isSet := os.LookupEnv("FLY_CONFIG_DIR"); isSet {
```
tvdfly, line 14: might be good to support the xdg spec env vars here too, while we're doing this. that could be done separately.

tvdfly: https://clig.dev/#configuration

tvdfly (commented):
Thanks for the pr! A couple things to consider:

1) the experience for developers that have been using flyctl for a while and then set the env var. What do you think about an opt-in migration when the config location changes, at least from the default location to somewhere else?

2) I suspect we may need to restart the agent when the config location changes between invocations of flyctl. I think there might be issues with the agent and flyctl using different configs. How about testing that out (manually is probably easiest way; preflight might work ok too)?

dangra:
> * the experience for developers that have been using flyctl for a while and then set the env var. What do you think about an opt-in migration when the config location changes, at least from the default location to somewhere else?
> *

T, I don't think we want this. I think the use case is to be able to run multiple flyctl agents on the same environment. This is common on CI and it is something I had to deal with myself when working on testflight builds long time ago.

tvdfly:
Oh yeah that's a good call Daniel.

I was thinking about the case where someone has one of those XDG env vars set by default and then when flyctl upgrades to start supporting them it loses its config. Might be a good reason to punt on XDG support for now, or at least handle that separately.

alexblackie:
> do you think `internal/update/update.go` need an update too

I did see that one but left it alone because it looked like it wasn't directly related to the configuration of the CLI as much as where the CLI binary itself was installed -- the install script just defaults to the same `~/.fly` path (but, of course, the config will always be in `~/.fly` currently, even if you set `FLY_INSTALL` to a different path).

> Might be a good reason to punt on XDG support for now

I agree, seems like it has a few potential side effects. Hopefully this work can pave the way 😉 

> I suspect we may need to restart the agent

Good call, will look into that

alexblackie:
> I suspect we may need to restart the agent

Actually, I am not sure how to go about this, since I don't think the CLI can know if there is an old agent running because the PID would be in the old (unknown to the CLI) config directory... This might just have to be a documented "gotcha"?

FWIW, since the PID is stored in the config directory, you can have two agents running at once and they don't _seem_ to conflict in my quick testing (Unless it's possible to get an IP conflict; not sure if subnet ranges are ever reused between accounts). Below I connected to two different apps in entirely different Fly accounts at the same time, each with its own agent, without issue:

Screenshot 2023-09-06 at 8 29 53 PM

image

dangra:
> Actually, I am not sure how to go about this, since I don't think the CLI can know if there is an old agent running because the PID would be in the old (unknown to the CLI) config directory... This might just have to be a documented "gotcha"?

Yes, we can ignore the path change. Multiple running agents are fine. At some point we should add a "shutdown-after-idle-timeout" setting to the agent.

alexblackie:
In that case, are there any other blockers for this?

alexblackie:
@dangra Updated!

PR #2786, opened 2023-09-04
Add LiteFS Cloud commands

benbjohnson:
### Change Summary

What and Why: This pull request adds CLI commands for managing LiteFS Cloud clusters. These commands include listing clusters, listing databases, import/exporting databases, & restoring databases.

How: Added a subcommand called `litefs-cloud` which is also aliased to `lfsc` for brevity.

Related to: Users have requested CLI access to LiteFS Cloud (https://github.com/superfly/litefs/issues/392).

---

### Documentation

- [ ] Fresh Produce

### DIFF
--- a/internal/command/lfsc/clusters.go
+++ b/internal/command/lfsc/clusters.go
@@ -1,0 +1,28 @@
+package lfsc
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/superfly/flyctl/internal/command"
+)
+
+func newClusters() *cobra.Command {
+	const (
+		long = `"Commands for managing LiteFS Cloud clusters"
+`
+		short = "Manage LiteFS Cloud clusters"
+		usage = "clusters <command>"
+	)
+
+	cmd := command.New(usage, short, long, nil,
+		command.RequireSession,
+	)
+
+	cmd.Aliases = []string{"clusters"}
+
+	cmd.AddCommand(
+		newClustersList(),
+	)
+
+	return cmd
+}
--- a/internal/command/lfsc/clusters.go
+++ b/internal/command/lfsc/clusters.go
@@ -21,6 +21,7 @@ func newClusters() *cobra.Command {
 	cmd.Aliases = []string{"clusters"}
 
 	cmd.AddCommand(
+		newClustersCreate(),
 		newClustersDestroy(),
 		newClustersList(),
 	)
--- a/internal/command/lfsc/clusters.go
+++ b/internal/command/lfsc/clusters.go
@@ -21,6 +21,7 @@ func newClusters() *cobra.Command {
 	cmd.Aliases = []string{"clusters"}
 
 	cmd.AddCommand(
+		newClustersDestroy(),
 		newClustersList(),
 	)
 
--- a/internal/command/lfsc/clusters.go
+++ b/internal/command/lfsc/clusters.go
@@ -18,8 +18,6 @@ func newClusters() *cobra.Command {
 		command.RequireSession,
 	)
 
-	cmd.Aliases = []string{"clusters"}
-
 	cmd.AddCommand(
 		newClustersCreate(),
 		newClustersDestroy(),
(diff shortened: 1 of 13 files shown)


benbjohnson:
@michaeldwan This is my first big PR against `flyctl` so I'd appreciate any feedback you have regarding conventions. I need to add docs and Fresh Produce and fix up CI but I wanted to push something up for review.

tvdfly (commented):
how long can the import/export commands take? if more than a couple seconds is possible, how about displaying a message on stderr about them potentially taking a while?

The `fly lfsc clusters list` command help was confusing to me. I ran:

```
% fly lfsc clusters list
Error: no org was provided, and none is available from the environment or fly.toml
```
then tried appending the org name:
```
fly lfsc clusters list tvd-testorg
Usage:
  flyctl usage clusters list <clustername> [flags]

Flags:
  -h, --help         help for list
  -j, --json         JSON output
  -o, --org string   The target Fly organization

Global Flags:
      --debug                 Print additional logs and traces
      --verbose               Verbose output

Error: unknown command "fly" for "flyctl usage clusters list"
```

I then figured out i was supposed to use `--org`. I think my intuition is to allow the org name to be passed in as an optional position argument. What if we provide a separate `fly lfsc clusters show` command for showing info on a single cluster (or fix the usage for the list subcommand)?

I noticed the `--json` output includes pagination info. What if we show 20 by default and allow devs to pass in a flag that enables showing more (e.g., a `--max-cluster` or `--limit` flag or something like that)?
```
% fly lfsc clusters list -o tvd-testorg --json
{
    "offset": 0,
    "limit": 20,
    "total": 0,
    "clusters": []
}```

**Comment by dangra** [comment]
"lfsc" is a bit cryptic for the non-initiated. Can we call the subcommand `litefs` instead?

File: internal/command/lfsc/clusters.go [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+`
+		short = "Manage LiteFS Cloud clusters"
+		usage = "clusters <command>"
+	)
+
+	cmd := command.New(usage, short, long, nil,
+		command.RequireSession,
+	)
+
+	cmd.Aliases = []string{"clusters"}
```
dangra, line 21: the alias is not needed as it matches the main subcommand name

internal/command/lfsc/lfsc.go
```diff
@@ -0,0 +1,122 @@
...
+	const (
+		long = `Commands for managing LiteFS Cloud databases`
+
+		short = "LiteFS Cloud management commands"
+
+		usage = "litefs-cloud <command>"
+	)
+
+	cmd := command.New("usage", short, long, nil)
+	cmd.Aliases = []string{"litefs-cloud", "lfsc"}
```
dangra, line 35: ah.. this was what my first comment was about. +1 to `litefs-cloud` but it doesn't need to be aliases. It is extracted from the usage string.

dangra (approved):

benbjohnson:
I'm finally getting back to this PR after having focus my attention elsewhere. I fixed up the PR and added cluster creation & deletion. 

> I then figured out i was supposed to use --org. I think my intuition is to allow the org name to be passed in as an optional position argument.

The `--org` flag is only needed if you're not in an app directory. It'll automatically determine the org if you are.

> What if we show 20 by default and allow devs to pass in a flag that enables showing more (e.g., a --max-cluster or --limit flag or something like that)?

I added `--offset` and `--limit` args to the `clusters list` command.

> "lfsc" is a bit cryptic for the non-initiated. Can we call the subcommand litefs instead?

I opted for `litefs-cloud` and `lfsc` instead of just `litefs` since LiteFS itself would refer to the users cluster of nodes (rather than the cloud service itself). I could see adding `litefs` commands for directly communicating with the user's LiteFS instances in the future and then `litefs export` and `litefs-cloud export` would be two separate concepts. At least, that's how I think of it in my head.

PR #2795, opened 2023-09-06
Show machine selection menu automatically

matttpt:
### Change Summary

It's tedious to type out `--select` for `fly machines` commands. As suggested in #2752 (selection prompts for volumes), show a selection menu automatically when there is no machine ID from the command line and we're running interactively.

The `--select` flag is now hidden, but it still exists for backward compatibility.

---

### Documentation

There are a few uses of `--select` in the docs that can be updated after merging, but in the meantime they won't stop working.

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -28,7 +28,7 @@ func newClone() *cobra.Command {
 		short = "Clone a Fly machine"
 		long  = short + "\n"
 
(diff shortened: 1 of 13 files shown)


alichay (approved):
didn't review the logic too closely, it's been a minute since I've been in `command/machine`, but it looks good to me

love this change btw

PR #2798, opened 2023-09-06
Rename `group` to `process-group` to make consistent with `scale count`

dejagersh:
### Change Summary

What and Why: 

Renamed `--group` parameter to `--process-group` to make consistent with `fly scale count`.

It annoys me that sometimes I have to write `--group` and sometimes I have to write `--process-group`. This PR makes them consistent across `fly scale` commands.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/scale/machines.go
+++ b/internal/command/scale/machines.go
@@ -29,7 +29,7 @@ func v2ScaleVM(ctx context.Context, appName, group, sizeName string, memoryMB in
 			return nil, err
 		}
 		if len(appConfig.Processes) > 1 {
-			return nil, fmt.Errorf("scaling an app with multiple process groups requires specifying a group with '--group <name>'\n * this app has the following process groups: %v", appConfig.FormatProcessNames())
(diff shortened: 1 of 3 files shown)


dangra (commented):
hey, this is good but we need to keep an alias to the old flag name. 

flag.String accepts a "Aliases: []string" field for that

dejagersh:
> hey, this is good but we need to keep an alias to the old flag name.
> 
> flag.String accepts a "Aliases: []string" field for that

Done! I moved it on separate lines because I thought it was getting a bit long. (Is there a formatting tool that I should run?)

dejagersh:
```
➜  flyctl git:(master) go run . scale memory -a *redacted* 2048
Error: scaling an app with multiple process groups requires specifying a group with '--group <name>'
 * this app has the following process groups: ['app', 'scheduler', 'worker']
exit status 1
```

Ah, here also `--group` is still mentioned.

dangra:
It looks good, if you still want to unify --process-group across commands, there is one more reference in internal/command/ssh/console.go

dangra:
well done and thanks a lot @dejagersh 🎉

Issue #1635, opened 2023-01-27
Deployment silently fails with malformed fly.toml


Plecra:
**Describe the bug**
With a duplicated service (I'm guessing it's the overlapping port?), deployment fails with a HTTP 504 and no details.

* Operating system
flyctl.exe v0.0.450 windows/amd64 Commit: 51325e4a BuildDate: 2023-01-13T21:57:29Z

```toml
# fly.toml file generated for reddit-assist on 2023-01-27T13:19:32Z

app = "reddit-assist"
kill_signal = "SIGINT"
kill_timeout = 5
processes = []

[env]

[experimental]
  auto_rollback = true

[[services]]
  http_checks = []
  internal_port = 8080
  processes = ["app"]
  protocol = "tcp"
  script_checks = []
  [services.concurrency]
    hard_limit = 25
    soft_limit = 20
    type = "connections"

  [[services.ports]]
    force_https = true
    handlers = ["http"]
    port = 80

  [[services.ports]]
    handlers = ["tls", "http"]
    port = 443

  [[services.tcp_checks]]
    grace_period = "1s"
    interval = "15s"
    restart_limit = 0
    timeout = "2s"

[[services]]
  http_checks = []
  internal_port = 8080
  processes = ["app"]
  protocol = "tcp"
[build]
  builder = "paketobuildpacks/builder:base"
  buildpacks = ["docker.io/paketocommunity/rust"]

[mounts]
source = "reddit_assist_store"
destination = "/data"
```

tvdfly:
Thanks for reporting this issue! flyctl should give a better error message in that case.

And you are correct, services need to have unique internal_port. Removing the second duplicate `[[services]]` block should fix the issue.

dangra:
@redjonzaci look at internal/appconfig/validation.go

redjonzaci:
What I was able to undestand is that I should return a validation error if 2 services have the same `internal_port`.
Is that correct?

dangra:
kind of, it may need to take protocol and processes into account. go for it and see what makes sense.

redjonzaci:
I tested with the `fly.toml` that @Plecra has provided. 🚀

PR #2810, opened 2023-09-10
Return validation error on duplicate service found

redjonzaci:
### Change Summary

Related to: #1635

### DIFF
--- a/internal/appconfig/testdata/validation-services-2-duplicate.toml
+++ b/internal/appconfig/testdata/validation-services-2-duplicate.toml
@@ -1,0 +1,14 @@
+app = "validation"
+
+[[services]]
+internal_port = 8080
+protocol = "tcp"
+processes = ["bash"]
+
+[[services]]
+internal_port = 8080
+protocol = "tcp"
+processes = ["bash"]
+
+[processes]
+bash = "bash"
--- a/internal/appconfig/testdata/validation-checks-invalid.toml
+++ b/internal/appconfig/testdata/validation-checks-invalid.toml
@@ -1,0 +1,11 @@
+app = "validation"
+
+[checks]
+[checks.test_http_check]
+grace_period = "30s"
+interval = "1s"
+method = "get"
+path = "/path/to/status"
+port = 5500
+timeout = "10s"
+type = "http"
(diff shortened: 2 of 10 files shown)


dangra:
The good thing about `internal/appconfig` package is that is easy to test and there are many examples.
The not so good thing is that there are no examples for config _validation_, but it is a great time to add `validation_test.go` and start by testing different service combinations.

Also, thinking on the possible port+protocol+processes combinations with help you realize how process groups work, and how your code is still buggy.

You have been contributing many small PRs lately, I think directions are good for starters, but at this point it is about time to level up and try to think on the edge cases by yourself before declaring your work as ready for review. That said, don't feel obligated to do any of this, I appreciate your contributions to this open source project but I also understand you do them on _best effort/your time_.

thanks!

redjonzaci:
Thanks for the feedback!

redjonzaci:
Making progress, but I understand that if I am to be more independent and efficient,
I have to go through both the docs and the code.

redjonzaci:
I added 6 test cases for `validateServicesSection` and 2 for `validateConsoleCommand`.

redjonzaci:
Will also try to cover: 
- `validateBuildStrategies`
- `validateDeploySection`
- `validateChecksSection`
- `validateMachineConversion`

internal/appconfig/testdata/validation-services-2-duplicate.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[[services]]
+internal_port = 8080
+protocol = "tcp"
+processes = ["bash"]
+
+[[services]]
+internal_port = 8080
+protocol = "tcp"
+processes = ["bash"]
```
dangra, line 11: define an additional process (i.e: "app") and add it only to this service.

redjonzaci: This would then produce the undefined process error (if added only there), maybe it should be another test case.

redjonzaci: I will be reading some more, you don't have to respond until I get back from the reading.

redjonzaci: [Processes run in their own Fly Machines](https://fly.io/docs/apps/processes/#:~:text=Every%20defined%20process). And services seem to be "decorators" of processes?
From this point of view, I don't see a problem with having multiple identical `[[services]]` sections besides those being redundant.

Issue #2658, opened 2023-08-07
`fly launch` doesn't work on nuxt 3 aplication.


Rohithgilla12:
*Please only report specific issues with `flyctl` behavior*. Anything like a support request for your application should go to https://community.fly.io. More people watch that space and can help you faster!

**Describe the bug**

Was following [this](https://fly.io/docs/languages-and-frameworks/nuxtjs/) guide, the `fly launch` command detected the `nuxt` app as a `node` app.

* Operating system `macOS`
* `fly version` : `0.1.7`
![image](https://github.com/superfly/flyctl/assets/19389850/de32daee-6a26-409d-add1-79ff83f88e26)

** Paste your `fly.toml`
Generates a node js docker file.

** Command output: **

![image](https://github.com/superfly/flyctl/assets/19389850/2002475b-6290-48fb-a071-e0cd1b683143)
```text
❯ fly launch
Creating app in /tmp/nuxt-app
Scanning source code
Detected a NodeJS app ?
```

redjonzaci:
https://github.com/superfly/flyctl/blob/64a31bf73c6e1509567cc6be0b2a19340e80e810/scanner/nuxtjs.go#L4
I think the issue is here. [Configuration · Get Started with Nuxt](https://nuxt.com/docs/getting-started/configuration) uses `nuxt.config.ts` by default.

redjonzaci:
I just made it work @Rohithgilla12! Opening a PR.

Rohithgilla12:
That's so cool
Thanks a lot 🫡

redjonzaci:
Hope this gets merged tomorrow. Otherwise, will tag a Fly.io team member.

Rohithgilla12:
@redjonzaci Looks like it's not reviewed yet 👀

redjonzaci:
Hey @Rohithgilla12, sorry about that, but I am sure Fly.io team members must be working on higher priority issues.

But to unblock you, you can rename `nuxt.config.ts` to `nuxt.config.js` and then when `fly launch` prompts you to deploy, deny and add these changes to the Dockerfile that is generated:
https://github.com/superfly/flyctl/pull/2812/commits/42f1aa8e49f4f257a4e5d2c12f78198a7f6449a5#diff-27c9f5cf2c02c33c96d414a8acc31092e11ebd1df8ba68884a55830b17523b9a

And then `fly deploy`.

Let me know if it works!

PR #2812, opened 2023-09-10
Update Nuxt scanner and Dockerfile

redjonzaci:
### Change Summary

Related to: #2658

### DIFF
--- a/scanner/nuxtjs.go
+++ b/scanner/nuxtjs.go
@@ -1,7 +1,7 @@
 package scanner
 
 func configureNuxt(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
-	if !checksPass(sourceDir, fileExists("nuxt.config.js")) {
+	if !checksPass(sourceDir, fileExists("nuxt.config.ts")) {
 		return nil, nil
 	}
 
--- a/scanner/templates/nuxtjs/Dockerfile
+++ b/scanner/templates/nuxtjs/Dockerfile
@@ -13,20 +13,14 @@ RUN yarn install \
 
 RUN yarn build
 
-RUN rm -rf node_modules && \
-  NODE_ENV=production yarn install \
-  --prefer-offline \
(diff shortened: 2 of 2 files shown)


Rohithgilla12:
Do we have any update when it might be merged 🙈

dangra:
will be shipped in v0.1.114, please @Rohithgilla12 test it and confirm it works as expected

Rohithgilla12:
@dangra Looks like it doesn't work :( 


Created a new `nuxt` project with `pnpm dlx nuxi@latest init fly-test`


`fly` version
image

`fly launch`
image

dangra:
The problem here seems to be that the [scanner's priority order for Nuxt](https://github.com/superfly/flyctl/blob/93001806ee467ed760df247337b31fa0d585d5b6/scanner/scanner.go#L119) is way below the generic [JsFramework](https://github.com/superfly/flyctl/blob/93001806ee467ed760df247337b31fa0d585d5b6/scanner/scanner.go#L108) scanner, so the project is detected as a "NodeJS" app.

@rubys my understanding of Javascript frameworks is very limited, no idea what is the best approach here, either add Nuxt support to JsFramework scanner or move Nuxt scanner up in the chain. What do you suggest?

Rohithgilla12:
@dangra I would suggest moving up the chain since they have a very particular config files per framework.
`Nuxt` -> `nuxt.config.ts`
`Next` -> `next.config.ts`
`Remix` -> `remix.config.js`

rubys:
Sorry, just noticing this now.  I would prefer to update the JSFramework scanner.

Basically the previous scanners would drop a precanned Dockerfile into your directory and you are on your own.  The current nuxt file assumes yarn, but perhaps you are using npm or pnpm or bun?  Perhaps you want caching?  Perhaps you are on Windows?  Perhaps you use puppeteer?  legacy-peer-deps?  sqlite3?  litefs?

PR #2816, opened 2023-09-11
Create volume and configure LITESTACK_DATA_PATH if `litestack` gem is present

bradgessler:
When a Rails application is deployed with Litestack enabled, it would assume Sqlite and configure the application accordingly.

### Change Summary

What and Why: Litestack makes it possible to deploy Rails applications that only need Sqlite (not Redis or database servers). This scanner streamlines the deployment by detecting the presence of the gem and configuring the application accordingly.

Related to: Rails Sqlite production deployments

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


bradgessler:
Going to implement this through the dockerfile-rails gem.

PR #2817, opened 2023-09-12
Validate machine limits before performing b/g deployment

gwuah:
### Change Summary
This PR adds machine limit validation to bluegreen deployments. It will allow us fail fast & reduce chances of creating zombie machines when a user hits their org limits in the middle of a deployment.

### DIFF
--- a/gql/schema.graphql
+++ b/gql/schema.graphql
@@ -6292,7 +6292,7 @@ input MigrateVolumeInput {
   """
   The target Volume ID to be migrated
   """
-  targetVolumeId: Int!
+  targetVolumeId: String!
 
   """
   Use the latest snapshot instead of creating a new one.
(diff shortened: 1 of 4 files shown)


gql/schema.graphql
```diff
@@ -6292,7 +6292,7 @@ input MigrateVolumeInput {
   """
   The target Volume ID to be migrated
   """
-  targetVolumeId: Int!
```
michaeldwan, line 6295: This seems unrelated?

gql/schema.graphql
```diff
@@ -8702,7 +8712,7 @@ input RegisterVolumeInput {
   Desired volume size, in GB
   """
   sizeGb: Int!
-  snapshotId: ID
```
michaeldwan, line 8705: also unrelated?

michaeldwan (approved):
your change looks good, but comments about possible unrelated changes. 

Also, is it possible to get this from flaps instead of graphql?

PR #2823, opened 2023-09-14
Remove . from `flyctl launch` long desc

redjonzaci:
### Change Summary

What and Why:
This might be the smallest change ever,
but I happened to notice the other commands in "Here's a few commands to get you started" didn't have a . in the end.


billyb2 (approved):
[banger pr](https://github.com/tailscale/ToBeReviewedBot/pull/29)

PR #2825, opened 2023-09-14
support separate metrics config per process group

wjordan:
Similar to other app-config constructs (e.g., #2086), this change allows specifying a separate Metrics config per process group with a `processes` field:

```toml
[[metrics]]
  port = 9998
  path = "/metrics"
  processes = ["web"]
```

### Documentation

- [ ] Fresh Produce
- [x] In superfly/docs (superfly/docs#1028)

### DIFF
--- a/internal/appconfig/machines.go
+++ b/internal/appconfig/machines.go
@@ -105,7 +105,10 @@ func (c *Config) updateMachineConfig(src *api.MachineConfig) (*api.MachineConfig
 	}
 
 	// Metrics
-	mConfig.Metrics = c.Metrics
+	mConfig.Metrics = nil
+	if c.Metrics[0] != nil {
+		mConfig.Metrics = c.Metrics[0].MachineMetrics
+	}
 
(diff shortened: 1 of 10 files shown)


dangra (commented):
This is great but we need a test case in `internal/appconfig/testdata/old-format.toml` to test backwards compatibility of `[metrics]`

internal/appconfig/machines.go
```diff
@@ -105,7 +105,10 @@ func (c *Config) updateMachineConfig(src *api.MachineConfig) (*api.MachineConfig
 	}
 
 	// Metrics
-	mConfig.Metrics = c.Metrics
+	mConfig.Metrics = nil
+	if c.Metrics != nil && len(c.Metrics) >= 1 {
```
dangra, line 109: just a nit: `len(c.Metrics)` is valid (and zero) even if `c.Metrics` is `nil`

dangra (approved):

PR #2833, opened 2023-09-18
legacy pg: don't try parsing plaintext str as json

alichay:
### Change Summary

What and Why: when migrating, we need to know a node's role. For some reason, the primary endpoint (`commands/admin/role`) returns errors for some people (404, `{"error":"no rows in result set"}`).

The fallback path is to use `flycheck/role`, but flyctl attempts to parse its plaintext response as a JSON-encoded string.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/flypg/http.go
+++ b/flypg/http.go
@@ -58,31 +58,51 @@ func newHttpClient(dialer agent.Dialer) *http.Client {
 	return &http.Client{Transport: logging}
 }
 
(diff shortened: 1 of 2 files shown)


redjonzaci:
Now it probably doesn't matter coming from me, but LGTM 😂

PR #2842, opened 2023-09-20
Support --image, --command, and --env on fly console

rubys:
### Change Summary

What and Why:

In order to support custom build and release processes: enable alternate image, command, and environment to be specified for fly console.

How:

Flags added, and used to populate/override the LaunchInput

Related to:

Feedback from the RemixJS team

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/console/console.go
+++ b/internal/command/console/console.go
@@ -326,6 +326,12 @@ func makeEphemeralConsoleMachine(ctx context.Context, app *api.AppCompact, appCo
 	}
 	machConfig.DNS.SkipRegistration = flag.GetBool(ctx, "skip-dns-registration")
 
+	machineFiles, err := command.FilesFromCommand(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed parsing filest: %w", err)
+	}
+	machine.MergeFiles(machConfig, machineFiles)
+
 	machConfig.Guest = guest
 
 	services, err := command.DetermineServices(ctx, machConfig.Services)
--- a/internal/command/console/console.go
+++ b/internal/command/console/console.go
@@ -294,6 +294,11 @@ func makeEphemeralConsoleMachine(ctx context.Context, app *api.AppCompact, appCo
 	if err != nil {
 		return nil, nil, err
 	}
+
+	if !flag.IsSpecified(ctx, "image") && flag.IsSpecified(ctx, "dockerfile") {
+		flag.SetString(ctx, "image", ".")
+	}
+
 	if currentRelease == nil && !flag.IsSpecified(ctx, "image") {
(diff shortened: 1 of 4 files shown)


internal/command/console/console.go
```diff
@@ -210,15 +233,29 @@ func makeEphemeralConsoleMachine(ctx context.Context, app *api.AppCompact, appCo
...
+	} else {
+		machConfig.Image = currentRelease.ImageRef
+	}
+
+	if env := flag.GetStringArray(ctx, "env"); len(env) > 0 {
+		parsedEnv, err := cmdutil.ParseKVStringsToMap(env)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed parsing environment: %w", err)
+		}
+		machConfig.Env = parsedEnv
```
matttpt, line 256: In the [`ToConsoleMachineConfig`](https://github.com/superfly/flyctl/blob/4e09fc60ad452dda368372a9767386d8a5c0549d/internal/appconfig/machines.go#L64) call, `machConfig.Env` is populated with environment variables set in the app's configuration and a few special ones, e.g. `FLY_PROCESS_GROUP`.

In light of this, I think you mean to add/overwrite the individual variables, rather than to replace the whole map?

rubys: Good catch!  In my case `FLY_PROCESS_GROUP` and `PRIMARY_REGION` were stomped on.  Fixed.

amo13:
Amazing!
Can you please make this work for the `flyctl ssh console` command too?

internal/command/console/console.go
```diff
@@ -56,6 +59,81 @@ func New() *cobra.Command {
...
+			Name:      "port",
+			Shorthand: "p",
+			Description: `Publish ports, format: port[:machinePort][/protocol[:handler[:handler...]]])
+		i.e.: --port 80/tcp --port 443:80/tcp:http:tls --port 5432/tcp:pg_tls
+		To remove a port mapping use '-' as handler, i.e.: --port 80/tcp:-`,
+		},
+		flag.Bool{
+			Name:        "skip-dns-registration",
+			Description: "Do not register the machine's 6PN IP with the internal DNS system",
+		},
```
matttpt, line 136: Console machines already have DNS registration skipped by default. Maybe add `Default: true` so that it gets documented in the help text?

rubys: Changing it to default true is a good idea; but I'm going to also mark it Hidden at the same time.  That way if there is a need it can be shared with others but more likely the topic won't come up.

Good catch!

michaeldwan (approved):

michaeldwan:
@rubys Looks good, works locally. I'm looking forward to seeing the use cases.

PR #2847, opened 2023-09-22
Installer script tests, install config fixes

michaeldwan:
- adds basic tests for the installation scripts on linux, macos, windows, and a few linux distros in Docker
- fixes an issue where the release channel passed to flyctl during installation never gets persisted.


billyb2 (approved):
banger pr

PR #2849, opened 2023-09-22
Support Dockerfile in a parent directory

kzys:
Having a Dockerfile in a parent directory, like below should work, but it didn't.

```
[build]
dockerfile = "../Dockerfile"
```

### Change Summary

What and Why: The example above doesn't work without this change.

How: By removing relDockerfile. It was copied from the original code without much thoughts.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/build/imgsrc/dockerfile_builder.go
+++ b/internal/build/imgsrc/dockerfile_builder.go
@@ -221,7 +221,7 @@ func (*dockerfileBuilder) Run(ctx context.Context, dockerFactory *dockerClientFa
 
 	build.SetBuilderMetaPart2(buildkitEnabled, serverInfo.ServerVersion, fmt.Sprintf("%s/%s/%s", serverInfo.OSType, serverInfo.Architecture, serverInfo.OSVersion))
 	if buildkitEnabled {
-		imageID, err = runBuildKitBuild(ctx, docker, opts, relDockerfile, buildArgs)
+		imageID, err = runBuildKitBuild(ctx, docker, opts, dockerfile, buildArgs)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/build/imgsrc/dockerfile_builder.go
```diff
@@ -221,7 +221,7 @@ func (*dockerfileBuilder) Run(ctx context.Context, dockerFactory *dockerClientFa
 
 	build.SetBuilderMetaPart2(buildkitEnabled, serverInfo.ServerVersion, fmt.Sprintf("%s/%s/%s", serverInfo.OSType, serverInfo.Architecture, serverInfo.OSVersion))
 	if buildkitEnabled {
-		imageID, err = runBuildKitBuild(ctx, docker, opts, relDockerfile, buildArgs)
+		imageID, err = runBuildKitBuild(ctx, docker, opts, dockerfile, buildArgs)
```
michaeldwan, line 224: do we need this fix on [line 234](https://github.com/superfly/flyctl/pull/2849/files#diff-c76cea6ab9d205c8b3838eecbf83e23ea3793acde3f112ca0f421dabd120b29eR234) as well?

michaeldwan (approved):
lgtm with a question about the classic docker builder

internal/build/imgsrc/dockerfile_builder.go
kzys: Seems non-Buildkit build is fine. https://flyio.discourse.team/t/specifying-a-non-pwd-path-to-dockerfile-in-fly-toml-spits-out-an-error/3975 cc @Roadmaster

Issue #2851, opened 2023-09-24
Removing the `fly.toml` `files` section should remove Machine files from config


catflydotio:
Customer report. https://community.fly.io/t/files-removed-from-fly-toml-persist-in-the-apps-vm-after-deploy/15582

flyctl v0.1.101

To reproduce, first add a file to the config with this in `fly.toml`. Deploy.

```toml
[[files]]
  guest_path = "/app/lookatthis"
  local_path = "localfile.txt"
```

Check Machine config with `fly m status -d`. Find a `files` entry, something like this:

```
"files": [
    {
      "guest_path": "/app/lookatthis",
[...]

PR #2853, opened 2023-09-25
reset files if none are passed in

senyosimpson:
### Change Summary

What and Why:

See #2851

How:

Just sets the `Files` var to an empty list if no files are passed in

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


jipperinbham (commented):
We can't easily do this because they may not pass the files as flags every time. We only need to clear it out if the toml files have been removed.

jipperinbham (approved):
Got clarification on behavior, LGTM

jipperinbham (approved):

PR #2854, opened 2023-09-25
Update docs to say that --remote-only is the default

andie787:
### Change Summary

What and Why: 
`--remote-only` is the default way to build, so added that to the docs.

How:
docs

Related to:
n/a

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/flag/flag.go
+++ b/internal/flag/flag.go
@@ -390,7 +390,7 @@ const remoteOnlyName = "remote-only"
 func RemoteOnly(defaultValue bool) Bool {
 	return Bool{
 		Name:        remoteOnlyName,
-		Description: "Perform builds on a remote builder instance instead of using the local docker daemon",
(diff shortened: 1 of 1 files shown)


internal/flag/flag.go
```diff
@@ -390,7 +390,7 @@ const remoteOnlyName = "remote-only"
 func RemoteOnly(defaultValue bool) Bool {
 	return Bool{
 		Name:        remoteOnlyName,
-		Description: "Perform builds on a remote builder instance instead of using the local docker daemon",
+		Description: "Perform builds on a remote builder instance instead of using the local docker daemon. This is the default. Use --local-only to build locally.",
```
andie787, line 393: @catflydotio do you think we should say "using a remote builder app" instead of "on a remote builder instance"? Or something like that?

catflydotio (approved):
lgtm

PR #2855, opened 2023-09-25
Add restart policy to fly.toml

rugwirobaker:
### Change Summary

What and Why:
Previously you could only set the restart policy for one machine at time with `fly m` command. Now it's possible to set a restart policy for each of your processes

How:

```toml
app = "app-name"
primary_region = "jnb"

+[[restart]]
  +policy = "no"
  +retries = 10
  +processes = ["app"]

[build]

[http_service]
  internal_port = 3000
  force_https = true
  auto_stop_machines = true
  auto_start_machines = true
  min_machines_running = 0
  processes = ["app"]
```

Options: 
* always
* no
* on-failure

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a


internal/appconfig/config.go
```diff
@@ -25,6 +25,14 @@ const (
 	DetachedPlatform = "detached"
 )
 
+type RestartPolicy string
+
+const (
+	RestartPolicyAlways    RestartPolicy = "always"
+	RestartPolicyNever     RestartPolicy = "no"
```
benwaffle, line 32: Can we make users write `policy = "never"` instead of `policy = "no"`?

fly.toml
```diff
@@ -0,0 +1,30 @@
+app = "test-config-21"
```
benwaffle, line 1: I don't think you meant to add this file

benwaffle (changes requested):

fly.toml
rugwirobaker: Good catch

internal/appconfig/config.go
rugwirobaker: I want to change it everywhere so it doesn't bite us in the future.

internal/appconfig/machines.go
```diff
@@ -190,6 +190,12 @@ func (c *Config) updateMachineConfig(src *api.MachineConfig) (*api.MachineConfig
 	// Files
 	machine.MergeFiles(mConfig, c.MergedFiles)
 
+	// Restart Policy
+	mConfig.Restart = api.MachineRestart{
+		Policy:     api.MachineRestartPolicy(c.Restart.Policy),
+		MaxRetries: c.Restart.MaxRetries,
+	}
```
dangra, line 197: that this block is not guarded with `if c.Restart != nil {` suggests it needs a test case. 

Follow the pattern of the other fields, nullify them and then set if set in appconfig.

```
mConfig.Restart = nil
if c.Restart != nil {
  mConfig.Restart = ...
}
```

internal/command/deploy/machines_launchinput_test.go
```diff
@@ -202,7 +202,7 @@ func Test_launchInputForUpdate_keepUnmanagedFields(t *testing.T) {
 	assert.Equal(t, "ord", li.Region)
 	assert.Equal(t, "24/7", li.Config.Schedule)
 	assert.Equal(t, true, li.Config.AutoDestroy)
-	assert.Equal(t, api.MachineRestart{Policy: api.MachineRestartPolicyNo}, li.Config.Restart)
```
rugwirobaker, line 205: @dangra  I feel like this test is counterintuitive to what I'm trying to do. What happens if you don't send a Restart policy to flaps? Does it automatically generate one?

dangra: > What happens if you don't send a Restart policy to flaps? Does it automatically generate one?

Yes, it defaults to "no" restart policy.

> I feel like this test is counterintuitive to what I'm trying to do.

That's true, the reason is that now the restart policy is set from fly.toml so it shouldn't preserve what was previously set for the machine with `fly m update`

internal/appconfig/config.go
benwaffle: Sorry, not sure what this means

benwaffle: I was thinking something like
```patch
diff --git a/internal/appconfig/config.go b/internal/appconfig/config.go
index 2e221321..4ff5d650 100644
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -29,7 +29,7 @@ type RestartPolicy string
 
 const (
 	RestartPolicyAlways    RestartPolicy = "always"
-	RestartPolicyNever     RestartPolicy = "no"
+	RestartPolicyNever     RestartPolicy = "never"
 	RestartPolicyOnFailure RestartPolicy = "on-failure"
 )
 
diff --git a/internal/appconfig/machines.go b/internal/appconfig/machines.go
index c0f5a9c8..c6135562 100644
--- a/internal/appconfig/machines.go
+++ b/internal/appconfig/machines.go
@@ -1,6 +1,7 @@
 package appconfig
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/google/shlex"
@@ -194,14 +195,31 @@ func (c *Config) updateMachineConfig(src *api.MachineConfig) (*api.MachineConfig
 	mConfig.Restart = nil
 
 	if c.Restart != nil {
+		policy, err := parseRestartPolicy(c.Restart.Policy)
+		if err != nil {
+			return nil, err
+		}
 		mConfig.Restart = &api.MachineRestart{
-			Policy:     api.MachineRestartPolicy(c.Restart.Policy),
+			Policy:     policy,
 			MaxRetries: c.Restart.MaxRetries,
 		}
 	}
 	return mConfig, nil
 }
 
+func parseRestartPolicy(policy RestartPolicy) (api.MachineRestartPolicy, error) {
+	switch policy {
+	case RestartPolicyAlways:
+		return api.MachineRestartPolicyAlways, nil
+	case RestartPolicyOnFailure:
+		return api.MachineRestartPolicyOnFailure, nil
+	case RestartPolicyNever:
+		return api.MachineRestartPolicyNo, nil
+	default:
+		return "", errors.New("invalid restart policy")
+	}
+}
+
 func (c *Config) tomachineSetStopConfig(mConfig *api.MachineConfig) error {
 	mConfig.StopConfig = nil
 	if c.KillSignal == nil && c.KillTimeout == nil {
```

benwaffle (approved):

api/machine_types.go
```diff
@@ -601,7 +601,7 @@ type MachineConfig struct {
 	Schedule string `json:"schedule,omitempty"`
 	// Optional boolean telling the Machine to destroy itself once it’s complete (default false)
 	AutoDestroy bool             `json:"auto_destroy,omitempty"`
-	Restart     MachineRestart   `json:"restart,omitempty"`
+	Restart     *MachineRestart  `json:"restart,omitempty"`
```
dangra, line 604: just to raise a warning, this change will need some changes in flaps side as it imports flyctl/api module

internal/appconfig/config.go
```diff
@@ -165,6 +174,10 @@ type Compute struct {
 	*api.MachineGuest `toml:",inline" json:",inline"`
 	Processes         []string `json:"processes,omitempty" toml:"processes,omitempty"`
 }
+type Restart struct {
+	Policy     RestartPolicy `toml:"policy,omitempty" json:"policy,omitempty"`
+	MaxRetries int           `toml:"retries,omitempty" json:"retries,omitempty"`
```
dangra, line 179: the thing missing here is the `processes` key as other section have. 

Also as `Restart` is a section, it must go after toplevel keys like `ConsoleCommand`

dangra: oh, the `processes` thing can be implemented in a following PR if you want to get this merged soonish

dangra (approved):

api/machine_types.go
rugwirobaker: How do we usually do that @dangra? I'd have to merge this first right?

dangra: Yes, that is one way of doing it but only if you plan to do it right after merging the PR. Otherwise pull the updated flyctl/api dependency from this branch, prepare the changes and merge once it is all ready.

benwaffle (approved):

api/machine_types.go
```diff
@@ -0,0 +1,764 @@
+package api
```
benbjohnson, line 1: This file has been moved to `fly-go`. Can you remove it from the PR?

benbjohnson (changes requested):

billyb2 (approved):
banger pr

benbjohnson (approved):

internal/command/postgres/barman.go
```diff
@@ -188,7 +188,10 @@ func runBarmanCreate(ctx context.Context) error {
 	}
 
 	// Restart policy
-	machineConfig.Restart = &fly.MachineRestart{Policy: fly.MachineRestartPolicyAlways}
+	machineConfig.Restart.Policy = fly.MachineRestartPolicyAlways
```
dangra, line 191: this line doesn't look good. It fails if `machine.Config.Restart` is `nil` but also it is no-op considering the next line assignment.

internal/appconfig/validation.go
```diff
@@ -320,3 +321,39 @@ func (cfg *Config) validateMounts() (extraInfo string, err error) {
...
+			if !slices.Contains(validGroupNames, processName) {
+				extraInfo += fmt.Sprintf("Restart policy specifies '%s' as one of its processes, but no processes are defined with that name; "+
+					"update fly.toml [processes] to add '%s' process or remove it from restart policy's processes list\n",
+					processName, processName,
+				)
+				err = ValidationError
+			}
+		}
+
+		switch restart.Policy {
```
dangra, line 344: you can reuse `parseRestartPolicy` function here to keep validation and parsing in sync.

PR #2861, opened 2023-09-27
`migrate-to-v2`: add flag to force standard migration on postgres apps

alichay:
### Change Summary

What and Why: Adds the flag `--force-standard-migration`, which forces the migration to proceed as if the app is not a Postgres app, even if it is. This allows us to (hopefully) migrate clusters that aren't perfectly healthy.

This is a last-resort, meant to be used in the final pass of automatic migrations as we approach the deadline.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


billyb2 (approved):
banger pr

PR #2864, opened 2023-09-27
Version package

michaeldwan:
### Change Summary

This PR adds a new package that handles parsing, comparisons, and types for the version numbers flyctl uses. It builds on the work @ndarilek did to prep us for a transition from semver to calver formats. The behavior should be the same as our current code for stable releases, but will support dev builds using our upcoming distribution system that uses calver. Eventually we'll switch over stable releases to calver. This package will let us handle both in parallel throughout the transition. 

As for why calver.... well, we'll say more about that later. But the tldr is that at the moment we're not following semantic versioning in a meaningful way, which creates confusion and false assumptions about upgrades and compatibility. flyctl tracks the platform, and the platform is continually moving forward independent of flyctl's release cadence. Changing to calendar based versions communicates this better. 

Here's an example of the version numbers this supports:
- 0.1.123 - semver stable releases
- 0.1.123-pre-1 - semver prereleases
- 0.0.500 - semver releases from aeons ago that nobody should use
- 2023.9.28-stable.1 - the first stable build on September 28, 2023
- 2023.9.28-stable.3 - the third stable build on September 28, 2023
- 2023.9.29-stable.1 - the first stable build on September 29, 2023
- 2023.9.28-prerelease.1 - the first prerelease build on September 28, 2023
- 2023.9.28-pr2862.1 - the first dev build for PR 2862 on September 28, 2023
- 2023.9.28-nightly.1 - the nightly build on September 28, 2023

and so on...


internal/buildinfo/env_dev.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"time"
+
+	"github.com/superfly/flyctl/internal/version"
+)
+
+var environment = "development"
+
+func loadBuildTime() error {
+	cachedBuildTime = time.Now()
```
alichay, line 14: This seems odd, dev builds report that they were built when they're run?

michaeldwan: Oh yes good eye! This is a temporary hack to avoid disrupting the build scripts while we test stuff out. A subsequent PR will use the build time which _will_ be passed in after build scripts are updated.

michaeldwan: Just for reference, using `time.Now()` is [the current behavior](https://github.com/superfly/flyctl/pull/2864/files#diff-4e852a88a07db777820550fedc019767dbb71b3fa558f4e7a59a6bfcdfe76d15L192-L216). This just moves it to another place for now. It's odd, but made sense in actual dev because it was only ran on a local machine. I'd like to create another "environment" for these in-dev preview builds. But I don't know what to call them. Let me know if you have suggestions :D

ndarilek: Maybe skunkworks, in keeping with the fly/R&D themes?

ndarilek (approved):

internal/buildinfo/env_dev.go
michaeldwan: oh I like that

billyb2 (approved):
banger pr

all seriousness, this is great. No noticeable issues and the testing seems solid

PR #2870, opened 2023-09-30
Extension provisioning updates

jsierles:
Redis users currently provision databases independently from apps. We want to support this behavior for all extensions, but we should default to associating an app should one be in the context of an app directory or have passed `-a`.

This PR also: 

* Removes extension auto-provisioning related code, since we don't plan to reintroduce that behavior
* Ensures that customers accept provider ToS agreements explicitly when provisioning

Note: normally these changes might be split into separate PRs, but the behavior across them would be too coupled to make it worthwhile. Also, the changes here are important for continued work from @ndarilek on FKS.


ndarilek (approved):

ndarilek:
LGTM. Would like to merge this soon if possible so I can integrate it with the fks branch.

PR #2872, opened 2023-10-02
`migrate-to-v2`: migrate volumes from maa to bom

alichay:
### Change Summary

What and Why: `maa` is deprecated. machines are moved to `bom`, but if they use volumes, those are not, which prevents migration from succeeding.

How: Migrate volumes remotely from `maa` -> `bom`

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


billyb2 (approved):
banger pr

PR #2876, opened 2023-10-03
add bin installation instructions

marcosnils:
Adds installation instructions to use bin (https://github.com/marcosnils/bin) as a simplified multi-platform method.


NullHypothesis:
Thanks for the PR! I'm closing because we prefer to limit the third-party tools we recommend for installation.

PR #2879, opened 2023-10-04
Add missing build tags for windows only file

dangra:
Follow up for #2726


alichay:
Do we need this? It was my understanding that the suffix `_windows` was enough to prevent building on a non-windows OS

dangra:
TIL https://pkg.go.dev/go/build#hdr-Build_Constraints

thanks @alichay

PR #2881, opened 2023-10-04
bluegreen: gracefully tear down blue machines

matttpt:
### Change Summary

**What and Why:**

Currently blue (old) machines are torn down at the end of a blue-green deployment by force-destroying them. This gives only 5 seconds before the machine is abruptly terminated. For users who rely on an explicitly specified `kill_timeout` to allow for graceful shutdowns, this is unexpected and can break their apps (see the forum post linked below). It's also inconsistent with rolling deployments, which do respect the `kill_timeout`.

This PR implements graceful tear-down of blue machines (and fixes an existing case where the hanging-machines message isn't shown).

**How:**

This inserts two steps immediately before destroying blue machines: (1) signaling them to stop and (2) waiting for them to stop. If either step fails for a machine, it's marked as "hanging" (as is done in the existing destruction code), and it's ignored in the subsequent steps.

**Related to:**

* https://community.fly.io/t/sigterm-sent-twice-not-respecting-kill-timeout-then-virtual-machine-exited-abruptly/15720

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/strategy_bluegreen.go
+++ b/internal/command/deploy/strategy_bluegreen.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/samber/lo"
 
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/flaps"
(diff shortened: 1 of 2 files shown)


pborzenkov:
@matttpt Sorry, I haven't seen this PR before and implemented something similar myself - https://github.com/superfly/flyctl/pull/3297

I suppose this PR can be closed.

matttpt:
@pborzenkov no worries! I forgot about this one, and yours handles cordoning too. 😃

PR #2884, opened 2023-10-05
Avoid agent restarts on each run by fixing the buildtime for dev

dangra:
The agent gets restarted everytime a dev version of flyctl is executed because its version differs due to per-execution timestamped version.


michaeldwan:
@dangra What problem are you trying to solve? For context, this behavior (while possibly unexpected) is intentional because the agent drifting in dev was causing hard to debug issues. At the time dev work fit into one of these groups:
- you don't care about the agent, so just let it get killed and restarted as needed
- you do care about maintaining agent connections, so run it manually in the foreground with `go run . agent run`, then `go run .` will pick it up

Most of the time dev work fell into the first group, so that was the default behavior. This does catch folks off now and again, and I'm all ears if you have a more elegant way to keep the agent from drifting automatically.

dangra:
@michaeldwan I rarely use `go run .`, I use the Makefile at the root of the project to build the flyctl binary. For me it is as simple as running `make`, let it build the binary which is symlinked from `~/bin/fly` (also in ~/bin is in $PATH).

My gripe with current *per-execution version* is that it forces an agent restart even if nothing changed, that shuts down any _ssh console_ session, in fact, it is not possible to have more than one ssh console open at the same time. I am looking for a *per-compilation version*, which is what this PR does.

As I see it, this PR is backwards compatible, it will behave exactly as before for `go run .` development style

michaeldwan (approved):

PR #2893, opened 2023-10-10
Flypkgs builds

michaeldwan:
This PR adds support for sending dev builds to flyio-pkgs. I'm trying to be careful to bring this over without impacting the production code or existing release process, but if there's any concerns let me know and I'll take care of it.

The diff is big, but it's mostly additions to config files and new binaries for build time. I've broken the PR down into commits:

- [Allow +build.meta suffixes in version numbers](https://github.com/superfly/flyctl/commit/3dee3b6112e46c970997849f01f9802384cee059): this lets us add `+whatever.we.want` to the new version number code for things like preflight snapshot builds. Includes tests.
- [distribute and version tools used for releases in CI](https://github.com/superfly/flyctl/commit/ba4718121fbe8ab570a7ae135292a518b51282ca): new binaries for build time. `tools/version` is used to calculate the next version number and to generate a JSON dump of the current release/repo state for build scripts. `tools/distribute` sends build artifacts (and that JSON file) to the flyio-pkgs backend.
- [remove old dev builds from CI](https://github.com/superfly/flyctl/commit/7b602cc7dc2809a325371844f1a60d8ed8c954f9): at one time we had support for dev builds, but it's not been used in a long time. I'm removing the workflow and config bits to keep things tidy and remove possible confusion.
- [new goreleaser config to isolate changes from the existing build](https://github.com/superfly/flyctl/commit/09704431e761537b7a963550da8d10a6d611b7e5): config for building new binaries. this is mostly copying what we have now but excludes things I don't want to touch yet like GH releases. This will go away later.
- [Fix next version command on stable](https://github.com/superfly/flyctl/commit/a5dbe79f8e7f04000fd25b7e7b4f5111799a691b): `./tools/version next` now returns the right "stable" build. This will let us replace the local `bump_version.sh` script soon without forcing a change in version number formats.
- [Comment out tests for now](https://github.com/superfly/flyctl/commit/c523bb6201cfde5c169144bdd86861f7b3b680fc): I was trying to write some tests for the version numer/repo state tool, but fabricating a realistic git repo to point at was more trouble than I had time for right now. will revisit later.

The rest of the commits are mucking around with GH action configs and fixing some logic issues that crept up after running this against a real PR on GHA.

Once this is merged, we'll be getting a steady stream of dev builds from PR activity which is enough data to start wiring up and testing the installation scripts and upgrade process in flyctl. GHA workflows will get cleaned up after that.


.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-go@v4
+        with:
+          go-version-file: "go.mod"
+      - shell: bash
+        run: |
+          echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+      # copy the cashes from prepare
```
billyb2, line 63: small typo, "cashes"

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+      - uses: actions/cache@v3
+        with:
+          path: dist/darwin
+          key: darwin-${{ env.sha_short }}
+      - uses: actions/cache@v3
+        with:
+          path: dist/windows
+          key: windows-${{ env.sha_short }}
+      - name: Write release meta
+        run: go run ./tools/version show > ./dist/release.json
```
billyb2, line 80: mostly just double checking, but `go run ./tools/version show > ./dist/release.json` only runs *after* a release, right?

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+        with:
+          name: build-artifacts
+          path: dist/*
+          retention-days: 1
+      - name: Upload to flypkgs
+        env:
+          FLYPKGS_API_TOKEN: ${{ secrets.FLYPKGS_API_TOKEN }}
+        run: go run ./tools/distribute upload ./dist
+
+  # preflight here
```
billyb2, line 92: just leaving a comment so you don't forget about this :)

.github/workflows/release.yml
michaeldwan: good catch!

.github/workflows/release.yml
michaeldwan: The previous job builds artifacts in parallel across a few runners. This job then downloads them into a directory, and from there the version tool generates a json file of the repo state. _That_ dir then gets sent over to flypkgs and uploaded as a GHA build artifact in the next few steps.

.github/workflows/release.yml
michaeldwan: Oh I can't implement that yet! This PR just gets builds working alongside the existing ci/release process. I'm going to consolidate them all once this is live, at which point we can (hopefully) safely disrupt the existing workflows :D

.goreleaser.2.yml
```diff
@@ -0,0 +1,84 @@
...
+      - -X github.com/superfly/flyctl/internal/buildinfo.commit={{.ShortCommit}}
+    tags:
+      - "{{.Env.BUILD_ENV}}"
+
+archives:
+  - id: linux
+    name_template: >-
+      {{ .ProjectName }}_{{ .Version }}_Linux_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}
+    builds: [linux]
+    files: [only-the-binary*]
```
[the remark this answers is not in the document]
michaeldwan: I think that came from one of their samples years ago. IIRC it forces an archive with no additional files because only-the-binary* will never match anything. Here's the latest docs that use `none*` instead. Same same https://goreleaser.com/customization/archive/#a-note-about-gzip

billyb2 (approved):
banger pr

.goreleaser.2.yml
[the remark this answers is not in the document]
billyb2: Cool cool

.github/workflows/release.yml
billyb2: ah i gotcha :)

tools/distribute/flypkgs/api.go
```diff
@@ -0,0 +1,85 @@
...
+			return errors.Wrap(err, "decoding error response")
+		}
+
+		return errRes
+	}
+
+	fullResponse := successResponse{
+		Data: v,
+	}
+	if err = json.NewDecoder(res.Body).Decode(&fullResponse); err != nil {
```
rugwirobaker, line 80: This is throwing me off a bit. Decoding into it after we just set `fullResponse.Data` to `v`. I thought we'd decide into v instead.

michaeldwan: The API returns a payload like `{"data": {...}}`, so we need to wrap the target struct in order to unmarshal the value of `"data"`  into the target.

PR #2903, opened 2023-10-13
Add a check to make sure phoenix compiles before continuing

jeregrine:
### Change Summary

What and Why: When running the phoenix launcher the scanner will attempt to build the project and compile it, if this fails it just continues silently.  

How:

Related to: While watching MD work through deploying livebeats I noticed that while his Elixir project failed to compile locally the scanner just kept trying to deploy and the error messages got lost in scrollback. I add a check here to die early if we can't even compile the elixir project. I think its a fair to assume the user has code working locally before they can launch. 

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


michaeldwan (approved):
Nice fix. That first early return if there's an error would have solved my problem yesterday. Well don.

PR #2905, opened 2023-10-13
Deploy: `rolling_one` support

alichay:
### Change Summary

What and Why: Adds support for `rolling_one` strategy from Nomad. The advantage here over setting the max-unavailable flag is that this can live in fly.toml, because having to type in special flags when deploying to avoid breaking a fragile system can be a bit of a footgun.

(the real motivation is automated migrations, though. we can't make people start deploying with flags, so this makes it possible to migrate them without breaking anything)

How: in the start of DeployMachinesApp, remaps "rolling_one" to "rolling" with a max unavail of 1

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -51,7 +51,7 @@ type Config struct {
 	// Sections that are typically short and benefit from being on top
 	Experimental *Experimental     `toml:"experimental,omitempty" json:"experimental,omitempty"`
 	Build        *Build            `toml:"build,omitempty" json:"build,omitempty"`
-	Deploy       *Deploy           `toml:"deploy, omitempty" json:"deploy,omitempty"`
+	Deploy       *Deploy           `toml:"deploy,omitempty" json:"deploy,omitempty"`
 	Env          map[string]string `toml:"env,omitempty" json:"env,omitempty"`
 
 	// Fields that are process group aware must come after Processes
@@ -88,8 +88,9 @@ type Config struct {
 }
 
 type Deploy struct {
-	ReleaseCommand string `toml:"release_command,omitempty" json:"release_command,omitempty"`
-	Strategy       string `toml:"strategy,omitempty" json:"strategy,omitempty"`
+	ReleaseCommand string  `toml:"release_command,omitempty" json:"release_command,omitempty"`
+	Strategy       string  `toml:"strategy,omitempty" json:"strategy,omitempty"`
+	MaxUnavailable float64 `toml:"max_unavailable,omitempty" json:"max_unavailable,omitempty"`
 }
 
 type File struct {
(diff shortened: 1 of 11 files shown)


dangra (approved):
SGTM but I'd prefer to expose max_unavailable under `[deploy]` section.  We can do that later anyways.

alichay:
> SGTM but I'd prefer to expose max_unavailable under `[deploy]` section. We can do that later anyways.

ahhh yeah that's a good call. I think I'll just rework this to do that, and skip the intermediate version of the config

internal/appconfig/config.go
```diff
@@ -88,8 +88,9 @@ type Config struct {
 }
 
 type Deploy struct {
-	ReleaseCommand string `toml:"release_command,omitempty" json:"release_command,omitempty"`
-	Strategy       string `toml:"strategy,omitempty" json:"strategy,omitempty"`
+	ReleaseCommand string  `toml:"release_command,omitempty" json:"release_command,omitempty"`
+	Strategy       string  `toml:"strategy,omitempty" json:"strategy,omitempty"`
+	MaxUnavailable float64 `toml:"max_unavailable,omitempty" json:"max_unavailable,omitempty"`
```
dangra, line 93: don't you prefer to use `*float64` here so we can check way down if it is set or not reliably. In a similar way the `--max-unavailable` cli flag should take precedence over the config setting if set. Using `*float64` serves there too.

internal/appconfig/validation.go
```diff
@@ -158,6 +158,11 @@ func (cfg *Config) validateDeploySection() (extraInfo string, err error) {
 			extraInfo += "error canary deployment strategy is not supported when using mounted volumes"
 			err = ValidationError
 		}
+
+		if s != "rolling" && cfg.Deploy.MaxUnavailable != 0.0 {
```
dangra, line 162: I wouldn't make it so strict, just ignore the field if not rolling or show a warning. But a validation error is too much. My personal take though.

internal/appconfig/config.go
dangra: Ignore all that, my only real concern is that --max-unavailable doesn't take precedence over appConfig.Deploy.MaxUnavailable.

internal/appconfig/validation.go
alichay: initially I couldn't see a way this could happen that wasn't an accident, but looking at it again there are places where it'd be helpful to have that flexibility.

internal/appconfig/config.go
alichay: I wrote this using *float64 initially, and halfway through changed it to just float64. I can't remember for the life of me why I changed it though!

fixed :)

dangra (approved):

PR #2926, opened 2023-10-19
Discharge 3p caveats in user's macaroons

btoews:
### Change Summary

What and Why:

Implements the client side of the protocol described in https://github.com/superfly/macaroon/pull/9. WIP

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


api/auth_test.go
```diff
@@ -1,22 +0,0 @@
-package api
```
kzys, line 1: Prefer to keep this test or something like that if possible.

internal/cmdutil/preparers/preparers.go
```diff
@@ -51,6 +51,32 @@ func LoadConfig(ctx context.Context) (context.Context, error) {
 	return config.NewContext(ctx, cfg), nil
 }
 
+// prune any invalid/expired macaroons and fetch needed third party discharges
+func UpdateMacaroons(ctx context.Context) (context.Context, error) {
```
kzys, line 55: Doc comments should start from the symbol name.

internal/tokens/tokens.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (t *Tokens) Docker() string {
+	return t.normalized(false)
+}
+
+func (t *Tokens) NATS() string {
+	return t.normalized(false)
+}
+
+func (t *Tokens) API() string {
```
kzys, line 52: - Is API means `web`? Flaps is also having API...
- Does it make sense to have `WithUserToken` instead? It is a bit odd to see the tokens package know all use cases inside flyctl.

kzys (approved):
LGTM! I only have non-blocking suggestions/questions.

internal/tokens/tokens.go
btoews: > Does API here means web? Flaps is also having API...

Good point. I can come up with a better name.

> Does it make sense to have WithUserToken instead? It is a bit odd to see the tokens package knows all use cases inside flyctl.

That's what I did originally. I felt like it would be easier for future developers to use the flaps token for flaps, the docker token for docker, etc... rather than asking them to think about whether they want the user token or not. I'm fine either way though if you think `WithUserToken` is a nicer API.

api/auth_test.go
btoews: It just moved to `internal/tokens/tokens_test.go` when the helper methods moved.

PR #2936, opened 2023-10-20
Feat/conc deploy

dangra:
### Change Summary

Current rolling deploy strategy uses a batching mechanism that pause until the whole batch is done before continuing with the next one. This approach was an huge improvement over the original and very slow sequential and one-at-a-time update before it, but it still feels slow for apps with 10+ machines. Also current batch base implementation doesn't consider process groups and it could take a whole group down if it fits in a batch.

This PR implements true concurrent updates respecting the max unavailable setting at any time and doing it per process group.

Related to: https://github.com/superfly/flyctl/pull/2679

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,6 +52,7 @@ require (
 	github.com/pkg/sftp v1.13.6
 	github.com/samber/lo v1.38.1
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966
+	github.com/sourcegraph/conc v0.3.0
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.17.0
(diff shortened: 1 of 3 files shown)


alichay (approved):
I didn't pull this down and run it, but it looks good (and very elegant!)

I'm assuming you checked that maxUnavailable still works like it did before? if so, ⛵

dangra:
> I didn't pull this down and run it, but it looks good (and very elegant!)
> 
> I'm assuming you checked that maxUnavailable still works like it did before? if so, 🚢 it

Thanks and Yes I did! 

There are two things that still bother me:
* No more "canary" (aka "SoloFirst")
* Still trying to figure if we want to stop the deploy after the first failure
* Only do it if updates machines are less than certain percentage of total?
* Rollback?

I mean, most of this stuff can be improved in following PRs but wanted to write them down.

PR #2937, opened 2023-10-20
Client-side JSON filtering of Machines (probably not!)

tqbf:
### Change Summary

What and Why:

Clientside filtering of machines.

How:

`-f [filter]` takes a JMESPath filter; JMESPath is like standardized `jq`. We already had it in the tree.

JMESPath is pretty inscrutable, and I'm not sure if this will ever get used. 

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


dangra:
I can't recommend enough to install pre-commit hooks:

```
brew install pre-commit 
# In flyctl dir:
pre-commit install
```

PR #2947, opened 2023-10-24
add a --shell argument to `flyctl m run`

tqbf:
### Change Summary

What and Why:

There's `flyctl console`, but it requires an existing app, which is more than you sometimes want to do to just pop a shell inside an org somewhere. There's `fly m run`, but it doesn't immediately launch a shell. Two great tastes.

How:

Add "shell", "user", and "command". If you don't specify an app name, we'll create a random one without prompting. 

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -221,11 +221,6 @@ func newRun() *cobra.Command {
 			Description: "Enable LSVD for this machine",
 			Hidden:      true,
 		},
-		flag.Bool{
-			Name:        "it",
-			Description: "Open a shell on the machine once created",
-			Hidden:      false,
-		},
 		flag.String{
 			Name:        "user",
 			Description: "Username, if we're shelling into the machine now.",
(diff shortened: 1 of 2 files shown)


internal/command/machine/run.go
```diff
@@ -186,25 +220,73 @@ func newRun() *cobra.Command {
 			Description: "Enable LSVD for this machine",
 			Hidden:      true,
 		},
+		flag.Bool{
+			Name:        "it",
```
dangra, line 224: What about naming it `--interactive` or just removing it in favor `--shell`? they are different but most people just want what --shell provides.

michaeldwan: `-it` is a confusing flag. in docker it's just 2 flags sharing a dash. `interactive` and `tty`

tqbf: Killed this arg. It's just `--shell` now.

PR #2955, opened 2023-10-25
Launch V2: validate app names, add errors that can be recovered by going into web UI

alichay:
### Change Summary

What and Why:
 * Sanitizes initial app names, lowercasing ascii characters, replacing special characters with dashes (collapsed dashes, so you don't get multiple at a time), and removing non-ascii characters
 * Validates app names passed via `--name`
 * Adds `recoverableInUiError`, a wrapper type for an error that indicates that the error can be corrected in the Web UI, so that naming validation failures funnel users towards correcting the problem instead of yelling at them
 * Add recovery conditions for many failure modes of Launch, such as invalid app name, org, guest, or region.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/launch/cmd.go
+++ b/internal/command/launch/cmd.go
@@ -149,9 +149,10 @@ func run(ctx context.Context) (err error) {
 
 		launchManifest, cache, err = buildManifest(ctx, canEnterUi)
 		if err != nil {
-			if errors.Is(err, recoverableInUiError{}) && canEnterUi {
+			var recoverableErr recoverableInUiError
+			if errors.As(err, &recoverableErr) && canEnterUi {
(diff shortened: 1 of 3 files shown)


dangra (approved):

internal/command/launch/cmd.go
```diff
@@ -193,6 +208,9 @@ func run(ctx context.Context) (err error) {
 			return err
 		}
 	}
+	if requireUi {
```
dangra, line 211: not completely following the logic but shouldn't it check err ?

alichay: yup, that's a bug!

also, this
> not completely following the logic

is probably a sign that it needed a tad of cleanup :)

PR #2956, opened 2023-10-25
Make Launch v2 the default 🎉

alichay:
## Change Summary

Replaces the `--ui` flag with a deprecated `--no-ui` flag. `fly launch` now defaults to Launch V2.

### Note: this removes `--reuse-app`! Merging this will remove the launch-into behavior entirely.

Personally, I feel like this is a good thing. Having `launch` work on existing apps is confusing, and we're actively working on better ways to work with existing apps that could replace this functionality while maintaining the separation between deploy and launch.

---

### Documentation

- [X] Fresh Produce (when it's time)
- [ ] n/a

### DIFF
--- a/internal/command/launch/cmd.go
+++ b/internal/command/launch/cmd.go
@@ -188,21 +188,23 @@ func run(ctx context.Context) (err error) {
 		summary,
 	)
 
-	confirm := false
-	prompt := &survey.Confirm{
-		Message: lo.Ternary(
-			incompleteLaunchManifest,
-			"Would you like to continue in the web UI?",
-			"Do you want to tweak these settings before proceeding?",
-		),
-	}
-	err = survey.AskOne(prompt, &confirm)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/command/launch/cmd.go
```diff
@@ -120,7 +121,9 @@ func getManifestArgument(ctx context.Context) (*LaunchManifest, error) {
 
 func run(ctx context.Context) (err error) {
 
-	if !flag.GetBool(ctx, "ui") {
+	// NOTE: We depend on legacy launcher behavior for Nomad support, which is needed for the MigrateToV2 tests
+	//       Once we rip out those tests, this can go with them.
+	if flag.GetBool(ctx, "no-ui") || flag.GetBool(ctx, "force-nomad") {
```
dangra, line 126: Can it resort to legacy "fly launch" if `--reuse-app` is passed and print a big deprecation warning if used?

dangra (approved):

PR #2957, opened 2023-10-26
Revert "simple "create" command for create-not-start"

tqbf:
Reverts superfly/flyctl#2935


tqbf:
This wasn't baked yet.

PR #2973, opened 2023-11-01
Use org's raw slug when streaming logs from NATS

btoews:
### Change Summary

What and Why:

Flaps can't resolve `personal` slug without a GraphQL query. When using macaroon auth, we try to avoid GraphQL queries in flaps.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


dangra (approved):
TIL about rawSlug

PR #2974, opened 2023-11-01
Block metrics on CI

billyb2:
The alternative is that metrics are never sent on CI, which is bad of course. There's still a timeout of at most 15 seconds, so worst case this won't block deploys for a considerable length of time.

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/agent/cmd_unix.go
+++ b/agent/cmd_unix.go
@@ -8,7 +8,7 @@ import (
 	"syscall"
 )
 
-func setSysProcAttributes(cmd *exec.Cmd) {
+func SetSysProcAttributes(cmd *exec.Cmd) {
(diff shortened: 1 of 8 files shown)


billyb2:
Not actually ready for review, I just need a build for testing

agent/cmd_unix.go
```diff
@@ -8,7 +8,7 @@ import (
 	"syscall"
 )
 
-func setSysProcAttributes(cmd *exec.Cmd) {
+func SetSysProcAttributes(cmd *exec.Cmd) {
```
rugwirobaker, line 11: I feel like these should go into a new package in internal `sys` or something

PR #2977, opened 2023-11-01
Tigris object storage provisioning

jsierles:
This PR adds initial provisioning commands for Tigris object storage buckets.


internal/command/extensions/tigris/dashboard.go
```diff
@@ -0,0 +1,42 @@
...
+		flag.App(),
+		flag.AppConfig(),
+	)
+	cmd.Args = cobra.MaximumNArgs(1)
+	return cmd
+}
+
+func runDashboard(ctx context.Context) (err error) {
+
+	extension, _, err := extensions_core.Discover(ctx, gql.AddOnTypeSupabase)
```
benwaffle, line 34: supabase?

PR #2980, opened 2023-11-02
appconfig: add 'http_options.h2_backend' port option

pborzenkov:
The option tells fly-proxy that it's Ok to forward HTTP/2 requests (h2c)
to an app and not downgrade them to HTTP/1.

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -520,8 +520,9 @@ type TLSOptions struct {
 }
 
 type HTTPOptions struct {
-	Compress *bool                `json:"compress,omitempty" toml:"compress,omitempty"`
-	Response *HTTPResponseOptions `json:"response,omitempty" toml:"response,omitempty"`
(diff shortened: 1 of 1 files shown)


internal/command/extensions/supabase/dashboard.go
```diff
@@ -43,7 +43,9 @@ func runDashboard(ctx context.Context) (err error) {
 		}
 
 		err = extensions_core.OpenDashboard(ctx, extension.Name)
-
+		if err != nil {
+			return err
+		}
```
dangra, line 48: no need since https://github.com/superfly/flyctl/pull/2998

dangra (approved):

internal/command/extensions/supabase/dashboard.go
pborzenkov: Aah, missed that PR. I'll drop the commit.

PR #2982, opened 2023-11-02
Revoke user's token when logging out

btoews:
### Change Summary

What and Why:

It seems wrong to me to not invalidate the user's token when they ask to log out. This PR calls the `logOut` mutation when the user runs `fly auth logout`. 

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


michaeldwan:
This is great. I think a token should be revoked when explicitly signing out considering signing in creates a token.

PR #2986, opened 2023-11-03
Trace flyctl deploy path

gwuah:
### Change Summary
We know things fail, but it's not **easily** clear to us why & where. This PR instruments the fly deploy path with traces to allow us narrow down issues better.

The general idea is to start from flyctl and work our way through all platform services that power "fly deploy". end2end tracing is the goal.

Todo
- [x] instrument all flaps & graphql calls
- [x] export to right collector (moved from stdout to dev/prod collector)
- [x] configure remote enable/disable of telemetry (this can be done from the remote collector)
- [x] test a few flows to ensure nothing is broken (works fine)
- [ ] instrument entire path

Notes
- [ ] filter out `machine_refresh_lease` spans. (keep only 5 of those spans)
- [x] scale up collector machines (currently they're using shared cpu)

### DIFF
--- a/flaps/flaps.go
+++ b/flaps/flaps.go
@@ -82,7 +82,9 @@ func NewWithOptions(ctx context.Context, opts NewClientOpts) (*Client, error) {
 	if opts.Logger != nil {
 		logger = opts.Logger
 	}
-	httpClient, err := api.NewHTTPClient(logger, httptracing.NewTransport(http.DefaultTransport))
+
+	transport := tracing.NewTransport(httptracing.NewTransport(http.DefaultTransport))
+	httpClient, err := api.NewHTTPClient(logger, transport)
 	if err != nil {
 		return nil, fmt.Errorf("flaps: can't setup HTTP client to %s: %w", flapsUrl.String(), err)
 	}
@@ -137,7 +139,9 @@ func newWithUsermodeWireguard(ctx context.Context, params wireguardConnectionPar
 		},
 	}
 
-	httpClient, err := api.NewHTTPClient(logger, httptracing.NewTransport(transport))
+	instrumentedTransport := tracing.NewTransport(httptracing.NewTransport(transport))
+
+	httpClient, err := api.NewHTTPClient(logger, instrumentedTransport)
 	if err != nil {
 		return nil, fmt.Errorf("flaps: can't setup HTTP client for %s: %w", params.orgSlug, err)
 	}
--- a/internal/cmdutil/preparers/preparers.go
+++ b/internal/cmdutil/preparers/preparers.go
@@ -20,6 +20,7 @@ import (
 	"github.com/superfly/flyctl/internal/instrument"
 	"github.com/superfly/flyctl/internal/logger"
 	"github.com/superfly/flyctl/internal/state"
+	"github.com/superfly/flyctl/internal/tracing"
 )
 
 // Preparers are split between here and `command/command.go` because
@@ -60,7 +61,7 @@ func InitClient(ctx context.Context) (context.Context, error) {
 	api.SetBaseURL(cfg.APIBaseURL)
 	api.SetErrorLog(cfg.LogGQLErrors)
 	api.SetInstrumenter(instrument.ApiAdapter)
-	api.SetTransport(httptracing.NewTransport(http.DefaultTransport))
+	api.SetTransport(tracing.NewTransport(httptracing.NewTransport(http.DefaultTransport)))
 
 	c := client.FromTokens(cfg.Tokens)
 	logger.Debug("client initialized.")
--- a/internal/tracing/transport.go
+++ b/internal/tracing/transport.go
@@ -1,0 +1,32 @@
+package tracing
+
+import (
+	"net/http"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+func NewTransport(inner http.RoundTripper) http.RoundTripper {
+	return &InstrumentedTransport{
+		inner: inner,
+	}
+}
+
+type InstrumentedTransport struct {
+	inner http.RoundTripper
+}
+
+func (t *InstrumentedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	gp := otel.GetTextMapPropagator()
+	req = req.Clone(req.Context())
+
+	gp.Inject(req.Context(), propagation.HeaderCarrier(req.Header))
+
+	resp, err := t.inner.RoundTrip(req)
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, err
+}


internal/cli/cli.go
```diff
@@ -58,6 +59,14 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 	httptracing.Init()
 	defer httptracing.Finish()
 
+	tp, err := tracing.InitTraceProvider(ctx)
+	if err != nil {
+		fmt.Fprintf(io.ErrOut, "failed to initialize tracing library: =%v", err)
+		return 127
+	}
+
+	defer tp.Shutdown(context.Background())
```
michaeldwan, line 68: Any chance this could block shutdown if the backend is slow/down?

api/client.go
```diff
@@ -14,6 +14,10 @@ import (
 
 	genq "github.com/Khan/genqlient/graphql"
 	"github.com/superfly/graphql"
+	"go.opentelemetry.io/otel"
```
michaeldwan, line 17: There's a lint error on this:

```
import of package `go.opentelemetry.io/otel` is blocked because the module is not in the allowed modules list. (gomodguard)
```

Any idea what that's about?

flaps/flaps_machines.go
```diff
@@ -263,13 +226,12 @@ func (f *Client) ListActive(ctx context.Context) ([]*api.Machine, error) {
 // Destroyed machines and console machines are excluded.
 // Unlike other List functions, this function retries multiple times.
 func (f *Client) ListFlyAppsMachines(ctx context.Context) ([]*api.Machine, *api.Machine, error) {
-	var allMachines []*api.Machine
+	allMachines := make([]*api.Machine, 0)
```
michaeldwan, line 229: The `var allMachines []*api.Machine` form is generally preferred if we're making a slice of an unknown size.

flaps/flaps_machines.go
```diff
@@ -17,64 +17,9 @@ import (
...
-			sendFlapsCallMetric(ctx, actionInfo.action, statusCode, time)
-		}
-	}()
-
-	if err != nil {
-		return err
-	}
-
-	return err
+func (f *Client) sendRequestMachines(ctx context.Context, action flapsAction, method, endpoint string, in, out interface{}, headers map[string][]string) error {
```
michaeldwan, line 20: The number of args is getting hard to track. Maybe you could cut down on one by doing what the stdlib http client does. Move the http method to the function name, like `sendMachinesRequestPost`, `sendMachinesRequestGet`, and those methods in turn pass the http method into `_sendRequest`.

go.mod
```diff
@@ -38,7 +38,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20
 	github.com/mattn/go-zglob v0.0.4
 	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d
-	github.com/miekg/dns v1.1.57
+	github.com/miekg/dns v1.1.56
```
michaeldwan, line 41: Why did this move to an older version?

go.mod
```diff
@@ -63,7 +63,9 @@ require (
 	github.com/superfly/macaroon v0.1.0
 	github.com/superfly/tokenizer v0.0.2
 	github.com/vektah/gqlparser v1.3.1
-	github.com/vektah/gqlparser/v2 v2.5.10
+	github.com/vektah/gqlparser/v2 v2.5.6
```
michaeldwan, line 66: Looks like this moved to an older version as well.

internal/build/imgsrc/docker.go
```diff
@@ -234,14 +254,30 @@ func newRemoteDockerClient(ctx context.Context, apiClient *api.Client, appName s
...
+		return nil, err
 	}
 
 	builderHostOverride, ok := os.LookupEnv("FLY_RCHAB_OVERRIDE_HOST")
 	if ok {
 		oldHost := host
 		host = builderHostOverride
+
+		span.SetAttributes(
+			attribute.String("builder.old_host", oldHost),
```
michaeldwan, line 276: I think "old" is just the name of the previous value, not actually "old". Do we want to burn "old" into the attribute name?

internal/build/imgsrc/dockerfile_builder.go
```diff
@@ -414,7 +415,15 @@ func runBuildKitBuild(ctx context.Context, docker *dockerclient.Client, opts Ima
 	return res.ExporterResponse[exptypes.ExporterImageDigestKey], nil
 }
 
-func pushToFly(ctx context.Context, docker *dockerclient.Client, streams *iostreams.IOStreams, tag string) error {
+func pushToFly(ctx context.Context, docker *dockerclient.Client, streams *iostreams.IOStreams, tag string) (err error) {
+	ctx, span := tracing.GetTracer().Start(ctx, "push_image_to_fly")
```
michaeldwan, line 419: Could this just be "push_image"? It defaults to registry.fly.io but technically it could be anything.

internal/build/imgsrc/resolver.go
```diff
@@ -46,6 +49,52 @@ type ImageOptions struct {
...
+	}
+
+	b, err = json.Marshal(io.BuildSecrets)
+	if err == nil {
+		attrs = append(attrs, attribute.String("imageoptions.buildSecrets", string(b)))
+	}
+
+	b, err = json.Marshal(io.BuiltInSettings)
+	if err == nil {
+		attrs = append(attrs, attribute.String("imageoptions.builtInSettings", string(b)))
```
michaeldwan, line 86: Some of the attributes in this diff are snake case and some camel case. What's the preferred form? Do they need to be consistent?

internal/build/imgsrc/resolver.go
```diff
@@ -108,10 +188,14 @@ func (r *Resolver) ResolveReference(ctx context.Context, streams *iostreams.IOSt
 		}
 		bld.BuildAndPushFinish()
 		bld.FinishImageStrategy(s, true /* failed */, nil, note)
+		span.AddEvent(fmt.Sprintf("failed to resolve image with strategy %s", s.Name()))
+
 	}
 
 	r.finishBuild(ctx, bld, true /* failed */, "no strategies resulted in an image", nil)
-	return nil, fmt.Errorf("could not find image \"%s\"", opts.ImageRef)
+	err = fmt.Errorf("could not find image \"%s\"", opts.ImageRef)
```
michaeldwan, line 196: Not a big deal but I recently discovered that the `%q` verb prints a quoted string. Much easier to read/type than escaped quotes :D

internal/cmdutil/preparers/preparers.go
```diff
@@ -148,3 +149,19 @@ func ApplyAliases(ctx context.Context) (context.Context, error) {
...
+// except it only came to my mind after writing this so let's stick with this for now.
+func SetOtelAuthenticationKey(ctx context.Context) (context.Context, error) {
+	token := config.Tokens(ctx).Flaps()
+	if token == "" {
+		token = os.Getenv("FLY_API_TOKEN")
+	}
+
+	os.Setenv("FLY_OTEL_AUTH_KEY", token)
+	return ctx, nil
+}
```
michaeldwan, line 167: Is there no way to directly set the token when we initialize the collector?

internal/command/deploy/deploy.go
```diff
@@ -205,11 +220,14 @@ func run(ctx context.Context) error {
...
 
 func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, forceYes bool, optionalGuest *api.MachineGuest) (err error) {
+	span := trace.SpanFromContext(ctx)
+
 	io := iostreams.FromContext(ctx)
 	appName := appconfig.NameFromContext(ctx)
 	apiClient := client.FromContext(ctx).API()
 	appCompact, err := apiClient.GetAppCompact(ctx, appName)
 	if err != nil {
+		tracing.RecordError(span, err, "get app compact")
```
michaeldwan, line 230: Do we need to emit an error here since the api client is also emitting errors?

flaps/actions.go
```diff
@@ -0,0 +1,91 @@
+package flaps
+
+type flapsAction int
+
+const (
+	appCreate flapsAction = iota
```
michaeldwan, line 6: These seem like they'll be hard to maintain. Would something like the stringer generator work so code gen handles the grunt work? https://pkg.go.dev/golang.org/x/tools/cmd/stringer

api/resource_domains.go
```diff
@@ -85,7 +85,7 @@ func (c *Client) CreateDomain(organizationID string, name string) (*Domain, erro
 	`
 
 	req := c.NewRequest(query)
-
+	req.Var("action", "create_domain")
```
michaeldwan, line 88: These are being passed to graphql as a var. Do they get used on the backend?

michaeldwan (approved):
Nice work. I left a bunch of comments with questions and small suggestions on implementation. Take it or leave it. But this is a huge PR that touches a lot, so I'd like a second opinion from some other folks with deep flyctl knowledge. And lots of time testing this before we go live.

api/client.go
dangra: This is `gomodguard` linter enforcing the dependency tree is kept to a minimal because then it is imported from flaps/flyd and things go bad. That said, adding `otel` is a safe bet here.

https://github.com/superfly/flyctl/blob/4a391eb3b88d1432520670b848154c4f41513fc7/api/.golangci.yml#L46-L51

internal/build/imgsrc/resolver.go
```diff
@@ -46,6 +49,52 @@ type ImageOptions struct {
 	Label           map[string]string
 }
 
+func (io ImageOptions) ToSpanAttributes() []attribute.KeyValue {
+	attrs := []attribute.KeyValue{
+		attribute.String("imageoptions.app_name", io.AppName),
+		attribute.String("imageoptions.work_dir", io.WorkingDir),
+		attribute.String("imageoptions.dockerfile_path", io.DockerfilePath),
+		attribute.String("imageoptions.ignorefile_path", io.IgnorefilePath),
+		// todo(gwuah): find out from security if this is fine
```
billyb2, line 58: don't forget about this :)

flaps/flaps.go
```diff
@@ -163,23 +162,32 @@ func (f *Client) CreateApp(ctx context.Context, name string, org string) (err er
...
+	err = f._sendRequest(ctx, appCreate, http.MethodPost, "/apps", in, nil, nil)
 	return
 }
 
-func (f *Client) _sendRequest(ctx context.Context, method, endpoint string, in, out interface{}, headers map[string][]string) (int, error) {
+func (f *Client) _sendRequest(ctx context.Context, action flapsAction, method, endpoint string, in, out interface{}, headers map[string][]string) error {
+	ctx, span := tracing.GetTracer().Start(ctx, fmt.Sprintf("flaps.%s", action.String()), trace.WithAttributes(
+		attribute.String("request.action", action.String()),
+		attribute.String("request.endpoint", endpoint),
+		attribute.String("request.method", method),
```
billyb2, line 173: some requests have specific headers that are pretty useful context for what the request is doing. Should we include that?

internal/appconfig/context.go
billyb2, line 1: is this a gofmt thing?

internal/build/imgsrc/local_image_resolver.go
```diff
@@ -59,21 +69,27 @@ func (*localImageResolver) Run(ctx context.Context, dockerFactory *dockerClientF
...
 		build.BuildFinish()
+		span.AddEvent("no image found and no error occurred")
 		return nil, "no image found and no error occurred", nil
 	}
 
 	build.BuildFinish()
 	fmt.Fprintf(streams.ErrOut, "image found: %s\n", img.ID)
 
+	// todo(@gwuah)
+	span.SetAttributes(attribute.String("image.id", img.ID))
```
billyb2, line 85: also dont forget this :)

billyb2 (approved):
banger pr

this is a very large pr, so take my ✅ with a grain of salt :). Aside from some todo's i noticed, the PR looks great. Well done.

PR #2992, opened 2023-11-06
fly deploy: skip launch for stopped autoscaled machines

wjordan:
#### What and Why:

Skips the launch behavior for stopped autoscaled machines during `fly deploy`. This provides better default behavior for autoscaling apps that want to leave a number of stopped instances in place for future scaling up, and that don't wish to scale up to their max instance count after every deploy.

We don't want to always skip launching stopped machines, because we still want a `fly deploy` to start up any crashed stopped machine that's not part of an autoscaled app.

#### How:

Sets `SkipLaunch` to `true` on the deploy/restart `LaunchInput` for machines that are currently stopped, and that have autoscaling enabled on one of their services.

Updates the `waitForMachine` logic to skip waiting for any machines with `SkipLaunch` set which already includes standby machines (but now may include other stopped machines as well).

#### Related to:

https://community.fly.io/t/deploying-a-multi-process-app-and-i-dont-want-any-machines-running-in-one-of-the-groups-after-a-deploy/16257

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -314,8 +314,8 @@ func suggestChangeWaitTimeout(err error, flagName string) error {
 
 func (md *machineDeployment) waitForMachine(ctx context.Context, e *machineUpdateEntry) error {
 	lm := e.leasableMachine
-	// Don't wait for Standby machines, they are updated but not started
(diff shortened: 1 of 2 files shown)


dangra (approved):
💯

PR #2993, opened 2023-11-06
Add work_mem and maintenance_work_mem to Postgres config

jvns:
### Change Summary

What and Why:

I wanted to set `maintenance_work_mem` in my Postgres cluster because it was set to 64MB, which is more than 25% of my instance's memory, and that's way too much. I thought `work_mem` would be a useful setting to have available too since it seems like a pretty core Postgres tuning setting.

I don't know what fly's philosophy about adding new Postgres settings is here but thought I'd make a PR to see if you're open to adding new settings.

How:

I just grepped for `shared-preload-libraries` and added `maintenance-work-mem` wherever I saw it.

### Documentation

Not sure where this is documented other than in `--help`. 

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/postgres/config.go
+++ b/internal/command/postgres/config.go
@@ -13,6 +13,8 @@ var pgSettings = map[string]string{
 	"max-wal-senders":            "max_wal_senders",
 	"max-replication-slots":      "max_replication_slots",
 	"max-connections":            "max_connections",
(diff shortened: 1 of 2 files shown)


lubien (approved):

dangra:
thanks @jvns !

PR #3001, opened 2023-11-07
fly status: optimize GetLatestImageDetails API calls

wjordan:
This is a narrow optimization to skip the `GetLatestImageDetails` API call for image refs that have already returned 'unknown repository' (reducing this to a single API call per unique image currently used by the app), or that have the default deploy-label prefix (`deployment-`).

There's probably a more thorough way to avoid these calls we can come up with later, but this was a simple, local way to handle most of the common use-cases as a quick improvement.

On an app with ~40 machine instances, this lowers the `fly status` duration from ~6 seconds to ~1 second by skipping many unnecessary API-call round-trips.

### DIFF
--- a/internal/command/status/machines.go
+++ b/internal/command/status/machines.go
@@ -107,12 +107,19 @@ func RenderMachineStatus(ctx context.Context, app *api.AppCompact, out io.Writer
 
 	var updatable []*api.Machine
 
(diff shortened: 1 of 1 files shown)


billyb2 (approved):
banger pr

PR #3009, opened 2023-11-08
`redis`: automatically determine plan

alichay:
### Change Summary

What and Why: Removes plan selection from Upstash Redis - automatically determined based on whether the org already has an existing Redis instance
  - It does? pay-as-you-go plan
  - It doesn't? free plan

Related to: new Upstash pricing model

---

@jsierles I know you said you'd like to simplify things by throwing some of the logic in the backend, but does this stopgap solution work for you? I'm hoping to get Launch v2 deployed, and this is the last remaining piece of it afaik

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


jsierles (commented):
I think this approach is fine for now.

matttpt (approved):

PR #3015, opened 2023-11-09
Speed up `fly volumes` commands by switching to `GetAppBasic`

matttpt:
#### What and Why:

`fly volumes` commands currently use the `GetApp` GraphQL query, which reports a lot of information, including a list of Machines. For apps with a lot of Machines (including destroyed ones), this can take a few seconds for the API to prepare. These commands don't actually need that information, though. They will be faster if we don't fetch it (and it may reduce a bit of load on the API too!).

#### How:

Use `GetAppBasic` instead.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


dangra (approved):
⏩ 🚢

PR #3016, opened 2023-11-09
Add [[compute]] section to fly.toml

dangra:
### Change Summary

We have been relying on command line flags to tackle this problems:
* Set VM size on first deploy
* Restore VM size when scaling up from zero
* Different process groups have different compute requirements (CPUs, GPUs, Dedications...)
* Default shared-cpu-1x is not a good fit for memory hungry apps  

But the use of the flags is not consistent, they are only required under specific circumstances and easily forgotten.
This PR adds consistency to deploys and scaling by extending fly.toml with a `[[compute]]` section customizable per process group.

```toml
app = "my-cool-assistant"
primary_region = "ord"

[processes]
webapi = "/api"
worker = "/worker"

[[mounts]]
source = "models"
destination = "/models"
initial_size = "100gb"
processes = ["worker"]

[[compute]]
cpus = 4
cpu_kind = "shared"
memory_mb = 1024
processes = ["webapi"]

[[compute]]
cpus = "16"
cpu_kind = "performance"
memory_mb = "64gb"
gpu_kind = "a100-pcie-40gb"
processes = ["worker"]

[http_service]
internal_port = 7860
processes = ["webapi"]
```

## DEMO (it works!)

compute-demo

## Follow ups

The following things are planned for upcoming PRs.
1. Volume's initial size
2. Add [[compute]] when passing `--vm-*` flags to `fly launch`

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/api/machine_types.go
+++ b/api/machine_types.go
@@ -325,13 +325,13 @@ type MachineMount struct {
 }
 
 type MachineGuest struct {
-	CPUKind          string `json:"cpu_kind,omitempty"`
-	CPUs             int    `json:"cpus,omitempty"`
-	MemoryMB         int    `json:"memory_mb,omitempty"`
-	GPUKind          string `json:"gpu_kind,omitempty"`
-	HostDedicationID string `json:"host_dedication_id,omitempty"`
+	CPUKind          string `json:"cpu_kind,omitempty" toml:"cpu_kind,omitempty"`
+	CPUs             int    `json:"cpus,omitempty" toml:"cpus,omitempty"`
+	MemoryMB         int    `json:"memory_mb,omitempty" toml:"memory_mb,omitempty"`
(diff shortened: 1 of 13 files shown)


michaeldwan:
Does this apply changes on deploy?

dangra:
> Does this apply changes on deploy?

Only if a `[[compute]]` section is added to fly.toml, otherwise it respects existing behavior of keeping whatever the machines are using.

I am still working on the edge cases but plan to keep it the surprise factor at a minimum, that is not affecting existing apps when they don't have compute requirements set. 

It is too early, but another idea I had was to add a "authoritative: false|true" flag to `[[compute]]` so `fly deploy` knows when to update the current machine guest or only use it for initial deploy.

dangra:
> Does this apply changes on deploy?

@michaeldwan how is the PR looking from your POV? any pushbacks or comments?

PR #3018, opened 2023-11-10
Revert to fix bluegreen deployments

benwaffle:
This reverts commit e2d15d62a963c6a86daeefcfb435788575ba98f1.

I've confirmed with a test app that this fixes bluegreen deployments.

https://community.fly.io/t/machine-doesnt-start-on-deployment/16375


tvdfly (approved):

wjordan:
Can you test the fix in #3019 as an alternative?

PR #3021, opened 2023-11-10
auto volume extension from fly.toml

lubien:
### Change Summary

What and Why:

How:

Related to:

Tested Cases:

- [x] New app with new volume with auto extension
- [x] Disable auto extension for volume that already had it
- [x] Deploy updates all volumes
- [x] New app with new volume WITHOUT auto extension
- [x] Old app with previous volume that will now have auto extension
- [x] App with auto extension but now disabling it
- [x] Clone machine with auto extension

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


internal/command/deploy/machines.go
```diff
@@ -367,9 +367,26 @@ func (md *machineDeployment) validateVolumeConfig() error {
...
+				case extendThresholdPercent == 0 && addSizeGb == 0 && sizeGbLimit == 0:
+					// not using this feature
+				case extendThresholdPercent != 0 && addSizeGb == 0 && sizeGbLimit == 0:
+					return fmt.Errorf("mounts extend_threshold_percent, add_size_gb and size_gb_limit must be all defined or none")
+				case extendThresholdPercent < 50, extendThresholdPercent > 99:
+					return fmt.Errorf("mounts extend_threshold_percent must be between 50 and 99")
+				case addSizeGb < 1, addSizeGb > 100:
+					return fmt.Errorf("mounts add_size_gb must be between 1 and 100")
+				case sizeGbLimit != 0 && (sizeGbLimit < 1 || sizeGbLimit > 500):
+					return fmt.Errorf("mounts size_gb_limit must be between 1 and 500")
```
dangra, line 388: A better place to catch invalid configurations which depends on static state is internal/appconfig/validation.go

lubien: Thanks for the tip, switched and seems better

lubien:
Note: will update `fly m create/update` in another PR, being able to do this via fly.toml already adds much value to users

internal/appconfig/config.go
```diff
@@ -132,10 +132,13 @@ type Static struct {
...
-	Destination string   `toml:"destination,omitempty" json:"destination,omitempty"`
-	InitialSize string   `toml:"initial_size,omitempty" json:"initial_size,omitempty"`
-	Processes   []string `toml:"processes,omitempty" json:"processes,omitempty"`
+	Source                 string   `toml:"source,omitempty" json:"source,omitempty"`
+	Destination            string   `toml:"destination,omitempty" json:"destination,omitempty"`
+	InitialSize            string   `toml:"initial_size,omitempty" json:"initial_size,omitempty"`
+	Processes              []string `toml:"processes,omitempty" json:"processes,omitempty"`
+	ExtendThresholdPercent int      `toml:"extend_threshold_percent,omitempty" json:"extend_threshold_percent,omitempty"`
+	AddSizeGb              int      `toml:"add_size_gb,omitempty" json:"add_size_gb,omitempty"`
+	SizeGbLimit            int      `toml:"size_gb_limit,omitempty" json:"size_gb_limit,omitempty"`
```
dangra, line 141: Can we come up with a naming scheme that reflects all this fields are related to size extension ?

* `extend_size_limit`
* `extend_size_increment`
* `extend_size_threshold`

Another nit is that we recently started supporting more human parseable values for sizes, see how it is done for `initial_size` and machine's `--vm-memory` flags for example.

I can do all these changes if you don't have time for it.

dangra (commented):
Functionality wise it looks clean and correct, the DX can be improved IMHO.

internal/appconfig/config.go
lubien: Since both limits are GB we might just append in the name `_gb`

How about:

- `auto_extend_add_gb`
- `auto_extend_limit_gb`
- `auto_extend_threshold`

dangra: nope, the idea is that the field type is a string and accepts units like in `10gb` or `1024mb` and defaults to GB if none is passed. See how it is done for the keys I mentioned.

lubien: I see. Will need to also validate there's no fractional GB too. You names make sense now

dangra: `autoextend` as a prefix makes sense tbh:

* `autoextend_size_limit`
* `autoextend_size_increment`
* `autoextend_size_threshold`

lubien: Just seen this now. How about `auto_extend_*`?

lubien: done

dangra (approved):

smorimoto:
Thanks a lot for this! I was surprised just a few days ago when I was unable to write to my development DB due to insufficient disk space 😵‍💫

PR #3034, opened 2023-11-14
Fail deploys quickly if a worker is down.

ndarilek:
### Change Summary

If a worker for a machine is down, fail deploys immediately rather than silently leaving apps in inconsistent states.

---

### Documentation

- [ ] Fresh Produce
- [X ] n/a

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -202,6 +202,9 @@ func NewMachineDeployment(ctx context.Context, args MachineDeploymentArgs) (Mach
 		volumeInitialSize:      args.VolumeInitialSize,
 		processGroups:          args.ProcessGroups,
 	}
+	if err := md.checkForDownWorkers(ctx); err != nil {
+		return nil, err
+	}
 	if err := md.setStrategy(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


dangra:
hey, I tend to disagree with the solution. I think fly deploy should replace the machine and leave a tombstone somewhere so it is destroyed once flyd on the lost worker comes back.

dangra:
Also, the lack of machine metadata, which is the cause for a 1970 date, can be mitigated by querying from info from corrosion when flyd doesn't respond.

ndarilek:
I agree, that sounds like a great eventual place to end up. Right now what I *think* we're doing is deploying an update to only part of the machines for an app, then when the other flyd comes back it's running an outdated image. What happens if the 2 successful deploys migrate a database to a version incompatible with the image running on the failed deploy? Suddenly parts of your app are working and others aren't.

If we can ensure that the failed flyd launches the newer version when it comes up, we could probably remove this. But I think failure might be better than partial success here.

dangra:
> But I think failure might be better than partial success here.

I don't think so. It blocks app deployments which is worse when you need to ship a fix but suddenly can't because a single disposable machine isn't reachable.  

btw, we update machines concurrently so this change will still left machines running inconsistent versions.

ndarilek:
> btw, we update machines concurrently so this change will still left machines running inconsistent versions.

Darn, somehow I figured it wouldn't be that easy. :) Where would I need to inject a check like that to not get machines into an inconsistent state?

Don't get me wrong, I'd like to eventually see a better solution for this. I'm just trying something that's good enough for now. I'm also hesitant about changing how we write to corrosion exactly when we're migrating to v2 and could probably benefit from leaving that alone for a bit. :)

Thanks for the review.

dangra:
> Darn, somehow I figured it wouldn't be that easy. :) Where would I need to inject a check like that to not get machines into an inconsistent state?

Your changes as they stand are doing what is intended because the check happens before any machine update is attempted. 
My itching with this approach is that it will block actual users from deploying and we can't offer a way to workaround it.

But once the machines start updating, a temporal error like a network hiccup will cancel the whole update leaving some machines running a mix of new and old versions.

ndarilek:
> But once the machines start updating, a temporal error like a network hiccup will cancel the whole update leaving some machines running a mix of new and old versions.

OK, so I guess a better question is, if one of an app's worker machines is down, how can we interrupt a deployment so they aren't running inconsistent versions?

It sounds like you might be saying we *shouldn't*?

Learning this as I go so please let me know if I'm misunderstanding how something is working. For context, this is in response to https://flyio.discourse.team/t/q-what-should-happen-when-a-host-goes-down-but-the-app-has-multiple-machines/3930.

PR #3039, opened 2023-11-15
When `--json`, only print valid JSON

### LINKED ISSUE #2809: certs check --json does not return valid json
`flyctl certs check host.tld --json` does not return valid json. The result starts with

```
The certificate for host.tld has been issued.

{
```

I guess the culprit is here:
https://github.com/superfly/flyctl/blob/64a31bf73c6e1509567cc6be0b2a19340e80e810/internal/command/certificates/root.go#L164
and should be wrapped in a check for the `--json` flag.

amo13:
The code change suppresses the output of a message to stdout before the actual certificate JSON. Without this change, if you try to parse the response as JSON, you get an error because the message and the two line breaks are not valid JSON.

Related to and fixes #2809

### DIFF
--- a/internal/command/certificates/root.go
+++ b/internal/command/certificates/root.go
@@ -163,7 +163,7 @@ func runCertificatesShow(ctx context.Context) error {
 	if cert.ClientStatus == "Ready" {
 		return nil
 	}
-	
+
 	return reportNextStepCert(ctx, hostname, cert, hostcheck)
 }
 
@@ -178,7 +178,7 @@ func runCertificatesCheck(ctx context.Context) error {
 	}
 
 	printCertificate(ctx, cert)
-	
+
 	if cert.ClientStatus == "Ready" {
 		return nil
 	}
--- a/internal/command/certificates/root.go
+++ b/internal/command/certificates/root.go
@@ -150,7 +150,6 @@ func runCertificatesList(ctx context.Context) error {
 
 func runCertificatesShow(ctx context.Context) error {
 	io := iostreams.FromContext(ctx)
-	colorize := io.ColorScheme()
 	apiClient := client.FromContext(ctx).API()
 	appName := appconfig.NameFromContext(ctx)
 	hostname := flag.FirstArg(ctx)
@@ -171,7 +170,6 @@ func runCertificatesShow(ctx context.Context) error {
 
 func runCertificatesCheck(ctx context.Context) error {
 	io := iostreams.FromContext(ctx)
-	colorize := io.ColorScheme()
 	apiClient := client.FromContext(ctx).API()
 	appName := appconfig.NameFromContext(ctx)
 	hostname := flag.FirstArg(ctx)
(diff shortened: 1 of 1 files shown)


internal/command/certificates/root.go
```diff
@@ -161,7 +161,9 @@ func runCertificatesShow(ctx context.Context) error {
 	}
 
 	if cert.ClientStatus == "Ready" {
-		fmt.Fprintf(io.Out, "The certificate for %s has been issued.\n\n", colorize.Bold(hostname))
+		if !flag.GetBool(ctx, "json") {
+			fmt.Fprintf(io.Out, "The certificate for %s has been issued.\n\n", colorize.Bold(hostname))
+		}
```
dangra, line 166: this is not the only place where "printCertificate" is called, and the messages about certificate issued/not issued are printed on all cases. Simpler to move them to the printCertificates function so it works for json in all cases.

dangra (changes requested):

internal/command/certificates/root.go
amo13: Thanks for your reply. I'm not sure how to go about this though. I'd need to make printCertificates know if it should say "has been issues" or "has not been issued". Should I simply make it take a boolean as third argument and adjust the calls of printCertificates?

dangra: no need to pass a third argument, `cert` is already passed to printCertificates and for _issued/not issued_ it checks `cert.ClientStatus == "Ready"` field

amo13: You are right, thank you. It should be better now.

amo13:
Do I need to do something about the failing tests? I don't see what exactly they are complaining about though...

dangra:
> Do I need to do something about the failing tests? I don't see what exactly they are complaining about though...

does it compile locally?

it is failing because of multiple compilation errors:

```
# github.com/superfly/flyctl/internal/command/certificates
Error: internal/command/certificates/root.go:153:2: colorize declared and not used
Error: internal/command/certificates/root.go:1[7](https://github.com/superfly/flyctl/actions/runs/6883176500/job/18723561100?pr=3039#step:4:8)4:2: colorize declared and not used
Error: internal/command/certificates/root.go:393:70: undefined: colorize
Error: internal/command/certificates/root.go:393:[8](https://github.com/superfly/flyctl/actions/runs/6883176500/job/18723561100?pr=3039#step:4:9)4: undefined: hostname
Error: internal/command/certificates/root.go:3[9](https://github.com/superfly/flyctl/actions/runs/6883176500/job/18723561100?pr=3039#step:4:10)5:78: undefined: colorize
Error: internal/command/certificates/root.go:395:94: undefined: hostname
```

dangra:
Check the PR diff, it is full of error annotations https://github.com/superfly/flyctl/pull/3039/files

dangra:
it's still failing to compile @amo13

dangra:
btw, we use [pre-commit](https://pre-commit.com/) hooks to help devs catch errors much earlier.

amo13:
Ok, thank you for the hint. Actually, I have no dev env set up for this repo and didn't plan on contributing, but suddenly I clicked on create pull request ;)
I looks to me like there is nothing I can do about `CI - Dev / tag_version (pull_request)` failing. Is that right?

dangra:
> I looks to me like there is nothing I can do about `CI - Dev / tag_version (pull_request)` failing. Is that right?

that's right.

I am merging your PR. Thanks for the contribution!

dangra (approved):

PR #3050, opened 2023-11-17
Update building instructions in CONTRIBUTING.md

IohannesArnold:
Running `go build .` alone returns the error:

```
go: build output "flyctl" already exists and is a directory
```

So tell users that if they want to use `go build` they need to explicitly name the output exe

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 
 ## Building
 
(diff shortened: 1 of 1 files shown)


billyb2:
banger pr

PR #3052, opened 2023-11-19
fly console: keep stdout clean

benwaffle:
This allows piping the output to other commands

https://community.fly.io/t/suppress-log-output-when-running-fly-console-c/16535

### DIFF
--- a/internal/machine/ephemeral.go
+++ b/internal/machine/ephemeral.go
@@ -40,7 +40,7 @@ func LaunchEphemeral(ctx context.Context, input *EphemeralInput) (*api.Machine, 
 	if input.What != "" {
 		creationMsg += " " + input.What
 	}
-	fmt.Fprintf(io.Out, "%s.\n", creationMsg)
+	fmt.Fprintf(io.ErrOut, "%s.\n", creationMsg)
 
 	sp := spinner.Run(io, fmt.Sprintf("Waiting for %s to start ...", colorize.Bold(machine.ID)))
 	defer sp.Stop()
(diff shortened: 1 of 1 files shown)


internal/machine/ephemeral.go
```diff
@@ -40,7 +40,7 @@ func LaunchEphemeral(ctx context.Context, input *EphemeralInput) (*api.Machine,
 	if input.What != "" {
 		creationMsg += " " + input.What
 	}
-	fmt.Fprintf(io.Out, "%s.\n", creationMsg)
+	fmt.Fprintf(io.ErrOut, "%s.\n", creationMsg)
 
 	sp := spinner.Run(io, fmt.Sprintf("Waiting for %s to start ...", colorize.Bold(machine.ID)))
```
mootari, line 45: Doesn't the spinner text also end up in STDOUT?

rubys: Perhaps only output the creationMsg and spinner if the output is going to a terminal?

See: https://rderik.com/blog/identify-if-output-goes-to-the-terminal-or-is-being-redirected-in-golang/#final-thoughts

rubys: Now that I know what to look for, I found an existing helper function:

https://github.com/superfly/flyctl/blob/5dcfd7514a6b3b94bb9038453960bd33b5035fdc/helpers/stdin.go#L11-L17

benwaffle: weird that it doesn't use `isatty()`

benwaffle: https://github.com/superfly/flyctl/blob/master/internal/cmdutil/terminal.go#L9-L11

rubys: See https://github.com/superfly/flyctl/pull/3053 for a pull request that uses cmdutil.IsTerminal

PR #3055, opened 2023-11-20
Lay foundations for xdg base dirs

IohannesArnold:
This commit prepares the codebase for a future implementation of xdg base dirs. Until now, the whole codebase assumed there was one "config dir" where anything flyctl-related went, even if it was not a config file. This commit creates two new concepts: the "state dir" and the "runtime dir", corresponding to XDG_STATE_HOME and XDG_RUNTIME_DIR, respectively. However, the commit keeps the same logic for determining the state dir and the runtime dir as the config dir, so the behavior of flyctl should not be different in any way.


billyb2 (approved):
Long overdue ✔️ . I just ask that you make sure that the PR passes preflight before merging :)

Firas2515 (commented):
Https:// docs.soliditylang.org

PR #3059, opened 2023-11-21
print stack traces on `LOG_LEVEL=debug`

lillianberryfly:
### Change Summary

What and Why:

Right now, only dev builds of flyctl print stack traces. I don't work primarily on flyctl so I don't usually have a dev build around, but I still want to see stack traces without checking Sentry.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


kzys:
How about removing the if condition and show the stacktrace regardless? That's what most Go programs do when they have a panic.

kzys:
Closing due to #3850.

PR #3060, opened 2023-11-21
Added check for Vite app

anniebabannie:
### Change Summary

With the release of `dockerfile-node` version 0.4.11, we now check for Vite apps so we tweak the Dockerfile to not use Node, but rather Nginx. As such, I've included the extra check in `flyctl`, as we've done for the other JS frameworks.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/scanner/jsFramework.go
+++ b/scanner/jsFramework.go
@@ -155,6 +155,10 @@ func configureJsFramework(sourceDir string, config *ScannerConfig) (*SourceInfo,
 	if !ok || devdeps == nil {
 		devdeps = make(map[string]interface{})
 	}
+	scripts, ok := packageJson["scripts"].(map[string]interface{})
+	if !ok || scripts == nil {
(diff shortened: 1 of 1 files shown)


scanner/jsFramework.go
```diff
@@ -198,6 +202,9 @@ func configureJsFramework(sourceDir string, config *ScannerConfig) (*SourceInfo,
 	} else if deps["gatsby"] != nil {
 		srcInfo.Family = "Gatsby"
 		srcInfo.Port = 8080
+	} else if scripts["dev"] == "vite" {
```
rubys, line 205: Move this after remix?

anniebabannie: oops you're right, my brain was thinking of the last change i made where it was a bunch of `if`s and not `else if`s

rubys (approved):

PR #3062, opened 2023-11-22
Consolidate usage of the process group flag

jsierles:
### Change Summary

What and Why:

This PR ensures all references to process groups can use `--group`,  `--process-group` or `-g`. This has bugged me, and similar [constructive feedback](https://www.lightward.guide/technical/fly/rough-edges) from a customer is good enough reason to do it.

How:

Use a common flag method with a configurable description.

### Documentation

- [ ] Fresh Produce
- [x] n/a


internal/flag/flag.go
```diff
@@ -529,9 +529,16 @@ func JSONOutput() Bool {
...
 
-func ProcessGroup() String {
+func ProcessGroup(desc string) String {
+
+	if desc == "" {
+		desc = "The target process group"
+	}
+
 	return String{
 		Name:        flagnames.ProcessGroup,
```
benwaffle, line 539: Does changing the constant from "process-group" to "group" break anything?

jsierles: It still has the `process-group` alias, so invocations will work. It will just show up as `group` in help. We had variations on this before. Now all three options work everywhere, and `group` is shown as the option.

benwaffle: You'll have to update the occurrences of `flag.GetString`, because it operates on the Name.

e.g.:
https://github.com/superfly/flyctl/blob/6cf80bccc8832537d4f50aa6c19738d53d308c18/internal/command/scale/count.go#L62C10-L62C10

jsierles: Ah right, good call

jsierles: Fixed that

internal/flag/flag.go
```diff
@@ -529,9 +529,16 @@ func JSONOutput() Bool {
...
-		Description: "The target process group",
+		Description: desc,
+		Shorthand:   "g",
+		Aliases:     []string{"process-group"},
```
dangra, line 542: Could we agree on using `--process-group` as main flag name, `--group` as alias and `-g` as shorthand? 

The reasoning is that in docs and forum topics we refer to this feature as _Process Groups_, calling them groups don't help the brain muscle.

jsierles: Done!

PR #3065, opened 2023-11-24
adjust --no-deploy desc for launch ui experience

andie787:
### Change Summary

What and Why:

Need to adjust the description for the `fly launch --no-deploy` flag to reflect the launch ui experience.

How:

Updated the help text

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/flag/flag.go
+++ b/internal/flag/flag.go
@@ -297,7 +297,7 @@ func (d Duration) addTo(cmd *cobra.Command) {
 func Org() String {
 	return String{
 		Name:         flagnames.Org,
-		Description:  "The target Fly organization",
+		Description:  "The target Fly.io organization",
 		Shorthand:    "o",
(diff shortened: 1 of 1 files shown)


internal/flag/flag.go
```diff
@@ -371,7 +371,7 @@ func Now() Bool {
 func NoDeploy() Bool {
 	return Bool{
 		Name:        "no-deploy",
-		Description: "Do not prompt for deployment",
+		Description: "Do not deploy the new app after initial creation and configuration",
```
andie787, line 374: My other idea was:
```suggestion
		Description: "Do not deploy the new app after fly launch creates and configures it",
```

catflydotio: Either works for me. I wonder about adding the word "immediately" in there.

PR #3073, opened 2023-11-29
Plumb `tokens.Tokens` struct into `api.Client`

btoews:
This PR plumbs the `tokens.Tokens` struct that is loaded from the user's config down into the `api.Client`. This allows the `tokens.Tokens` struct to be updated without having to replace the `api.Client` in the context. It also allows the `api.Client` to use different tokens for different requests without having to re-parse the token string it was configured with.

For example, this PR also fixes a bug where we were sending oauth+macaroon tokens to bubblegum-api when fetching old logs when that app only knows how to handle oauth _or_ macaroon tokens in a request.

I was careful not to break existing APIs in the `api` package. This makes for less-than-ideal APIs and some deprecated fields. If we're okay with breaking changes, I'd restructure things a bit.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	genq "github.com/Khan/genqlient/graphql"
+	"github.com/superfly/flyctl/internal/tokens"
 	"github.com/superfly/graphql"
 )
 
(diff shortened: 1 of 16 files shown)


internal/command/macaroons.go
```diff
@@ -0,0 +1,97 @@
+package command
```
btoews, line 1: The contents of this file came from `internal/tokens/tokens.go`

api/tokens/tokens.go
```diff
@@ -0,0 +1,112 @@
+package tokens
```
btoews, line 1: The contents of this file came from `internal/tokens/tokens.go`

btoews:
Weee. Lots of rebases and force pushes! I had to move some stuff around to fix some tricky dependency issues. This should be ready for :eyes: now.

PR #3074, opened 2023-11-29
deploy: hint that machines may time out due to region issue

alichay:
### Change Summary

What and Why: if your deploy times out while waiting for a machine to reach "started," it suggests changing the wait timeout *only*. This is misleading, as this error can sometimes be caused by hosts that are having issues.

How: This shares the blame a bit, telling users that this could *also* be due to a region issue, giving them more to work off of.

Related to: https://github.com/superfly/reflyability/issues/12

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


benwaffle:
Maybe there's a way to pass the cause through the context here? https://flyio.discourse.team/t/death-to-context-cancelled/4461/3

alichay:
> Maybe there's a way to pass the cause through the context here? https://flyio.discourse.team/t/death-to-context-cancelled/4461/3

sure. I was just wary of getting into refactoring territory

PR #3076, opened 2023-11-29
Do not attempt to leases machines on scale operation if too slow

dangra:
### Change Summary

What and Why: Attempting to scale apps with a large number of machines is a slow operation because of lease grabbing.
The lease grabbing is performed sequentially and requires two api calls, one to grab the lease and a second one to update the machine config. 

How: Add a `--do-not-acquire-leases` flag to `fly scale count` and disable lease grabbing on apps with 30+ machines.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


dangra:
> Can we do them all in parallel instead? Ideally we could make this faster without adding more flags.

Totally agree!  (the flag was suppose to be a escape hatch for air.ai)

I can rework it (tomorrow) to run multiple lease calls concurrently, but this is a case where server side support for batched leases would be even faster.

PR #3079, opened 2023-11-30
Acquire and release machine leases concurrently

dangra:
### Change Summary

What and Why: A follow up of https://github.com/superfly/flyctl/pull/3076 to grab leases concurrently instead of not doing it.

How:

Related to: https://github.com/superfly/flyctl/pull/3076

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -107,7 +107,10 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
 		}
 	}
 
-	machines, releaseFunc, err := mach.AcquireLeases(ctx, machines)
+	// XXX: Don't acquire the leases until the user confirms it wants to execute any action
+	//      The downside is that AcquireLeases has the side effect of fetching an updated copy of machine config
+	//      that we don't use here, but it also updates the `LeaseNonce` field of the original machine which we rely on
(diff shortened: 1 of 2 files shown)


benwaffle:
I tested this on an app with 200 machines by running `fly deploy`, and it doesn't seem to be any faster. Is that expected?

dangra:
> I tested this on an app with 200 machines by running `fly deploy` before the lease changes and after, and it doesn't seem to be any faster. Is that expected?

Yes, it is. `fly deploy` uses a different mechanism for leasing that is parallelized already ([MachineSet](https://github.com/superfly/flyctl/blob/master/internal/machine/machine_set.go)). The only recent improvements to deploy logic was to concurrently and continuously update machines, before it was done semi concurrently and in batches. Leasing wasn't a problem.

The changes here affects all other commands like `fly scale`, `fly apps restart`, `fly machine restart` which are backed by https://github.com/superfly/flyctl/blob/master/internal/machine/lease.go

Why do we have two leasing mechanism? no idea, they seems very similar but one is tuned a lot towards deployments. Ideally we should have only one, or one backing the other.

benwaffle:
got it, just tried `fly scale count` and it was literally unusable with 500 machines before your changes. nice job!

PR #3087, opened 2023-12-01
Add `fly volume snapshot create`

kzys:
This new command is used to snapshot volumes on-demand.

### Change Summary

What and Why: Adding a new command to snapshot volumes on-demand. In addition to daily snapshots, customers are now able to take arbitrary number of snapshots.

How: By adding the functionality on the server-side. In other words, we cannot merge the PR until we update the servers.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/volumes/snapshots/create.go
+++ b/internal/command/volumes/snapshots/create.go
@@ -1,0 +1,57 @@
+package snapshots
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/superfly/flyctl/client"
+	"github.com/superfly/flyctl/flaps"
+	"github.com/superfly/flyctl/internal/appconfig"
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/flag"
+)
+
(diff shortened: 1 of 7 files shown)


kzys:
Not ready for merge (because the backend is not there) but ready for review.

internal/command/volumes/snapshots/create.go
```diff
@@ -0,0 +1,60 @@
...
+func newCreate() *cobra.Command {
+	const (
+		short = "Snapshot a volume"
+		long  = "Snapshot a volume\n"
+		usage = "create <volume-id>"
+	)
+
+	cmd := command.New(usage, short, long, create, command.RequireSession)
+
+	cmd.Aliases = []string{"create"}
```
dangra, line 25: no need to alias the command name. It is "extracted" from usage field.

dangra (approved):

internal/command/volumes/snapshots/create.go
kzys: TIL! Fixed in the last revision.

PR #3090, opened 2023-12-04
Detect Defaults for Laravel Scanner: DB, Redis

KTanAug21:
### Change Summary

**What and Why:**
We need to set defaults for apps deployed in Fly.io, [see here!](https://community.fly.io/t/the-new-launch/16490). This is also [true for Laravel apps](https://flyio.slack.com/archives/C042W39VAMB/p1700229728827279). So this PR contains changes to detect DBs and redis connection set in the local .env file, and set those as defaults for the Laravel Fly App, closely following the changes already available in the [rails scanner.](https://github.com/superfly/flyctl/blob/master/scanner/rails.go#L80-L110)

**How:**
Checks the .env file of a Laravel directory and scans for db connection strings like "mysql", "pgsql", and "sqlite". If any of the connections above are available,  sets it as the default DesiredDatabase value. 
Additionally, detects the "redis" keyword and sets the configuration value of "RedisDesired" as true.

**Related to:**
1. Setting of [defaults Fresh Produce](https://community.fly.io/t/the-new-launch/16490)
2. The Rails Scanner( this PR would take inspiration on the [rails scanner](https://github.com/superfly/flyctl/blob/master/scanner/rails.go#L80-L110) for detecting defaults )

---

### Documentation
(will add docs in the laravel section of our docs) 
- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/scanner/laravel.go
+++ b/scanner/laravel.go
@@ -115,7 +115,7 @@ func extractPhpVersion() (string, error) {
 	return "", fmt.Errorf("could not find php version")
 }
 
-var dbRegStr = "^ *(DB_CONNECTION|DATABASE_URL) *= *[a-zA-Z]+"
+var dbRegStr = "^ *(DB_CONNECTION|DATABASE_URL) *=(\"|')? *[a-zA-Z]+(\"|')?"
 var redisRegStr = "^[^#]*redis"
 
 // extractConnections detects the database connection of a laravel fly app
@@ -129,6 +129,7 @@ var redisRegStr = "^[^#]*redis"
 //	skipDb - reports whether a connection or redis was detected
 func extractConnections(path string) (db DatabaseKind, redis bool, skipDb bool) {
 	// Get File Content
+
 	file, err := os.Open(path)
 	if err != nil {
 		return 0, false, true
--- a/scanner/laravel.go
+++ b/scanner/laravel.go
@@ -115,6 +115,9 @@ func extractPhpVersion() (string, error) {
 	return "", fmt.Errorf("could not find php version")
 }
 
+
+var dbRegStr = "^ *DB_CONNECTION *= *[a-zA-Z]+"
+var redisRegStr = "^[^#]*redis"
 // extractConnections detects the database connection of a laravel fly app
 // By checking the .env file in the project's base directory for connection keywords.
 // This ignores commented out lines and prioritizes the first connection occurance over others
@@ -133,9 +136,9 @@ func extractConnections(path string) (db DatabaseKind, redis bool, skipDb bool) 
 
 	// Set up Regex to match
 	// -not commented out, with DB_CONNECTION
-	dbReg := regexp.MustCompile("^ *DB_CONNECTION *= *[a-zA-Z]+")
+	dbReg := regexp.MustCompile( dbRegStr )
 	// -not commented out with redis keyword
-	redisReg := regexp.MustCompile("^[^#]*redis")
+	redisReg := regexp.MustCompile( redisRegStr )
 
 	// Default Return Variables
 	db = 0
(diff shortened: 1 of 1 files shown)


dangra (approved):
It can certainly be improved but looks good as first iteration!

scanner/laravel.go
```diff
@@ -104,3 +114,54 @@ func extractPhpVersion() (string, error) {
...
+	// Get File Content
+	file, err := os.Open(path)
+	if err != nil {
+		return 0, false, true
+	}
+	defer file.Close()
+
+	// Set up Regex to match
+	// -not commented out, with DB_CONNECTION
+	dbReg := regexp.MustCompile("DB_CONNECTION *= *[a-zA-Z]+")
```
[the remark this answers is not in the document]
KTanAug21: Nice catch!! The regex above is incorrect! It ain't gonna ignore commented out DB_CONNECTIONS, thanks for the eagle eyes on this!

Fixed this by only accepting whitespace before the DB_CONNECTION and reject variations with other characters before the string, ultimately ignoring lines where comment characters are added!

dangra (approved):

KTanAug21:
> It can certainly be improved but looks good as first iteration!

Hi @dangra! Thank you for taking the time check my PR, I appreciate it! I've fixed the issue you've pointed above, and I'm eager to know how I can improve it, so when you're free, please do list some notes I can use to improve my PR

scanner/laravel.go
```diff
@@ -104,3 +114,54 @@ func extractPhpVersion() (string, error) {
 
 	return "", fmt.Errorf("could not find php version")
 }
+
+func extractConnections(path string) (DatabaseKind, bool, bool) {
+	/*
+		Determine the default db
+		Determine whether redis connection is desired
+			returns ( db, redis, skipDB )
+	*/
```
dangra, line 123: See how to document Functions https://tip.golang.org/doc/comment#func

scanner/laravel.go
```diff
@@ -104,3 +114,54 @@ func extractPhpVersion() (string, error) {
...
+	if err != nil {
+		return 0, false, true
+	}
+	defer file.Close() //skipcq: GO-S2307
+
+	// Set up Regex to match
+	// -not commented out, with DB_CONNECTION
+	dbReg := regexp.MustCompile("^ *DB_CONNECTION *= *[a-zA-Z]+")
+	// -not commented out with redis keyword
+	redisReg := regexp.MustCompile("^[^#]*redis")
```
dangra, line 136: static regular patterns (and its regex.MustCompile() calls) are ideal as global variables as they can catch regex syntax errors earlier. More details at https://stackoverflow.com/a/72185182

scanner/laravel.go
```diff
@@ -104,3 +114,54 @@ func extractPhpVersion() (string, error) {
...
+		} else if db == 0 && dbReg.MatchString(text) {
+			if strings.Contains(text, "mysql") {
+				db = DatabaseKindMySQL
+				skipDb = false
+			} else if strings.Contains(text, "pgsql") {
+				db = DatabaseKindPostgres
+				skipDb = false
+			} else if strings.Contains(text, "sqlite") {
+				db = DatabaseKindSqlite
+				skipDb = false
```
dangra, line 161: text contains is very loose matching, a better approach would URI parsing. See https://gobyexample.com/url-parsing

fideloper: Laravel's config isn't a connection string that'll work with parseURL, - in this case `DB_CONNECTION` is one of `mysql, `pgsql`, `sqlite`, `sqlsrv` and in rarer cases, something custom. It's not a full DB connection string tho.

fideloper: image

KTanAug21: Team thanks!! I added the DATABSE_URL case in the regex to check connections from!

PR #3100, opened 2023-12-07
Require unique zone for volumes during v2 migration

benbjohnson:
### Change Summary

What and Why: Volumes should typically be migrated to unique hosts for improved reliability/durability.

How: Mark the `RequireUniqueZone` as `true` in the `CreateVolume()` call.

Related to:

Shout out to @matttpt for spotting the issue 🎉 

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/migrate_to_v2/volumes.go
+++ b/internal/command/migrate_to_v2/volumes.go
@@ -105,6 +105,7 @@ func (m *v2PlatformMigrator) migrateAppVolumes(ctx context.Context) error {
 				Name:                vol.Name,
 				ComputeRequirements: m.machineGuests[processGroup],
 				Region:              region,
+				RequireUniqueZone:   api.Pointer(true),
 			})
 			if err != nil && strings.HasSuffix(err.Error(), " is not a valid candidate") {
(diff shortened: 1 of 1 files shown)


alichay:
as long as machine placement is dependent on volume placement (not sure if it is, can't check atm), lgtm

benbjohnson:
@alichay Yes, it looks like the machines are created based on the volumes:

https://github.com/superfly/flyctl/blob/850acb13a38853b3931aa5090b8b794f3417e0c4/internal/command/migrate_to_v2/machines.go#L172-L181

alichay:
> @alichay Have you seen this CI error before? It looks like the migration was successful but it's erroring out. I tried rerunning the test to see if it was flakey but it's still happening:

No, I don't have any clue what's happening to be honest. I looked through migrate-to-v2 failures in sentry, and while I could find a *few* that happened during roughly that time period, I couldn't find any that matched the app ID or machine IDs from the test output. Totally stumped.

benbjohnson:
It passes now. 🤷 I ran this code for a bunch of Consul cluster migrations so I'm going merge. 👍

PR #1643, opened 2023-01-30
`launch` and `deploy`: support both `docker` and `podman` when do local image build

ttys3:
implement as upstream: https://github.com/buildpacks/pack/pull/988

add `--buildpacks-docker-host` flag

this will allow using local build with tools other than docker, like `podman`, we need to add this flag because https://buildpacks.io/docs/app-developer-guide/building-on-podman/#pack-build  
with this PR, flyctl build image will just works like buildpacks:

```shell
fly deploy --local-only --buildpacks-docker-host=inherit
```

-------------------------------------------------

we need some env setup (only once for a local machine) for this to work (for detailed document, please see https://buildpacks.io/docs/app-developer-guide/building-on-podman/ ):

```shell
# enable podman user socket
systemctl enable --user podman.socket
systemctl start --user podman.socket

# add below to .zshrc or .bashrc
export DOCKER_HOST="unix://$(podman info -f "{{.Host.RemoteSocket.Path}}")"
```

PR #3108, opened 2023-12-11
buildpacks builder: add volume mount support

ttys3:
### Change Summary

add volume mount support for the`buildpacks` builder

What and Why:

currently there's no way to mount local volume to buildpacks. like we need to support private modules on github.

ref https://paketo.io/docs/howto/go/#import-private-go-modules

for example, we need bind git credentails:

we will need to bind the `git-binding` dir to `/platform/bindings/git-binding`

How:

by add a `--buildpacks-volume` cli flag, user can now mount volume to buildpacks builder containers.

like this:

prepare the service bindings dirs:

```shell
echo git-credentials > $HOME/.git-binding-type


cat << EOF > $HOME/.git-binding-credentials

url=https://github.com
username=foo
password=github_pat_xxxxxxxxxx

EOF
```

now we can build success:

```shell
flyctl --verbose --debug deploy --local-only \
--env "CGO_ENABLED=0" \
--env GOPRIVATE="github.com/foo/private-go-mod" \
--buildpacks-volume $HOME/.git-binding-type:/platform/bindings/git-binding/type \
--buildpacks-volume $HOME/.git-binding-credentials:/platform/bindings/git-binding/credentials
```

Related to:

this PR is a continued enhancement for the`buildpacks` builder after https://github.com/superfly/flyctl/pull/1643

---

### Documentation

- [x] n/a

### DIFF
--- a/internal/build/imgsrc/buildpacks_builder.go
+++ b/internal/build/imgsrc/buildpacks_builder.go
@@ -88,6 +88,14 @@ func (*buildpacksBuilder) Run(ctx context.Context, dockerFactory *dockerClientFa
 	}
 	build.ContextBuildFinish()
 
+	if opts.BuildpacksDockerHost != "" {
+		cmdfmt.PrintDone(streams.ErrOut, fmt.Sprintf("BuildpacksDockerHost=%v", opts.BuildpacksDockerHost))
+	}
+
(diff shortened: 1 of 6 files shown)


dangra:
thanks again @ttys3

PR #3115, opened 2023-12-11
disable many v2 regions commands

rubys:
Issue more helpful error message if somebody attempts to
run regions commands that are not supported on v2.


andie787 (commented):
OK! You added it in nomad; that makes sense I was trying to figure out how to make it work in root. :)

This lgtm!

rubys:
Adding to root is possible, but harder.  Each command would have to route to a new function (like is done today for list), and that function would have to extract the app, and determine whether or not to forward the request to the nomad function or produce an error.

I took the lazy way out.  If you get to the nomad function but your app is v2, abort.

PR #3131, opened 2023-12-18
Skip launch UI if --yes was passed on fly launch command

rubys:
See https://community.fly.io/t/pass-in-cli-arg-to-skip-do-you-want-to-tweak-these-settings-before-proceeding-during-fly-launch/17308


alichay (approved):

nicholasoxford:
Thank you! 🤩

PR #3146, opened 2024-01-02
attach trace-id to sentry events

gwuah:
### Change Summary
Attach a trace-id to sentry events so we can investigate weird issues popping up in sentry using honeycomb.

### Documentation
- [ ] Fresh Produce
- [x] n/a


billyb2 (approved):
👍🏾

PR #3150, opened 2024-01-03
Replace `GetApp` in `fly tokens list` and `fly pg attach`

matttpt:
### Change Summary

**What and Why:**

`GetApp` returns a lot of information from the GraphQL API, including a list of all of an app's Machines (including destroyed ones!). For large apps or apps with a lot of history, this can be very slow.

https://github.com/superfly/flyctl/pull/3015 sped up `fly volumes` commands by replacing `GetApp`. This PR is a follow-on that (finally) gives the same treatment to `fly tokens list` and `fly pg attach`.

The remaining uses of `GetApp` are in `fly turboku` and `fly migrate-to-v2`, neither of which seem worth updating.

**How:**

* For `fly tokens list`, create and use a new `GetAppLimitedAccessTokens` query.
* For `fly pg attach`, use `GetAppBasic` and `GetIPAddresses` as appropriate.

**Related to:**

* https://github.com/superfly/flyctl/pull/3015

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/api/resource_tokens.go
+++ b/api/resource_tokens.go
@@ -4,6 +4,33 @@ import (
 	"context"
 )
 
+func (c *Client) GetAppLimitedAccessTokens(ctx context.Context, appName string) ([]LimitedAccessToken, error) {
+	query := `
+		query ($appName: String!) {
+			app(name: $appName) {
+				limitedAccessTokens {
+					nodes {
+						id
+						name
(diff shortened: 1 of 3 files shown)


dangra (approved):
LGTM!

PR #3172, opened 2024-01-09
use vm.memory_mb from fly.toml (if present) as a lower bound

rubys:
use process group 'console' to make this determination

### DIFF
--- a/internal/command/console/console.go
+++ b/internal/command/console/console.go
@@ -218,7 +218,7 @@ func selectMachine(ctx context.Context, app *api.AppCompact, appConfig *appconfi
 	} else if flag.IsSpecified(ctx, "machine") {
 		return getMachineByID(ctx)
 	} else {
-		guest, err := determineEphemeralConsoleMachineGuest(ctx)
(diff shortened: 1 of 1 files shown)


billyb2 (approved):
LGTM

internal/command/console/console.go
```diff
@@ -393,6 +393,15 @@ func determineEphemeralConsoleMachineGuest(ctx context.Context) (*api.MachineGue
 			return nil, fmt.Errorf("invalid CPU kind '%s'; this is a bug", desiredGuest.CPUKind)
 		}
 
+		// use vm.memory_mb from fly.toml (if present) as a lower bound
+		appConfig.Flatten("console")
```
dangra, line 397: hmm, this is a NOOP line. `AppConfig.Flatten()` returns a copy of appConfig flatted for the process group.

honestly, much of the code in this function could use a simplification by reusing `internal.flag.GetMachineGuest()` and `internal.appconfig.Config.toMachineGuest()` functions.

Likely as part of `ToConsoleMachineConfig()` function from `internal/appconfig/machines.go`  file

PR #3174, opened 2024-01-10
Validate services have at least one port exposed

dangra:
### Change Summary

What and Why:

At least one service port must be defined for a service to be valid, check the invalid config earlier.  
image

How: Enforce the rule as a config validation rule.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/appconfig/testdata/validate-groups.toml
+++ b/internal/appconfig/testdata/validate-groups.toml
@@ -15,10 +15,16 @@ internal_port = 9999
 protocol = "udp"
 processes = ["vpn"]
 
+[[services.ports]]
+port = 80
+
 [[services]]
 internal_port = 1111
 protocol = "tcp"
 processes = ["vpn", "app", "foo"]
+
+[[services.ports]]
+port = 81
 
 [checks.listening]
 port = 8080
--- a/internal/appconfig/testdata/validate-services.toml
+++ b/internal/appconfig/testdata/validate-services.toml
@@ -1,0 +1,20 @@
+app = "foo"
+primary_region = "ord"
+
+[processes]
+app = ""
+foo = ""
+success = ""
+
+# Check:
+# * missing [[services.ports]]
+# * no processes specified
+[[services]]
+internal_port = 8080
+
+[[services]]
+internal_port = 8080
+processes = ["success"]
+
+[[services.ports]]
+port = 80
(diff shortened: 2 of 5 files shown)


matttpt (approved):

MaxSchaefer:
Hey @dangra 👋. seams like that this change was breaking for some of us.
- https://community.fly.io/t/service-must-expose-at-least-one-port-since-when/17668
- https://github.com/helpwave/services/actions/runs/7505977583/job/20436402897

dangra:
hi @MaxSchaefer, sorry to hear that. Not exactly something to deprecate, as a service with ports is invalid but on second thoughts releasing a flyctl version that warns instead of fail for a few releases was indeed a good idea. The damage is already done tho.

isker:
> The damage is already done tho.

The damage is not already done.  More damage is done until everyone with such an invalid configuration deploys for the first time following this PR, finds this PR, and fixes the problem in their config.

You can choose to stop the damage by reverting this change.

dangra:
@isker @MaxSchaefer something like https://github.com/superfly/flyctl/pull/3182 would have been better in your opinion?  My worry about warning is that people using automated deployments won't notice it until we turn it into a hard fail. 

The "missing service ports" has always been a bug, it gives the fake sense of having a service when in reality there isn't one when `[[services.ports]]` are not set.

isaacbowen:
fwiw, I experienced this as surprise, but not as damage. I'm pleased with the changes that were prompted as a result.

The fact that flyctl doesn't have a major semver number yet (1) keeps me mindful of the deploy logs (I _would_ have noticed a new warning), and (2) makes me feel like a non-destructively breaking change like this is okay.

This comment is about me and my experience; it is not a generalization or advice or an ask. :)

khuezy:
> The "missing service ports" has always been a bug, it gives the fake sense of having a service when in reality there isn't one unless `[[services.ports]]` are set.

Thanks for explaining. I understand/get that it's a bug on fly's side, but on the user's side, it's been a feature. It would have been nice to have a deprecation warning first, but the Band-Aid is off.
(Just my 2c, not a rant)

dangra:
https://github.com/superfly/flyctl/pull/3182 is live now since flyctl v0.1.143. Thanks everyone for the feedback.

grapereader:
I get this error even with an [http_service] section and a valid internal port. Is this intentional?

The docs suggest I shouldn't need a [[services]] section if I have an [http_service] section.

PR #3182, opened 2024-01-15
Warn instead of fail on service definitions without ports

dangra:
### Change Summary

What and Why: Hard failing on missing service ports was too aggressive and had alienated users that found their automated deploys suddenly failing without notice.    

How: Switch to a one Warning notice with a one month deadline. 

Related to: https://github.com/superfly/flyctl/pull/3174#issuecomment-1891090082

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/appconfig/validation.go
+++ b/internal/appconfig/validation.go
@@ -238,7 +238,8 @@ func (cfg *Config) validateServicesSection() (extraInfo string, err error) {
 			//      fix fly.toml configuration -- 2024-01-15
 			extraInfo += fmt.Sprintf(
 				"WARNING: Service must expose at least one port. Add a [[services.ports]] section to fly.toml; " +
-					"Check docs at https://fly.io/docs/reference/configuration/#services-ports \n ",
+					"Check docs at https://fly.io/docs/reference/configuration/#services-ports \n " +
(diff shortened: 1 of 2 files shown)


khuezy:
Hi, why is the external port required? Shouldn't a configuration with an internal port be sufficient? Eg my use case is an internal service (temporal worker and UI) I don't want to expose those services to the public nor do I want to specify an ad hoc public port. Those services are accessed via .internal and .fly cast.

benwaffle (approved):

dangra:
> Hi, why is the external port required?

Oh, it is not an "external port" but a "service port". If the *service* is _internal_ or _external_ depends whether the app has a public or private IP address assigned.

`appname.flycast` won't work unless there is a `[[services.ports]]` definition in fly.toml and a private ipv6 assigned (`fly ips allocate-v6 --private` 

`.internal` will always work, whether you add a `[[services]]` section or not. The connections to `.internal` uses the machine's 6pn network address to communicate directly; while the `.flycast` addresses are proxied by Fly-Proxy service but requires a port to listen for incoming connections that then are balanced and forwarded to the internal_port of the service.

The documentation about private networking has more info https://fly.io/docs/reference/private-networking/

PR #3184, opened 2024-01-17
Set isFirstDeploy if there are no launch machines

billyb2:
### Change Summary

What and Why:
Currently, if someone tries to deploy on an app with no managed machines (such as scaling down to 0), we'll just return an answer. IMO the correct answer is just to create a new machine.

Related to:
https://flyio.discourse.team/t/machines-apps-with-both-fly-launch-and-unmanaged-machines/4661/8

https://github.com/superfly/docs/pull/1297

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/flaps/flaps_machines.go
+++ b/flaps/flaps_machines.go
@@ -222,7 +222,7 @@ func (f *Client) ListActive(ctx context.Context) ([]*api.Machine, error) {
 	return machines, nil
 }
 
-// ListFlyAppsMachines returns apps that are part of the Fly apps platform.
+// ListFlyAppsMachines returns apps that are part of Fly Launch.
 // Destroyed machines and console machines are excluded.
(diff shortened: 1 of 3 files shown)


billyb2:
Actually before merging this, I need to add some kind of message to the user explaining the situation

billyb2:
> @billyb2 @catflydotio When you run `fly status` on an app with both managed and unmanaged Machines, you get a message about it: https://github.com/superfly/flyctl/blob/6d5058a4bc9359d19288fb9f066cd495d00c1d6e/internal/command/status/machines.go#L210C1-L213C1
> 
> That message should probably say "Machines that aren't managed by Fly Launch" rather than "Machines that aren't part of the Fly Apps Platform", and then the same message could be used when you run `fly deploy` and have unmanaged Machines?

I changed the message and added a warning :)

PR #3195, opened 2024-01-19
Add a check for the fly's DNS address being ready

billyb2:
### Change Summary

What and Why:
Ensuring that the DNS address is ready helps to mitigate users not being able to access their fly apps, despite flyctl saying it's "ready".

How:
We just do a simple DNS query to make sure the address is reachable.

Related to:

---

### Documentation
maybe fresh produce?
- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -70,8 +70,10 @@ func (md *machineDeployment) DeployMachinesApp(ctx context.Context) error {
 		}
 	}
 
-	if err := md.checkDNS(ctx); err != nil {
-		return err
+	if !md.skipDNSChecks {
+		if err := md.checkDNS(ctx); err != nil {
+			return err
+		}
 	}
 
 	return err
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -12,16 +12,19 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/miekg/dns"
 	"github.com/samber/lo"
 	"github.com/sourcegraph/conc/pool"
 	"github.com/superfly/flyctl/api"
+	"github.com/superfly/flyctl/client"
 	"github.com/superfly/flyctl/flaps"
 	"github.com/superfly/flyctl/helpers"
 	machcmd "github.com/superfly/flyctl/internal/command/machine"
 	"github.com/superfly/flyctl/internal/flyerr"
 	"github.com/superfly/flyctl/internal/machine"
 	"github.com/superfly/flyctl/internal/statuslogger"
+	"github.com/superfly/flyctl/iostreams"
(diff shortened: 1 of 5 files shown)


internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -962,3 +973,64 @@ func (md *machineDeployment) doSmokeChecks(ctx context.Context, lm machine.Leasa
...
 }
+
+func (md *machineDeployment) checkDNS(ctx context.Context) error {
+	client := client.FromContext(ctx).API()
+	ipAddrs, err := client.GetIPAddresses(ctx, md.appConfig.AppName)
+	if err != nil {
+		return err
+	}
+
+	if url := md.appConfig.URL(); url != nil && len(ipAddrs) > 0 {
```
benbjohnson, line 984: I typically try to avoid shadowing stdlib package names (e.g. `url`). A better name might be `u` (if you're feeling terse) or `appURL`.

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -962,3 +973,64 @@ func (md *machineDeployment) doSmokeChecks(ctx context.Context, lm machine.Leasa
...
+				ip := net.ParseIP(ipAddr.Address)
+				if ip.To4() != nil {
+					numIPv4 += 1
+				} else {
+					numIPv6 += 1
+				}
+			}
+
+			m.SetQuestion(fqdn, dns.TypeA)
+			answerv4, _, err := c.Exchange(m, "8.8.8.8:53")
```
benbjohnson, line 1015: It looks like we use `9.9.9.9` in `flyctl doctor`. It might be better to use a consistent DNS resolver? https://github.com/superfly/flyctl/blob/7cb1642ee5931c99292c69ad67eaa7fa3b1dbfda/internal/command/doctor/app_checks.go#L199

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -962,3 +973,64 @@ func (md *machineDeployment) doSmokeChecks(ctx context.Context, lm machine.Leasa
...
+			Dialer:       &net.Dialer{Timeout: time.Minute},
+			Timeout:      time.Minute,
+			DialTimeout:  time.Minute,
+			ReadTimeout:  time.Minute,
+			WriteTimeout: time.Minute,
+		}
+
+		b := backoff.NewExponentialBackOff()
+		b.InitialInterval = 1 * time.Second
+		b.MaxElapsedTime = 10 * time.Second
```
benbjohnson, line 999: Do we want to extend this to `30s`? It should hopefully happen well before that but it'd be nice to give a little more leeway in case DNS is slow.  /cc @michaeldwan

benbjohnson (changes requested):

internal/command/deploy/machines_deploymachinesapp.go
billyb2: yea 30s sounds good

benbjohnson (approved):

PR #3200, opened 2024-01-22
Use stringer instead of maintaining the switch statement

kzys:
golang.org/x/tools/cmd/stringer is Go team's official tool to implement String() for enum-like types.

### Change Summary

What and Why:

Use stringer to remove the switch statement.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


kzys:
@gwuah This actually changes the strings from snake_case to lowerCamelCase. I don't know how we use these traces right now. 

Does it make sense to keep them backward-compatible? It is technically possible with [-linecomment](https://pkg.go.dev/golang.org/x/tools/cmd/stringer).

dangra (approved):
💯 TIL!

flaps/flaps.go
```diff
@@ -203,9 +204,18 @@ waiting:
 	}
 }
 
+func snakeCase(s string) string {
+	pat := regexp.MustCompile("[A-Z]")
```
dangra, line 208: not sure if Go optimizes it somehow but this is better placed as toplevel var so it is compiled once.

kzys: I think it won't and I was lazy. Updated in the last revision.

PR #3202, opened 2024-01-23
From toml to toml

dangra:
### Change Summary

What and Why: While adding `gpus` to `[[vm]]` section, I found that [BurntSushi/toml](https://github.com/BurntSushi/toml) lib serializes zero values of integer fields even if `omitempty` tag is set for the struct field.

Sounds like a minor thing, but it is very awkward for fly.toml files created by `fly launch` whose `[[vm]]` section looks like:

```toml
[[vm]]
size = "shared-cpu-1x"
memory_mb  = 1024
gpus = 0
```

How: Unfortunately BurntSushi won't fix it and suggests using an integer pointer (`*int` instead of `int`), so the other alternative was to switch to https://github.com/pelletier/go-toml that works like we expect and claims to be more `encoding/json` alike.

As a big plus, this PR adds a test case to ensure reading a toml file and writing it back generates an equal configuration without missing sections nor new ones.

Related to: 
* https://github.com/superfly/flyctl/pull/3199
* https://github.com/pelletier/go-toml/pull/923
* https://github.com/pelletier/go-toml/pull/924

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


billyb2:
banger pr

billyb2 (approved):
banger pr 👍🏾

PR #3209, opened 2024-01-25
Don't check for private ipv6 address in DNS check

billyb2:
Dumb bug, this was incredibly stupid to overlook

### Change Summary

What and Why:

The way we were checking which IPs the DNS AAAA query would return also expected IPv6 addresses.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


dangra (approved):
🚢

PR #3213, opened 2024-01-30
Add new Tigris bucket options

jsierles:


internal/command/extensions/tigris/update.go
```diff
@@ -0,0 +1,126 @@
...
+			"name":          shadowName,
+			"endpoint":      endpoint,
+			"write_through": writeThrough,
+		}
+	}
+
+	_, err = gql.UpdateAddOn(ctx, client, addOn.Id, addOn.AddOnPlan.Id, []string{}, options)
+	if err != nil {
+		return
+	}
```
benbjohnson, line 123: @jsierles I tried passing the `addOn.ReadRegions` but it gave me an error. I swapped it to `[]string{}` and it worked. Do we use read regions for Tigris?

jsierles: Nope, Tigris buckets are global!

PR #3219, opened 2024-01-31
`launch`: Use appconfig.Compute for Plan guest information.

alichay:
### Change Summary

What and Why: The Launch Plan now uses `appconfig.Compute` instead of the disparate fields like `CPUs` or `MemoryMB`. We still support reading those fields from the UI, though, and throw them into the default process' Compute definition when modifying the plan.

This means that the UI could be changed to write the Compute field instead of these individual fields, which would allow defining multiple guests for different groups (although I'm not sure we want that), but importantly it would allow defining things like launching GPU machines.

How: When we create the plan, we fill in both the Compute field, and all the deprecated individual fields. When we read the Plan from the UI, if the Compute field is empty (meaning the UI has not transmitted it back to us), we'll reconstruct it by taking the Compute from the appConfig, and modifying the default group's Compute with the values the UI passed back to us.

This should be totally transparent to users, should make fly.toml import a little bit better, and should future-proof Launch for things to come :)

Related to: conversation with Daniel in #fly-launch

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -3,7 +3,6 @@ package launch
 import (
 	"context"
 	"fmt"
-	"slices"
 
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/client"
@@ -90,6 +89,8 @@ func (state *launchState) updateComputeFromDeprecatedGuestFields(ctx context.Con
 		// If the UI returns a compute field, then we don't need to do any forward-compat patching.
 		return nil
 	}
+	// Fallback for versions of the UI that don't support a Compute field in the Plan.
+
 	defer func() {
 		// Set the plan's compute field to the calculated compute field.
 		// This makes sure that code expecting a compute definition in the plan is able to find it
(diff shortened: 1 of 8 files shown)


internal/command/launch/launch.go
```diff
@@ -69,6 +73,59 @@ func (state *launchState) Launch(ctx context.Context) error {
...
+		//       The only real solution to this is updating the UI to return a Compute field.
+		switch {
+		case matchedComputeStrong != nil:
+			applyGuestToCompute(matchedComputeStrong, state.Plan.Guest())
+		case matchedComputeWeak != nil:
+			applyGuestToCompute(matchedComputeWeak, state.Plan.Guest())
+		default:
+			state.appConfig.Compute = append(state.appConfig.Compute, guestToCompute(state.Plan.Guest()))
+		}
+	}
```
alichay, line 124: I'm going to rewrite this. This is 100% "I am very tired" code

dangra (approved):

PR #3240, opened 2024-02-07
Remove App v2 migration tests

kzys:
Since we no longer have Nomad apps, we can remove these flaky tests.

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


test/preflight/apps_v2_integration_test.go
```diff
@@ -1,494 +0,0 @@
...
-	"testing"
-	"time"
-
-	"github.com/stretchr/testify/require"
-	"github.com/superfly/flyctl/api"
-	"github.com/superfly/flyctl/internal/appconfig"
-	"github.com/superfly/flyctl/test/preflight/testlib"
-)
-
-func TestAppsV2Example(t *testing.T) {
```
billyb2, line 23: I don't think this is a nomad related test, this just tests some apps v2 functionality. We could maybe rename it to TestAppExample though

billyb2: i think that goes for every test in this file

dangra:
I agree with @billyb2 ,  the tests are not migration related but can use a renaming now that AppsV2 is the only option.

test/preflight/apps_v2_integration_test.go
kzys: Oh they were removed by e80f21b6cffca9bc2db1e1b65409c146f0d1defc already

PR #3241, opened 2024-02-07
Agent auth improvements

btoews:
When using macaroons, flyctl's auth tokens change periodically as old tokens expire and new ones are issued. The agent doesn't pick up on these changes though.

This PR teaches the agent to periodically reload its tokens from the config file so long as the tokens didn't originally come from an environment variable or CLI flag. When auth tokens expire, the agent also tries to fetch new ones from web. If that update is successful, the agent writes its updates tokens back to the config file.

Eventually, it would be nice to have flyctl tell the agent what tokens to use so that `FLY_API_TOKENS=whatever flyctl ssh console`  would work more reliably.


api/tokens/tokens.go
```diff
@@ -1,7 +1,17 @@
 package tokens
```
btoews, line 1: The changes in this file are mostly things moving from `internal/command/macaroons.go`.

dangra (approved):

PR #3249, opened 2024-02-08
Laravel: Allow Octane-flavored Frankenphp

fideloper:
### Change Summary

What and Why:

For the Laravel template, we ignore the user's local `frankenphp` binary which may be built for their local dev machine's architecture. We'll be updating the `Dockerfile` to install the correct `frankenphp` binary in the remote builder.


dangra:
merge?

fideloper:
I'll merge this for now but there's more work to do to make Franken actually work!

PR #3263, opened 2024-02-09
Run go without buildpacks

fideloper:
### Change Summary

What and Why:

Buildpack builds for Golang are failing, with an odd error about linux and CPU architecture.

How:

Here we use out own Dockerfile setup instead of a build pack.

1. We attempt to get the correct Go version
2. We attempt to name the resulting binary someting sensible based on the project `go.mod`'s module name.
    - even tho Dockerfile shortcomings force us to `ln -s run-app the-actual-binary-name` 

Related to:

https://community.fly.io/t/stuck-flyctl-deploy/16612/9

---

### Documentation

- [ ] Fresh Produce: https://gist.github.com/fideloper/4b0da90f8ac3c4d6815d6ca262109f57
- [ ] docs update: https://github.com/superfly/docs/pull/1344


btoews (approved):

dangra:
@fideloper run `go mod tidy` to fix https://github.com/superfly/flyctl/actions/runs/7848014572/job/21418212546?pr=3263

fideloper:
If anyone else could run a quick test against a go app, that'd be useful too. This worked locally for me on a simple webs server.

This afternoon I'll run a few more tests as well.

scanner/go.go
```diff
@@ -1,19 +1,87 @@
...
+		return false
+	}
+
+	gomod = f
+
+	return true
+}
+
+func extractGoVersion(modfileParsed bool) string {
+	// todo: Can we get current latest stable via web request?
```
dangra, line 57: what about using `"1"` which according to [golang offiicial image](https://hub.docker.com/_/golang) always point to latest Go v1 release.

scanner/go.go
```diff
@@ -1,19 +1,87 @@
 package scanner
 
+import (
+	"golang.org/x/mod/modfile"
+	"os"
+	"path"
+)
+
+var gomod *modfile.File
```
dangra, line 9: honestly, I don't see the need for a global variable here 🤔

dangra (commented):
May I suggest generating a Dockerfile like this:

```
ARG GO_VERSION=1
FROM golang:${GO_VERSION}-alpine AS builder
WORKDIR /usr/src/app
COPY go.mod go.sum ./
RUN go mod download && go mod verify
COPY . .
RUN go build -v -o /run-app .

FROM alpine:latest
COPY --from=builder /run-app /usr/local/bin
CMD ["/usr/local/bin/run-app"]
```

Or If you think "alpine" is not a good option for some reason, even if it generates a 4x smaller image, then a debian based image looks like:

```
ARG GO_VERSION=1
FROM golang:${GO_VERSION} AS builder
WORKDIR /usr/src/app
COPY go.mod go.sum ./
RUN go mod download && go mod verify
COPY . .
RUN go build -v -o /run-app .

FROM debian:bookworm-slim
COPY --from=builder /run-app /usr/local/bin
CMD ["/usr/local/bin/run-app"]
```


The key takeaways are:
* $GO_BIN_NAME is not needed, it is "/run-app" no matter what.
* COPYing go.mod and go.sum and downloading mods uses caching layers better so faster builds
* alpine based builds are 20MBs vs bookworm 87MBs (and my Go app binary is 12MBs alone)

fideloper:
thanks! (totally forgot caching the go mod/sum files).

That all makes sense. 

Have you had issues with ipv6/dns in Alpine? For some reason I remember that being an issue on Fly sometimes. Definitely happy to use a smaller container.

fideloper:
Updated! Thanks, your feedback (and better sleep 👦👦) made this a ton simpler/better.

scanner/templates/go/Dockerfile
```diff
@@ -0,0 +1,15 @@
+ARG GO_VERSION=1
+FROM golang:${GO_VERSION} as builder
```
dangra, line 2: if the runner is alpine, the builder has to be alpine based too
```suggestion
FROM golang:${GO_VERSION}-alpine as builder
```

scanner/templates/go/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+COPY . .
+RUN go build -v -o /run-app .
+
+
+FROM alpine:latest
+
+COPY --from=builder /run-app /usr/local/bin/
+
```
dangra, line 14: nitpick: the extra empty lines doesn't add any value. It is good to separate FROM sections, but within a section it is noisy.

scanner/go.go
```diff
@@ -1,19 +1,51 @@
...
 func configureGo(sourceDir string, config *ScannerConfig) (*SourceInfo, error) {
-	if !checksPass(sourceDir, fileExists("go.mod", "Gopkg.lock")) {
+	if !checksPass(sourceDir, fileExists("go.mod")) {
 		return nil, nil
 	}
 
+	gomod, parseErr := parseModfile()
+
+	version := "1"
+	if parseErr == nil && len(gomod.Go.Version) > 0 {
```
dangra, line 17: nitpick: an empty string is usually compared as `gomod.Go.Version != ""` rather than on its length 🤷

dangra: on another note, parseErr is swallowed, no idea how other scanners handles this but not showing a warning with the error may lead to broken image builds  as go.mod isn't valid anyway.

dangra (approved):
:shipit:

fideloper:
Made one last quick tweak

ClayShentrup:
just curious, why not just fork the go buildpack and fix whatever you find problematic with it, rather than use a dockerfile?

https://www.youtube.com/watch?v=2Xorbt3TaP4

dangra:
@ClayShentrup: building from a Dockerfile tends to generate smaller images, more customizable and are faster to build.

PR #3264, opened 2024-02-09
Go 1.22

dangra:
### Change Summary

flyctl builds are fixed to version 1.21.0, it should be 1.21.7 but 1.22 is already here so let's use that.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


dangra:
Env is not ready for 1.22 yet (DeepSource and Brew lacks support for it)

PR #3266, opened 2024-02-09
Update jsFramework.go to include Astro

anniebabannie:
Added Astro framework detection

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/scanner/jsFramework.go
+++ b/scanner/jsFramework.go
@@ -215,7 +215,13 @@ func configureJsFramework(sourceDir string, config *ScannerConfig) (*SourceInfo,
 	// confusion for many when the framework detected is listed as "NodeJS"
 	// See flyapps/dockerfile-node for the actual framework detction.
 	// Also change PlatformMap in core.go if this list ever changes.
-	if deps["@adonisjs/core"] != nil {
+	if (deps["astro"] != nil) && (deps["@astrojs/node"] != nil) {
+		srcInfo.Family = "Astro with SSR"
+		srcInfo.Port = 4321
+	} else if deps["astro"] != nil {
+		srcInfo.Family = "Astro Static Site"
+		srcInfo.Port = 80
+	} else if deps["@adonisjs/core"] != nil {
 		srcInfo.Family = "AdonisJS"
 	} else if deps["gatsby"] != nil {
 		srcInfo.Family = "Gatsby"


scanner/jsFramework.go
```diff
@@ -217,6 +217,8 @@ func configureJsFramework(sourceDir string, config *ScannerConfig) (*SourceInfo,
 	// Also change PlatformMap in core.go if this list ever changes.
 	if deps["@adonisjs/core"] != nil {
 		srcInfo.Family = "AdonisJS"
+	} else if deps["astro"] != nil {
+		srcInfo.Family = "Astro"
```
rubys, line 221: If static, what port should it listen on?

anniebabannie: @rubys It will listen on port 80, I believe this is already set somewhere?

rubys: Looking a few lines down from this point:

https://github.com/superfly/flyctl/blob/4f62f27b182ca45f9c35a1c5c1945e866058c58a/scanner/jsFramework.go#L233-L236

My guess is that you will want to match what you did with vite, but only for static Astro?  And does Astro SSR use port 3000?  It looks like it might use 4321: https://docs.astro.build/en/reference/configuration-reference/#serverport

anniebabannie: Hmm, weird, I was able to launch an Astro app just fine with `fly launch` even without setting this – I guess this is for setting the port in `fly.toml` ?

And yes Astro SSR uses 4321, it's set in the Dockerfile generator. Does it need to be set here too?

rubys: >  Does it need to be set here too?

I think so.  The first question is: what is shown if you say `fly launch`, then say you want to tweak the settings?  The second question is what happens if you *do* tweak the settings?

image

The above should match what we think is appropriate for the app, and should the user change the value, that value should be respected.

The dockerfile generator predated this fly launch ui; which is why there is a need for dual maintenance.  Perhaps longer term the dockerfile generator should have a `--dry-run` option that returns a description of the app (likely in JSON).

anniebabannie: @rubys aahhh gotcha gotcha, ok, I pushed some changes that should fix this for both static and SSR versions of Astro.

Also do we have tests for our framework checks? wasn't sure where to find those if they exist

rubys: > Also do we have tests for our framework checks? wasn't sure where to find those if they exist

Not that I'm aware of :-(

rubys (approved):

PR #3278, opened 2024-02-13
track gpu and cpu kinds for deployments

gwuah:
### Change Summary
Allow us to group deployments by CPU and GPU kinds


billyb2 (approved):
banger pr

Issue #2557, opened 2023-07-12
fly scale count requires process group argument if "app" process missing


estephinson:
*Please only report specific issues with `flyctl` behavior*. Anything like a support request for your application should go to https://community.fly.io. More people watch that space and can help you faster!

**Describe the bug**

When running `fly scale count` with process groups, the [docs imply](https://fly.io/docs/apps/scale-count/#change-the-number-of-machines-in-a-process-group) that you shouldn't need to supply the `--process-group` argument if you're scaling by names. This doesn't seem to play ball if you haven't got a process with the name "app".

Looking at the code, the `groupName` parsed into `parseGroupCounts` is only used in the case where there is a single arg supplied, so the validation beforehand isn't required with args > 1. Also, if you supply the process group in the `<group>=<count>` format, I don't think it's needed either.

I may be missing some context, but I've read through the docs extensively to try and make sure it's not me being silly! 

My rough guesses based [on reading here](https://github.com/superfly/flyctl/blob/7981f99ff550f66def5bbd9374db3d413310954f/internal/command/scale/count.go#L58C1-L86C63)
- `parseGroupCounts` needs to be updated to parse the args differently (not just the number of args passed, but if a single arg then `=` indicates a process group is passed)
- The default value of `"app"` should only be checked and used if there are no process groups passed in the args

* Operating system: MacOS
* `fly version`: `fly v0.1.55 darwin/arm64 Commit: ee5296409507fa87dc9b0c1c8aee6b411172c81e-dirty BuildDate: 2023-07-12T15:19:59Z`

** Paste your `fly.toml`

```toml
[processes]
worker = "/blah -worker -billing-worker"
edge= "/blah -edge-api"
```

** Command output: **

```text
$ fly scale count worker=1 edge=2

Error: --process-group flag is required when no group named 'app' is defined
```

yet
```text
$ fly scale count worker=1 edge=2 --process-group worker
```

works fine

andie787:
Introduced by https://github.com/superfly/flyctl/pull/2210 which maybe doesn't take into account the ability to specify process groups without the `--process-group` option in the `fly scale count` command.

dangra:
closed by #3280

PR #3280, opened 2024-02-13
do not require --process-group on every case of fly scale count

dangra:
### Change Summary

What and Why: `fly scale count` requires --process-group when 'app' process group isn't present but it shouldn't in every case.

How: Deal with multi process group syntax and group name passed by group name

Related to: https://github.com/superfly/flyctl/issues/2557

I am testing this on a GPU app that have 3 different process group and none is named "app". That is why the groups resembles GPU kinds so closely. 

```
~
❯ fly scale count 4
Error: Attempting to scale unknown process groups [app] but valid names are [40gb 80gb l40s].
 Use `fly scale count COUNT --process-group=NAME`
 or multi group syntax `fly scale count NAME1=COUNT1 NAME2=COUNT2 ...`

~ took 2s
❯ fly scale count foo=4
Error: Attempting to scale unknown process groups [foo] but valid names are [40gb 80gb l40s].
 Use `fly scale count COUNT --process-group=NAME`
 or multi group syntax `fly scale count NAME1=COUNT1 NAME2=COUNT2 ...`

~
❯ fly scale count foo=4 bar=3
Error: Attempting to scale unknown process groups [foo bar] but valid names are [40gb 80gb l40s].
 Use `fly scale count COUNT --process-group=NAME`
 or multi group syntax `fly scale count NAME1=COUNT1 NAME2=COUNT2 ...`

~
❯ fly scale count foo=4 bar=3 l40s=5
Error: Attempting to scale unknown process groups [foo bar] but valid names are [40gb 80gb l40s].
 Use `fly scale count COUNT --process-group=NAME`
 or multi group syntax `fly scale count NAME1=COUNT1 NAME2=COUNT2 ...`
```

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/scale/count.go
+++ b/internal/command/scale/count.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/samber/lo"
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/api"
 	"github.com/superfly/flyctl/flaps"
@@ -14,6 +15,7 @@ import (
 	"github.com/superfly/flyctl/internal/command"
 	"github.com/superfly/flyctl/internal/flag"
 	"github.com/superfly/flyctl/internal/flag/completion"
+	"golang.org/x/exp/maps"
 )
 
 func newScaleCount() *cobra.Command {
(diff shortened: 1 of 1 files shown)


andie787:
I like this helpful output! 

The multi group syntax is in docs, but not in flyctl output or --help anywhere (I'll look at that next).

PR #3288, opened 2024-02-14
Sort extension secrets by key

jsierles:
Extension secrets should always show up in the same order. For now, that will be alphabetical.

Based on feedback from this article: https://benhoyt.com/writings/flyio-and-tigris/?moving

### DIFF
--- a/internal/command/extensions/core/core.go
+++ b/internal/command/extensions/core/core.go
@@ -455,15 +455,18 @@ func setSecretsFromExtension(ctx context.Context, app *gql.AppData, extension *E
 		setSecrets = false
 	}
 
+	keys := lo.Keys(secrets)
+	sort.Strings(keys)
+
 	if setSecrets {
 		extension.SetsSecrets = true
 		fmt.Fprintf(io.Out, "Setting the following secrets on %s:\n", app.Name)
 		input := gql.SetSecretsInput{
 			AppId: app.Id,
 		}
-		for key, value := range secrets {
-			input.Secrets = append(input.Secrets, gql.SecretInput{Key: key, Value: value.(string)})
-			fmt.Println(key)
+		for _, key := range keys {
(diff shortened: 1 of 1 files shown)


internal/command/extensions/core/core.go
```diff
@@ -472,12 +473,19 @@ func setSecretsFromExtension(ctx context.Context, app *gql.AppData, extension *E
...
+		var keys []string
+
+		for key := range secrets {
+			keys = append(keys, key)
 		}
-	}
+		sort.Strings(keys)
 
+		for _, key := range keys {
+			fmt.Println(key + ": " + secrets[key].(string))
```
dangra, line 485: should it use `fmt.Fprintf(io.Out, "%s: %s\n", key, secrets[key].(string))` ?

internal/command/extensions/core/core.go
```diff
@@ -472,12 +473,19 @@ func setSecretsFromExtension(ctx context.Context, app *gql.AppData, extension *E
...
 	} else {
-		fmt.Fprintf(io.Out, "Set one or more of the following secrets on your target app.\n")
-		for key, value := range secrets {
-			fmt.Println(key + ": " + value.(string))
+		fmt.Fprintf(io.Out, "Set the following secrets on your target app.\n")
+		var keys []string
+
+		for key := range secrets {
+			keys = append(keys, key)
 		}
```
dangra, line 481: you can replace all of that with `keys := lo.Keys(secrets)`

internal/command/extensions/core/core.go
jsierles: 👍

internal/command/extensions/core/core.go
jsierles: 🧇

dangra (approved):

PR #3289, opened 2024-02-14
Include whether or not we're using GPUs in command/stats

billyb2:
### Change Summary

What and Why:
WE're now including whether or not a command is using a GPU in metrics, so we can more easily do some useful queries.

How:
We just check fly.toml to see if it contains gpu_kind anywhere. This should work for every command, but it'll be especially useful for deploys

Related to:

https://github.com/superfly/flyctl-metrics/pull/23

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


internal/command/command.go
```diff
@@ -694,3 +696,22 @@ func ChangeWorkingDirectory(ctx context.Context, wd string) (context.Context, er
...
+func setUsingGPU(ctx context.Context) (context.Context, error) {
+	appConfig := appconfig.ConfigFromContext(ctx)
+	if appConfig == nil {
+		instrument.UsingGPU = false
+		return ctx, nil
+	}
+
+	usingGPU := false
+	for _, compute := range appConfig.Compute {
+		if compute != nil && compute.GPUKind != "" {
```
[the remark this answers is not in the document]
dangra: alternatively you can:

```go
instrument.UserGPU = lo.SomeBy(appConfig.Compute, func(x *appconfig.Compute) bool {
  return x != nil && x.MachineGuest != nil && x.MachineGuest.GPUKind != ""
})
```

dangra (changes requested):

internal/cli/cli.go
```diff
@@ -86,7 +86,7 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 
 	cmd, err = cmd.ExecuteContextC(ctx)
 
-	if err != nil {
+	if cmd != nil {
```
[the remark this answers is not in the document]
billyb2: And we want to always show that a command finish, even if it errored. We should probably incldue whether or not the command finishing was an error or not in command/stats though

dangra (approved):

internal/command/command.go
```diff
@@ -694,3 +697,17 @@ func ChangeWorkingDirectory(ctx context.Context, wd string) (context.Context, er
...
+func setUsingGPU(ctx context.Context) (context.Context, error) {
+	appConfig := appconfig.ConfigFromContext(ctx)
+	if appConfig == nil {
+		instrument.UsingGPU = false
+		return ctx, nil
+	}
+
+	instrument.UsingGPU = lo.SomeBy(appConfig.Compute, func(x *appconfig.Compute) bool {
+		return x != nil && x.MachineGuest != nil && x.MachineGuest.GPUKind != ""
+	})
```
dangra, line 710: ignore me but had to suggest a simplification nitpick:

```suggestion
	if appConfig != nil {
		instrument.UsingGPU = lo.SomeBy(appConfig.Compute, func(x *appconfig.Compute) bool {
			return x != nil && x.MachineGuest != nil && x.MachineGuest.GPUKind != ""
		})
	}
```

internal/command/command.go
```diff
@@ -694,3 +697,18 @@ func ChangeWorkingDirectory(ctx context.Context, wd string) (context.Context, er
...
+	appConfig := appconfig.ConfigFromContext(ctx)
+	if appConfig != nil {
+		instrument.UsingGPU = lo.SomeBy(appConfig.Compute, func(x *appconfig.Compute) bool {
+			return x != nil && x.MachineGuest != nil && x.MachineGuest.GPUKind != ""
+		})
+	}
+
+	instrument.UsingGPU = lo.SomeBy(appConfig.Compute, func(x *appconfig.Compute) bool {
+		return x != nil && x.MachineGuest != nil && x.MachineGuest.GPUKind != ""
+	})
```
dangra, line 711: remove this

billyb2: forgot to stage removing this :(

PR #3297, opened 2024-02-15
Cordon and gracefully shutdown old machines during blue-green deploy

pborzenkov:
If flyctl destroys the old machines immediatelly after uncordoning the
new ones, fly-proxy doesn't have any time to learn about the new machines
and has nowhere to send the requests.

So this patch modifies the procedure to first cordon the old machines.

Even with this in place, we need to gracefully shutdown the old machines
in order to let the app properly terminate existing connections by handling
SIGTERM.

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -143,6 +143,11 @@ var CommonFlags = flag.Set{
 		Name:        "process-groups",
 		Description: "Deploy to machines only in these process groups",
 	},
+	flag.String{
+		Name:        "signal",
+		Shorthand:   "s",
+		Description: "Signal to stop the machine with for bluegreen strategy (default: SIGINT)",
+	},
 }
 
 func New() (cmd *cobra.Command) {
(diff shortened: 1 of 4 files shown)


pborzenkov:
Preflight fail seems to be unrelated to this PR.

pborzenkov:
> * We now have a new failure mode where blue-machines are cordoned but all the green machines have been destroyed by the rollback logic.

True. `CordonBlueMachines` and `StopBlueMachines` will return an error (and lead to a rollback) only in case the deployment is manually aborted (the rest of the errors are just logged and never returned), but we should probably handle this case.

> This adds a static 20 seconds to the deployment time, which is the trade off we are making for reliability, but I wonder if there was a way we could proactively tell if the proxy is aware of a machine.

I've thought about this. There are a couple of ways:
  - query metrics: might be even slower as scrape duration is something like every 15 seconds
  - add some API to proxy and query it. But, we will have to query *all* edge proxies, not just one.

pborzenkov:
> * The app will have machines running alright, but it will get no traffic.

Yeah, ideally, the rollback procedure should be the same as the deployment but with `green` and `blue` machines swapped. It has to cordon the newly deployed machines (as they might already be receiving traffic), stop them (to gracefully terminate the connections) and only after that destroy them. I think it should be possible to abstract these steps into a function the can be used during deploy and rollback so we don't have to duplicate the code.

pborzenkov:
Ok, I made an attempt to implement fully functional rollback, it's proving to be more complicated than I though with the current code :) So I'm gonna merge this one as is and leave proper rollback for later unless there are objections.

PR #3312, opened 2024-02-20
[golang] Use a base image that sets CGO_ENABLED=1 

fideloper:
[golang] Use a base image that sets CGO_ENABLED=1 

### Change Summary

What and Why:

How:

Related to:

Fly discourse: /t/non-buildpack-go-behavior-differences/4975/8

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/scanner/templates/go/Dockerfile
+++ b/scanner/templates/go/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /usr/src/app
 COPY go.mod go.sum ./
 RUN go mod download && go mod verify
 COPY . .
-RUN CGO_ENABLED=1 go build -v -o /run-app .
+RUN go build -v -o /run-app .
 
 
 FROM debian:bookworm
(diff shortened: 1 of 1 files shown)


scanner/templates/go/Dockerfile
```diff
@@ -1,14 +1,14 @@
 ARG GO_VERSION=1
-FROM golang:${GO_VERSION}-alpine as builder
+FROM golang:${GO_VERSION}-bookworm as builder
 
 WORKDIR /usr/src/app
 COPY go.mod go.sum ./
 RUN go mod download && go mod verify
 COPY . .
-RUN go build -v -o /run-app .
+RUN CGO_ENABLED=1 go build -v -o /run-app .
```
dangra, line 8: CGO_ENABLED=1 is the default for non-alpine images

fideloper: oh, ironically I put that there just so I could copy/paste it for somewhere else! bah - I'll keep it off.

uniphil:
Really small thing but figured I'd note it: for golang apps that make TLS client connections, this change makes a difference.

- Alpine: just works
- Debian: requires a `RUN apt-get update && apt-get install -y ca-certificates` to be added

Easy enough to figure out, but it's a small stumbling block, maybe especially for people who aren't authoring Dockerfiles frequently and hoping the automatic template will just work for them.

PR #3337, opened 2024-03-01
command to save kubeconfig to disk

senyosimpson:
### Change Summary

What and Why:

Adds a command to save a FKS kubeconfig to disk. Right now, you only get it on creation and if you lose it/delete it, it's game over.

How:

Fetches the kubeconfig from the API

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/extensions/kubernetes/kubeconfig.go
+++ b/internal/command/extensions/kubernetes/kubeconfig.go
@@ -1,0 +1,53 @@
+package kubernetes
+
+import (
+	"context"
+	"fmt"
+	"os"
+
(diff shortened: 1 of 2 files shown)


senyosimpson:
@jsierles this doesn't actually work yet because i need a way to derive the extension name from the cluster name. it's non-obvious how to do that because the cluster name has the format `fks-<org name>-<ext name>` and its not easy to know where the org name ends and the ext name starts

jsierles (approved):

PR #3350, opened 2024-03-06
Add Sentry extension option to `fly launch`

andreolvr:
Adds an option to add the Sentry extension during `fly launch`. It is opt-in, so by default, the extension is disabled, but can be enabled on the Launch web UI if the user chooses to tweak the default configuration.


billyb2 (approved):
banger pr

PR #3351, opened 2024-03-06
Add experimental `kernel_args` field in fly.toml.

ndarilek:
### Change Summary
Add a `kernel_args` field to the `[experimental]` section in fly.toml.

This is useful, for instance, when debugging an issue that requires bumping init's log level but without having to update the machine to add it in later.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x ] n/a

### DIFF
--- a/internal/appconfig/config.go
+++ b/internal/appconfig/config.go
@@ -27,7 +27,7 @@ func NewConfig() *Config {
 }
 
 // Config wraps the properties of app configuration.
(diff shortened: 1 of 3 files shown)


dangra:
@ndarilek there is no need for this, `appconfig.Compute` embeds machine's `Guest` which has the `KernelArgs` field already exposed as:

```
[[vm]]
kernel_args = ["foo", "bar"]
```

ndarilek:
Thanks a bunch, missed that!

PR #3295, opened 2024-02-15
[bluegreen] detect & delete zombie machines from failed deployments

gwuah:
### Change Summary
Supersedes https://github.com/superfly/flyctl/pull/2836

### Documentation

- [x] Fresh Produce
- [x] In superfly/docs, or asked for help from docs team
- [x] n/a

PR #3360, opened 2024-03-11
Disable zombie reaping for blue/green deployments

benbjohnson:
### Change Summary

What and Why: Previously, we attempted to reap old machines from partial blue/green deployments, however, the logic was removing all newer machines which isn't always what we want. It can also remove all previously deployed machines if any machines are missing a `fly_bluegreen_deployment_tag`.

How: I'm disabling the reaping functionality until we can go through it more thoroughly.

Related to: https://github.com/superfly/flyctl/pull/3295

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/deploy/strategy_bluegreen.go
+++ b/internal/command/deploy/strategy_bluegreen.go
@@ -602,12 +602,14 @@ func (bg *blueGreen) Deploy(ctx context.Context) error {
 		return ErrOrgLimit
 	}
 
-	fmt.Fprintf(bg.io.ErrOut, "\nCleanup Previous Deployment\n")
+	/*
+		fmt.Fprintf(bg.io.ErrOut, "\nCleanup Previous Deployment\n")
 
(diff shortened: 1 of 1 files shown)


internal/command/deploy/strategy_bluegreen.go
```diff
@@ -734,7 +736,7 @@ func getZombies(ids map[string]bool) (map[string]bool, error) {
 
 	sort.Ints(numbers)
 
-	delete(ids, fmt.Sprint(numbers[0]))
+	delete(ids, fmt.Sprint(numbers[len(numbers)-1]))
```
benbjohnson, line 739: Previously, it would only retain the oldest deployment. However, there's two cases where this is a problem:

1. The deployment finished except for a partial blue destroy step. For example, we could end up with only one machine in the oldest deployment.

2. The machine list contains a machine without the `fly_bluegreen_deployment_tag` so the code defaults it to -1 and it's the only machine that will be retained.

PR #3365, opened 2024-03-12
`fly launch`: support launching Supabase databases

alichay:
### Change Summary

What and Why: Adds support for launching Supabase databases to `fly launch`. They aren't currently the default, but this allows future updates to the Launch UI to provide Supabase as an option for Postgres databases.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a - Self documenting, as this is currently only exposed via UI.

### DIFF
--- a/internal/command/launch/state.go
+++ b/internal/command/launch/state.go
@@ -105,7 +105,7 @@ func (state *launchState) PlanSummary(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	postgresStr, err := describePostgresPlan(ctx, state.Plan.Postgres, org)
+	postgresStr, err := describePostgresPlan(state.Plan)
 	if err != nil {
 		return "", err
 	}
--- a/internal/command/launch/state.go
+++ b/internal/command/launch/state.go
@@ -2,13 +2,19 @@ package launch
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/samber/lo"
 	fly "github.com/superfly/fly-go"
+	"github.com/superfly/flyctl/gql"
+	extensions_core "github.com/superfly/flyctl/internal/command/extensions/core"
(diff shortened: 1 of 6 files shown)


internal/command/launch/state.go
```diff
@@ -157,3 +163,76 @@ func (state *launchState) PlanSummary(ctx context.Context) (string, error) {
...
+
+	io := iostreams.FromContext(ctx)
+	noConfirm := !io.IsInteractive() || flag.GetBool(ctx, "now")
+
+	org, err := state.Org(ctx)
+	if err != nil {
+		return err
+	}
+
+	validateSupabase := func() error {
```
kzys, line 181: Could it be a top-level function, so that you can write some tests?

kzys (approved):
Looks good to me. Some nits.

internal/command/launch/state.go
alichay: not sure how easy/viable to test this would be!

essentially the only logic here that isn't prompting to continue is
```go
		response, err := gql.ListAddOns(ctx, client, "supabase")
		if err != nil {
			return fmt.Errorf("failed to list Supabase databases: %w", err)
		}

		// TODO: We'd like to be able to query the user's plan to see if they're on a paid plan.
		//       For now, we'll just nag when they create their second database, every time.

		if len(response.AddOns.Nodes) != 1 {
			// If we're at zero databases, we're within the free plan.
			// If we're at >=2 databases, we know we're on a paid plan.
			// It's only 1 existing database where we need to validate the plan.
			return nil
		}
```
and the test would have to make a synthetic `ListAddOnsResponse` anyway, so the only thing being tested would be `if len(response.AddOns.Nodes) != 1`.

Not sure if that's worth restructuring the code (at this point)
Definitely worth keeping in mind though if we get a richer API that can tell us the plan they're on!

alichay:
the failing test is fixed in #3401, so it doesn't matter here c:

PR #3367, opened 2024-03-13
Remove AppCompact.ImageDetails

benbjohnson:
### Change Summary

What and Why: The `image_details` field in the GraphQL call can be slow if it needs to resolve from the underlying registry. We also don't use the field in `flyctl`.

How: Upgrade `fly-go` to use https://github.com/superfly/fly-go/pull/16

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


michaeldwan (approved):
lgtm

PR #3369, opened 2024-03-14
Build overlaybd images

benwaffle:
Enable with

```
[experimental]
lazy_load_images = true
```


internal/build/imgsrc/dockerfile_builder.go
```diff
@@ -288,11 +293,94 @@ func (*dockerfileBuilder) Run(ctx context.Context, dockerFactory *dockerClientFa
...
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", fmt.Sprintf("flyctl/%s", buildinfo.Version().String()))
+	req.SetBasicAuth(appName, config.Tokens(ctx).Docker())
+
+	res, err := docker.HTTPClient().Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to post to /buildOverlaybdImage: %w", err)
+	}
+	defer res.Body.Close()
```
matttpt, line 365: `//skipcq: GO-S2307` if you'd like to silence DeepSource here

internal/build/imgsrc/dockerfile_builder.go
```diff
@@ -288,11 +293,94 @@ func (*dockerfileBuilder) Run(ctx context.Context, dockerFactory *dockerClientFa
...
+		return nil, fmt.Errorf("building lazy image returned status %d: %s", res.StatusCode, buf.String())
+	}
+	hash := buf.String()
+
+	terminal.Info("Lazy-loading image built successfully!")
+
+	return &DeploymentImage{
+		ID:   hash,
+		Tag:  repo + ":" + version + "-obd",
+		Size: 0,
```
matttpt, line 380: Would it make sense to set the `Size` to that of the original image since (I think) it's printed to output in a few places? (I suspect the original image size is still pretty close to the overlaybd image, though I could be wrong?)

internal/build/imgsrc/dockerfile_builder.go
benwaffle: thanks!

internal/build/imgsrc/dockerfile_builder.go
benwaffle: Oh, forgot about this. The image sizes can vary by over 30%. I'd rather have a 0 here than a closer wrong number, to make it easier to detect issues. 

I can still deploy an overlaybd image without setting the size here, so I'm inclined not to worry about it at the moment.

PR #3373, opened 2024-03-15
change laravel scanner to install and invoke the dockerfile generator via compose

rubys:


scanner/laravel.go
```diff
@@ -27,8 +25,6 @@ func configureLaravel(sourceDir string, config *ScannerConfig) (*SourceInfo, err
 		return nil, nil
 	}
 
-	files := templates("templates/laravel")
```
KTanAug21, line 30: @rubys The `templates/laravel` directory contain the `/.fly` directory. This directory contains default( _like cache.sh that configures the app's cache config file, etc_ ), and [custom startup scripts](https://fly.io/docs/laravel/the-basics/customizing-deployments/#startup-scripts) alike that the user can create, that are run by entrypoint.sh. 

I'm not sure if this should be generated by flyctl or by the dockerfile-laravel command

KTanAug21: I think that yes, it should be generated from the dockerfile-laravel command, I'ma try to list why in [this issue](https://github.com/orgs/fly-apps/projects/6?pane=issue&itemId=56706705).

Johannes-Werbrouck (approved):

KTanAug21 (approved):
Looks great @rubys ! Thanks for integrating the dockerfile-laravel with flyctl!

PR #3383, opened 2024-03-19
Adding the user who created a token to the flyctl token list command

moss-fly:
### Change Summary

What and Why:
Added a `created by` column to the `fly tokens list` app and org scope so someone can figure out who created the tokens in an org if they ever need to revoke them. This info is currently exposed in `ui-ex` but the UX of going through all org/app tokens in an org an revoking them is pretty painful.

How:
Changed the user field in `LimitedAccessTokens` to public

Related to:
https://github.com/superfly/fly-go/pull/25
https://github.com/superfly/web/pull/2909
https://flyio.slack.com/archives/C035ZMAM206/p1710767524505839?thread_ts=1710737001.723769&cid=C035ZMAM206

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.18.2
 	github.com/stretchr/testify v1.9.0
-	github.com/superfly/fly-go v0.1.2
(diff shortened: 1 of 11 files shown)


btoews (approved):
As the test failures' 18 line comments subtly hint at, you'll have to update the GraphQL query too 😆 Otherwise, LGTM

PR #3387, opened 2024-03-20
Make the DNS checks warn, rather than fail the deploy

billyb2:
### Change Summary

What and Why:
We sometimes fail deploys unnecessarily when DNS checks fail. It's better to just warn the user that there may be an issue, so that they can continue debugging if need be, rather than failing the deployment completely.

### Documentation

- [ ] Fresh Produce
- [x] n/a


matthewlehner (approved):
Looks good to me

PR #3405, opened 2024-03-26
Allow choosing a leader that isn't in the primary region

billyb2:
### Change Summary

What and Why:
john bumped into an app who's leader is in a primary region, but the other machines are in other regions. In this case, it's better to be able to failover to another region, then to not failover at all.

How:
We allow selecting a leader from any machine, but we prefer machines that are in the primary region.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/postgres/failover.go
+++ b/internal/command/postgres/failover.go
@@ -152,7 +152,8 @@ func flexFailover(ctx context.Context, machines []*fly.Machine, app *fly.AppComp
 	fmt.Fprintf(io.Out, "Performing a failover\n")
 
 	primaryRegion := ""
-	candidates := make([]*fly.Machine, 0)
+	primaryCandidates := make([]*fly.Machine, 0)
+	secondaryCandidates := make([]*fly.Machine, 0)
 
 	for _, machine := range machines {
 		machinePrimaryRegion, ok := machine.Config.Env["PRIMARY_REGION"]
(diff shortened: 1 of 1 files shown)


internal/command/postgres/failover.go
```diff
@@ -40,6 +40,17 @@ func newFailover() *cobra.Command {
 		cmd,
 		flag.App(),
 		flag.AppConfig(),
+		flag.Bool{
+			Name:        "force",
+			Description: "Force a failover even if we can't connect to the active leader",
+			Default:     false,
+			Shorthand:   "F",
```
benbjohnson, line 47: A capital `F` seems a little unusual. I'd suggest either using `-f` or if that's already taken then just skip the shorthand. I think it's fine for people to write `--force`.

internal/command/postgres/failover.go
```diff
@@ -339,10 +360,19 @@ func handleFlexFailoverFail(ctx context.Context, machines []*fly.Machine) (err e
...
 
-	for _, machine := range machinesWithinPrimaryRegion {
+	// We should go for the primary canddiates first, but the secondary candidates are also valid
+	var candidates []*fly.Machine
+	switch allowSecondaryRegion {
+	case true:
+		candidates = append(primaryCandidates, secondaryCandidates...)
+	case false:
+		candidates = primaryCandidates
+	}
```
benbjohnson, line 373: `switch` on a boolean seems unusual. Why not just use an `if`/`else`?

benbjohnson (approved):

internal/command/postgres/failover.go
billyb2: genuinely don't know why i wrote it like this. if else is a few less lines so i'll switch it to that

internal/command/postgres/failover.go
billyb2: i'll just skip the shorthand :)

PR #3408, opened 2024-03-27
Run the laravel dockerfile generator even if a Dockerfile already exists

rubys:
The --skip option will be passed to ensure that the Dockerfile isn't overwritten

Also don't fail if the generator can't be installed and a Dockerfile exists.

### DIFF
--- a/scanner/laravel.go
+++ b/scanner/laravel.go
@@ -132,10 +132,8 @@ func LaravelCallback(appName string, srcInfo *SourceInfo, plan *plan.LaunchPlan)
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 
-		if err := cmd.Run(); err != nil {
-			if !dockerfileExists {
-				return fmt.Errorf("Dockerfile doesn't exist and failed to install fly-apps/dockerfile-laravel: %w", err)
-			}
+		if err := cmd.Run(); err != nil && !dockerfileExists {
+			return fmt.Errorf("Dockerfile doesn't exist and failed to install fly-apps/dockerfile-laravel: %w", err)
 		}
 	}
 
--- a/scanner/laravel.go
+++ b/scanner/laravel.go
@@ -96,54 +96,61 @@ func LaravelCallback(appName string, srcInfo *SourceInfo, plan *plan.LaunchPlan)
 	}
 
 	// generate Dockerfile if it doesn't already exist
+	dockerfileExists := true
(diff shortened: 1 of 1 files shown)


scanner/laravel.go
```diff
@@ -96,56 +96,63 @@ func LaravelCallback(appName string, srcInfo *SourceInfo, plan *plan.LaunchPlan)
...
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 
 		if err := cmd.Run(); err != nil {
-			return fmt.Errorf("failed to generate Dockerfile: %w", err)
+			if !dockerfileExists {
+				return fmt.Errorf("Dockerfile doesn't exist and failed to install fly-apps/dockerfile-laravel: %w", err)
+			}
 		}
 	}
```
benbjohnson, line 140: Not terribly important but you could merge these two `if` statements into one:

```go
if err := cmd.Run(); err != nil && !dockerfileExists {
    ...
}
```

scanner/laravel.go
```diff
@@ -96,56 +96,63 @@ func LaravelCallback(appName string, srcInfo *SourceInfo, plan *plan.LaunchPlan)
...
 
 	// generate Dockerfile if it doesn't already exist
+	dockerfileExists := true
 	_, err = os.Stat("Dockerfile")
 	if errors.Is(err, fs.ErrNotExist) {
+		dockerfileExists = false
+	}
```
benbjohnson, line 103: Do you want to handle other types of errors here (e.g. permissions errors)? Or just delegate that to the Dockerfile generator later?

benbjohnson (approved):

KTanAug21 (approved):
Tested! And it works wonders--got to deploy an app with an existing Dockerfile  that still needed the .fly directory( for its caches.sh script )!

scanner/laravel.go
rubys: Fixed.  Thanks!

scanner/laravel.go
rubys: I'm not that worried about that case.  Delegated to even later.... If the Dockerfile exits and can't be read, the generator will skip over it and deploy will later fail, and to be honest I'm OK with that... it is no different than if your source files were unreadable.

Put another way, the biggest thing I'm concerned about is us not stomping over existing Dockerfiles.  If you provide one, and it is broken, then that's your problem.  We will provide you with the tools to regenerate a new one, but how you proceed at that point is entirely up to you.

PR #3409, opened 2024-03-27
Switch to using Google DNS

billyb2:
### Change Summary

What and Why:
9.9.9.9 has been pretty inconsistent, which isn't good for our CLI.

How:
Switch all references to 9.9.9.9 to 8.8.8.8

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


matttpt (approved):

smorimoto:
I think the Cloudflare one is better for performance and privacy. Both Cloudflare and Google keep logs around 24 hours, but Google keeps some logs permanently.

https://www.dnsperf.com/#!dns-resolvers

PR #3411, opened 2024-03-28
Have fly launch stop parsing for arguments at a double dash

rubys:
The purpose of this is to pass the remainder of the arguments to the language or framework specific scanner to do with as it wishes. To start with, the framework scanners that include dockerfile generators will pass the remaining arguments onto their respective dockerfile generator.  Other scanners may find other uses.

Note that the chosen argument scanner, cobra, does not support this at this time so the arguments are removed and retained before being passed to cobra.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -8,6 +8,7 @@ import (
 	"html/template"
 	"os"
 	"runtime/debug"
+	"slices"
 	"strings"
 	"time"
 
@@ -18,6 +19,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/superfly/fly-go/flaps"
 	"github.com/superfly/flyctl/internal/env"
+	"github.com/superfly/flyctl/internal/flag"
 	"github.com/superfly/flyctl/internal/flag/flagnames"
 	"github.com/superfly/flyctl/internal/flyerr"
 	"github.com/superfly/flyctl/internal/metrics"
@@ -71,6 +73,19 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 	cmd := root.New()
 	cmd.SetOut(io.Out)
 	cmd.SetErr(io.ErrOut)
+
+	// Special case for the launch command, support `flyctl launch args -- [subargs]`
+	// Where the arguments after `--` are passed to the scanner/dockerfile generator.
+	// This isn't supported natively by cobra, so we have to manually split the args
+	// See: https://github.com/spf13/cobra/issues/739
+	if len(args) > 0 && args[0] == "launch" {
+		index := slices.Index(args, "--")
+		if index >= 0 {
+			ctx = flag.WithExtraArgs(ctx, args[index+1:])
+			args = args[0:index]
+		}
+	}
+
 	cmd.SetArgs(args)
 	cmd.SilenceErrors = true
 
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -82,7 +82,7 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 		index := slices.Index(args, "--")
 		if index >= 0 {
 			ctx = flag.WithExtraArgs(ctx, args[index+1:])
-			args = args[0:index]
+			args = args[:index]
 		}
 	}
 
--- a/internal/command/launch/launch_frameworks.go
+++ b/internal/command/launch/launch_frameworks.go
@@ -128,7 +128,7 @@ func (state *launchState) scannerRunCallback(ctx context.Context) error {
 		return nil
 	}
 
-	err := state.sourceInfo.Callback(state.Plan.AppName, state.sourceInfo, state.Plan)
+	err := state.sourceInfo.Callback(state.Plan.AppName, state.sourceInfo, state.Plan, flag.ExtraArgsFromContext(ctx))
 
 	if state.sourceInfo.MergeConfig != nil {
 		if err == nil {
--- a/internal/flag/flag.go
+++ b/internal/flag/flag.go
@@ -3,6 +3,7 @@ package flag
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"time"
 
@@ -10,6 +11,8 @@ import (
 	"github.com/superfly/flyctl/internal/flag/completion"
 	"github.com/superfly/flyctl/internal/flag/flagnames"
 )
+
+type extraArgsContextKey struct{}
 
 func makeAlias[T any](template T, name string) T {
 	var ret T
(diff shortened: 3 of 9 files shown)


internal/cli/cli.go
```diff
@@ -71,6 +73,19 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
...
+
+	// Special case for the launch command, support `flyctl launch args -- [subargs]`
+	// Where the arguments after `--` are passed to the scanner/dockerfile generator.
+	// This isn't supported natively by cobra, so we have to manually split the args
+	// See: https://github.com/spf13/cobra/issues/739
+	if len(args) > 0 && args[0] == "launch" {
+		index := slices.Index(args, "--")
+		if index >= 0 {
+			ctx = flag.WithExtraArgs(ctx, args[index+1:])
+			args = args[0:index]
```
benbjohnson, line 85: Minor nit: omitting the `0` is typically more common (e.g. `args[:index]`)

internal/flag/flag.go
```diff
@@ -588,3 +591,19 @@ Repeat for each volume in order (comma-separated lists not accepted)
...
+	return context.WithValue(ctx, extraArgsContextKey{}, extraArgs)
+}
+
+// ExtraArgsFromContext returns the extraArgs ctx carries.
+func ExtraArgsFromContext(ctx context.Context) []string {
+	if extraArgs, ok := ctx.Value(extraArgsContextKey{}).([]string); ok {
+		return extraArgs
+	}
+
+	fmt.Printf("unable to get extra args from context\n")
```
benbjohnson, line 606: It seems like error handling should be moved to the caller. Either change the function signature to return a bool for existence:

```go
func ExtraArgsFromContext(ctx context.Context) ([]string, bool)
```

or use `nil` to indicate non-existent extra args and a zero-length `[]string` to indicate empty extra args.

benbjohnson (approved):
Overall I think this looks good. Error handling in the context helper is my only concern.

internal/command/launch/launch_frameworks.go
```diff
@@ -128,7 +128,7 @@ func (state *launchState) scannerRunCallback(ctx context.Context) error {
 		return nil
 	}
 
-	err := state.sourceInfo.Callback(state.Plan.AppName, state.sourceInfo, state.Plan)
+	err := state.sourceInfo.Callback(state.Plan.AppName, state.sourceInfo, state.Plan, flag.ExtraArgsFromContext(ctx))
```
KTanAug21, line 131: Quick question! 
This new arg is passed to a framework-scanner callback right? 
Right now I can see that only 4 framework scanners' callbacks where updated to receive this new argument. Will other scanners need their callback functions to be updated to receive the new arg?

internal/flag/flag.go
rubys: First, I merely copied the idea from other places:

```
% git grep 'ctx.Value' | grep ok
internal/appconfig/context.go:	if cfg, ok := ctx.Value(configContextKey).(*Config); ok {
internal/appconfig/context.go:	if name, ok := ctx.Value(nameContextKey).(string); ok {
internal/flag/flag.go:	if extraArgs, ok := ctx.Value(extraArgsContextKey).(*[]string); ok {
internal/metrics/helpers.go:	val, ok := ctx.Value(disableFlushMetricsKey{}).(bool)
```

Are those other cases wrong?

Second, I'm already returning a zero length `[]string` to indicate empty extra args.

My feeling overall is no extra args is not an error, it actually is the normal case.  There is no reason for the caller to care.

Let me know if you disagree.

Finally, the`fmt.Printf` is leftover debug.  I'll remove it.

internal/cli/cli.go
rubys: Will fix on my next commit (as well as remove the leftover debug statement) after we come to agreement on error handling.

rubys: Fixed

internal/flag/flag.go
rubys: I'll leave it as is as it is consistent with other sections of the code.

internal/cli/cli.go
```diff
@@ -71,6 +73,19 @@ func Run(ctx context.Context, io *iostreams.IOStreams, args ...string) int {
 	cmd := root.New()
 	cmd.SetOut(io.Out)
 	cmd.SetErr(io.ErrOut)
+
+	// Special case for the launch command, support `flyctl launch args -- [subargs]`
+	// Where the arguments after `--` are passed to the scanner/dockerfile generator.
+	// This isn't supported natively by cobra, so we have to manually split the args
+	// See: https://github.com/spf13/cobra/issues/739
+	if len(args) > 0 && args[0] == "launch" {
```
dangra, line 81: I think we should make it the default, it is useful for `fly m run` too and probably others.

rubys: Can you say more?  `fly m run` doesn't run a scanner, so it isn't clear to me what subcommand would be the consumer of these arguments.

dangra: Sorry Sam, I think what I want to do is a different issue after all.
 
`fly m run` accepts a command after the image name, think of it as the analogous of `docker run`

```
fly m run -a dan-sandbox alpine:latest sleep inf --aflag
Usage:
  flyctl machine run <image> [command] [flags]
...
Error: unknown flag: --aflag

...
 
but it works with `--` 🤯 

```
❯ fly m run -a dan-sandbox alpine:latest -- sleep inf --aflag
Searching for image 'alpine:latest' remotely...
image found: img_lj9x4d8n6repwo1k
Image: docker-hub-mirror.fly.io/library/alpine:latest
Image size: 3.4 MB

Success! A Machine has been successfully launched in app dan-sandbox
 Machine ID: 6e82740b42e287
 Instance ID: 01HTT5JDR0R7X7F21GWZP0C5AH
 State: created

 Attempting to start machine...

==> Monitoring health checks
No health checks found

Machine started, you can connect via the following private ip
  fdaa:0:5ea6:a7b:d33b:3fe3:7b83:2
```

and the flag was passed over:
```
❯ fly m exec 6e82740b42e287 'ps axf' |grep aflag
  305 root      0:00 sleep inf --aflag

```

Comment (rubys) [reply internal/cli/cli.go:81]: Hmmm...

```
% docker run --help

Usage:  docker run [OPTIONS] IMAGE [COMMAND] [ARG...]
```

vs:

```
% fly machine run --help
Run a machine

Usage:
  flyctl machine run <image> [command] [flags]
```

Note with Docker, options is before command, and with flyctl flags is after command, which would lead to the somewhat awkward:

```
  flyctl machine run <image> [command] [flags] -- [args]
```

That being said, what you tried already works (though not clearly documented):

```
  flyctl machine run <image> [flags] -- command [args]
```

Other than documentation, it isn't clear to me how to improve on that, while retaining backwards compatibility.

PR #3418, opened 2024-04-01
Watch config file for token changes

btoews:
### Change Summary

What and Why:

The macaroons that flyctl gets have a short lifespan. Commands with `RequireSession` refresh macaroons that are about to expire before the command runs. Similarly, the agent checks every minute if macaroons need to be refreshed soon. This setup works great for commands that run quickly, but I've had some slower deploys fail because my macaroons expired partway through. I've also had issues with long running `fly logs` where the NATS client has to reconnect but no longer has a valid token. This PR fixes that.

How:

This PR adds an API for watching for changes to the config file. Both the agent and other commands load changes to auth tokens that are written to the config. The agent still checks periodically if tokens need to be updated, and writes those updates to the config. This way, a long running command will start using the new macaroons that were refreshed by the running agent.

I had to submit a PR to the NATS library to allow dynamically changing tokens to be used when reconnecting after an error. I'm hoping https://github.com/nats-io/nats.go/pull/1599 will merge soon so I don't need to point flyctl at my fork.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/agent/server/server.go
+++ b/agent/server/server.go
@@ -137,17 +137,19 @@ func (s *server) serve(parent context.Context, l net.Listener) (err error) {
 		return nil
 	})
 
-	eg.Go(func() error {
-		if f := config.Tokens(ctx).FromConfigFile; f == "" {
-			s.print("monitoring for token expiration")
-			s.updateMacaroonsInMemory(ctx)
-		} else {
-			s.print("monitoring for token changes and expiration")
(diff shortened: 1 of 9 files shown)


btoews:
I'm going to sit on this for another day or two to see if the nats PR merges.

PR #3421, opened 2024-04-02
Make `fly launch --from` clone to current dir

benwaffle:


benbjohnson:
@benwaffle Do you have some context about why the change to pwd?

benwaffle:
I felt the experience was better, especially if launching results in an error

PR #3430, opened 2024-04-05
Machine Based health Checks

### LINKED ISSUE #3233: Allow Using Ephemeral Machines For Health Checks
We could allow trust checks by running a custom ephemeral machine, where if the machine returns 0 the trust check passes, and otherwise it failed. 

This would be useful for doing more complex health checks, like making sure signup functionality continues to work after a deployment.

billyb2:
We basically take the command provided in [[services.machine_checks]],
and run it. If we get an exit code that isn't 0, we say that we failed
the health check. This is only implemented for the rolling strategy
right now, and needs to be implemented for bluegreen, canary, etc.

Fixes #3233 

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a


billyb2:
A huge caveat is that this doesn't support bluegreen deploys yet. It's quiet a bit more involved then i initially thought to get that support moving

internal/command/deploy/machinebasedtest.go
```diff
@@ -0,0 +1,240 @@
...
+		image := lo.Ternary(machineCheck.Image == "", md.img, machineCheck.Image)
+		defer func() {
+			if err != nil {
+				statuslogger.Failed(ctx, err)
+			}
+		}()
+
+		mach, err := md.createTestMachine(ctx, machineCheck.Command, image, machineCheck.Entrypoint, machineToTest)
+		if err != nil {
+			err = fmt.Errorf("error running test machine %s: %w", machineCheck.Command, err)
```
benbjohnson, line 57: It might be easier to read if you use a `%q` instead of `%s` since that will quote the whole command.

internal/command/deploy/machinebasedtest.go
```diff
@@ -0,0 +1,240 @@
...
+	}))
+
+	// FIXME: consolidate this wait stuff with deploy waits? Especially once we improve the output
+	err = md.waitForTestMachinesToFinish(ctx, machineSet)
+	if err != nil {
+		tracing.RecordError(span, err, "failed to wait for test cmd machine")
+		return err
+	}
+
+	time.Sleep(2 * time.Second) // Wait 2 secs to be sure logs have reached OpenSearch
```
benbjohnson, line 86: Is there a better way to check for this rather than sleeping?

benbjohnson (approved):

internal/appconfig/service.go
```diff
@@ -21,6 +21,7 @@ type Service struct {
 	Concurrency        *fly.MachineServiceConcurrency `json:"concurrency,omitempty" toml:"concurrency"`
 	TCPChecks          []*ServiceTCPCheck             `json:"tcp_checks,omitempty" toml:"tcp_checks,omitempty"`
 	HTTPChecks         []*ServiceHTTPCheck            `json:"http_checks,omitempty" toml:"http_checks,omitempty"`
+	MachineChecks      []*ServiceMachineCheck         `json:"machine_checks,omitempty" toml:"machine_checks,omitempty"`
```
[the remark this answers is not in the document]
billyb2: This should be a service level check, so that in theory you can have a test per service

internal/command/deploy/machinebasedtest.go
billyb2: we're now waiting for logs using backoff :)

internal/appconfig/machines.go
```diff
@@ -62,6 +62,53 @@ func (c *Config) ToReleaseMachineConfig() (*fly.MachineConfig, error) {
...
+			fly.MachineConfigMetadataKeyFlyProcessGroup:    fly.MachineProcessGroupFlyAppTestMachineCommand,
+		},
+		Env: lo.Assign(c.Env),
+	}
+
+	if c.Experimental != nil {
+		mConfig.Init.Entrypoint = c.Experimental.Entrypoint
+	}
+
+	mConfig.Env["TEST_COMMAND"] = "1"
```
dangra, line 99: should this be called FLY_TEST_COMMAND ?

internal/appconfig/machines.go
```diff
@@ -62,6 +62,53 @@ func (c *Config) ToReleaseMachineConfig() (*fly.MachineConfig, error) {
...
+	mConfig.Env["FLY_PROCESS_GROUP"] = fly.MachineProcessGroupFlyAppTestMachineCommand
+	if c.PrimaryRegion != "" {
+		mConfig.Env["PRIMARY_REGION"] = c.PrimaryRegion
+	}
+	mConfig.Env["FLY_TEST_MACHINE_IP"] = machineIP
+
+	// StopConfig
+	c.tomachineSetStopConfig(mConfig)
```
dangra, line 107: considering machine checks could be a completely different process than the app process, I am not sure it is practical to set the same stop configuration than the main app.  I think this is a call out to have a kill signal and. timeout as part of MachineCheck

internal/appconfig/machines.go
```diff
@@ -62,6 +62,53 @@ func (c *Config) ToReleaseMachineConfig() (*fly.MachineConfig, error) {
 	return mConfig, nil
 }
 
+func (c *Config) ToTestMachineConfig(machineCommand, machineImage, machineEntrypoint, machineIP string) (*fly.MachineConfig, error) {
+	// for whatever reason, we need to split the entry point, but not the command
+	entrypoint, err := shlex.Split(machineEntrypoint)
+	if err != nil {
+		return nil, err
+	}
+	command := []string{machineCommand}
```
dangra, line 71: Overriding the entrypoint has to be optional, if not set, it has to use the image's default entrypoint.

In fact, shell splitting machineCommand is a must, otherwise this is inconsistent and only works because the test is overriding the image's entrypoint by `/bin/sh -c`

internal/appconfig/machines.go
billyb2: definitely best, i just used the same naming convention as RELEASE_COMMAND

internal/appconfig/machines.go
billyb2: Done!

internal/appconfig/service.go
```diff
@@ -44,6 +45,14 @@ type ServiceHTTPCheck struct {
 	HTTPHeaders       map[string]string `json:"headers,omitempty" toml:"headers,omitempty"`
 }
 
+type ServiceMachineCheck struct {
+	Command     []string      `json:"command,omitempty" toml:"command,omitempty"`
+	Image       string        `json:"image,omitempty" toml:"image,omitempty"`
+	Entrypoint  []string      `json:"entrypoint,omitempty" toml:"entrypoint,omitempty"`
+	KillSignal  string        `json:"kill_signal,omitempty" toml:"kill_signal,omitempty"`
```
dangra, line 52: KillSignal is `*string` on top level, I would use the same here as it makes propagating the value much easier.

internal/appconfig/machines.go
```diff
@@ -62,6 +62,72 @@ func (c *Config) ToReleaseMachineConfig() (*fly.MachineConfig, error) {
...
+		mConfig.Env["PRIMARY_REGION"] = c.PrimaryRegion
+	}
+	mConfig.Env["FLY_TEST_MACHINE_IP"] = lo.Ternary(origMachine == nil, "", origMachine.PrivateIP)
+
+	// Use the stop config from the app config by default
+	c.tomachineSetStopConfig(mConfig)
+
+	mConfig.StopConfig = &fly.StopConfig{
+		Timeout: lo.Ternary(svc.KillTimeout != nil, svc.KillTimeout, c.KillTimeout),
+		Signal:  lo.Ternary(svc.KillSignal != "", &svc.KillSignal, c.KillSignal),
```
dangra, line 125: I think we want to default to top level KillSignal/Timeout values only when `svc.Image` isn't set.

internal/appconfig/machines.go
```diff
@@ -62,6 +62,72 @@ func (c *Config) ToReleaseMachineConfig() (*fly.MachineConfig, error) {
 	return mConfig, nil
 }
 
+func (c *Config) ToTestMachineConfig(svc *ServiceMachineCheck, origMachine *fly.Machine) (*fly.MachineConfig, error) {
+	var machineEntrypoint []string
+	if len(svc.Entrypoint) > 0 {
```
dangra, line 67: there is subtle difference bettwen `len(svc.Entrypoint) > 0` and `svc.Entrypoint != nil`, the former includes the later, but by checking for the later is more accurate way to check if an entrypoint wasn't set vs set to an empty entrypoint.

internal/appconfig/machines_test.go
```diff
@@ -175,6 +175,44 @@ func TestToReleaseMachineConfig(t *testing.T) {
...
+		Restart:     &fly.MachineRestart{Policy: fly.MachineRestartPolicyNo},
+		DNS:         &fly.DNSConfig{SkipRegistration: true},
+		StopConfig: &fly.StopConfig{
+			Timeout: fly.MustParseDuration("10s"),
+			Signal:  fly.Pointer("SIGTERM"),
+		},
+	}
+
+	check := cfg.HTTPService.MachineChecks[0]
+	got, err := cfg.ToTestMachineConfig(check, nil)
```
dangra, line 211: we have to test passing a non-nil second argument here. In fact, I don't think the second argument could be optional for the feature tested.

internal/command/deploy/machinebasedtest.go
```diff
@@ -0,0 +1,256 @@
...
+		origMachineRaw = &fly.Machine{
+			Region: md.appConfig.PrimaryRegion,
+		}
+	}
+
+	mConfig, err := md.appConfig.ToTestMachineConfig(svc, origMachineRaw)
+	if err != nil {
+		return nil, err
+	}
+	mConfig.Guest = md.inferTestMachineGuest()
```
dangra, line 181: having another guest inferring function here sucks (sorry), what about using the same machine size than the canary? or even add a `appconfig.Compute` sub section to machine checks.

internal/command/deploy/machinebasedtest.go
```diff
@@ -0,0 +1,256 @@
...
+		io    = md.io
+	)
+	defer func() {
+		if err != nil {
+			tracing.RecordError(span, err, "failed to run test machine")
+		}
+		span.End()
+	}()
+
+	machineChecks := lo.FlatMap(md.appConfig.AllServices(), func(svc appconfig.Service, _ int) []*appconfig.ServiceMachineCheck {
```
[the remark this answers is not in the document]
billyb2: woah this is a huge catch!

internal/appconfig/machines.go
billyb2: Hm, how come?

internal/appconfig/machines.go
billyb2: true! thank you for pointing this out

internal/appconfig/machines.go
dangra: let say the application needs a minute to drain connections and gracefully stop; also it stop the shutdown procedure on SIGUSR1 signal (not SIGTERM which is the default). 

Now the app also has a few machine checks, one uses the application image and others an external image like `curlimages/curl` 


```toml
app = "myapp"
kill_timeout = "1m"
kill_signal = "SIGUSR1"

[http_service]
internal_port = 8080

[[http_service.machine_checks]]
command = "/my-app runs-my-check"

[[http_service.machine_checks]]
image = "curlimages/curl"
command = ["-sf", "http://$FLY_TEST_MACHINE_IP:8080"]

```

Now, think about this:
1. Would you want the curl check to wait for a minute because that is what the app needs to gracefully drain and stop?
2. Is sending SIGUSR1 to curl's machine check what the user would expect us doing? I think this is a big gotcha moment.
3. Even for the first check that reuses the application image, does it makes sense to wait for a minute?
4. Do we need a way to replace FLY_TEST_MACHINE_IP by its actual value before launching the machine?
5. It is possible to rely on shell variable expansion but only for images that pipes the command trough a shell first

billyb2: that makes total sense, you've convinced me :)

internal/command/deploy/machinebasedtest.go
billyb2: yea i don't mind just using `inferCanaryGuest` here

dangra (commented):
I was expecting to see a test case added for this complex app case https://github.com/superfly/flyctl/pull/3430#discussion_r1596026508

It pretty much checks expectations around stop configuration, but in general it could be the perfect test case for all "machine check" related edge cases. 

I'd add it as `tomachine-machinechecks.toml`, include multiple process groups, services and machine checks.

dangra (approved):
I can keep throwing nitpicks all day but It is in much better shape billy! well done.

PR #3432, opened 2024-04-05
Create machines for bluegreen deploy concurrently

tvdfly:
Previously, we created the green (new) machines sequentially when running a bluegreen deployment. That works ok when there are a few machines, and it becomes a bit slow when there are 70 or 100 or 500+ machines in the app.

Update the CreateGreenMachines() func to use the pool pattern we have in a few places.

I did the same thing in the MarkGreenMachinesAsReadyForTraffic(), CordonBlueMachines(), StopBlueMachines() funcs.

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


tvdfly:
The `TestConfigWatch` test is failing on macos. It doesn't seem related to this change and I wasn't able to repro. I think we can take care of it in a separate pr.

internal/command/deploy/machines.go
```diff
@@ -191,6 +193,11 @@ func NewMachineDeployment(ctx context.Context, args MachineDeploymentArgs) (Mach
 		maxUnavailable = *appConfig.Deploy.MaxUnavailable
 	}
 
+	bluegreenMaxConcurrent := args.BlueGreenMaxConcurrent
+	if bluegreenMaxConcurrent < 1 {
+		bluegreenMaxConcurrent = 1
+	}
+
```
benbjohnson, line 200: Can we consolidate this down to a single `--max-concurrent` flag? Or do we need a separate bluegreen & immediate flag? For backwards compatibility, we could have `--immediate-max-concurrent` still work but maybe just hide it from the docs?

benbjohnson (approved):
Overall lgtm. Just a question about the flag name.

internal/command/deploy/machines.go
tvdfly: i like that, i'll update

PR #3442, opened 2024-04-10
Unify token updating code

btoews:
There's a lot of stuff we need to do to make macaroon tokens work well in flyctl:

- Prune expired/invalid tokens
- Fetch and refresh the user's discharge tokens
- Fetch new macaroons when the user is added to an org (we weren't doing this)
- Prune macaroons when the user is removed from an org (we were just letting them expire after 10 minutes)

This was complicated further by wanting to keep tokens in sync between the flyctl agent and any foreground processes. Long running commands were also winding up with expired tokens because we were only doing a lot of this work when the command started.

This PR centralizes all of this logic and runs it both in foreground flyctl processes as well as in the background agent. When flyctl makes changes to its set of tokens, it writes those back to the config file (so long as flyctl wasn't started with `FLY_AUTH_TOKEN`). This should allow everything to stay nicely up to date and synced between processes.

There was also some annoyance (/cc @jipperinbham) when running commands with `FLY_AUTH_TOKEN`. The background agent and foreground commands would often end up running with a different set of tokens and one or the other would run into authorization errors. I'm adding a new `set-token` RPC to the fly agent in this PR. Clients send this command before sending other commands, letting the agent know what tokens to use for the session.

TODO:
- Make sure `send-token` command fails nicely when sent to old agent
- Store agent tunnels by "raw slug" instead of "slug" to avoid confusion between multiple `personal` orgs when using different tokens.

PS: There's been a lot of churn in the tokens code in flyctl. Sorry about that. I'd been trying to avoid writing this big PR, but half measure weren't cutting it.

### DIFF
--- a/agent/server/session.go
+++ b/agent/server/session.go
@@ -20,11 +20,14 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	fly "github.com/superfly/fly-go"
+	"github.com/superfly/fly-go/tokens"
 	"github.com/superfly/flyctl/agent"
 	"github.com/superfly/flyctl/agent/internal/proto"
 	"github.com/superfly/flyctl/wg"
 
 	"github.com/superfly/flyctl/internal/buildinfo"
+	"github.com/superfly/flyctl/internal/config"
+	"github.com/superfly/flyctl/internal/flyutil"
 )
 
 type id uint64
@@ -38,6 +41,7 @@ type session struct {
 	conn   net.Conn
 	logger *log.Logger
 	id     id
+	tokens *tokens.Tokens
 }
 
 var errUnsupportedCommand = errors.New("unsupported command")
@@ -81,6 +85,10 @@ func runSession(ctx context.Context, srv *server, conn net.Conn, id id) {
 		return
 	}
 
+	s.runCommand(ctx)
+}
+
+func (s *session) runCommand(ctx context.Context) {
 	buf, err := proto.Read(s.conn)
 	if len(buf) > 0 {
 		s.logger.Printf("<- (% 5d) %q", len(buf), redact(buf))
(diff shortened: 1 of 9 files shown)


internal/config/tokens.go
```diff
@@ -0,0 +1,364 @@
...
+
+	ticker := time.NewTicker(15 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			currentStr, err := ReadAccessToken(t.FromConfigFile)
```
timflyio, line 105: why is `t.FromConfigFile` accessed directly here but under lock (`tokensConfigFile()`) earlier?

internal/config/tokens.go
```diff
@@ -0,0 +1,364 @@
...
+			m.Lock()
+			// don't clobber config file if it changed out from under us. the
+			// consequences of a race here (agent and foreground command both
+			// fetching updates simultaneously) are low, so don't bother with an
+			// extra lock file.
+			if t.All() == beforeUpdate {
+				t.Replace(localCopy)
+
+				if fromConfigFile != "" {
+					if err := SetAccessToken(fromConfigFile, t.All()); err != nil {
```
timflyio, line 167: is `fromConfigFile` here still valid after `t.Replace` above?

timflyio (commented):
I looked over the code, had some questions and have some minor things to point out. I'm still a little short on the big picture and context around this.

agent/server/session.go
```diff
@@ -533,6 +548,43 @@ func (s *session) ping6(ctx context.Context, args ...string) {
...
+		s.tokens.FromConfigFile = args[1]
+	case "str":
+		s.tokens = tokens.Parse(args[1])
+	}
+
+	go s.srv.UpdateTokensFromClient(s.tokens)
+
+	s.ok()
+
+	s.runCommand(ctx)
```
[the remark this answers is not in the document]
btoews: Each command from the client is it's own session. Since we want to set the token per-session, we have to do it on each command. So a session might look like

```
-> set-token cfg ~/.fly/config.toml
<- ok
-> ping whatever
<- ok whatever
```

agent/server/session.go
```diff
@@ -533,6 +548,43 @@ func (s *session) ping6(ctx context.Context, args ...string) {
...
+}
+
+// getClient returns an API client that uses any API tokens sent by the client.
+// If not have been sent, it falls back to using the server's tokens.
```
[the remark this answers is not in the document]
btoews: Good catch. I didn't realize I had both and it was breaking stuff. I just pushed a fix :-)

internal/config/tokens.go
btoews: I need to make another fly-go change to put all the Tokens struct fields behind accessor functions that use a mutex (I'll go do that now). The mutex that is used elsewhere in this file is created in this function and avoids concurrent access by the two goroutines that this file launches.

internal/config/tokens.go
btoews: It'll be the same value.

internal/config/tokens.go
btoews: Doing this in https://github.com/superfly/fly-go/pull/35 and 94f715a89cc9871f0442e714f5877705e7d95470

PR #3446, opened 2024-04-12
Add `fly tokens create ssh` command

btoews:
This allows creating a token that can only be used for SSH'ing to a specific app (and also connecting to the wg network).

/cc @moss-fly @timflyio @benwaffle @nbw @asib

### DIFF
--- a/internal/command/tokens/create.go
+++ b/internal/command/tokens/create.go
@@ -93,7 +93,7 @@ func newSSH() *cobra.Command {
 			Name:        "name",
 			Shorthand:   "n",
 			Description: "Token name",
-			Default:     "flyctl deploy token",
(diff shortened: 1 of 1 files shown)


internal/command/tokens/create.go
```diff
@@ -72,6 +73,39 @@ func newOrg() *cobra.Command {
...
+
+	flag.Add(cmd,
+		flag.App(),
+		flag.AppConfig(),
+		flag.JSONOutput(),
+		flag.String{
+			Name:        "name",
+			Shorthand:   "n",
+			Description: "Token name",
+			Default:     "flyctl deploy token",
```
asib, line 96:
```suggestion
			Default:     "flyctl ssh token",
```

PR #3454, opened 2024-04-17
Make single volume warnings scarier

benwaffle:


internal/command/volumes/create.go
```diff
@@ -184,7 +184,7 @@ func confirmVolumeCreate(ctx context.Context, appName string) (bool, error) {
 	io := iostreams.FromContext(ctx)
 	colorize := io.ColorScheme()
 
-	const msg = "Warning! Every volume is pinned to a specific physical host. You should create two or more volumes per application to avoid downtime. Learn more at https://fly.io/docs/reference/volumes/"
+	const msg = "Warning! Every volume is pinned to a specific physical host. You should create two or more volumes per application to avoid downtime and permanent data loss. Learn more at https://fly.io/docs/reference/volumes/"
```
andie787, line 187: Do you think we should say something about replication so we're not implying two volumes will solve the problem? "...replicate your data between two or more volumes per app..." or something.

Also I wonder if people just zone out after the second sentence which isn't saying the important thing? 

Warning! Volumes are pinned to physical hosts and hosts can have outages or fail completely.

benwaffle: How about something like
```
Warning! You should replicate your data across two or more volumes per application to avoid downtime and the risk of permanent data loss (Learn more at https://fly.io/docs/reference/volumes).
```

andie787: LGTM!

andie787:
@benwaffle we should probably merge this?

PR #3459, opened 2024-04-18
[tracing] change otel exporter from grpc to http

gwuah:
### Change Summary
We are having some reliability/consistency issues with traces we are ingesting. Some spans never make it to the collector, mostly because grpc on the open internet is a bit of a weird story and so it breaks for some users. (speculation)

This PR swaps out the old exporter for a new one, which will hopefully improve delivery reliability.
I also moved the trace provider outside the CLI package because the otelhttp exporter doesn't support dynamic headers and it's really hard to decently get access to the token (no pun intended) in the cli.go package.

The trace provider is now closer to whichever command is being run and is stored in a global variable internally so duplicate calls don't create 2 providers :))

Improvments
- [x]  change flyctl exporter from using grpc to http (this will massively improve delivery reliability for the traces)
- [x]  sometimes traces/spans never make it to honeycomb (i suspect grpc exporter and uptime of collector)
- [x]   [log](https://github.com/superfly/fly-go/pull/40) trace-id on errors (for easier lookups)
- [ ]  heavily sample machine lease spans (they make up about 80% of spans in a trace and are mostly useless)
- [ ]  run more instances of the otel collector (availability)
- [ ] add metrics to the collector (so we can monitor it's health and behaviour, eg how many spans it's dropping)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -71,7 +71,7 @@ require (
 	github.com/vektah/gqlparser/v2 v2.5.11
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.50.0
 	go.opentelemetry.io/otel v1.25.0
-	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.25.0
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.21.0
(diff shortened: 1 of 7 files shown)


rugwirobaker (approved):
LGTM

PR #3460, opened 2024-04-18
Log trace-id from flaps error

gwuah:
### Change Summary
Log trace-id from flaps error so we can easily debug stuff with honeycomb.
Can't be merged till https://github.com/superfly/fly-go/pull/40 is merged and i've updated fly-go's version.


rugwirobaker (approved):
Nice

PR #3462, opened 2024-04-18
store app name on all spans

gwuah:
### Change Summary
store app name on all spans

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -193,6 +193,7 @@ func New() (cmd *cobra.Command) {
 
 func run(ctx context.Context) error {
 	io := iostreams.FromContext(ctx)
+	appName := appconfig.NameFromContext(ctx)
 
 	hook := ctrlc.Hook(func() {
 		metrics.FlushMetrics(ctx)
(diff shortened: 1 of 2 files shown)


internal/command/deploy/deploy.go
```diff
@@ -193,22 +193,22 @@ func New() (cmd *cobra.Command) {
...
+	appName := appconfig.NameFromContext(ctx)
 
 	hook := ctrlc.Hook(func() {
 		metrics.FlushMetrics(ctx)
 	})
 
 	defer hook.Done()
 
-	tp, err := tracing.InitTraceProvider(ctx)
+	tp, err := tracing.InitTraceProvider(ctx, appName)
```
rugwirobaker, line 204: Could you make this optional? We don't obtain the app name for the launch command until a couple of lines late into the command, so it doesn't work either.

internal/tracing/tracing.go
```diff
@@ -95,7 +95,11 @@ func getToken(ctx context.Context) string {
 	return token
 }
 
-func InitTraceProvider(ctx context.Context) (*sdktrace.TracerProvider, error) {
+func InitTraceProviderWithoutApp(ctx context.Context) (*sdktrace.TracerProvider, error) {
```
gwuah, line 98: you can use this @rugwirobaker

rugwirobaker (approved):

PR #3464, opened 2024-04-18
Start tracing fly launch

rugwirobaker:
Instrumenting `fly launch` the same way we did `fly deploy`

### DIFF
--- a/internal/command/launch/cmd.go
+++ b/internal/command/launch/cmd.go
@@ -194,11 +194,6 @@ func run(ctx context.Context) (err error) {
 		metrics.LaunchStatus(ctx, "launch", status)
 	}()
 
-	user, err := client.GetCurrentUser(ctx)
-	if err != nil {
-		return fmt.Errorf("failed retrieving current user: %w", err)
-	}
-	span.SetAttributes(attribute.String("user.id", user.ID))
 
 	if err := warnLegacyBehavior(ctx); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


internal/command/launch/cmd.go
```diff
@@ -172,10 +189,17 @@ func run(ctx context.Context) (err error) {
...
+		status.TraceID = span.SpanContext().TraceID().String()
 		status.Duration = time.Since(startTime)
 		metrics.LaunchStatus(ctx, "launch", status)
 	}()
 
+	user, err := client.GetCurrentUser(ctx)
+	if err != nil {
+		return fmt.Errorf("failed retrieving current user: %w", err)
+	}
+	span.SetAttributes(attribute.String("user.id", user.ID))
```
gwuah, line 201: the collector [already](https://github.com/superfly/fly-otel-collector/blob/4ec6179058683c20934d444d0b5b16d930a88531/processors/span/processor.go#L125) sets this
```suggestion
```

internal/tracing/tracing.go
```diff
@@ -74,19 +74,20 @@ func SpanContextFromHeaders(res *http.Response) trace.SpanContext {
...
 	}
 
 	startOpts = append(startOpts, opts...)
 
 	return GetTracer().Start(ctx, spanName, startOpts...)
 }
 
+func CmdSpanWithApp(ctx context.Context, spanName string, appName string) (context.Context, trace.Span) {
+	return CMDSpan(ctx, spanName, trace.WithAttributes(attribute.String("app.name", appName)))
+}
```
gwuah, line 89:
```suggestion
we don't need this helper function then, whoever calls cmdSpan can pass the attribute directly as opts.
```

gwuah (approved):

PR #3469, opened 2024-04-19
Enhanced Windows Installation via Chocolatey

ecsistem:
### Change Summary

What and Why:

- Added documentation for installing flyctl via Chocolatey on Windows to enhance installation options and provide clear instructions for users.

How:

- Updated the Windows installation section to include instructions for installing flyctl via Chocolatey package manager.
- Provided a clear command for users to execute in PowerShell as an administrator to install flyctl.

Related to:

- Windows installation options for flyctl.
- Improving user experience on Windows systems.

---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Run the Powershell install script:
 iwr https://fly.io/install.ps1 -useb | iex
 ```
 
-##### Installation via Chocolatey
+#### Installation via Chocolatey
 
 You can also install the `flyctl` package via Chocolatey. Run the following command in PowerShell as an administrator:
 
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ During the transition, note that where you see `flyctl` as a command it can be r
 ```bash
 brew install flyctl
 ```
+
 To upgrade to the latest version:
 
 ```bash
@@ -48,10 +49,17 @@ curl -L https://fly.io/install.sh | sh -s 0.0.200
 
 Run the Powershell install script:
 
-```
+```powershell
 iwr https://fly.io/install.ps1 -useb | iex
 ```
 
+##### Installation via Chocolatey
+
+You can also install the `flyctl` package via Chocolatey. Run the following command in PowerShell as an administrator:
+
+```powershell
+choco install flyctl
+```
 
 ## Downloading from GitHub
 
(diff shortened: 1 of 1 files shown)


dangra:
hi @ecsistem, thanks for the contribution. I have no idea how Chocolatey works but I imagine there is some package definition somewhere that tells the choco client how to install an app like flyctl.

My only concern is whether auto-updates keep working after installing with Chocolatey. Do you have any experience with that? have you seen your flyctl updating after a new release is released or updates are disabled?

ecsistem:
Hi @dangra , thank you for your questions and feedback! Let me clarify a bit about how Chocolatey works and address your concerns:

Chocolatey functions similarly to Snap on Linux. It manages packages by downloading the installer directly from this repository. The package is then passed through a virus checker to ensure safety before being made available on Chocolatey, which is a widely used package manager for Windows.

Regarding updates, Chocolatey packages are typically updated whenever a new release is published, as long as the package maintainer pushes the latest version. This means the installed application, in this case flyctl, will be updated whenever you run choco upgrade flyctl. Automatic updates for the application itself (outside of Chocolatey) should still function unless explicitly disabled during installation.

From my experience, flyctl updates have continued working normally after installation through Chocolatey.

To provide more context, I’ll include a screenshot showing how the installer is downloaded directly from the repository and a version history log to illustrate the update process.

I hope this clears things up! Let me know if you have any further concerns or questions.
Captura de tela de 2024-11-20 13-18-11
Captura de tela de 2024-11-20 13-21-35

dangra:
That looks good. I have a few more questions if you don't mind. 

1. Who maintains the Chocolatey package definition for flyctl? 
2. What about linking the Choco recipe in the README and adding a note clariflying it is community maintained? I expect it to be public, similar to [brew's formula](https://formulae.brew.sh/formula/flyctl)
3. Is the Choco recipe something Fly devs can update? Flyio publishes Brew packages daily as part of the release process, if Chocolatey gets traction I think we should do the same.

ecsistem:
@dangra

dangra:
> Adding a link to the recipe in the README, along with a note explaining that it is community-maintained, would be a good idea to clarify responsibilities and guide users. This approach is similar to what's done with the Brew formula.

hi @ecsistem. Sorry for the sleeping on this pull request. I think addressing this point would make it mergeable.

PR #3477, opened 2024-04-23
Add support for "tokens special exec" command.

timflyio:
- Add "tokens special exec" command for building attenuated tokens that restrict which commands can be executed.
- Several items left TBD, will discuss in draft PR.

### Change Summary

What and Why: Adds a command to flyctl to create attenuated tokens that has a commands caveat for restricting which commands can be executed.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


timflyio:
This is a draft to solicit feedback on what the command arguments should look like and exactly what caveat we want to construct.  Some questions:

1. Right now it wraps the commands caveat in an if-exists.  The thought here is that in order to use the `flyctl machine exec` command, the token will need graphql and access to read the machine.  So right now the created token can still do everything it could do before except when executing commands, in which case the commands are restricted appropriately.  But, perhaps we want to support restricting to ONLY allow command execution, with a minimal set of permissions needed to do that (still read machines, maybe still query graphql? or maybe we get rid of the need for graphql for machines exec?)
2. The if-present caveat right now doesnt do any further restrictions (else is `resset.ActionAll`). It could lock down to `resset.ActionRead` to minimize what else can be done besides command execution.
3. I just made up argument names. I'm open to suggestsions for improvements.  Right now `--command` for an exact command, and `--command-prefix` for prefix matching of a command.
4. I added a `--all-commands` flag for adding a wildcard command caveat.  This doesnt really make sense if we're always wrapping the caveat in an if-present.  It might make sense if we have an option to not wrap the caveat in an if-present.

Looking for feedback and thoughts, @btoews and @tqbf .

btoews:
Looks like flyctl can hit the exec endpoint without talking to GraphQL first. It does need to hit the machine-read endpoint first though. I wonder if we can remove even that.

```
FLY_API_TOKEN=$(fly tokens deploy -x1h -a machinetest-btoews) LOG_LEVEL=debug fly m exec 5683d299b7e978 whoami -a machinetest-btoews
DEBUG --> GET https://api.machines.dev/v1/apps/machinetest-btoews/machines/5683d299b7e978
DEBUG --> POST https://api.machines.dev/v1/apps/machinetest-btoews/machines/5683d299b7e978/exec
```

A good balance might be something like

```
Machine(id=5683d299b7e978); IfPresent(else=READ, ifs: [Command(...)])
```

> 3. I just made up argument names. I'm open to suggestsions for improvements. Right now --command for an exact command, and --command-prefix for prefix matching of a command.
> 4. I added a --all-commands flag for adding a wildcard command caveat. This doesnt really make sense if we're always wrapping the caveat in an if-present. It might make sense if we have an option to not wrap the caveat in an if-present.

I haven't gotten to the code yet, but one thing to keep in mind is that the flyctl UX doesn't need to exactly mirror the caveats. We can expose the subset of functionality that we think people actually want. We can also expose multiple common workflows via separate commands instead of with arguments if that's easier for users to understand.

internal/command/tokens/special.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/superfly/macaroon/resset"
+)
+
+func newSpecial() *cobra.Command {
+	const (
+		short = "Create special attenuated Fly.io API tokens"
+		long  = `Attenuate a Fly.io API token for special purposes. The
+				token to be attenuated may either be passed in the -t argument
+				or in FLY_API_TOKEN.`
+		usage = "special"
```
btoews, line 22: My 2¢: `fly tokens create` is already our place to shove whatever wacky token-generation stuff we want. I don't think we need a sub-folder for special tokens. Things that are really top-level concerns can be aliased (`fly tokens deploy` is an alias to `fly tokens create deploy`).

btoews:
I think the improved UX of an if-present-command-else-read is worth the decreased security. Reading the app/machine isn't scary functionality. That way, you should be able to run `fly m list -a <app>` and `fly m exec <mid> <cmd>`.

As for the command DX, I think it would be nice to be able to run 

```
fly tokens create machine-exec -- ps aux
```

to get a token that allows exactly `ps aux`.

Or

```
fly tokens create machine-exec --command ps aux -- command whoami
```

to allow exactly `ps aux` or exactly `whoami`

Or

```
fly tokens create machine-exec --command-prefix ps
```

to allow `ps *`

Or

```
fly tokens create machine-exec
```

to allow any execing with any command.

internal/command/tokens/special.go
timflyio: yah, this is much better. Also it creates new tokens with arbitrary lifetimes vs. attenuating based on the login tokens flyctl holds (which have 3rd party caveats). Updated in latest commit.

timflyio:
Ok, I reworked the PR with all the feedback above.  For testing I'm using:

```
FLY_API_TOKEN="$(./flydbg tokens create machine-exec -a timflyio-go-example "$@")" \
	./flydbg tokens debug
```

and

```
FLY_API_TOKEN="$(./flydbg tokens create machine-exec -a timflyio-go-example --command ls --command-prefix echo -- id)" \
	./flydbg machine exec --app timflyio-go-example 784e692c222598 "$*"
```

btoews (approved):
Looks great!

PR #3482, opened 2024-04-24
Suggest `--wg=false` when users run into wireguard issues

gwuah:
### Change Summary
This is to drive adoption of the wg-less feature and more importantly, unblock users.
When we detect they're running into wireguard issues during deploys, we suggest the "--wg=false" flag

### DIFF
--- a/internal/build/imgsrc/docker.go
+++ b/internal/build/imgsrc/docker.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


benbjohnson:
How does adding the heartbeat cause `flyctl` to suggest to use `--wg=false`?

gwuah:
@benbjohnson this PR is not ready for review haha. I've not pushed the actual work.

benbjohnson (approved):

PR #3483, opened 2024-04-24
update error message for bluegreen unblocking

gwuah:
### Change Summary
Add `--force` flag since most zombie machines will be started and use `(s)` since it might only be 1 broken machine.


saleemrashid (approved):
SOC2

PR #3485, opened 2024-04-24
Show worker status in `fly m ls`, `fly vol ls`

matttpt:
### Change Summary

**What and Why:**

Expose worker status information in `fly m ls` and `fly vol ls`, so that users have better visibility into host issues. This also suppresses unset fields for Machines with unreachable workers in `fly m ls` so that the (potentially confusing) default values are not shown. (Most notably, this eliminates the strange 1970 dates shown in the "created" and "last updated" columns.)

If a worker is unreachable, then it looks like this:

```
ID                   	STATE  	NAME    	SIZE	REGION	ZONE	ENCRYPTED	ATTACHED VM	CREATED AT   
vol_rd9kkvvz39ox01wn*	created	test_vol	1GB 	ord 	ed69 	false    	           	22 hours ago	

* The workers hosting these volumes could not be reached.
```

**How:**

Use the `worker_status` field recently added to Machines and volumes in the Machines API.

---

### Documentation

- [x] Fresh Produce (one has been drafted)
- [ ] n/a

### DIFF
--- a/internal/command/machine/list.go
+++ b/internal/command/machine/list.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/samber/lo"
 	"github.com/spf13/cobra"
 	fly "github.com/superfly/fly-go"
 	"github.com/superfly/fly-go/flaps"
@@ -91,6 +92,8 @@ func runMachineList(ctx context.Context) (err error) {
 		}
 		_ = render.Table(io.Out, "", rows)
 	} else {
+		unreachableMachines := false
+
 		for _, machine := range machines {
 			var volName string
 			if machine.Config != nil && len(machine.Config.Mounts) > 0 {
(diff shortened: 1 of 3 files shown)


dangra (approved):

internal/command/machine/list.go
```diff
@@ -115,16 +118,23 @@ func runMachineList(ctx context.Context) (err error) {
...
 			rows = append(rows, []string{
-				machine.ID,
+				machine.ID + note,
 				machine.Name,
 				machine.State,
 				machine.Region,
-				machine.ImageRefWithVersion(),
-				machine.PrivateIP,
+				lo.Ternary(unreachable, "", machine.ImageRefWithVersion()),
+				lo.Ternary(unreachable, "", machine.PrivateIP),
```
dangra, line 134: no idea why we are nullifying it but this information can come from corrosion if the worker isn't responding

PR #3491, opened 2024-04-25
Rename "workers" to "hosts"

matttpt:
Follow-on to #3485 after internal discussion about what to call this in the public UI.


billyb2 (approved):
banger pr

PR #3493, opened 2024-04-26
If there is a VOLUME statement in the Dockerfile, mount it

rubys:


scanner/dockerfile.go
```diff
@@ -53,5 +53,18 @@ func ScanDockerfile(dockerfilePath string, config *ScannerConfig) (*SourceInfo,
 		s.Port = defaultPort
 	}
 
+	// extract volume - handle both plain string and JSON format, but only allow one path
+	re := regexp.MustCompile(`(?m)^VOLUME\s+(\[\s*")?(\/[\w\/]*?(\w+))("\s*\])?\s*$`)
```
benbjohnson, line 57: It's usually best to define this as a global variable so that the `MustCompile` runs once and panics on startup rather than only if this code path is used.

e.g.

```go
var extractVolumeRegex := regexp.MustCompile(`(?m)^VOLUME\s+(\[\s*")?(\/[\w\/]*?(\w+))("\s*\])?\s*$`)
```

benbjohnson (approved):

scanner/dockerfile.go
rubys: This is a regular expression that will rarely be used (only if the scanner is Dockerfile), and even then, will only be used once.

PR #3498, opened 2024-04-29
write kubeconfig file to disk instead

senyosimpson:
### Change Summary

What and Why:

Writes the kubeconfig file to disk instead of printing it to screen (which is annoying to have to copy and paste).

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


jipperinbham (commented):
Is there anyway we can add a flag to override the output path and name?

PR #3499, opened 2024-04-29
Explain Redis billing at provisioning time and allow provisioning in all regions

jsierles:
```

Your database bold-wave-318 is ready. Apps in the fly-ephemeral org can connect to Redis at redis://default:a0c10d31340141d6ab4a84030bc6b059@fly-bold-wave-318.upstash.io:6379

If you have redis-cli installed, use fly redis connect to get a Redis console.

Your database is billed at $0.2 per 100K commands. If you're using Sidekiq or BullMQ, which poll Redis frequently, consider switching to a fixed-price plan. See https://fly.io/docs/reference/redis/#plans
```

### DIFF
--- a/internal/command/redis/create.go
+++ b/internal/command/redis/create.go
@@ -86,7 +86,7 @@ func runCreate(ctx context.Context) (err error) {
 		return err
 	}
 
-	primaryRegion, err := prompt.Region(ctx, !org.PaidPlan, prompt.RegionParams{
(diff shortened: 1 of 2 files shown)


benbjohnson:
lgtm but should we show the money as `$0.20` instead of `$0.2`? The formatting looks a little odd to me but maybe that's how we do it elsewhere?

jsierles:
So this is temporary, since billing is about to happen as we want people to start to be aware of these other plans.

Issue #3509, opened 2024-05-01
fly deploy -r changes primary region


tvdfly:
We have an app running machines in three regions: iad, sea, syd. The primary region is iad. We ran `fly deploy -r sea` today, and the release included `primary_region = "sea"`.

It only updated the machine in sea before failing, and that meant that some machines thought iad was the primary region while the one that got updated thought sea was the primary region. That caused issues for our services, and was a surprising side effect of using `-r`.

Maybe we change `-r` to be an alias for `--only-regions` to make it less confusing?

tvdfly:
We had a similar issue ~9months ago, and I added the `--only-regions` flag at that time. https://github.com/superfly/flyctl/commit/1bdedefef34dfe45da88a706fd22f42597586411

benbjohnson:
@tvdfly I think my preference would be to remove the shorthand `-r` flag entirely. I also think we should rename `--region` to `--set-region`.

Another option would be to alias `-r` and `--region` to `--only-regions` and change the existing `--region` to something more explicit like `--set-region`.

tvdfly:
@benbjohnson any of those sound good to me! I think changing primary region based on a flag is surprising behavior, and that's the main issue we've run into.

PR #3514, opened 2024-05-02
Rename --only-regions, remove primary regions flag

benbjohnson:
### Change Summary

This pull request removes the previous `--region` flag on `fly deploy` as it overrode the `primary-region` setting in the config, which was surprising to say the least. I was originally going to rename it to `--primary-region` but I don't see a good use case for overriding the primary region for a single deploy. We can add it back in if it's useful.

I also renamed the `--only-regions` to `--regions` but retained the `--only-regions` alias for backwards compatibility. This flag can also use the shorthand of `-r`.

Fixes: https://github.com/superfly/flyctl/issues/3509

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -30,7 +30,6 @@ import (
 var defaultMaxConcurrent = 16
 
 var CommonFlags = flag.Set{
-	flag.Region(),
 	flag.Image(),
 	flag.Now(),
 	flag.RemoteOnly(false),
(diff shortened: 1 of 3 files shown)


tvdfly (approved):
💖

natewallis:
Hi Ben, I think you should update your example usages so that people aren't confused.   This just broke our deploy and the example usage still shows... 

`--regions strings                  Deploy to machines only in these regions. Multiple regions can be specified with comma separated values or by providing the flag multiple times. --region iad,sea --regions syd will deploy to all three iad, sea, and syd regions. Applied before --exclude-regions. V2 machines platform only.`

jroes:
Just FYI - this kind of change is breaking, and can bust CI/CD workflows, which I just experienced. Might be nice to alias the old behavior or something. (I'm personally using a fork of https://github.com/superfly/fly-pr-review-apps)

wweaver:
So this means `--region` is no longer a valid flag. Does that mean this build action needs to be updated? https://github.com/superfly/fly-pr-review-apps. We started getting failed builds using a build action that extends that action.

alexspeller:
This also broke our CI workflow - and this is the second time this has happened in the past couple of months - please can you consider some kind of formal deprecation process with release notes to users. It's really hard to rely on fly for our production environments at the moment.

supriyo-biswas:
Adding another comment about breakage to our CI deployment process as well.

In fact, in this case `--regions` behaves almost the same as `--region`, so I don't see why the deprecation needed to happen at all.

On my side, I guess I'm just gonna have to pin the deployment version.

benbjohnson:
Thanks for the feedback everyone and I apologize for the breakage. I'll do a better job handling this in the future. 

> in this case --regions behaves almost the same as --region, so I don't see why the deprecation needed to happen at all.

@supriyo-biswas The `--region` flag on `fly deploy` changes the primary region that the application runs in whereas `--regions` specifies a list of regions to deploy to. The change was made because updating the primary region can cause very unexpected changes in some apps when all you meant to do was just deploy to a subset of regions.

remorses:
This change is very dangerous, people may simply replace `--region` with `--regions` but the behaviour is completely different and will cause having machines with older image versions running side by side with the new machines.

Imagine what someone that was previously using --region may do when the flag is no longer working:

1. User tries to deploy, `--region` no longer works
2. Finds the new `--regions` flag, just replaces it
3. The user passes a new region in this flag thinking the command will simply deploy machines in that region, like it did previously
4. The command instead only runs the deployment ignoring regions not passed to `--regions`, causing the old machines to keep running with the old image version, possibly breaking production 

This just happened to me 2 different times breaking prod, I didn't really understand why at first

PR #3516, opened 2024-05-02
Add the --http-failover flag for remote Docker builders

billyb2:
### Change Summary

What and Why:
I added the --http-failover flag, which will automatically retry to determine an image if we have any issues connecting to our org's builder.

How:

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -37,6 +37,7 @@ var CommonFlags = flag.Set{
 	flag.LocalOnly(),
 	flag.Push(),
 	flag.Wireguard(),
+	flag.HttpFailover(),
 	flag.Detach(),
 	flag.Strategy(),
 	flag.Dockerfile(),
@@ -248,10 +249,12 @@ func run(ctx context.Context) error {
 }
 
 func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, forceYes bool) (err error) {
+	ctx, span := tracing.GetTracer().Start(ctx, "deploy_with_config")
 	io := iostreams.FromContext(ctx)
 	appName := appconfig.NameFromContext(ctx)
 	apiClient := fly.ClientFromContext(ctx)
 	appCompact, err := apiClient.GetAppCompact(ctx, appName)
+
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


internal/command/deploy/deploy.go
```diff
@@ -248,6 +249,9 @@ func run(ctx context.Context) error {
 }
 
 func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, forceYes bool) (err error) {
+	ctx, span := tracing.GetTracer().Start(ctx, "deploy_with_config")
+	defer span.End()
```
gwuah, line 253:
```suggestion

```
You don't need to create a new span. you can just pull the existing span from context and add your stuff on it.
https://pkg.go.dev/go.opentelemetry.io/otel/trace#SpanFromContext

internal/command/deploy/deploy.go
```diff
@@ -263,8 +267,17 @@ func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, forceYes
...
+	httpFailover := flag.GetHTTPFailover(ctx)
+	usingWireguard := flag.GetWireguard(ctx)
+
 	// Fetch an image ref or build from source to get the final image reference to deploy
-	img, err := determineImage(ctx, appConfig)
+	img, err := determineImage(ctx, appConfig, usingWireguard)
+	if err != nil && usingWireguard && httpFailover {
+		span.SetAttributes(attribute.String("builder.failover_error", err.Error()))
+		span.AddEvent("using http failover")
+		img, err = determineImage(ctx, appConfig, false)
```
gwuah, line 278: this is really interesting. 

but, there are so many reasons why `determineImage` could fail, not just because of wireguard.

does it make sense to only fail over if a specific error happens(tunnel unavailble, hearbeat issues, etc) or can we just proceed as is? (i'm just wondering)

billyb2: hm we could go more in depth to only fail over on specific issues, but i think this is fine for two reasons:
1. determineImage is pretty atomic, I can't think of any "in-between" states an image could be in
2. because of that, even if the error isn't related to wireguard, retrying really can't hurt, to see if the error was just transient.

if you disagree though, I have no issue only failing over on specific errors :)

gwuah: no hard opinions actually, lets ship.

gwuah (approved):

PR #3517, opened 2024-05-02
Add the Arcjet extension commands

danni-popova:
Adding Arcjet extension commands to enable installations of Arcjet through the Fly CLI


internal/command/extensions/arcjet/list.go
```diff
@@ -0,0 +1,58 @@
...
+
+	for _, addon := range response.AddOns.Nodes {
+		rows = append(rows, []string{
+			addon.Name,
+			addon.Organization.Slug,
+			addon.PrimaryRegion,
+		})
+	}
+
+	_ = render.Table(out, "", rows, "Name", "Org", "Primary Region")
```
jsierles, line 55: Arcjet is region-less, so we should remove the `Primary Region` here.

internal/command/extensions/arcjet/arcjet.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/superfly/flyctl/internal/command"
+)
+
+func New() (cmd *cobra.Command) {
+
+	const (
+		short = "Protect your applications with Arcjet"
```
jsierles, line 12: We should keep this language functional, similar to what we have now:

"Provision and manage Arcjet WAF sites" or similar. We can add a longer description in the long help which should display with `fly ext arcjet`.

jsierles: Note that we want to improve visibility of extensions, but will do that in a separate project we have going on.

jsierles (changes requested):

internal/command/extensions/arcjet/arcjet.go
davidmytton: Let’s use “Provision and manage Arcjet”. We don’t want to use “WAF” because that’s only a small component of what we’re doing, and not what we lead with anywhere.

jsierles (approved):

jsierles:
I think one more thing required - adding the missing commands under `fly ext arcjet`

danni-popova:
@jsierles Added the missing `status` and `destroy` commands

PR #3521, opened 2024-05-03
Add snapshot retention to mount section of fly.toml

tvdfly:
### Change Summary

Add snapshot_retention to the mounts section of fly.toml, so you can customize how long we retain volume snapshots from the app.

---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/appconfig/validation.go
+++ b/internal/appconfig/validation.go
@@ -275,6 +275,14 @@ func (cfg *Config) validateMounts() (extraInfo string, err error) {
 			}
 		}
 
+		switch {
+		case m.SnapshotRetention == 0:
+			m.SnapshotRetention = 5
+		case m.SnapshotRetention < 1 || m.SnapshotRetention > 60:
+			extraInfo += fmt.Sprintf("mount '%s' has a snapshot_retention value which is not between 1 and 60 days\n", m.Source)
+			err = ValidationError
+		}
+
 		var autoExtendSizeIncrement, autoExtendSizeLimit int
 		var vErr error
 		if m.AutoExtendSizeIncrement != "" {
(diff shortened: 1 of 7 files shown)


internal/appconfig/validation.go
```diff
@@ -275,6 +275,14 @@ func (cfg *Config) validateMounts() (extraInfo string, err error) {
 			}
 		}
 
+		switch {
+		case m.SnapshotRetention == 0:
+			m.SnapshotRetention = 5
+		case m.SnapshotRetention < 1 || m.SnapshotRetention > 60:
+			extraInfo += fmt.Sprintf("mount '%s' has a snapshot_retention value which is not between 1 and 60 days\n", m.Source)
```
benbjohnson, line 282: Unless you specifically want single-quotes, you can also use `%q` to print a quoted value.

benbjohnson (approved):
:+1

internal/appconfig/validation.go
```diff
@@ -275,6 +275,14 @@ func (cfg *Config) validateMounts() (extraInfo string, err error) {
 			}
 		}
 
+		switch {
+		case m.SnapshotRetention == 0:
+			m.SnapshotRetention = 5
```
dangra, line 280: hmm, mutating the configuration while validating it is not the right place IMHO.

Isn't `0` a valid retention value in case the use wants to disable snapshots?

If the problem is that CreateVolumeRequest uses an `*int` type for SnapshotRetention, I would vote to use the same type for `Mount.SnapshotRetention`. It makes things much easier to propagate and let the API service decide what value to use by default.

internal/appconfig/validation.go
tvdfly: thanks for the tip! all the other validation errors were formatted like this so I'll to keep the format consistent.

internal/appconfig/validation.go
tvdfly: thx, i like that better!

PR #3522, opened 2024-05-03
Prevent users from deploying fly tomls with duplicate service definitions

gwuah:
### Change Summary
Users sometimes end up defining multiple services for the same protocol & port. 
This harmless mistake would cause health-checks to fail because of only one of these services get picked by flaps.
To fix this, I'm introducing some client-side validation to flag these issues.

They would now see this warning/help text 👇 
```
Service [tcp-1738] has 2 duplicate definitions. To resolve this, merge them into 1 service.
```

References
https://community.fly.io/t/health-checks-failing-with-failed-to-get-vm/18077
https://community.fly.io/t/unable-to-perform-health-checks/19604
https://community.fly.io/t/health-checks-always-failing/19641/2

### DIFF
--- a/internal/appconfig/testdata/validate-duplicate-services-diff-processgroup.toml
+++ b/internal/appconfig/testdata/validate-duplicate-services-diff-processgroup.toml
@@ -1,22 +1,18 @@
 app = "foo"
 
-[[services]]
+[processes]
+web = "/run1"
+app = "/run2"
+
+
+[http_service]
   internal_port = 1738
-  protocol = "tcp"
-
-  [[services.ports]]
-    port = 80
-    handlers = ['http']
-    force_https = true
-
-  [services.concurrency]
-    type = "requests"
-    hard_limit = 22
-    soft_limit = 13
+  processes = ["app"]
 
 [[services]]
   internal_port = 1738
   protocol = "tcp"
+  processes = ["web"]
 
   [[services.ports]]
     port = 80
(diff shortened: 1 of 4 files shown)


lubien:
@gwuah do you know if this validation also checks if a http_service and a regular service use the same port? I feel like this could be a common source of misconception.

gwuah:
@lubien yes it does. The thing is, http services are actually tcp services. 
They are converted to tcp services in flyctl before we send them to flaps.

lubien (approved):
LGTM!

dangra (commented):
+1 to adding more service validations

A few Q:
1. What happens when the similar services are for different process groups? 
2. Should it validate for external port duplication too?

gwuah:
@dangra I think the process group remark is valuable, so I've added support for that. https://github.com/superfly/flyctl/pull/3522/commits/3a0b951309d0e3de6037db41101c8c2d974afc22
Duplicate external ports however, I think we can ignore them for now and merge this.

internal/appconfig/testdata/validate-duplicate-services-diff-processgroup.toml
```diff
@@ -0,0 +1,25 @@
...
+
+
+[http_service]
+  internal_port = 1738
+  processes = ["app"]
+
+[[services]]
+  internal_port = 1738
+  protocol = "tcp"
+  processes = ["web"]
```
dangra, line 15: I would add a more complete case,  for example a service that belongs to two process groups, and then another service that only belong to one of them.  Both services listen on tcp-1738, and the validation must fail only for the process group that is common to both services.

PR #3531, opened 2024-05-07
More flags for fine-tuning deploy

btoews:
### Change Summary

What and Why:

This PR adds `--skip-release-command` and `--only-machines`/`--exclude-machines` flags to `fly deploy`.

Being able to deploy to a subset of machines allows users to conduct their own blue/green deploys at their own pace. I will find it very useful to be able to deploy to a single machine in a big app for half an hour to see if things break.

Being able to skip release commands is useful in this situation also, where you wouldn't, for example, want to run database migrations when testing a small change.

How:

Added flags.

Related to:

Apropos of nothing

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -121,11 +121,23 @@ var CommonFlags = flag.Set{
 	flag.StringSlice{
 		Name:        "regions",
 		Aliases:     []string{"only-regions"},
-		Description: "Deploy to machines only in these regions. Multiple regions can be specified with comma separated values or by providing the flag multiple times. --region iad,sea --regions syd will deploy to all three iad, sea, and syd regions. Applied before --exclude-regions. V2 machines platform only.",
+		Description: "Deploy to machines only in these regions. Multiple regions can be specified with comma separated values or by providing the flag multiple times.",
 	},
 	flag.StringSlice{
 		Name:        "exclude-regions",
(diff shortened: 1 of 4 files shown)


internal/command/deploy/deploy.go
```diff
@@ -121,11 +121,23 @@ var CommonFlags = flag.Set{
 	flag.StringSlice{
 		Name:        "regions",
 		Aliases:     []string{"only-regions"},
-		Description: "Deploy to machines only in these regions. Multiple regions can be specified with comma separated values or by providing the flag multiple times. --region iad,sea --regions syd will deploy to all three iad, sea, and syd regions. Applied before --exclude-regions. V2 machines platform only.",
```
btoews, line 124: > --region iad,sea --regions syd will deploy to all three iad, sea, and syd regions.

I thought this part was kind of obvious. I can add it back if others find it helpful.

> Applied before --exclude-regions.

It isn't clear what this means. All the filters are always applied. Order shouldn't matter.

> V2 machines platform only

We shouldn't need language like this now that everything's on machines, right?

internal/command/deploy/deploy.go
```diff
@@ -365,19 +378,29 @@ func deployToMachines(
 		return err
 	}
 
-	excludeRegions := make(map[string]interface{})
-	for _, r := range flag.GetStringSlice(ctx, "exclude-regions") {
-		reg := strings.TrimSpace(r)
-		if reg != "" {
-			excludeRegions[reg] = struct{}{}
-		}
+	excludeRegions := make(map[string]bool)
```
btoews, line 381: `map[string]bool` is slightly wasteful compared to `map[string]interface{}`, but the ergonomics are so much better for this kind of work.

dangra (approved):

PR #3532, opened 2024-05-08
Remove homebrew bump workflow

tvdfly:
Our homebrew version bump is failing because we're now in the autobump list. afaict this is a good thing.

Every 3 hours, homebrew runs an action to update all formulae in the autobump list. That's better than the daily update we were doing, yay!
https://github.com/Homebrew/homebrew-core/pull/168397
https://github.com/Homebrew/homebrew-core/blob/ef39119111a7a4786b66af3c4921cd05d5c642fc/.github/workflows/autobump.yml

We no longer need our own homebrew bump workflow. 🎉

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


clouvet:
> Every 3 hours, homebrew runs an action to update all formulae in the autobump list.

How's that work with Homebrew's [throttle](https://github.com/Homebrew/homebrew-core/blob/42d0306dd6dad595426f0358971c0bfbb369d686/Formula/f/flyctl.rb#L15C1-L19C6)? 
```
  livecheck do
    url :stable
    strategy :github_latest
    throttle 5
  end
  ```
I wasn't sure what the interval indicated and couldn't find it from the docs, so I asked ChatGPT:
> The `throttle` value in the `livecheck` block does not directly translate to real-time hours but rather acts as a counter for the number of `brew bump` runs. If `brew bump` runs every 3 hours, and your `livecheck` block has `throttle 5`, the throttle isn't counting hours directly but instead counts these `brew bump` executions.

clouvet:
I've been keeping an eye on when Homebrew updates and now I think I understand how it works. It seems like the throttle value corresponds to patch levels, so that means flyctl will get bumped on Homebrew after every 5th patch-level update.

PR #3537, opened 2024-05-09
fix: fly deploy --regions should be honored

kzys:
The flag seems ignored and only primary_region in fly.toml is considered.

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


kzys:
The motivation of the original change was discussed in #3509 and #3514. I'm gonna discuss about it with others internally.

PR #3539, opened 2024-05-10
The default is now rolling

matthewlehner:
### Change Summary

What and Why:
Updated the flag text for `fly deploy --strategy` to be correct now that we're in a world without nomad.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


billyb2 (approved):
lgtm

PR #3541, opened 2024-05-11
Fix Postgres plan display on `fly launch`

andreolvr:
### Change Summary

What and Why:

On `fly launch`, we were suggesting Postgres instances with 1GB of RAM, which is not an option on the Launch UI or on `fly postgres create`. This was causing confusing on the text being displayed as seen on the image below, and was causing the Launch UI to open with the wrong Postgres configs, as the UI is not aware of a possible Postgres config with 1GB of RAM.

Before:
Screenshot 2024-05-08 at 20 30 37

\* Notice that we were displaying "256MB (1GB RAM)", as the guest machine has 256MB, but Postgres was configured as 1GB. Just a confusing experience overall. The postgres RAM for `fly launch` is now matching the guest machine's RAM.

After:
Screenshot 2024-05-10 at 22 41 13

---

### Documentation

- [X] Fresh Produce (soon, with a couple other related changes)
- [ ] n/a


alichay (approved):
nice catch!

PR #3544, opened 2024-05-12
Make fly config validate detect unknown fields

rubys:
While we have a number of validators that will check to make sure that field _values_ are correct, we surprisingly do not check to make sure theat field _names_ are correct.  Instead we silently ignore fields that have unknown names, often leading to difficult to debug situations.

This change makes use of `decoder.DisallowUnknownFields` to enable reporting on these errors by fly config validate.

### DIFF
--- a/internal/appconfig/patches.go
+++ b/internal/appconfig/patches.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/spf13/viper"
+	"github.com/superfly/flyctl/terminal"
 )
 
 type patchFuncType func(map[string]any) (map[string]any, error)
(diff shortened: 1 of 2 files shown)


rubys:
Amusingly, one of the CI tests is now failing due to an invalid `fly.toml`:

image

What is the current thinking on validation checks?  Adding a validation check can break future deploys.  Should it (in general)?  Is this specific case different?  We can make unknown field validation messages a warning only by capturing and resetting `err` inside of `mapToConfig` when running validation as a part of `fly deploy`.

dangra:
> What is the current thinking on validation checks? Adding a validation check can break future deploys.

that is my main concern, every time we change something to be more strict, it breaks users' automated deployments in some way. I think we have to start by showing big warnings first.

internal/appconfig/patches.go
```diff
@@ -37,7 +39,21 @@ func mapToConfig(cfgMap map[string]any) (*Config, error) {
 	if err != nil {
 		return cfg, err
 	}
-	return cfg, json.Unmarshal(newbuf, cfg)
+
+	decoder := json.NewDecoder(bytes.NewReader(newbuf))
+	decoder.DisallowUnknownFields()
+	err = decoder.Decode(cfg)
```
dangra, line 45: can we add a test case and its test file at `internal/appconfig/testdata/` ? 

I wonder if `cfg` is "filled" with the rest of the values even when there is an error due to unknown fields.

PR #3545, opened 2024-05-13
support fly config in json and yaml formats

rubys:
`fly launch` using the configuration of your choice -- toml remains the default, but `--json` or `--yaml` can be specified to select an alternate format.

Or convert at any time using `fly config show --local` and specify `--yaml` or `--toml` and redirect the output to a file with the correct file extension.  (json remains the default for this command).  Just remember to delete the other representation -- the search order is fly.toml, then fly.json, then fly.yaml.

`fly config save` can also be used (it defaults to toml).


dangra (commented):
Come to say I love the work you are doing here, adding json and yaml support is awesome but please add unit tests. TOML is well tested now, not perfect but covers many legacy stuff that we don't want to break. I'd like to see a similar level of test coverage for yaml and json.

rubys:
@dangra If you look in [testdata](https://github.com/superfly/flyctl/tree/74d6937680e00b4bf90c4259abc57616fc4537f9/internal/appconfig/testdata), you will now see full-reference in all three formats.  There were produced using `bin/flyctl config show --local`.  And there now are [tests](https://github.com/superfly/flyctl/blob/74d6937680e00b4bf90c4259abc57616fc4537f9/internal/appconfig/serde_test.go#L606-L634) that loading these configurations produce identical configurations as the original toml file.

This demonstrates that configurations can round-trip in all three formats.

Any other suggesttions

internal/appconfig/serde.go
```diff
@@ -66,7 +86,7 @@ func (c *Config) WriteToFile(filename string) (err error) {
 		}
 	}()
 
-	_, err = c.WriteTo(file)
+	_, err = c.WriteTo(file, filepath.Ext(filename))
```
dangra, line 89: honest question: how does it work on Windows these days? is a filename like MYDATA.JSON possible ? thinking on uppercase extension names

internal/command/config/save.go
```diff
@@ -65,6 +74,12 @@ func runSave(ctx context.Context) error {
 		return err
 	}
 
+	if flag.GetBool(ctx, "json") {
+		configfilename = strings.TrimSuffix(configfilename, ".toml") + ".json"
+	} else if flag.GetBool(ctx, "yaml") {
+		configfilename = strings.TrimSuffix(configfilename, ".toml") + ".yaml"
```
dangra, line 80: is there a chance that `--config path/to/fly.yaml` generates `path/to/lfy.yaml.yaml` ?

internal/command/launch/launch.go
```diff
@@ -79,8 +80,14 @@ func (state *launchState) Launch(ctx context.Context) error {
 	}
 
 	// Finally write application configuration to fly.toml
-	state.appConfig.SetConfigFilePath(state.configPath)
-	if err := state.appConfig.WriteToDisk(ctx, state.configPath); err != nil {
+	configPath := state.configPath
+	if flag.GetBool(ctx, "json") {
+		configPath = strings.TrimSuffix(configPath, ".toml") + ".json"
+	} else if flag.GetBool(ctx, "yaml") {
+		configPath = strings.TrimSuffix(configPath, ".toml") + ".yaml"
```
dangra, line 87: does it suffer from the same double extension bug?  fly.yaml.yaml ?

internal/appconfig/serde_test.go
```diff
@@ -602,3 +602,33 @@ func TestIsSameTOMLAppConfigReferenceFormat(t *testing.T) {
...
 	require.Equal(t, cfg, actual)
 }
+
+// bin/flyctl config show --local -c internal/appconfig/testdata/full-reference.toml > internal/appconfig/testdata/full-reference.json
+func TestIsSameJSONAppConfigReferenceFormat(t *testing.T) {
+	const TOMLpath = "./testdata/full-reference.toml"
+	TOMLcfg, err := LoadConfig(TOMLpath)
+	require.NoError(t, err)
+
+	const JSONpath = "./testdata/full-reference.json"
```
dangra, line 612: Sam, feel free to ignore this nit, here it goes:

As far as I see, we are not testing `WriteTo()` and (I know I asked for it) by adding the json and yaml versions of full-reference, we have another set of files to maintain which is good and even after below suggestion I don't think we have to change that. 

All that to say we should be serializing the unmarshaled toml into toml, json and yaml; and then loading from this serialized files to compare the configs.

dangra (approved):
💯

internal/appconfig/serde_test.go
rubys: Not a nit.

I removed those new files, and changed the existing tests to start from the full-reference.toml file and write the config out and read it back in for both json and yaml formats.... and found a bug.  I've been focusing mostly on `fly config show` and visual inspection, but `WriteTo` will add an extra comment at the top - and JSON does not allow comments.   So by adding a test, I found and fixed a but.

Thanks!

dangra (approved):
:shipit:

PR #3546, opened 2024-05-14
Set HTTPS Failover to true by defualt

billyb2:
### Change Summary

What and Why:
This commit does two things:
1. We're now using HTTP failover by default, and seeing if it helps with remote builder connections. The original plan was to use flyctl feature flags, but plans change ¯\_(ツ)_/¯
2. We're renaming HTTP failover to HTTPS failover, and keeping http-failover as an alias. It's a bit more accurate to what's happening

How:

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a


dangra (approved):
Looks good!

PR #3548, opened 2024-05-14
Configure Postgres for Phoenix apps on `fly launch`

andreolvr:
### Change Summary

I created a new Phoenix app with `mix phx.new app --database postgres` and `fly launch` failed since we are not putting Postgres on the launch plan.

This PR sets the `s.DatabaseKind` property, which puts Postgres on the launch plan and makes the launch succeed.

---

### Documentation

- [X] Fresh Produce
- [ ] n/a

### DIFF
--- a/scanner/phoenix.go
+++ b/scanner/phoenix.go
@@ -108,11 +108,16 @@ a Postgres database.
 `
 	}
 
-	// Add migration task if we find one of the dependencies that would run migrations.
-	// They are listed here: https://github.com/elixir-ecto/ecto?tab=readme-ov-file#usage
-	if checksPass(sourceDir, dirContains("mix.exs", "postgrex", "myxql", "tds")) {
+	if checksPass(sourceDir, dirContains("mix.exs", "postgrex")) {
+		s.DatabaseDesired = DatabaseKindPostgres
+		s.ReleaseCmd = "/app/bin/migrate"
+	} else if checksPass(sourceDir, dirContains("mix.exs", "myxql")) {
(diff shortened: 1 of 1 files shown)


scanner/phoenix.go
```diff
@@ -108,11 +108,16 @@ a Postgres database.
...
-	// They are listed here: https://github.com/elixir-ecto/ecto?tab=readme-ov-file#usage
-	if checksPass(sourceDir, dirContains("mix.exs", "postgrex", "myxql", "tds")) {
+	if checksPass(sourceDir, dirContains("mix.exs", "postgrex")) {
+		s.DatabaseDesired = DatabaseKindPostgres
+		s.ReleaseCmd = "/app/bin/migrate"
+	} else if checksPass(sourceDir, dirContains("mix.exs", "myxql")) {
+		s.DatabaseDesired = DatabaseKindMySQL
+		s.ReleaseCmd = "/app/bin/migrate"
+	} else if checksPass(sourceDir, dirContains("mix.exs", "tds")) {
 		s.ReleaseCmd = "/app/bin/migrate"
```
andreolvr, line 118: I'm not sure if we should set `s.ReleaseCmd` for MySQL and MSSQL databases, since we don't support them and the release command will fail. I just did this to keep the previous behavior, but I guess we should remove it?

benbjohnson: Yeah, if we don't support them then we probably shouldn't add handling for them.

benbjohnson (approved):
lgtm minus the mysql/mssql comment.

PR #3555, opened 2024-05-16
Change fly m clone default to unique zone

IohannesArnold:
Previously, `fly machines clone` would default to allowing the clone Machine to be placed on the same host as another Machine in the app. This keeps the flag but changes the default to forbidding this.

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


dangra:
dupe https://github.com/superfly/flyctl/pull/3554

PR #3558, opened 2024-05-17
Help docs: updates to volumes and apps command help

andie787:
### Change Summary

What and Why:

- volumes and apps help needed a few updates/fixes

How:

- remove "min 5" from snapshot retention help
- clean up apps command help
- make volumes help consistent

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/apps/apps.go
+++ b/internal/command/apps/apps.go
@@ -17,11 +17,8 @@ import (
 // New initializes and returns a new apps Command.
 func New() *cobra.Command {
 	const (
-		long = `The APPS commands focus on managing your Fly applications.
-Start with the CREATE command to register your application.
-The LIST command will list all currently registered applications.
-`
-		short = "Manage apps"
+		long = "Manage your Fly applications."		
+		short = "Manage apps."
 	)
 
 	// TODO: list should also accept the --org param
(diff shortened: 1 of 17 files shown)


internal/command/apps/apps.go
```diff
@@ -17,11 +17,8 @@ import (
 // New initializes and returns a new apps Command.
 func New() *cobra.Command {
 	const (
-		long = `The APPS commands focus on managing your Fly applications.
-Start with the CREATE command to register your application.
-The LIST command will list all currently registered applications.
-`
-		short = "Manage apps"
+		long  = "Manage your Fly applications."
+		short = "Manage apps."
```
dangra, line 21: I don't know what is our stance regarding period at the end of sentences but we recently saw this happening https://github.com/superfly/flyctl/pull/3552

andie787: ok good to know thanks! the proposed style guide says to have a period for the short desc, but the way of least resistance might be to just go with whatever is already there 😆

andie787: @dangra I looked at this again, and it doesn't make sense to remove the period from short descriptions for top level or sub-commands. Especially when the long desc is "short + <more words>". The PR you linked to is not a good change for this reason: https://fly.io/docs/flyctl/machine-clone/

Having no period for option descriptions is fine though.

PR #3559, opened 2024-05-20
Some Remote Builder Heartbeat Improvements

billyb2:
### Change Summary

What and Why:
This PR extends some timeouts, removes some unnecessary operations, and retries ones that can fail an entire deploy.

How:
We extended some timeouts to make sure that a blip in internet doesn't kill a deployment. Also, for whatever reason we were doing two heartbeat operations that could possibly fail, instead of just one. Just removing the second 'could fail' heartbeat operation should help deployment success rates. Finally, we're now retrying the first heartbeat up to 3 times, to try and make sure that we can actually communicate with the remote builder.

In the future, I'd like to either
a. understand why remote builders just randomly fail to respond to heartbeats or
b. automatically destroy remote builders if they aren't functioniong.

Related to:

---

### Documentation

- [x] Fresh Produce (soon, I want to make sure these changes reduce failure rates first)
- [ ] n/a

### DIFF
--- a/internal/build/imgsrc/resolver.go
+++ b/internal/build/imgsrc/resolver.go
@@ -590,7 +590,7 @@ func heartbeat(ctx context.Context, client *dockerclient.Client, req *http.Reque
 	ctx, span := tracing.GetTracer().Start(ctx, "heartbeat")
 	defer span.End()
 
-	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
 
 	resp, err := client.HTTPClient().Do(req.Clone(ctx))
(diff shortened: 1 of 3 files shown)


internal/build/imgsrc/resolver.go
```diff
@@ -145,6 +146,7 @@ func (di DeploymentImage) ToSpanAttributes() []attribute.KeyValue {
 type Resolver struct {
 	dockerFactory *dockerClientFactory
 	apiClient     flyutil.Client
+	heartbeatFn   func(ctx context.Context, client *dockerclient.Client, req *http.Request) error
```
rugwirobaker, line 149: Paused here for a moment on the use of `ctx` vs `req.Context()` but I'm guessing we want to use ctx on purpose. LGTM

rugwirobaker (approved):
LGTM

PR #3561, opened 2024-05-21
make "make test" pass again on MacOS

rubys:
lo.Ternary doesn't "short circuit" evaluation - all operands are evaluated, then one is selected.

As an alternative, [lo.TernaryF](https://github.com/samber/lo?tab=readme-ov-file#ternaryf) can be used, but I opted to make this explicit instead.

### DIFF
--- a/internal/appconfig/machines.go
+++ b/internal/appconfig/machines.go
@@ -150,7 +150,12 @@ func (c *Config) ToTestMachineConfig(svc *ServiceMachineCheck, origMachine *fly.
 	if c.PrimaryRegion != "" {
 		mConfig.Env["PRIMARY_REGION"] = c.PrimaryRegion
 	}
-	mConfig.Env["FLY_TEST_MACHINE_IP"] = lo.Ternary(origMachine == nil, "", origMachine.PrivateIP)
+
+	if origMachine == nil {
+		mConfig.Env["FLY_TEST_MACHINE_IP"] = ""
+	} else {
+		mConfig.Env["FLY_TEST_MACHINE_IP"] = origMachine.PrivateIP
+	}
 
 	// Use the stop config from the app config by default
 	c.tomachineSetStopConfig(mConfig)


internal/appconfig/machines.go
```diff
@@ -150,7 +150,12 @@ func (c *Config) ToTestMachineConfig(svc *ServiceMachineCheck, origMachine *fly.
 	if c.PrimaryRegion != "" {
 		mConfig.Env["PRIMARY_REGION"] = c.PrimaryRegion
 	}
-	mConfig.Env["FLY_TEST_MACHINE_IP"] = lo.Ternary(origMachine == nil, "", origMachine.PrivateIP)
+
+	if origMachine == nil {
+		mConfig.Env["FLY_TEST_MACHINE_IP"] = ""
+	} else {
```
dangra, line 156: no need for the `nil` case, an unset FLY_TEST_MACHINE_IP is better imo /cc @billyb2 

I can't understand why the test doesn't fail on linux 🤷

rubys: The test will fail if you do that:

```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,5 +1,6 @@
        	            	 (*fly.MachineConfig)({
        	            	- Env: (map[string]string) (len=3) {
        	            	+ Env: (map[string]string) (len=4) {
        	            	   (string) (len=17) "FLY_PROCESS_GROUP": (string) (len=28) "fly_app_test_machine_command",
        	            	   (string) (len=16) "FLY_TEST_COMMAND": (string) (len=1) "1",
        	            	+  (string) (len=19) "FLY_TEST_MACHINE_IP": (string) "",
        	            	   (string) (len=14) "PRIMARY_REGION": (string) (len=3) "mia"
        	Test:       	TestToTestMachineConfigWKillInfo
```

dangra: @billyb2 is the empty FLY_TEST_MACHINE_IP expected when there is no machine to point to?

billyb2: @dangra yea it is

dangra: and what is the point of an empty value compared to not setting it?

rubys: Perhaps this can be resolved in a subsequent PR?  It is annoying to have the tests fail... and the underlying issue is not in test code but in code that runs in production.

billyb2: > and what is the point of an empty value compared to not setting it?

i don't have strong opinions either way. I just thought explicitly setting it to empty was the right move at the time?

billyb2: > Perhaps this can be resolved in a subsequent PR? It is annoying to have the tests fail... and the underlying issue is not in test code but in code that runs in production.

that sounds good (btw I haven't announced the feature yet, so unsetting this will likely not cause any harm)

rubys: Merged as is, which leaves the current behavior (setting `mConfig.Env["FLY_TEST_MACHINE_IP"] = ""`) intact.  I have no opinion on that subject, all I was proposing was that that changing that behavior could be the subject of a separate pull request.

PR #3568, opened 2024-05-23
name wg peers "static-" when created manually.

timflyio:
- when a wg peer name is not specified, name it "interactive-*" or "static-*" based on how it was created.

### Change Summary

What and Why: When a wg peer name is not specified, name it "interactive-*" or "static-*" based on how it was created. We were previously giving default names of "interactive-" even for peers explicitly created by customers on the command line. We want to reserve the "interactive-" prefix for peers that are auto-generated and can be deleted at will without causing disruption.

How: Internally select the prefix for default names to be "interactive-" for automatically generated peers or "static-" for explicitly generated peers.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/wireguard/wireguard.go
+++ b/internal/command/wireguard/wireguard.go
@@ -144,7 +144,7 @@ func runWireguardCreate(ctx context.Context) error {
 	// TODO: allow custom network
 	network := ""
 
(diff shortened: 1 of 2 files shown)


btoews (approved):
LGTM. Thanks for doing this!

PR #3572, opened 2024-05-24
Ensure that we only run tracing.RecordError when error is nil

billyb2:
### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


matthewlehner (approved):
🤩

PR #3574, opened 2024-05-27
Reset the client before re-launching

rugwirobaker:


benbjohnson:
@rugwirobaker Can you add a description to the PR about why the change was made to give context?

PR #3579, opened 2024-05-28
wg: reuse peer from any region if not specific region is requested

btoews:
Every time the agent was restarted we'd create a new wg peer (unless `FLY_AGENT_WG_REGION` was explicitly set). This isn't necessary.

### DIFF
--- a/internal/wireguard/wg.go
+++ b/internal/wireguard/wg.go
@@ -41,12 +41,12 @@ func generatePeerName(ctx context.Context, apiClient flyutil.Client) (string, er
 	return name, nil
 }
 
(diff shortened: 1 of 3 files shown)


timflyio (approved):
had to dig for context to understand what the patch does, but looks good to me.

internal/wireguard/wg.go
```diff
@@ -46,7 +46,7 @@ func StateForOrg(ctx context.Context, apiClient flyutil.Client, org *fly.Organiz
 	if err != nil {
 		return nil, err
 	}
-	if state != nil && !recycle && state.Region == regionCode {
+	if state != nil && !recycle && (regionCode == "" || state.Region == regionCode) {
```
timflyio, line 49: `recycle` here really means `re-establish`

PR #3580, opened 2024-05-28
Set the default PG certificate lifespan to 100 years

davissp14:
### Change Summary

**What and Why:**
The current default PG certificate lifespan is only a couple of days.  This causes issues when users attempt to issue a manual failover, or use internal repmgr commands that require cross-member ssh access.  This PR extends the default certificate lifespan to 100 years.


DAlperin (approved):
Seems sane

PR #3581, opened 2024-05-28
Adding pg command that will allow certificate renewal

davissp14:
### Change Summary

**What and Why:**
Added a new PG command that allows users to renew their SSH certificates.  The default certificate expiration was set to a couples of days and once expired will prevent manual failovers and some internal repmgr commands from working as expected.  

Renewing certificates is pretty straight forward and just requires running the following command, followed by a app deploy.

```
Usage:
  fly postgres renew-certs [flags]

Flags:
  -a, --app string       Application name
  -c, --config string    Path to application configuration file
  -h, --help             help for renew-certs
      --valid-days int   The number of days the certificate should be valid for. (default 36525)
```  

To verify whether your certificates have expired or not, you can run the following:

```
fly ssh console --app <app-name>

ssh-keygen -L -f ~/.ssh/id_rsa-cert.pub

...
Valid: from 2024-05-28T18:35:56 to 2025-05-28T19:35:56
...
```

### DIFF
--- a/internal/command/postgres/postgres.go
+++ b/internal/command/postgres/postgres.go
@@ -32,6 +32,7 @@ func New() *cobra.Command {
 		newDb(),
 		newDetach(),
 		newList(),
+		newRenewSSHCerts(),
 		newRestart(),
 		newUsers(),
 		newFailover(),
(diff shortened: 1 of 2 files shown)


btoews (approved):
LGTM. Should we use a longer default expiration when creating a new pg cluster too?

davissp14:
> Should we use a longer default expiration when creating a new pg cluster too?

Yep!  I added that here: https://github.com/superfly/flyctl/pull/3580

PR #3594, opened 2024-06-04
Even more remote heartbeat improvements

billyb2:
### Change Summary

What and Why:
Having the logic for creating a remote builder be in `web` doesn't make sense. The majority of the work can be done in `flyctl`, and as such, remove another point of failure in a different component

How:
Move all of the logic from EnsureMachineRemoteBuilder from web to flyctl

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a


internal/build/imgsrc/ensure_builder.go
```diff
@@ -0,0 +1,325 @@
...
+		span.AddEvent("builder app already exists and is valid")
+		return builderMachine, builderApp, nil
+	}
+
+	var validateBuilderErr ValidateBuilderError
+	if !errors.As(err, &validateBuilderErr) {
+		return nil, nil, err
+	}
+
+	if validateBuilderErr != NoBuilderApp {
```
benbjohnson, line 64: I believe this block won't occur because `NoBuilderApp` is an instance of `ValidateBuilderError` which is checked with the `errors.As()` above, correct?

internal/build/imgsrc/ensure_builder.go
```diff
@@ -0,0 +1,325 @@
...
+
+	if app == nil {
+		tracing.RecordError(span, NoBuilderApp, "no builder app")
+		return nil, NoBuilderApp
+	}
+
+	flapsClient := flapsutil.ClientFromContext(ctx)
+
+	var volumes []fly.Volume
+	numRetries := 0
```
benbjohnson, line 122: We use a "backoff" library elsewhere for retries. Maybe that's better to use than rolling your own?

internal/build/imgsrc/ensure_builder.go
```diff
@@ -0,0 +1,325 @@
...
+			if numRetries >= 3 {
+				tracing.RecordError(span, err, "error listing machines")
+				return nil, err
+			}
+			time.Sleep(1 * time.Second)
+		} else {
+			tracing.RecordError(span, err, "error listing machines")
+			return nil, err
+		}
+	}
```
benbjohnson, line 178: The volume handling and the machine handling could be split off into their own smaller functions as this function is getting kinda large.

internal/build/imgsrc/ensure_builder.go
```diff
@@ -0,0 +1,325 @@
...
+			tracing.RecordError(span, err, "error creating volume")
+			return nil, nil, err
+		}
+	}
+
+	defer func() {
+		if err != nil {
+			span.AddEvent("cleaning up new volume due to error")
+			flapsClient.DeleteVolume(ctx, volume.ID)
+		}
```
benbjohnson, line 267: It might be better to name the return error something like `retErr` so we don't accidentally introduce a bug by shadowing the `err` elsewhere in the future.

internal/build/imgsrc/ensure_builder.go
billyb2: the only issue with the backoff library is that i need to exit early if we don't get a server error, instead of retrying multiple times (for example, if we get a 404, it's likely a user error)

internal/build/imgsrc/ensure_builder.go
billyb2: i don't believe so. errors.As just checks if it's a ValidateBuilderError or not, and we only return before this check if it isn't a validation error.

https://go.dev/play/p/I5_iTNzBwiQ

internal/build/imgsrc/ensure_builder.go
billyb2: done :)

benbjohnson (approved):

Issue #3584, opened 2024-05-31
Feature suggestion: allow launch command to accept -c as in deploy


justin-hackin:
On the forum, [I explained how I did some scripting of configs for multi-env deploys](https://community.fly.io/t/fly-config-in-your-choice-of-formats-toml-yaml-or-json/20017/10?u=justinb). 

An issue with this approach is that it requires the developer to create a temporary `fly.yaml` file, pasting in one of the env-specific configs, then, after the launch runs, copy the modified config back into the env-specific config. This is why I suggest providing a `-c` arg to the launch command from which it will read and to which it will write.

_Aside: I'm not sure if feature requests belong here but there didn't seem to be a section for it in the community forum._

PR #3615, opened 2024-06-11
Config file flag

fliepeltje:
### Change Summary

What and Why:
Adds a `--config` flag to launch to enable managing multiple environments easier.

```sh
fly launch --config fly.dev.toml
```

this also works for `yaml` and `json` files

```sh
fly launch --config fly.dev.json
```

and with the shorthand

```sh
fly launch -c fly.dev.json
```

How:

Related to: #3584

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -82,7 +82,8 @@ func (state *launchState) Launch(ctx context.Context) error {
 	}
 
 	// Finally write application configuration to fly.toml
-	configPath := state.configPath
+	configDir := filepath.Dir(state.configPath)
+	configPath := filepath.Join(configDir, flag.GetString(ctx, "config-file"))
 	if flag.GetBool(ctx, "json") {
 		configPath = strings.TrimSuffix(configPath, filepath.Ext(configPath)) + ".json"
 	} else if flag.GetBool(ctx, "yaml") {
--- a/internal/command/launch/launch.go
+++ b/internal/command/launch/launch.go
@@ -84,9 +84,9 @@ func (state *launchState) Launch(ctx context.Context) error {
 	// Finally write application configuration to fly.toml
 	configDir := filepath.Dir(state.configPath)
 	configPath := filepath.Join(configDir, flag.GetString(ctx, "config-file"))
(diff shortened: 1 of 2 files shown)


internal/command/launch/launch.go
```diff
@@ -82,10 +84,15 @@ func (state *launchState) Launch(ctx context.Context) error {
 	}
 
 	// Finally write application configuration to fly.toml
-	configPath := state.configPath
-	if flag.GetBool(ctx, "json") {
+	configDir := filepath.Dir(state.configPath)
```
rubys, line 87: Minor suggestion: 

```go
configDir, configFile := filepath.Split(state.configPath)
```

And then later:

```go
configPath = filepath.Join(configPath, configFile)
```

The concern is that at some point in the future, the default format may change, and the person making that change may miss the `fly launch --config dirname` path...

rubys (approved):
I've left a suggestion for you to consider; but approving.  In retrospect an obvious enhancement, and even one that was requested by a user.  Excellent!

internal/command/launch/launch.go
fliepeltje: I think I get what you mean, but I'm not 100% sure :sweat_smile: 

I've pushed an updated version that is a bit cleaner and I think resolves the issue you describe; close the issue if this is the case and otherwise maybe you can give me an example of where this might fail :) 

thanks for the :eyes: !

internal/command/launch/launch.go
```diff
@@ -84,17 +83,20 @@ func (state *launchState) Launch(ctx context.Context) error {
 	}
 
 	// Finally write application configuration to fly.toml
-	configDir := filepath.Dir(state.configPath)
-	configPath := filepath.Join(configDir, flag.GetString(ctx, flagnames.AppConfigFilePath))
-	if fileInfo, err := os.Stat(configPath); err == nil && fileInfo.IsDir() {
-		configPath = filepath.Join(configPath, "fly.toml")
+	configDir, configFile := filepath.Split(state.configPath)
```
rubys, line 86: even better!

rubys (approved):

PR #3623, opened 2024-06-12
Poetry scanner

fliepeltje:
### Change Summary

What and Why:
Building on the stuff used for scanning FastAPI, this now does a similar thing for more python projects using poetry:

- `flask`
- `streamlit`
- `FastAPI`

In addition, tweaks the fastapi config to be more compatible with the framework docs of how to setup the app. 

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


rubys (changes requested):
Very cool.

You still have some cleanup to do to make the build pass.
  * Add a comment to your `defer file.Close()` call to get it not to be flagged.  Example here: https://github.com/superfly/flyctl/blob/96b6ebcded2e939cf445cb414bd6c642b6847b7f/scanner/phoenix.go#L157
  * `listSuffixedFiles` is no longer used so it needs to be removed; you can run lint locally by running `make lint`.  You may need to install `golangci-lint` first.

fliepeltje:
Great catch on the `listSuffixedFiles` - totally forgot about it! 

I have `golangci-lint` installed, but `make lint` is not working well for me; running it directly without `make` is fine though :upside_down_face:

rubys (approved):

PR #3633, opened 2024-06-13
Add `fly machines suspend`

matttpt:
### Change Summary

**What and Why:**

As part of the new suspend/resume feature, add a new `fly machines` command to suspend a Machine.

**How:**

Use the new `Suspend` flaps client method from fly-go, which in turn uses the new suspend endpoint in flaps, and support for the `suspended` state added to the wait endpoint.

This PR upgrades fly-go to v0.1.17.

**Related to:**

* Depends on https://github.com/superfly/fly-go/pull/71

---

### Documentation

I will be in touch with the docs team, and there will be a Fresh Produce when we're ready.

### DIFF
--- a/internal/command/machine/suspend.go
+++ b/internal/command/machine/suspend.go
@@ -58,10 +58,10 @@ func runMachineSuspend(ctx context.Context) (err error) {
 	}
 
 	machines, release, err := mach.AcquireLeases(ctx, machines)
+	defer release()
 	if err != nil {
 		return err
 	}
-	defer release()
 
 	for _, machine := range machines {
 		if err = suspend(ctx, machine, waitTimeout); err != nil {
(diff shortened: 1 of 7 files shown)


internal/command/machine/suspend.go
```diff
@@ -0,0 +1,100 @@
...
+	machines, ctx, err := selectManyMachines(ctx, args)
+	if err != nil {
+		return err
+	}
+
+	machines, release, err := mach.AcquireLeases(ctx, machines)
+	if err != nil {
+		return err
+	}
+	defer release()
```
dangra, line 64: I know it is counterintuitive but `release` func must be called even in case acquiring leases fail to do so.  Otherwise in case of a partial error acquiring a set of leases, the non-failed leases won't be released.

dangra (approved):
Only a small fix required. pre-approving.

PR #3638, opened 2024-06-15
Abort proxy when stdin gets closed

jonatanklosko:
There is no guarantee that a OS process spawning `fly proxy ...` will terminate it, however the stdin is always closed when the parent terminates. In order to avoid zombie processes, we can watch stdin and when it gets closed we terminate the program. The original motivation is spawning fly proxy from Erlang VM.

For more context see a similar PR on esbuild: https://github.com/evanw/esbuild/pull/1449.

### DIFF
--- a/internal/command/proxy/proxy.go
+++ b/internal/command/proxy/proxy.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"os"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -14,6 +16,7 @@ import (
 	"github.com/superfly/flyctl/internal/flag/flagnames"
 	"github.com/superfly/flyctl/internal/flyutil"
 	"github.com/superfly/flyctl/internal/prompt"
+	"github.com/superfly/flyctl/iostreams"
 	"github.com/superfly/flyctl/proxy"
 )
 
@@ -131,5 +134,37 @@ func run(ctx context.Context) (err error) {
 		params.RemoteHost = fmt.Sprintf("%s.internal", appName)
 	}
 
+	watchStdinAndAbortOnClose(ctx)
+
 	return proxy.Connect(ctx, params)
+}
+
+// Asynchronously watches stdin and abort when it closes.
+//
+// There is no guarantee that a OS process spawning the proxy will
+// terminate it, however the stdin is always closed whtn the parent
+// terminates. This way we make sure there are no zombie processes,
+// especially that they hold onto TCP ports.
+//
+// Note that we don't do this when stdin is TTY, because that prevents
+// the process from being moved to a background job on Unix.
+// See https://github.com/brunch/brunch/issues/998.
+func watchStdinAndAbortOnClose(ctx context.Context) {
+	ios := iostreams.FromContext(ctx)
+
+	if !ios.IsStdinTTY() {
+		go func() {
+			// We don't expect any input, but if there is one, we ignore it
+			// to avoid allocating space unnecessarily
+			buffer := make([]byte, 1)
+			for {
+				_, err := ios.In.Read(buffer)
+				if err == io.EOF {
+					os.Exit(0)
+				} else if err != nil {
+					os.Exit(1)
+				}
+			}
+		}()
+	}
 }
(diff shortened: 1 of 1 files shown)


internal/command/proxy/proxy.go
```diff
@@ -131,5 +134,37 @@ func run(ctx context.Context) (err error) {
...
+//
+// There is no guarantee that a OS process spawning the proxy will
+// terminate it, however the stdin is always closed whtn the parent
+// terminates. This way we make sure there are no zombie processes,
+// especially that they hold onto TCP ports.
+//
+// Note that we don't do this when stdin is TTY, because that prevents
+// the process from being moved to a background job on Unix.
+// See https://github.com/brunch/brunch/issues/998.
+func watchStdinAndAbortOnClose(ctx context.Context) {
```
jonatanklosko, line 152: I changed only the `fly proxy` specifically, but there are a few that could probably use the same logic, like `fly machine api-proxy` and `fly redis proxy`. I can enable the watching there too, just let me know what is the best place to put this function to share across commands.

jonatanklosko:
Kindly ping :)

internal/command/proxy/proxy.go
dangra: hi @jonatanklosko, sorry I haven't replied, I saw it days ago and took a step back because my first reaction was "what a gross hack" and preferred to give it some time to settle in my mind 😄 

I mean, it is kind of clever thing to do but also could be a source of gotcha moment for someone that doesn't expect the proxy to terminate out of nowhere. Are you aware of any tool that does this trick?

I think we can work towards merging it anyways but:
* Instead of making it the default, can this terminate-on-stdin-close mode be behind a command line flag? 
* Instead of calling `os.Exit()`, can it cancel the context used by the proxy to gracefully stop it?

jonatanklosko: @dangra no worries, it seems like a hack, but many command line tools actually do that, like `cat`, `fsevents`, `esbuild`, `webpack`, `postcss`.

A flag sounds good to me, that's what webpack does.

Should I share and apply this to other proxies, or do you prefer to keep it for this one specifically?

> can it cancel the context used by the proxy to gracefully stop it?

How would the cancel look like? :)

dangra: > A flag sounds good to me, that's what webpack does.

sgtm!

> Should I share and apply this to other proxies, or do you prefer to keep it for this one specifically?

what other proxies do you refer to?

> How would the cancel look like? :)

I realize this a nitpick and won't make much difference for this case, but I'd derive the context passed to `proxy.Connect()` using [context.WithCancelCause()](https://pkg.go.dev/context#WithCancelCause); and then call the cancel function instead of `os.Exit()`.

That should allow the proxy to gracefully shutdown instead of abruptly exit.

jonatanklosko: > what other proxies do you refer to?

This is `fly proxy`, I noticed the proxy is also used in `fly machine api-proxy` and `fly redis proxy`, so I wonder if all of these should have the flag for consistency.

jonatanklosko: FWIW I only need this for `fly proxy`, so I'm happy having the flag just here.

dangra: only this one is fine

jonatanklosko: Ok! I added the flag and context cancel, let me know if there's anything else :)

internal/command/proxy/proxy.go
```diff
@@ -131,5 +138,43 @@ func run(ctx context.Context) (err error) {
...
+// There is no guarantee that a OS process spawning the proxy will
+// terminate it, however the stdin is always closed whtn the parent
+// terminates. This way we make sure there are no zombie processes,
+// especially that they hold onto TCP ports.
+//
+// Note that we don't do this when stdin is TTY, because that prevents
+// the process from being moved to a background job on Unix.
+// See https://github.com/brunch/brunch/issues/998.
+func watchStdinAndAbortOnClose(ctx context.Context) context.Context {
+	ctx, cancel := context.WithCancelCause(ctx)
```
dangra, line 159: nit: no need to derive the context when stdin is not a tty

dangra: hmm, should it check for stdin tty after all? it is behind a cli flag now.

jonatanklosko: Good call, I don't think we need the check anymore!

dangra:
Good stuff @jonatanklosko . thanks :)

jonatanklosko:
Fantastic, thanks for the help!

PR #3640, opened 2024-06-16
Add scheduled machines to fly.toml

lillian-b:
Needed this for something, thought I'd throw it upstream.

### DIFF
--- a/internal/appconfig/machines_test.go
+++ b/internal/appconfig/machines_test.go
@@ -62,7 +62,7 @@ func TestToMachineConfig(t *testing.T) {
 	// Update a machine config
 	got, err = cfg.ToMachineConfig("", &fly.MachineConfig{
 		Guest:       &fly.MachineGuest{CPUs: 3},
-		Schedule:    "24/7",
+		Schedule:    "",
 		AutoDestroy: true,
 		Restart:     &fly.MachineRestart{Policy: "always"},
 		DNS:         &fly.DNSConfig{SkipRegistration: true},
@@ -76,7 +76,7 @@ func TestToMachineConfig(t *testing.T) {
 	assert.Equal(t, want.Services, got.Services)
 	assert.Equal(t, want.Checks, got.Checks)
 	assert.Equal(t, &fly.MachineGuest{CPUs: 3}, got.Guest)
-	assert.Equal(t, "24/7", got.Schedule)
+	assert.Equal(t, "", got.Schedule)
 	assert.Equal(t, true, got.AutoDestroy)
 	assert.Equal(t, &fly.MachineRestart{Policy: "always"}, got.Restart)
 	assert.Equal(t, &fly.DNSConfig{SkipRegistration: true}, got.DNS)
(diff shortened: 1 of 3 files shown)


internal/appconfig/processgroups.go
```diff
@@ -113,6 +113,11 @@ func (c *Config) Flatten(groupName string) (*Config, error) {
 		return dst.flattenGroupMatches(groupName, k)
 	})
 
+	// [schedules]
+	dst.Schedules = lo.PickBy(dst.Schedules, func(k, v string) bool {
+		return k == groupName
```
dangra, line 118: please, use `flattenGroupMatches` function as others fields do around this line

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -406,8 +406,8 @@ func suggestChangeWaitTimeout(err error, flagName string) error {
 
 func (md *machineDeployment) waitForMachine(ctx context.Context, e *machineUpdateEntry) error {
 	lm := e.leasableMachine
-	// Don't wait for SkipLaunch machines, they are updated but not started
-	if e.launchInput.SkipLaunch {
+	// Don't wait for SkipLaunch or scheduled machines, they are updated but not started
+	if e.launchInput.SkipLaunch || e.launchInput.Config.Schedule != "" {
```
dangra, line 410: setting SkipLaunch=true is the right way to do it unless there is something else special about Schedule, see https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines_launchinput.go#L205

dangra (changes requested):
Hey, thanks for taking the time to add support for schedules to fly.toml. I wonder if you have the willing to finish this PR considering the feedback I am about to throw at you :)

_fly.toml_ (aka appconfig) and its translation to _machine config_ is one of the best unit tested part of flyctl and we would like to keep it that way because it also at the core of all flyctl operations.

that said, see how to modify and add new tests starting with:
* https://github.com/superfly/flyctl/blob/master/internal/appconfig/testdata/full-reference.toml
* https://github.com/superfly/flyctl/blob/master/internal/appconfig/machines_test.go
* https://github.com/superfly/flyctl/blob/master/internal/appconfig/serde_test.go

thanks!

dangra:
> there is one failing one that checks the schedule field of a non-scheduled machine is `24/7`,but I can't find any reference to that value anywhere, nor is that value actually set on a real non-scheduled machine. what is the correct value there?

the '24/7' value is a placeholder, the idea is that any machine config field that isn't directly configurable from fly.toml must remain unchanged when updating an existing machine with that field set.  Your change is breaking this assumption because it makes `MachineConfig.Schedule` depend on a value configurable from fly.toml.

Before, for an application to set a machine schedule, it has to `fly deploy` to create new machines and then `fly machine update --schedule hourly MACHINEID` to set a schedule; as schedules weren't configurable with fly.toml, flyctl had to keep its value on the next machine updates caused by `fly deploy`. I must say this is fragile; `fly deploy` replaces (not updates) machines on some conditions so schedules may get lost in the process. This is why I think we can make schedules always depends on fly.toml and reset `MachineConfig.Schedule` if no schedule is defined in the application config.

That said, please remove the `24/7` and add a *new* testcase for the `[[schedule]]` section like in `internal/appconfig/testdata/tomachine*.toml`

dangra:
> in that case, I would probably be worried that people with existing scheduled machines are going to have their schedules removed. should there be a warning printed in that case?

This is a valid concern, I did a quick check on existing machines and there are a few (20 something) that relies on the old behavior from keep working. 

there is a way to know when not to reset the machine Schedule is by checking the flyctl's version set as machine metadata. The metadata key we are looking for is `fly_flyctl_version` and its value looks like `0.1.119`, `0.2.72` and so on (there are a few 2024.06.20-dev* that can be ignored)

Point is, after we detect the flyctl version is pre the version your patch is going in, another metadata key should be added so following `fly deploy`s keep respecting it, because `fly_flyctl_version` is updated to current flyctl version on deploy.

makes sense?

Issue #3642, opened 2024-06-18
panic when running deploy_test.go with tags=production


RaghavSood:
*Please only report specific issues with `flyctl` behavior*. Anything like a support request for your application should go to https://community.fly.io. More people watch that space and can help you faster!

**Describe the bug**

I maintain the flyctl package in [nixpkgs](https://github.com/NixOS/nixpkgs).

Our builds on darwin are failing due to an issue with how flyctl initialized the sentry-go SDK. This results in a panic within sentry-go that is triggered in `TestCommand_Execute` in [deploy_test.go](https://github.com/superfly/flyctl/blob/e1491edef1d9b04d94d54bfc893fef59c4d5bed5/internal/command/deploy/deploy_test.go#L25).

Specifically, flyctl [initializes the sentry-go](https://github.com/superfly/flyctl/blob/e1491edef1d9b04d94d54bfc893fef59c4d5bed5/internal/sentry/sentry.go#L26) SDK as follows:

```go
func init() {
	opts := sentry.ClientOptions{
		Dsn: "https://89fa584dc19b47a6952dd94bf72dbab4@sentry.io/4492967",
		// TODO: maybe set Debug to buildinfo.IsDev?
		// Debug:       true,
		Environment: buildinfo.Environment(),
		Release:     "v" + buildinfo.Version().String(),
		Transport: &sentry.HTTPSyncTransport{
			Timeout: 3 * time.Second,
		},
		BeforeSend: func(event *sentry.Event, _ *sentry.EventHint) *sentry.Event {
			if buildinfo.IsDev() {
				return nil
			}

			return event
		},
	}

	initError = sentry.Init(opts)
}
```

By initializing the `sentry.HTTPSyncTransport` struct directly, the `ratelimit.Map` under `HTTPSyncTransport.limits` (non-exported) receives a nil zero-value.

When the test above is run, it results in a panic within sentry-go's `ratelimit.Map.Merge`:

```
go test -tags=production -ldflags "-s -w -X github.com/superfly/flyctl/internal/buildinfo.buildDate=1970-01-01T00:00:00Z -X github.com/superfly/flyctl/internal/buildinfo.buildVersion=0.2.71" -v ./internal/command/deploy/deploy_test.go 
```

<details><summary>Stacktrace</summary>
<pre>
github.com/getsentry/sentry-go/internal/ratelimit.Map.Merge(...)
[...]

PR #3643, opened 2024-06-18
bugfix: build upon default sentry-go transport config

RaghavSood:
Directly initializing the HTTPSyncTransport struct results in the non-exported limits map being nil, which can cause a panic in sentry-go under certain conditions

### Change Summary

Fixes: #3642

What and Why: By setting the timeout on the default `HTTPSyncTransport` returned by `NewHTTPSyncTransport`, we avoid a panic within the sentry-go SDK

How: 

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/sentry/sentry.go
+++ b/internal/sentry/sentry.go
@@ -23,15 +23,20 @@ var initError error // set during init
 type Context = sentry.Context
 
 func init() {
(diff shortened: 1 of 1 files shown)


dangra:
thanks @RaghavSood !

PR #3652, opened 2024-06-18
fix: update 'failed to probe' error message handler

gwuah:
### Change Summary
Error message changed in https://github.com/superfly/flyctl/pull/3620


billyb2 (approved):
banger PR

PR #3657, opened 2024-06-20
Switch to using superfly's autorelease action

billyb2:
### Change Summary

What and Why:
The old action workflow was catching the PR tags (like v2024.x.y-z). My workflow just catches v0.*.* and bumps those

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -2,7 +2,7 @@ name: Automatically release a new version of flyctl
 
 on:
   schedule:
-    - cron: '0 20 * * MON-THU'  # Runs at 3 PM Eastern Standard Time Monday Through Thursday (8 PM UTC)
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           fetch-depth: '0'
 
       - name: Bump version
-        uses: anothrNick/github-tag-action@1.67.0
+        uses: billyb2/github-tag-action
```
dangra, line 18: hey, if we are going to use a fork of github-tag-action. Can it be under superfly org so everyone in Fly is able to maintain it?

billyb2: yep, sounds good!

dangra (approved):

PR #3671, opened 2024-06-25
Retry more flaps calls and classify errors better

billyb2:
### Change Summary

What and Why:
Sorry, this is a two part PR:
1. We're now retrying every flaps call where it seems relevant. I noticed in tracing that some builder creations were failing unnecessarily
2. We're now catching more errors. For whatever reason, flaps can return "App not found" or "Could not find app" when it has an issue finding an app. We're doing string comparisons :vomit: to find both

How:
I created a new function that retries flaps errors that are > 500 and < 600. There are likely other errors we can retry (408, for example)

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/build/imgsrc/ensure_builder.go
+++ b/internal/build/imgsrc/ensure_builder.go
@@ -368,7 +368,7 @@ func createBuilder(ctx context.Context, org *fly.Organization, region, builderNa
 }
 
 func retryFlapsCall[T *fly.Machine | *fly.Volume | []fly.Volume | []*fly.Machine](ctx context.Context, maxNumRetries int, flapsCallFn func() (T, error)) (returnValue T, err error) {
-	ctx, span := tracing.GetTracer().Start(ctx, "retry_flaps_call")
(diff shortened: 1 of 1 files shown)


billyb2:
> Could the retry code be added to the client code itself? I think it would be more clean.
yea that's a good idea

internal/build/imgsrc/ensure_builder.go
```diff
@@ -66,11 +67,21 @@ func EnsureBuilder(ctx context.Context, org *fly.Organization, region string, re
...
-				return nil, nil, err
+			_, err := retryFlapsCall(ctx, 5, func() (*fly.Machine, error) {
+				err := flapsClient.Restart(ctx, fly.RestartMachineInput{
+					ID: builderMachine.ID,
+				}, "")
+				return nil, err
+			})
+
+			if err != nil {
+				if strings.Contains(err.Error(), "machine still restarting") {
```
rugwirobaker, line 78: Small neat but switch statements are more readable

billyb2:
Closing, superseded by #3785

PR #3679, opened 2024-06-27
Extended python support

fliepeltje:
### Change Summary

What and Why: Extends the existing fastapi/streamlit/flask scanner to also support:
- [PEP621](https://peps.python.org/pep-0621/) Compliant `pyproject.toml`; this means that management tools like `rye` or `hatch` are also supported
- Scanner for Pipifile
- Scanner for `requirements.txt` and `requirements.in`

With that, most modern and common variations of package/project management are integrated into the tool

How: :sparkle: 

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


scanner/python.go
```diff
@@ -1,8 +1,261 @@
...
+	dep = strings.Split(dep, "==")[0]
+	dep = strings.Split(dep, ">")[0]
+	dep = strings.Split(dep, "<")[0]
+	dep = strings.Split(dep, "~=")[0]
+	return dep
+}
+
+func fromPoetry(pyProject map[string]interface{}) (PyProjectCfg, error) {
+	// Parse pyproject.toml managed with poetry
+	deps := pyProject["tool"].(map[string]interface{})["poetry"].(map[string]interface{})["dependencies"].(map[string]interface{})
```
rubys, line 79: I may be needlessly worrying (and feel free to ignore me if so), but this series of type assertions concern me.  If the toml file does not contain this exact series of nested maps/dictionaries, flyctl will panic and halt:

https://go.dev/ref/spec#Type_assertions

As annoying as it might be, you might want to code this as a series of assignment statements, and return an error if any of the assignment statements returns a not ok.

rubys (approved):
I left a comment that *probably* should be addressed, but I'll leave that up to you.

scanner/python.go
fliepeltje: I got some additional advice to rewrite this into structs and use the unmarshal feature. I'll implement that and cleanup the code a bit more - I think the result will be a bit more robust; thanks for the feedback!

PR #3680, opened 2024-06-27
Allow `fly pg connect` to work with custom images

benwaffle:
This is relevant in development - e.g. if you run `fly deploy -a <my-db>` in postgres-flex


davissp14:
You can currently specify `FLY_DEV=1` to bypass the custom image constraint.

benwaffle:
Oh, didn't notice that. This PR makes it unnecessary. Any thoughts about getting rid of FLY_DEV? I don't see any risk here.

davissp14:
The `pg connect` calls  https://github.com/fly-apps/postgres-flex/blob/master/bin/connect

The main problem is that we can't assume custom images have this binary.

benwaffle:
Today if you run `fly pg connect` on a custom image, you get this error: `Error: Malformed version: ustom`.

Anyone who's hacking on postgres-flex is likely to have the `connect` binary. In the rare case that they have removed it, with this PR they will see:
```
❯ f pg connect -a dbdbdb
Connecting to fdaa:1:a290:a7b:2f1:f87a:2dab:2... complete
exec: "connect": executable file not found in $PATH
Error: ssh shell: wait: remote command exited without exit status or exit signal
```

davissp14 (approved):

davissp14:
I'm don't think we should remove the FLY_DEV check, but I cool opening up the `fly pg connect` command.

benwaffle:
For the record, this changes `hasRequiredVersionOnMachines`, which is used in nearly all `fly pg` commands.

PR #3685, opened 2024-06-27
`fly machines`: release leases even on error

matttpt:
### Change Summary

The returned release function should be called even on error, because it will release any leases that *were* successfully obtained before the error happened (see review comments on #3633).

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


dangra (approved):
🚢

PR #3686, opened 2024-06-28
Make the actual release creation workflow happen when we finish tagging

billyb2:
Also allow manually creating a new release

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,8 @@ on:
 
   workflow_run:
     workflows: ["Automatically release a new version of flyctl"]
+    types:
+      - completed
 
   workflow_dispatch:
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,11 @@ on:
     tags:
       - v**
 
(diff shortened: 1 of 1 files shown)


matthewlehner (approved):

.github/workflows/release.yml
```diff
@@ -5,6 +5,11 @@ on:
     tags:
       - v**
 
+  workflow_run:
+    workflows: ["Automatically release a new version of flyctl"]
```
matthewlehner, line 9: I think this might need `types`, like this:
```suggestion
    workflows: ["Automatically release a new version of flyctl"]
    types:
      - completed
```

matthewlehner: We might also want to limit this to the main branch?

billyb2: hm, i think that the tag workflow only runs on the main branch, so it should be fine?

billyb2: or at least, it should by default i think

PR #3692, opened 2024-07-01
`fly launch` from a template repo and clone submodules

kylemclaren:
### Change Summary

What and Why:

Used in conjunction with the `--from` flag when running `fly launch`, a user can optionally specify a bool flag of `--submodules` to recursively clone a git repo's submodules.

How:

```
fly launch --submodules --from <REPO_URL>
```

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/launch/cmd.go
+++ b/internal/command/launch/cmd.go
@@ -77,11 +77,6 @@ func New() (cmd *cobra.Command) {
 			Description: "A github repo URL to use as a template for the new app",
 		},
 		flag.Bool{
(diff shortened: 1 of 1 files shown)


benwaffle:
Can we just include `--recurse-submodules every time? Any downsides?

kylemclaren:
Maybe? Can think of a scenario where a repo has lots of submodules, causing an IO issue? But that should be super rare so this is probably fine.

benwaffle (approved):

PR #3693, opened 2024-07-01
Create and manage Tigris bucket for PG/Barman backups

ndarilek:
### Change Summary

Integrates the new Tigris WAL storage into flyctl.

* If backups are enabled, provisions a new Tigris bucket on cluster creation and sets the barman secret appropriately.
* Adds a `--enable-backup` flag to `fly pg create` command to enable bucket/secret provisioning on cluster creation.
* Adds `--restore-target-app`, `--restore-target-name`, `--restore-target-time` and `--restore-target-inclusive` flags to `fly pg create` which behave as documented [here](https://www.postgresql.org/docs/15/runtime-config-wal.html#RUNTIME-CONFIG-WAL-RECOVERY-TARGET). When a new instance with a cluster size of 1 is created with an app and optional target set, the specified backup is restored into the newly-created instance.
* Adds a `fly pg backup enable` command for provisioning buckets and secrets into clusters where WAL backup is disabled.
* Adds `fly pg backup create` for creating backups, optionally named.
* Adds `fly pg backup list` to list existing backups.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


benwaffle:
If the bucket name is already taken, we should generate a unique name

ndarilek:
> If the bucket name is already taken, we should generate a unique name instead of erroring out

https://github.com/superfly/flyctl/blob/cd52698b475e7edb8e47f1ab24cb02ebffabceeb/internal/command/extensions/core/core.go#L100 Looking at this naming logic, if a name isn't explicitly specified via `OverrideName`, it's automatically set to the target app name. I think the extension creation process in web will autogenerate random names but I don't see this easily exposed by flyctl's extension code. So in practice this means that PGs created with `my-name` will have buckets also named `my-name`.

I agree that the ideal case is not to bomb out on unique bucket names, but how likely is this in practice? I imagine:
1. Someone runs `fly pg create -n my-pg ...`
2. They then run `fly app rm -y my-pg`, followed by another create which would crash.

The crash isn't ideal but it isn't entirely unexpected and easy enough to work around by removing the PG storage bucket.

Either way, I don't want to spend much time overthinking this, so possible approaches if this needs to be addressed:

1. Always autogenerate silly-name-1234 and similar. I know we do this in various places now, but I'm not immediately finding it in flyctl and wonder if it's happening in flaps/web?
2. Append `-postgres` or similar to the extension names--that doesn't necessarily fix conflicts but at least minimizes clashes with someone creating `appname` as a regular file storage bucket.
3. Append a suffix and retry reation if provisioning fails.

If this definitely needs addressing then I think I'd prefer 2, or 1 as long as we aren't pulling in another dependency. 2 is simple enough to just do and is probably a good idea anyway, so I'll just do that and we can decide what additional needs doing.

ndarilek:
> how likely is this in practice?

Apparently somewhat, we hold onto recently-deleted bucket names so create/destroy/create cycles will fail even if the bucket itself is removed. Going to try scraping the error text for naming conflicts and appending a suffix, unless there's a better approach.

davissp14:
Awesome start on this!

Couple notes:

> Adds a --disable-backups flag to fly pg create command to disable bucket/secret provisioning on cluster creation.

I think we should actually make this opt-in to start.  There's a cost implication that some users may not be cool with.

> Adds fly pg backup list to list existing backups.

It looks like this currently breaks if the bucket name doesn't match the App name.  

> Adds a currently non-functional fly pg backup enable command for provisioning buckets and secrets into clusters where WAL backup is disabled.

Once the release is cut, we need to ensure that these commands can't be run against older Flex versions.

ndarilek:
> • The log output after enabling is no longer accurate now that we've consolidated down to a single secret.
> • The S3_ARCHIVE_CONFIG secret is the only one we need to inject.

Sorry, which log output do you mean specifically, and which additional secret am I injecting? Looks like I'm only injecting the one as per our docs.

Thanks for the feedback.

ndarilek:
> • Once the release is cut, we will need to ensure that these commands can't be run against older Flex versions.

Agreed. Any thoughts about the easiest way to implement this? Version checks seem like the most immediately obvious, but that feels wrong and brittle. Seems like the better way is checking for the existence of this specific feature in the image, either with a command we can exec in and run, or something in the machine metadata. Thoughts?

Unless of course there's something I'm missing with the version check, but semver version comparison feels like something we'd either need another dependency for or would have to be very careful about.

davissp14:
>Agreed. Any thoughts about the easiest way to implement this? 

There's already logic in place for this: 
https://github.com/superfly/flyctl/blob/master/internal/command/postgres/postgres.go#L48

internal/command/postgres/backup.go
```diff
@@ -0,0 +1,585 @@
...
+		return fmt.Errorf("No active machines")
+	}
+
+	// Ensure the the app has the required flex version.
+	if err := hasRequiredVersion(appName, machines); err != nil {
+		return err
+	}
+
+	// Resolve the leader
+	leader, err := pickLeader(ctx, machines)
```
benwaffle, line 110: this will fail if the DB is having problems, or a host is down.

davissp14 (commented):
Overall looks good, just a few comments.

internal/command/postgres/create.go
```diff
@@ -271,20 +285,27 @@ func CreateCluster(ctx context.Context, org *fly.Organization, region *fly.Regio
...
+		ImageRef:      params.PostgresConfiguration.ImageRef,
+		Region:        region.Code,
+		Manager:       params.Manager,
+		Autostart:     params.Autostart,
+		ForkFrom:      params.ForkFrom,
+		BackupEnabled: flag.GetBool(ctx, "enable-backup"),
+		// Eventually we populate this with a full S3 endpoint, but use the
+		// restore app target for now.
+		BarmanRemoteRestoreConfig: flag.GetString(ctx, "restore-target-app"),
+		RestoreTargetName:         flag.GetString(ctx, "restore-target-name"),
```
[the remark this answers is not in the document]
ndarilek: Not anymore, looks like those were left over from the command refactor. Good catch, removing.

internal/command/postgres/create.go
```diff
@@ -34,6 +34,10 @@ func newCreate() *cobra.Command {
 		flag.Region(),
 		flag.Org(),
 		flag.Detach(),
+		flag.Bool{
+			Name:        "enable-backup",
```
davissp14, line 38: Let's change this to `enable-backups` and update the description to indicate that it will include the provisioning of a new Tigris bucket.

internal/command/postgres/backup.go
```diff
@@ -0,0 +1,585 @@
...
+	if out.StdErr != "" {
+		fmt.Fprint(io.ErrOut, out.StdErr)
+	}
+
+	return nil
+}
+
+func newBackupEnable() *cobra.Command {
+	const (
+		short = "Enable backups on a Postgres cluster"
```
davissp14, line 263: We need to make this super clear that it will create a new Tigris bucket.

davissp14 (changes requested):
Some command clarifications

davissp14 (approved):
LGTM!

PR #3709, opened 2024-07-05
Better recovery

billyb2:
### Change Summary

What and Why:
This PR focuses on trying to change how machine updates work. Instead of failing whenever one machine isn't able to update, we instead try again, making sure to not redo any work that we'd already completed. We do this by keeping track of 'app state', which is basically a list of all the machines we have in our app. This works great since it includes machine states, machine configs, and mounts.

Sometimes, certain errors aren't worth retrying. For example, if we aren't able to acquire a lease on your machine because it's already being held, then it's more likely than not that retrying won't do any good. In that case, we return an 'unrecoverable error', and choose to completely fail the deploy. In certain cases, we could try rolling back to how we were (for example, if health checks fail we could rollback), though I'm saving that implementation for a future PR

How:
We will continue to try to finish a deploy a few times, until either:
a. the deploy completes successfully, meaning we transitioned from the old app state to the new app state
b. we've exhausted a certain number of attempts to complete a deploy without success
c. we encountered an unrecoverable error, meaning that flyctl doesn't think it's worth attempting a retry

In the first case, flyctl is much more resilient to intermittent platform errors, which can happen pretty often. In the second case, it's likely that there's something wrong with either the user's environment or the platform itself. If it's the former, then @rugwirobaker 's work to move orchestration logic into a new fly machine will help with those cases in the future. If it's the latter, we're working on setting up alerting to try and learn about these cases sooner. In the third case, the goal in the future is to add suggestions to the user to help recover from these sorts of issues.

The bulk of the code is in `internal/command/deploy/plan.go`

Related to:
https://flyio.discourse.team/t/deployments-roadmap-discussion/6326
https://flyio.discourse.team/t/deployments-roadmap-redux/6451
https://flyio.discourse.team/t/deployment-recoverability/6441
---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -167,10 +167,10 @@ var CommonFlags = flag.Set{
 		Shorthand:   "s",
 		Description: "Signal to stop the machine with for bluegreen strategy (default: SIGINT)",
 	},
-	flag.Int{
+	flag.String{
 		Name:        "deploy-retries",
 		Description: "Number of times to retry a deployment if it fails",
-		Default:     0,
+		Default:     "auto",
 	},
 }
 
@@ -486,6 +486,28 @@ func deployToMachines(
 
 	status.FlyctlVersion = buildinfo.Info().Version.String()
 
+	retriesFlag := flag.GetString(ctx, "deploy-retries")
+	deployRetries := 0
+
+	switch retriesFlag {
+	case "auto":
+		ldClient := launchdarkly.ClientFromContext(ctx)
+		retries := ldClient.GetFeatureFlagValue("deploy-retries", 0.0).(float64)
+		deployRetries = int(retries)
+
+	default:
+		var invalidRetriesErr error = fmt.Errorf("--deploy-retries must be set to a positive integer, 0, or 'auto'")
+		retries, err := strconv.Atoi(retriesFlag)
+		if err != nil {
+			return invalidRetriesErr
+		}
+		if retries < 0 {
+			return invalidRetriesErr
+		}
+
+		deployRetries = retries
+	}
+
 	md, err := NewMachineDeployment(ctx, MachineDeploymentArgs{
 		AppCompact:            app,
 		DeploymentImage:       img.Tag,
@@ -513,7 +535,7 @@ func deployToMachines(
 		MaxConcurrent:         maxConcurrent,
 		VolumeInitialSize:     flag.GetInt(ctx, "volume-initial-size"),
 		ProcessGroups:         processGroups,
-		DeployRetries:         flag.GetInt(ctx, "deploy-retries"),
+		DeployRetries:         deployRetries,
 	})
 	if err != nil {
 		sentry.CaptureExceptionWithAppInfo(ctx, err, "deploy", app)
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -170,7 +170,7 @@ var CommonFlags = flag.Set{
 	flag.Int{
 		Name:        "deploy-retries",
 		Description: "Number of times to retry a deployment if it fails",
-		Default:     5,
+		Default:     0,
 	},
 }
 
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -167,6 +167,11 @@ var CommonFlags = flag.Set{
 		Shorthand:   "s",
 		Description: "Signal to stop the machine with for bluegreen strategy (default: SIGINT)",
 	},
+	flag.Int{
+		Name:        "deploy-retries",
+		Description: "Number of times to retry a deployment if it fails",
+		Default:     5,
+	},
 }
 
 type Command struct {
@@ -504,6 +509,7 @@ func deployToMachines(
 		MaxConcurrent:         maxConcurrent,
 		VolumeInitialSize:     flag.GetInt(ctx, "volume-initial-size"),
 		ProcessGroups:         processGroups,
+		DeployRetries:         flag.GetInt(ctx, "deploy-retries"),
 	})
 	if err != nil {
 		sentry.CaptureExceptionWithAppInfo(ctx, err, "deploy", app)
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -321,10 +321,14 @@ func DeployWithConfig(ctx context.Context, appConfig *appconfig.Config, userID i
 
 	// Fetch an image ref or build from source to get the final image reference to deploy
 	img, err := determineImage(ctx, appConfig, usingWireguard, recreateBuilder)
-	if err != nil && usingWireguard && httpFailover {
-		span.SetAttributes(attribute.String("builder.failover_error", err.Error()))
-		span.AddEvent("using http failover")
-		img, err = determineImage(ctx, appConfig, false, recreateBuilder)
+	if err != nil {
+		noBuilder := strings.Contains(err.Error(), "Could not find App")
+		recreateBuilder = recreateBuilder || noBuilder
+		if noBuilder || (usingWireguard && httpFailover) {
+			span.SetAttributes(attribute.String("builder.failover_error", err.Error()))
+			span.AddEvent("using http failover")
+			img, err = determineImage(ctx, appConfig, false, recreateBuilder)
+		}
 	}
 
 	if err != nil {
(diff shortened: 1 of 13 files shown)


internal/command/deploy/plan.go
```diff
@@ -0,0 +1,504 @@
...
+		})
+	}
+
+	if updateErr := group.Wait(); updateErr != nil {
+		if !pushForward {
+			return updateErr
+		}
+
+		var unrecoverableErr unrecoverableError
+		if strings.Contains(updateErr.Error(), "context canceled") || errors.As(updateErr, &unrecoverableErr) || strings.Contains(updateErr.Error(), "Unrecoverable error") {
```
billyb2, line 128: this is gross, i'll fix it

billyb2:
^ btw i'm working on adding unit testing to plan.go now, I'll make sure to have it up and reviewed before I merge

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -290,12 +279,6 @@ func (md *machineDeployment) deployMachinesApp(ctx context.Context) error {
 		}
 	}
 
-	if err := md.machineSet.AcquireLeases(ctx, md.leaseTimeout); err != nil {
-		return err
-	}
-	defer md.machineSet.ReleaseLeases(ctx) // skipcq: GO-S2307
-	md.machineSet.StartBackgroundLeaseRefresh(ctx, md.leaseTimeout, md.leaseDelayBetween)
```
dangra, line 297: I can't find where this functionality was moved to or replaced by 🤔

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -485,261 +475,6 @@ func (md *machineDeployment) updateUsingBlueGreenStrategy(ctx context.Context, u
...
-
-	var poolSize int
-	switch mu := md.maxUnavailable; {
-	case mu >= 1:
-		poolSize = int(mu)
-	case mu > 0:
-		poolSize = int(math.Ceil(float64(len(entries)) * mu))
-	default:
-		return fmt.Errorf("Invalid --max-unavailable value: %v", mu)
-	}
```
dangra, line 604: I can't find this functionality on the new code 👎  seems it is gone

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -485,261 +475,6 @@ func (md *machineDeployment) updateUsingBlueGreenStrategy(ctx context.Context, u
...
-
-	// Rolling strategy
-	slices.SortFunc(updateEntries, func(a, b *machineUpdateEntry) int {
-		return cmp.Compare(a.leasableMachine.Machine().ID, b.leasableMachine.Machine().ID)
-	})
-
-	// Group updates by process group
-	entriesByGroup := lo.GroupBy(updateEntries, func(e *machineUpdateEntry) string {
-		return e.launchInput.Config.ProcessGroup()
-	})
```
dangra, line 568: the whole point here is to do controlled concurrent updates per process group. I can't find that in new code

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -101,13 +96,6 @@ func (md *machineDeployment) restartMachinesApp(ctx context.Context) error {
 	ctx, span := tracing.GetTracer().Start(ctx, "restart_machines")
 	defer span.End()
 
-	if err := md.machineSet.AcquireLeases(ctx, md.leaseTimeout); err != nil {
-		tracing.RecordError(span, err, "failed to acquire lease")
-		return err
-	}
-	defer md.machineSet.ReleaseLeases(ctx) // skipcq: GO-S2307
-	md.machineSet.StartBackgroundLeaseRefresh(ctx, md.leaseTimeout, md.leaseDelayBetween)
```
dangra, line 109: why is it removing lease grabbing here?

internal/command/deploy/plan.go
```diff
@@ -171,7 +172,7 @@ func (md *machineDeployment) updateMachines(ctx context.Context, oldAppState, ne
 					span.RecordError(updateErr)
 					return err
 				}
-				fmt.Println("Failed to update machines:", err, ". Retrying...")
+				fmt.Fprintln(os.Stderr, "Failed to update machines:", err, ". Retrying...")
```
rugwirobaker, line 174: Why not use iostreams here?

internal/command/deploy/machines_deploymachinesapp.go
billyb2: yea that's a good catch!

internal/command/deploy/plan.go
billyb2: good catch, fixed

internal/command/deploy/machines_deploymachinesapp.go
billyb2: added it back :)

internal/command/deploy/plan.go
```diff
@@ -120,8 +121,18 @@ func (md *machineDeployment) updateMachines(ctx context.Context, oldAppState, ne
...
+		poolSize = int(mu)
+	case mu > 0:
+		poolSize = int(math.Ceil(float64(len(machineTuples)) * mu))
+	default:
+		return fmt.Errorf("Invalid --max-unavailable value: %v", mu)
+	}
+
 	group := errgroup.Group{}
-	group.SetLimit(md.maxConcurrent)
+	group.SetLimit(poolSize)
```
dangra, line 135: the max unavailable setting is per-group, not across groups if that makes sense.

billyb2: oh that makes sense. i'll fix it

dangra (commented):
@benbjohnson @michaeldwan this PR touches the hot path of Fly deployments, a small screw up here can make a big wave. It deserves as many eyes as it can get on it 👀 🙏

internal/command/deploy/deploy.go
```diff
@@ -477,6 +486,28 @@ func deployToMachines(
...
 	status.FlyctlVersion = buildinfo.Info().Version.String()
 
+	retriesFlag := flag.GetString(ctx, "deploy-retries")
+	deployRetries := 0
+
+	switch retriesFlag {
+	case "auto":
+		ldClient := launchdarkly.ClientFromContext(ctx)
+		retries := ldClient.GetFeatureFlagValue("deploy-retries", 0.0).(float64)
+		deployRetries = int(retries)
```
benbjohnson, line 496: I don't see a `launchdarkly.NewClient()` to instantiate the client. Am I missing it?

Also, there doesn't seem to be anywhere that waits for `updateFeatureFlags()` to fetch the flags before grabbing the feature flag value here. So there's probably a good chance that the flag won't be initialized by the time you get to this point in the code.

benbjohnson (changes requested):

internal/command/deploy/deploy.go
billyb2: we instantiate the ld client in the DeployWithConfig function:
https://github.com/superfly/flyctl/blob/better_recovery/internal/command/deploy/deploy.go#L301

i purposely do this before we start messing with the remote builder. that takes a few seconds at minimum, so by the time we try to get the feature flag from launchdarkly, it's very likely that the flag will be instantiated. I also confirmed this with some testing. While there's a chance that the flag isn't fetched from LD if the user's internet is being weird, I don't want to slow down deploys too much.

Would you prefer if we added a timeout option to the GetFeatureFlag function?

benbjohnson: I think I didn't have an updated view of the code locally. Sorry about that. I see it now.

As for the fetch, I think that having the remote builder step in between is good enough. No need to add a timeout.

benbjohnson (approved):

internal/command/deploy/plan.go
```diff
@@ -0,0 +1,691 @@
...
+			} else if machineTuple.newMachine != nil {
+				machine = machineTuple.newMachine
+			} else {
+				return nil
+			}
+			sl.LogStatus(statuslogger.StatusRunning, fmt.Sprintf("Acquiring lease for %s", machine.ID))
+
+			if machine.LeaseNonce == "" {
+				sl.LogStatus(statuslogger.StatusRunning, fmt.Sprintf("Waiting for job %s", machine.ID))
+			}
```
btoews, line 324: Do you remember why we show `Waiting for job` in this case? It seems like we should still say `Acquiring lease for` if the machine isn't current leased, right?

billyb2: that's a great question, this seems like a bug.  I feel like we should check if LeaseNonce != "", and if that's the case, we just return (since we already have an acquired lease)

btoews: The lease nonce could be set if someone else has it too though, which which case we still want to wait. I'm going to remove this LogStatus though.

PR #3720, opened 2024-07-09
Allow machine checks without a service

billyb2:
### Change Summary

What and Why:
Some customers want to be able to use machine checks for apps that only use private communications, which is a very fair use case.

How:
 In order to allow for this, we just allow putting machine_checks at the top level of a fly.toml

Related to:
https://github.com/superfly/customer-help/issues/55

---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/test/preflight/fly_deploy_test.go
+++ b/test/preflight/fly_deploy_test.go
@@ -250,6 +250,24 @@ func TestFlyDeploy_DeployMachinesCheck(t *testing.T) {
 	require.Contains(f, output, "Test Machine")
 }
 
+func TestFlyDeploy_NoServiceDeployMachinesCheck(t *testing.T) {
+	f := testlib.NewTestEnvFromEnv(t)
+	appName := f.CreateRandomAppName()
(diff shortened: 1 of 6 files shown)


test/preflight/fly_deploy_test.go
```diff
@@ -250,6 +250,24 @@ func TestFlyDeploy_DeployMachinesCheck(t *testing.T) {
...
 }
 
+func TestFlyDeploy_NoServiceDeployMachinesCheck(t *testing.T) {
+	f := testlib.NewTestEnvFromEnv(t)
+	appName := f.CreateRandomAppName()
+	f.Fly("launch --org %s --name %s --region %s --image nginx --internal-port 80 --ha=false", f.OrgSlug(), appName, f.PrimaryRegion())
+	appConfig := f.ReadFile("fly.toml")
+	appConfig += `
+		[[machine_checks]]
+            image = "curlimages/curl"
```
lubien, line 260: mixed indent here

lubien (approved):

test/preflight/fly_deploy_test.go
billyb2: good catch!

PR #3724, opened 2024-07-10
Meteor 3 scanner

kylemclaren:
### Change Summary

What and Why: A scanner for [Meteor](https://meteor.com) apps. Much anticipated Meteor 3 launch is coming in the next week or two. This is a (formerly) huge community so it would be good to cater to these apps in the scanner.

Related:

- https://github.com/fly-apps/dockerfile-node/pull/92

### Test

```console
git clone github.com/fly-apps/meteor-3-demo
rm Dockerfile fly.toml
```

```console
git clone github.com/superfly/flyctl
cd flyctl
git checkout feat/meteor-scanner
make build
```

```console
cd meteor-3-demo
/path/to/built/flyctl launch
```

---

### Documentation

- [ ] Fresh Produce
- [ ] In superfly/docs, or asked for help from docs team
- [ ] n/a

PR #3731, opened 2024-07-12
Meteor `ROOT_URL` fix

kylemclaren:
### Change Summary

What and Why:

when setting FQDN for Meteor `ROOT_URL`, include `https` scheme (or Meteor will error)

Related to:

- https://github.com/superfly/flyctl/pull/3724


internal/command/extensions/core/core.go
kylemclaren, line 1: Honestly not sure how this popped up as it was merged in the last PR https://github.com/superfly/flyctl/pull/3724/files

PR #3732, opened 2024-07-12
Replace the machine if the host isn't accpepting inplace updates

dangra:
### Change Summary

What and Why: Machines whose hosts are under maintenance can't be updated in-place causing the whole deployment to fail

```
failed to update VM 148e290xxxx948: unavailable: deploys to this host are temporarily disabled, please try again later or check the status page: https://status.flyio.net/
```

How: Replace the whole machine instead of updating it in-place, it is a bit slower but gets the deployment going. 

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


billyb2 (approved):
banger pr

PR #3733, opened 2024-07-12
Implement autosuspend

matttpt:
### Change Summary

**What and Why:**

Add autosuspend support:

* `auto_suspend_machines` in fly.toml now accepts `"off"`, `"stop"`, and `"suspend"` in addition to boolean values.
* The `--autostop` flag for `fly machines create`/`fly machines run`/`fly machines update` accepts `off`, `stop`, and `suspend` in addition to boolean values.

Note that these changes should be backward compatible.

**How:**

* Upgrade fly-go to use the new `fly.MachineAutostop` type, and make the necessary changes to use it.
* Convert the `fly machines` commands' `--autostop` flag to a string, and parse it so as to accept the new values as well as the old boolean values.

**Related to:**

* Depends on superfly/fly-go#80.

---

### Documentation

- [x] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -18,6 +18,7 @@ import (
 	"github.com/sourcegraph/conc/pool"
 	fly "github.com/superfly/fly-go"
 	"github.com/superfly/flyctl/helpers"
+	"github.com/superfly/flyctl/internal/appconfig"
 	machcmd "github.com/superfly/flyctl/internal/command/machine"
 	"github.com/superfly/flyctl/internal/flapsutil"
 	"github.com/superfly/flyctl/internal/flyerr"
@@ -199,6 +200,7 @@ func (md *machineDeployment) deployCanaryMachines(ctx context.Context) (err erro
 // Create machines for new process groups
 func (md *machineDeployment) deployCreateMachinesForGroups(ctx context.Context, processGroupMachineDiff ProcessGroupsDiff) (err error) {
 	groupsWithAutostopEnabled := make(map[string]bool)
+	groupsWithAutosuspendEnabled := make(map[string]bool)
 	groups := maps.Keys(processGroupMachineDiff.groupsNeedingMachines)
 	total := len(groups)
 	slices.Sort(groups)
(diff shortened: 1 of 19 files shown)


internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -223,11 +224,18 @@ func (md *machineDeployment) deployCreateMachinesForGroups(ctx context.Context,
...
 		services := groupConfig.AllServices()
+		autostopSetting := fly.MachineAutostopSuspend
 		for _, s := range services {
-			if s.AutoStopMachines != nil && *s.AutoStopMachines {
-				groupsWithAutostopEnabled[name] = true
+			if s.AutoStopMachines != nil && *s.AutoStopMachines < autostopSetting {
+				autostopSetting = *s.AutoStopMachines
 			}
 		}
+		switch autostopSetting {
```
dangra, line 233: I think there is an edge case when `s.AutoStopMachines == nil` that defaults to `MachineAutostopSuspend` when it should be `MachineAutostopOff`

internal/command/machine/run.go
```diff
@@ -774,7 +774,11 @@ func determineMachineConfig(
 		}
 
 		if flag.IsSpecified(ctx, "autostop") {
-			s.Autostop = fly.Pointer(flag.GetBool(ctx, "autostop"))
+			if flag.GetBool(ctx, "autostop") {
```
dangra, line 777: looks like this needs to be "extended" to support new values :)

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -223,11 +224,22 @@ func (md *machineDeployment) deployCreateMachinesForGroups(ctx context.Context,
 		}
 
 		services := groupConfig.AllServices()
+		autostopSetting := fly.MachineAutostopSuspend
```
dangra, line 227: why isn't `fly.MachineAutostopOff` the default? below logic doesn't play well when there are no services; it will default to include the group in groupsWithAutosuspendEnabled.

I don't know follow why you insist in using a out-of-loop variable (the `autostopSetting`) instead of something like:

```go
for _, s := range services {
  switch v := s.AutostopMachines; {
  case v == nil || *v == fly.MachineAutostopOff:
    // skip
  case *v == fly.MachineAutostopStop:
    groupsWithAutostopEnabled[name] = true
  case *v == fly.MachineAutostopSuspend:
    groupsWithAutosuspendEnabled[name] = true
  }
}

Comment (matttpt) [reply internal/command/deploy/machines_deploymachinesapp.go:227]: Agh, thanks, I goofed this again!

I'm trying to take the minimum of the autostop settings across all of the process group's services, because that is what the proxy will use to decide how to downscale the group. E.g. if the process group has three services, two with `auto_stop_machines = true` and one with `auto_stop_machines = false`, then the proxy will take no action because of the third, not-autostoppable service. But (if I understand correctly) the existing logic would print that autostop is enabled anyway.

File: internal/command/deploy/machines_deploymachinesapp.go [hunk 4]
```diff
@@ -223,9 +225,22 @@ func (md *machineDeployment) deployCreateMachinesForGroups(ctx context.Context,
 		}
 
 		services := groupConfig.AllServices()
-		for _, s := range services {
-			if s.AutoStopMachines != nil && *s.AutoStopMachines {
+		if len(services) > 0 {
+			// The proxy will use the most restrictive (which, in terms
+			// of the fly.MachineAutostop type, is the least) autostop
+			// setting across all of the group's services.
```
dangra, line 231: ❤️

dangra:
Ready to approve and merge?

matttpt:
* Updated go.mod to reference the main-branch commit of fly-go
* Updated the help text for the `--autostop` flag
* Converted unit-test TOML files to the new `"off"`/`"stop"`/`"start"` format, but also added `auto_stop_machines = false` to the old-format test to check that booleans still parse correctly

@dangra apologies for the delay—I think this is ready now!

dangra (approved):
💯 :shipit:

PR #3753, opened 2024-07-18
Disable backup/cluster creation/enabling on clusters with less than 512 MB of RAM

ndarilek:
### Change Summary

* Disallow creation of Postgres clusters with backups enabled if specified RAM is less than 512 MB
* Disallow creating or enabling backups on clusters whose leaders have less than 512 MB, and give instructions on how to scale up

---

### Documentation

- [ ] Fresh Produce
- [X] n/a


flypg/launcher.go
```diff
@@ -79,6 +79,12 @@ func (l *Launcher) LaunchMachinesPostgres(ctx context.Context, config *CreateClu
 		client   = flyutil.ClientFromContext(ctx)
 	)
 
+	machineConf := l.getPostgresConfig(config)
+	// Can't use `hasRequiredMemoryForBackup` because of circular import.
+	if config.BackupsEnabled && machineConf.Guest.MemoryMB < 512 {
```
davissp14, line 84: Hrmm, so this is a little weird since the 1x shared VM setup can actually be scaled up to 2Gib.

I think if we are going to error, we may want to consider offering a pre-configured plan that comes with 512mib of memory.  Otherwise, we could just log a warning.

davissp14 (changes requested):
Overall looks good, just one issue we will need to consider.

flypg/launcher.go
ndarilek: > Hrmm, so this is a little weird since the 1x shared VM setup can actually be scaled up to 2Gib.

What's the relevance? Wouldn't the RAM check then detect the correct amount and pass, or is that not accurate here for some reason?

davissp14 (approved):
LGTM!

PR #3754, opened 2024-07-18
Feature Flag Support

billyb2:
### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


billyb2:
just waiting on https://github.com/superfly/flyctl/pull/3733 to get merged before putting this up for review, since i'm building off the new fly-go

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
+package fflag
```
benbjohnson, line 1: If this package is specifically an implementation of LaunchDarkly then we should just name the package `launchdarkly`.

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/superfly/flyctl/internal/tracing"
+)
+
+const clientSideID string = "6557a71bbffb5f134b84b15c"
+
+type FeatureFlagClient struct {
+	ldContext  ldcontext.Context
+	flags      map[string]FeatureFlag
+	flagsMutex sync.Mutex
+}
```
benbjohnson, line 24: You can just call this `Client` since "feature flag" is implied from the package name.

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+	Version     int         `json:"version"`
+	Variation   int         `json:"variation"`
+}
+
+func (ffClient *FeatureFlagClient) updateFeatureFlags() (map[string]FeatureFlag, error) {
+	ldContextJSON := ffClient.ldContext.JSONString()
+	ldContextB64 := base64.URLEncoding.EncodeToString([]byte(ldContextJSON))
+
+	request, err :=
+		http.NewRequest("GET", fmt.Sprintf("https://clientsdk.launchdarkly.com/sdk/evalx/%s/contexts/%s", clientSideID, ldContextB64), http.NoBody)
```
benbjohnson, line 111: TIL about `http.NoBody`. I've always just set it to `nil`.

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+	resp, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	var flags map[string]FeatureFlag
+	err = json.Unmarshal(resp, &flags)
+	if err != nil {
+		return nil, err
+	}
```
benbjohnson, line 130: It's good to try to avoid `io.ReadAll()` since you could potentially load a lot of data into memory. It's probably not likely here but a better approach is to use `json.Decoder` directly:

```go
var flags map[string]FeatureFlag
if err := json.NewDecoder(response.Body).Decode(&flags); err != nil {
	return err
}
```

benbjohnson: Also, if the body contained `null` then you could potentially return a `nil` map. Again, unlikely but something to consider for safety if the caller is expecting a non-nil value.

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+		http.NewRequest("GET", fmt.Sprintf("https://clientsdk.launchdarkly.com/sdk/evalx/%s/contexts/%s", clientSideID, ldContextB64), http.NoBody)
+	if err != nil {
+		return nil, err
+	}
+	client := &http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
```
benbjohnson, line 120: You can probably just use `http.Get()` since you're not setting headers and you're using the default client. https://pkg.go.dev/net/http#Get

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+
+}
+
+type FeatureFlag struct {
+	FlagVersion int         `json:"flagVersion"`
+	TrackEvents bool        `json:"trackEvents"`
+	Value       interface{} `json:"value"`
+	Version     int         `json:"version"`
+	Variation   int         `json:"variation"`
+}
```
benbjohnson, line 104: Minor nit. Go allows the use of `any` instead of `interface{}` which is a little easier to read.

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+
+func NewClient(ctx context.Context, userInfo UserInfo) (*FeatureFlagClient, error) {
+	_, span := tracing.GetTracer().Start(ctx, "new_feature_flag_client")
+	defer span.End()
+
+	orgID, err := strconv.Atoi(userInfo.OrganizationID)
+	if err != nil {
+		orgID = 0
+	}
```
benbjohnson, line 52: I feel like this should probably error if we're passing in something unparseable for an org ID. Maybe default to `0` if a blank string is passed in but otherwise return an error?

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+	ffClient := &FeatureFlagClient{
+		ldContext:  launchDarklyContext,
+		flagsMutex: sync.Mutex{},
+	}
+
+	go func() {
+		for {
+			if ffClient == nil {
+				return
+			}
```
benbjohnson, line 68: When can `ffClient` be `nil`?

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+
+			ffClient.flagsMutex.Lock()
+			ffClient.flags = flags
+			ffClient.flagsMutex.Unlock()
+
+			// the launchdarkly docs recommend polling every 30 seconds
+			time.Sleep(30 * time.Second)
+		}
+	}()
```
benbjohnson, line 82: Can you move this to a regular function (e.g. `FeatureFlagClient.monitor()`) and call that? It's crowding the constructor and it doesn't seem to need the scope it's defined in.

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+	go func() {
+		for {
+			if ffClient == nil {
+				return
+			}
+
+			flags, err := ffClient.updateFeatureFlags()
+			if err != nil {
+				return
+			}
```
benbjohnson, line 73: We should log this error or else it's going to be difficult to figure out when there's an issue. You could even do it on `DEBUG` logging only.

internal/fflag/fflag.go
```diff
@@ -0,0 +1,133 @@
...
+
+			ffClient.flagsMutex.Lock()
+			ffClient.flags = flags
+			ffClient.flagsMutex.Unlock()
```
benbjohnson, line 77: I feel like this code should be in the `updateFeatureFlags()` code since the function is called "update".

benbjohnson (changes requested):

internal/fflag/fflag.go
billyb2: yea deep go was just bugging me about it lol

internal/fflag/fflag.go
billyb2: oh i had no idea, that's way nicer

internal/fflag/fflag.go
billyb2: oh that's way better

internal/fflag/fflag.go
billyb2: it can't right now, though i noticed these sorts of guard clauses when initializing the flaps client too. should i remove it?

billyb2: yea good point. i was thinking that the object could just get freed at any time, but that's not how go works 🤦🏾

benbjohnson (approved):

PR #3765, opened 2024-07-23
[fix] support files in release command machines

jipperinbham:
### Change Summary

What and Why:

Currently, if a user wants to rely on any `files` defined in `fly.toml` for the release command it fails because it is not included in the machine config.

How:

Use the same logic for defining files on the machine config for release command machines as we do for "normal" machines.

Related to:

Customer Ticket

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


dangra (commented):
What an opportunity to add it to this test case:
https://github.com/superfly/flyctl/blob/e4111c5640ccd758dfd4a3419cba09218693e69c/internal/appconfig/machines_test.go#L149

dangra (approved):

PR #3777, opened 2024-07-25
Restrict synthetic monitoring targets to Fly infra hosts

aschiavo:
### Change Summary

What and Why:

Because of the concerns raised after [Fresh Produce](https://community.fly.io/t/we-are-building-synthetic-monitoring-would-you-help-us/20929) we are restricting the synthetic monitoring to Fly infra by now ([context](https://flyio.discourse.team/t/pitch-synthetic-monitoring/6334/13?u=aschiavo))

TL;DR We are enabling the synthetics agent by default and restricting the probes to Fly hosts that are already being used by the CLI.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -62,6 +62,7 @@ require (
 	github.com/pkg/sftp v1.13.6
 	github.com/prometheus/blackbox_exporter v0.25.0
 	github.com/prometheus/client_golang v1.19.1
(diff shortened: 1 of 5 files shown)


kzys (approved):

btoews (approved):
LGTM

PR #3778, opened 2024-07-25
fix error handling and chaining in bluegreen deployments

gwuah:
### Change Summary
`github.com/pkg/errors.Wrap` doesn't attach anything to the error chain, so performing an `errors.Is(err, b)` against b always returns false. It's sad but that's life.


billyb2 (approved):
banger pr

PR #3779, opened 2024-07-25
fix: wait for builder machine to be started to prevent sending a bunch of restart requests

jipperinbham:
### Change Summary

What and Why:

Currently, we create the builder machine but do not wait for it to be in a started state before continuing which then results in the validateBuilder function seeing it as not started which then sends a bunch of restart requests resulting in the builder never getting into a steady state for use.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/build/imgsrc/ensure_builder.go
+++ b/internal/build/imgsrc/ensure_builder.go
@@ -394,6 +394,12 @@ func createBuilder(ctx context.Context, org *fly.Organization, region, builderNa
 		return nil, nil, retErr
 	}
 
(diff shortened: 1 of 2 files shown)


billyb2 (approved):
banger pr

thank you for this

jipperinbham:
Thanks! There appears to still be some weird race condition customers (and testflight) are seeing where the app is created successfully but then the call to create the volume returns "Could not find App" even though it's going by to the central API to validate/auth the request.

PR #3785, opened 2024-07-26
Make sure to wait before restarting the remote builder

billyb2:
### Change Summary

What and Why:

If the remote builder is in the 'created' or 'migrating' state, then we won't be able to restart it. By making sure to wait for one of those two states, we can avoid errors like:

`failed to fetch an image or build from source: failed to restart VM asfijsdoif: failed_precondition: unable to restart machine, not currently started or stopped`

How:
make sure that the builder machine is either in the started or stopped state before calling the 'restart' API

Related to:

---

### Documentation

- [x] Fresh Produce (i'm gonna fix a few more remote builder issues before FP)
- [ ] n/a

### DIFF
--- a/internal/build/imgsrc/ensure_builder.go
+++ b/internal/build/imgsrc/ensure_builder.go
@@ -15,8 +15,6 @@ import (
 	"github.com/superfly/flyctl/internal/flyutil"
 	"github.com/superfly/flyctl/internal/haikunator"
 	"github.com/superfly/flyctl/internal/tracing"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/trace"
 )
 
 func EnsureBuilder(ctx context.Context, org *fly.Organization, region string, recreateBuilder bool) (*fly.Machine, *fly.App, error) {
@@ -148,7 +146,7 @@ const (
 )
 
 func validateBuilder(ctx context.Context, app *fly.App) (*fly.Machine, error) {
-	ctx, span := tracing.GetTracer().Start(ctx, "validate_builder", trace.WithAttributes(attribute.String("bulder_app", app.Name)))
+	ctx, span := tracing.GetTracer().Start(ctx, "validate_builder")
 	defer span.End()
 
 	if app == nil {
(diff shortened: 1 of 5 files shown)


internal/build/imgsrc/ensure_builder_test.go
```diff
@@ -275,16 +276,24 @@ func TestRestartBuilderMachine(t *testing.T) {
 			return nil
 		},
 		WaitFunc: func(ctx context.Context, machine *fly.Machine, state string, timeout time.Duration) (err error) {
+			if state == "started" || state == "stopped" {
+				waitedForStartOrStop = true
```
jipperinbham, line 280: I don't understand how this test is asserting the logic correctly since the initial state of `waitedForStartOrStop` is `true`.

internal/build/imgsrc/ensure_builder.go
```diff
@@ -409,6 +410,8 @@ func restartBuilderMachine(ctx context.Context, builderMachine *fly.Machine) err
 
 	flapsClient := flapsutil.ClientFromContext(ctx)
 
+	mach.WaitForAnyMachineState(ctx, builderMachine, []string{"started", "stopped"}, 60*time.Second, nil)
```
jipperinbham, line 413: Should we be checking the return values? Or are we restarting the machine even if it's already in `started` state which seems to be the desired state?

internal/build/imgsrc/ensure_builder_test.go
billyb2: oh, yea i should set that to false by default 🤦🏾

internal/build/imgsrc/ensure_builder.go
billyb2: yes we absolutely should. thank you for pointing this out :)

internal/command/deploy/plan.go
```diff
@@ -493,11 +494,16 @@ func (md *machineDeployment) updateMachineWChecks(ctx context.Context, oldMachin
 	if !healthcheckResult.regularChecksPassed {
 		sl.LogStatus(statuslogger.StatusRunning, fmt.Sprintf("Checking health of machine %s", machine.ID))
 		err = lm.WaitForHealthchecksToPass(ctx, md.waitTimeout)
-		if err != nil {
+		switch {
+		case errors.Is(err, context.DeadlineExceeded):
```
billyb2, line 498: didn't mean to commit this, but this is a small fix to retry health checks correctly.

PR #3786, opened 2024-07-26
return after a successful builder machine restart

jipperinbham:
### Change Summary

What and Why:

We're currently checking if the `restartBuilderMachine` errored but if it doesn't then we continue on to replace the builder with an entirely new app. The changes here returns if the machine builder restart was successful and in a known good state.

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/build/imgsrc/ensure_builder.go
+++ b/internal/build/imgsrc/ensure_builder.go
@@ -65,15 +65,16 @@ func EnsureBuilder(ctx context.Context, org *fly.Organization, region string, re
 
 		if validateBuilderErr == BuilderMachineNotStarted {
 			err := restartBuilderMachine(ctx, builderMachine)
-			if err != nil {
-				if errors.Is(err, ShouldReplaceBuilderMachine) {
-					span.AddEvent("recreating builder due to resource reservation error")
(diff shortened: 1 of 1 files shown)


internal/build/imgsrc/ensure_builder.go
```diff
@@ -73,6 +73,7 @@ func EnsureBuilder(ctx context.Context, org *fly.Organization, region string, re
 					return nil, nil, err
 				}
 			}
+			return builderMachine, builderApp, nil
```
matttpt, line 76: If I'm reading this right, this also returns if `err != nil && errors.Is(err, ShouldReplaceBuilderMachine)`, which doesn't seem intended?

clouvet (approved):

internal/build/imgsrc/ensure_builder.go
jipperinbham: Yea, I was doing my own review and caught this too! Addressed via https://github.com/superfly/flyctl/pull/3786/commits/f3f196ad2cc4fa8b90246c8829516674fcca1cd8

PR #3789, opened 2024-07-27
Add option to configure fly by env

fliepeltje:
### Change Summary

What and Why:
Allow config to be set by env vars for most flags.

Let's say you're managing multiple infrastructure setups, it can be be convenient to have certain flags/options set in the environment; this is a "one-size" solution for that; the precedence is still any flags you might specify, but if no flag is specified, we check for the corresponding env var (which is all-caps, `FLY_` prefixed stuff). 

This is a draft/idea for now - I'd love to have a discussion about whether this makes sense or not :)

### DIFF
--- a/internal/flag/context.go
+++ b/internal/flag/context.go
@@ -3,6 +3,7 @@ package flag
 import (
 	"context"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rubys (approved):
Even though you described what you were doing quite clearly, it took me a minute or two to figure out what this was doing.  If would be a good idea for Michael and/or Ben to review this too.

We currently have FLY_ORG.  This change would generalize this to pretty much any flag on any command.

internal/flag/context.go
```diff
@@ -42,23 +41,29 @@ func FirstArg(ctx context.Context) string {
 	return ""
 }
 
-func CmdParentName(cmd *cobra.Command) cobra.Command {
-	parent := cmd.Parent()
-	if strings.ToUpper(parent.Name()) == "FLY" {
-		return *cmd
+func EnvNameFromCmd(cmd *cobra.Command) string {
+	if cmd.Parent() != nil {
```
rubys, line 45: Does this break existing users of FLY_ORG?

fliepeltje: No, I broke that elsewhere; this is the "fix", but with a caveat:
`FLY_ORG` by itself as an env var will do what you expect, but if you for instance do `FLY_LAUNCH_ORG`, that will get precedence. Some examples:

list apps for org xxx:
```sh
FLY_ORG=xxx fly apps list 
```

list apps for org yyy:
```sh
FLY_APPS_ORG=yyy fly apps list
```

list apps for org yyy:
```sh
FLY_ORG=xxx FLY_APPS_ORG=yyy fly apps list
```

PR #3790, opened 2024-07-27
Include Fly infra domains in the allowed domains list for synthetic checks

aschiavo:
### Change Summary

What and Why:

Update the target check to include all Fly infra domains

### DIFF
--- a/internal/metrics/synthetics/agent.go
+++ b/internal/metrics/synthetics/agent.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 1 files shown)


internal/metrics/synthetics/agent.go
```diff
@@ -139,7 +139,11 @@ func isFlyInfraTarget(target string) bool {
 		host = parsedURL.Host
 	}
 
-	_, existsInFlyInfraDomains := flyInfraHosts[host]
+	for _, flyInfraDomain := range flyInfraDomains {
+		if strings.HasSuffix(host, "."+flyInfraDomain) || host == flyInfraDomain {
```
davidgraig, line 143: Could just return the evaluation here I think?

davidgraig (approved):
Lgtm

internal/metrics/synthetics/agent.go
aschiavo: Hmmm, not sure I fully get the comment 😅 I _think_ I am . If the target host ends with any of the allowed domains or is the naked domain itself I'm returning the evaluation as positive, if the loop ends and it haven't returned for any of the allowed domains I return a nope ... What am I missing 🤔

davidgraig: oh. nevermind i was looking at this on my phone. NEVERMIND ME.

PR #3791, opened 2024-07-28
fly launch --attach:

rubys:
* Org of new app is based on current app
  * --into defaults to fly/apps/${basename(into)}
  * only a private ipv6 address is allocated for this app
  * exported secrets are set on current app


fliepeltje:
I have one idea for a DX improvement to consider for you:

Prefix the app-name with the parent app (by default). Unless a `--name` arg is passed, take the value of `--into` or default `--from` and prefix it. So:

- `fly launch --from xxx int- yyy` -> app name is *parent-yyy*
- `fly launch --from xxx` -> app name is *parent-xxx*
- `fly launch --from xxx --name zzz` -> app name is *zzz*
- `fly launch --from xxx --into yyy --name zzz` -> app name is *zzz*

rubys:
@fliepeltje sounds good to me.  Done.

PR #3794, opened 2024-07-29
add retry for socket scanning with delay

### LINKED ISSUE #3358: FlyCTL is eager to say "The app is not listening on the expected address" when the app takes longer to bind
**Describe the bug**

Myself and a [customer](https://community.fly.io/t/app-deployment-warning-despite-successful-access/18673) have been seeing flyctl report "The app is not listening on the expected address and will not be reachable by fly proxy..." very fast because our apps have slow boot times, mine being a NodeJS+puppeteer and theirs a rails app.

Deployment works as expected this might be a non-urgent display issue.

* Operating system; Mac M1 2022
* `fly version`: fly v0.2.15 darwin/arm64 Commit: 846630217aff135b32ec0d6a018cf6bdde0f1762-dirty BuildDate: 2024-03-10T09:52:28Z

** Paste your `fly.toml`

```toml
# fly.toml app configuration file generated for myapp on 2024-03-09T09:06:35+09:00
#
# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
#

app = 'myapp'
primary_region = 'nrt'
console_command = '/rails/bin/rails console'
swap_size_mb = 512

[build]

[deploy]
  release_command = './bin/rails db:prepare'

[http_service]
  internal_port = 3000
  force_https = true
  auto_stop_machines = true
  auto_start_machines = true
  min_machines_running = 0
  processes = ['app']

[[vm]]
  memory = 256
  cpu_kind = 'shared'
  cpus = 1
```

** Command output: **

```text
2024-03-11T02:41:27.581 app[784e667b274628] nrt [info] => Booting Puma

2024-03-11T02:41:27.581 app[784e667b274628] nrt [info] => Rails 7.0.4.3 application starting in production

2024-03-11T02:41:27.581 app[784e667b274628] nrt [info] => Run `bin/rails server --help` for more startup options

2024-03-11T02:41:30.007 app[784e667b274628] nrt [info] Puma starting in single mode...

2024-03-11T02:41:30.007 app[784e667b274628] nrt [info] * Puma version: 5.6.5 (ruby 3.2.2-p53) ("Birdie's Version")

2024-03-11T02:41:30.007 app[784e667b274628] nrt [info] * Min threads: 5

2024-03-11T02:41:30.007 app[784e667b274628] nrt [info] * Max threads: 5

2024-03-11T02:41:30.007 app[784e667b274628] nrt [info] * Environment: production

2024-03-11T02:41:30.008 app[784e667b274628] nrt [info] * PID: 305

2024-03-11T02:41:30.009 app[784e667b274628] nrt [info] * Listening on http://0.0.0.0:3000

2024-03-11T02:41:30.028 app[784e667b274628] nrt [info] Use Ctrl-C to stop
```

### LINKED ISSUE #3761: `fly launch` with `python` occasionally generating an (incorrect) address warning
**Describe the bug**

When launching a python app, `flyctl` sometimes throws out a warning that the app should be listening on the specified address/port, however the app is listening on the correct address/port and functions fine. This warning should not happen:

```sh
WARNING The app is not listening on the expected address and will not be reachable by fly-proxy.
You can fix this by configuring your app to listen on the following addresses:
  - 0.0.0.0:8080
 ```

### PR DESCRIPTION
**Pull request by fliepeltje**
possibly closes #3761 and possibly closes #3358

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -1141,7 +1141,7 @@ func (md *machineDeployment) warnAboutIncorrectListenAddress(ctx context.Context
 				}
 			}
 		}
-		if foundSockets > 0 {
+		if len(tcpServices) == 0  {
 			break
 		} else {
 			retries -= 1
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -1119,27 +1119,33 @@ func (md *machineDeployment) warnAboutIncorrectListenAddress(ctx context.Context
 	}
 
 	var foundSockets int
-	for _, proc := range processes {
-		for _, ls := range proc.ListenSockets {
-			foundSockets += 1
-
-			host, portStr, err := net.SplitHostPort(ls.Address)
-			if err != nil {
-				continue
+	retries := 1
+	for retries >= 0 {
+		for _, proc := range processes {
+			for _, ls := range proc.ListenSockets {
+				foundSockets += 1
+				host, portStr, err := net.SplitHostPort(ls.Address)
+				if err != nil {
+					continue
+				}
+				port, err := strconv.Atoi(portStr)
+				if err != nil {
+					continue
+				}
+				ip := net.ParseIP(host)
+				// We don't know VM's internal ipv4 which is also a valid address to bind to.
+				// Let's assume that whoever binds to a non-loopback address knows what they are doing.
+				// If we expose this address to flyctl later, we can revisit this logic.
+				if !ip.IsLoopback() {
+					delete(tcpServices, port)
+				}
 			}
-			port, err := strconv.Atoi(portStr)
-			if err != nil {
-				continue
-			}
-
-			ip := net.ParseIP(host)
-
-			// We don't know VM's internal ipv4 which is also a valid address to bind to.
-			// Let's assume that whoever binds to a non-loopback address knows what they are doing.
-			// If we expose this address to flyctl later, we can revisit this logic.
-			if !ip.IsLoopback() {
-				delete(tcpServices, port)
-			}
+		}
+		if foundSockets > 0 {
+			break
+		} else {
+			retries -= 1
+			time.Sleep(3 * time.Second)
 		}
 	}
 

### CODE REVIEW

File: internal/command/deploy/machines_deploymachinesapp.go [hunk 1]
```diff
@@ -1119,28 +1119,34 @@ func (md *machineDeployment) warnAboutIncorrectListenAddress(ctx context.Context
...
+				ip := net.ParseIP(host)
+				// We don't know VM's internal ipv4 which is also a valid address to bind to.
+				// Let's assume that whoever binds to a non-loopback address knows what they are doing.
+				// If we expose this address to flyctl later, we can revisit this logic.
+				if !ip.IsLoopback() {
+					delete(tcpServices, port)
+				}
 			}
 		}
+		if foundSockets > 0 {
```
KTanAug21, line 1144:
```suggestion
		if len(tcpServices) == 0  {
```

Notes on the suggested change above:
1. The variable `tcpServices` seems to contain the list of configured services for the app, [see lines](https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines_deploymachinesapp.go#L1150C2-L1157)

2. [In this line](https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines_deploymachinesapp.go#L1197), we can see the comment that if the tcpServices map reaches 0 length, it indicates all services are covered

3. Further, the warning returned by the function warnAboutIncorrectListenAddress, returns the warning that relies on getting values from the tcpServices map in order to list addresses that the app needs to be configured. [See lines](https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines_deploymachinesapp.go#L1204)


So, in conclusion,  the tcpServices is used to list the needed services to be configured for the app.  

So, since we want to do retries to recheck whether all services are indeed run by the app's machine, we can replace the [foundSockets > 0  condition](https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines_deploymachinesapp.go#L1193C4-L1193C16), with the[ len(tcpServices) == 0](https://github.com/superfly/flyctl/blob/master/internal/command/deploy/machines_deploymachinesapp.go#L1198C6-L1198C26) condition to stop the retry when all services are found to be covered. Hence the suggested change above!

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -1119,28 +1119,34 @@ func (md *machineDeployment) warnAboutIncorrectListenAddress(ctx context.Context
...
+			break
+		} else {
+			retries -= 1
+			time.Sleep(3 * time.Second)
```
KTanAug21, line 1148: Also! @fliepeltje,  something to ponder about: would you consider reducing the sleep time to 1 second for each retry? 
1. **Pro**: This will reduce wait time in case the services becomes reachable earlier than 3 seconds
2. **Con**: You'll have to increase the number of retries though! Since the intended logic is that most services are reachable by 3 seconds, when reducing the sleep time to 1 second, the retries have to be increased to reach that 3 second mark. So that's a drawback, since that would increase requests made to the app right. 

Also this is a very experimental question, with the aim of taking into consideration the wait time of users. But if 3 seconds is not that long, and is not any different from waiting 1 second, then I guess this comment don't need to considered even.

PR #3800, opened 2024-07-30
add FLY_STANDBY_FOR environment variable to standby machine

wjordan:
### Change Summary

What and Why:

Adds a `FLY_STANDBY_FOR` environment variable to standby machines based on the `--standby-for` option provided to `fly machine run`, `fly machine clone`, or the `fly launch` default configuration.

Applications wishing to support custom standby-machine logic (e.g., to perform custom operations to failover/recover state from a target machine that went down) may read this environment variable to find out that the current-running machine is an activated standby, and the machine ID(s) of its target(s).

(Though less common, a standby machine can have >1 comma-delimited target machines listed and the machine will be started when only one of them is down. An application could confirm which machine is down through 6pn in this case.)

How:

Sets/deletes the `MachineConfig.Env["FLY_STANDBY_FOR"]` value anywhere `MachineConfig.Standbys` is currently set/cleared.

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/machines_launchinput.go
+++ b/internal/command/deploy/machines_launchinput.go
@@ -3,6 +3,7 @@ package deploy
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/samber/lo"
 	fly "github.com/superfly/fly-go"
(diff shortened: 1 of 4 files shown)


kzys (approved):
The change looks good to me.

I've checked "In superfly/docs, or asked for help from docs team" because, why not? :)

PR #3803, opened 2024-07-31
\\Better handling for Postgres image versions and features on Stolon clusters

ndarilek:
### Change Summary

* If a backup-related function is attempted on a Stolon cluster, explicitly indicate that Flex is required rather than throwing a generic "Malformed version" error.
* If a "Malformed version" error is received, rethrow it with a message suggesting an image update.

---

### Documentation

- [ ] Fresh Produce
- [X] n/a


internal/command/postgres/backup.go
```diff
@@ -111,17 +102,30 @@ func runBackupRestore(ctx context.Context) error {
...
 		return err
 	}
 
+	if !enabled {
+		return fmt.Errorf("backups are not enabled. Run `fly pg backup enable -a %s` to enable them", appName)
+	}
+
+	// Ensure the the app has the required flex version.
+	if err := hasRequiredVersionOnMachines(appName, machines, "", backupVersion, ""); err != nil {
+		return handleMalformedVersionErrors(appName, err)
```
davissp14, line 126: We should just handle this within `hasRequiredVersionOnMachines`:

E.G. https://github.com/superfly/flyctl/blob/f4ae6f03c144e4d5ccbe454172c4d3bea6b1d85b/internal/command/postgres/postgres.go#L69

ndarilek: Ah, good call. Fixing.

davissp14: An empty version string is an indication of incompatibility, right?

ndarilek: The line I linked to isn't actually checking the image version, though. It's checking the version of the empty string being passed in as an argument. So this "Malformed version" error isn't coming from the image version at all--it comes from the fact that we're passing empty strings in as the versions to check against. For this PR it doesn't strictly matter since we're already checking the cluster type first, but it's definitely a bit misleading when it does get hit.

#3805 should make this a bit clearer. If that seems like a good path forward, I can merge it and rewrite this PR to use those version checks instead. But I think this shows how the logic here is getting harder to reason about and needs a bit of cleanup.

ndarilek: We're now routing all the Flex-related version checks through `hasRequiredFlexVersionOnMachines` which, in theory, should only return "Malformed version" errors if the cluster is actually Flex. It also returns a better error message for this case, so the old helper is gone.

Agreed that we could handle errors better for Stolon and...whatever the other option we offered was, but that feels like a bigger effort, and I'd want to be sure we're comfortable committing to it before putting more time into it.

davissp14 (approved):
As long as everything has been tested, lgtm.

PR #3805, opened 2024-07-31
Add helper to simplify Flex version checks.

ndarilek:
### Change Summary

* Simplify Postgres cluster version checks when comparing Flex versions, and assume malformed versions are likely resolvable by image updates.

---

### Documentation

- [ ] Fresh Produce
- [X] n/a

### DIFF
--- a/internal/command/postgres/postgres.go
+++ b/internal/command/postgres/postgres.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/hashicorp/go-version"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


internal/command/postgres/postgres.go
```diff
@@ -115,6 +116,22 @@ func hasRequiredVersionOnMachines(appName string, machines []*fly.Machine, clust
...
+		return fmt.Errorf("no machines provided")
+	}
+
+	if !IsFlex(machines[0]) {
+		return fmt.Errorf("not a Flex cluster")
+	}
+
+	err := hasRequiredVersionOnMachines(appName, machines, "", flexVersion, "")
+	if strings.Contains(err.Error(), "Malformed version") {
+		return fmt.Errorf("This image is not compatible with this feature. Please attempt an update with `fly image update -a %s`", appName)
```
davissp14, line 130: I don't think an image update would not address the malformed version error.  You'd typically hit a malformed version error when a non-compatible App tries to use the feature, or someone's running a forked version of Flex.

ndarilek: Fair point, I can remove the update CTA.

davissp14 (approved):

PR #3813, opened 2024-08-02
Bug fix: Ensure err isn't nil

davissp14:
Related to `fly pg` commands to perform version checks.


ndarilek:
Ah, thought this had been merged. Anyhow, circled back and fixed as part of #3803 so I'll close it. Thanks again.

PR #3817, opened 2024-08-03
fix downtime causing bug during rolling deployments

gwuah:
### Change Summary
Previously, we were lumping started & stopped machines during rolling deployments.  When the number of started machines is larger than that of stopped machines, deployments worked fine with 0 downtime. However, when stopped > started, you could run into cases where all your started machines end up in an update pool, causing downtime.

This PR fixes this by updating started & stopped machines differently. 
- Started machines are updated in batches of 33%, which means at any point, you'll have at most 33% unavailable machines.
- stopped machines are updated in batches >= 30. 

With these changes, deployments with many stopped machines will see a reduction in deployment time, since we are now concurrently updating all stopped machines in large batches.

Fixes https://github.com/superfly/customer-help/issues/73

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -771,15 +771,19 @@ func (md *machineDeployment) updateUsingRollingStrategy(parentCtx context.Contex
 			defer close(errChan)
 
 			go func() {
-				// for cold machines, we can update all of them at once.
-				// there's no need for protection against downtime since the machines are already stopped
-				startIdx += len(coldMachines)
-				poolSize := len(coldMachines)
-				if poolSize >= 50 {
-					poolSize = 50
+				var err error
+				if len(coldMachines) > 0 {
+					// for cold machines, we can update all of them at once.
+					// there's no need for protection against downtime since the machines are already stopped
+					startIdx += len(coldMachines)
+					poolSize := len(coldMachines)
+					if poolSize >= 50 {
+						poolSize = 50
+					}
+					err = md.updateEntriesGroup(ctx, group, coldMachines, sl, startIdx-len(coldMachines), poolSize)
 				}
 
-				errChan <- md.updateEntriesGroup(ctx, group, coldMachines, sl, startIdx-len(coldMachines), poolSize)
+				errChan <- err
 			}()
 
 			go func() {
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -758,9 +758,49 @@ func (md *machineDeployment) updateUsingRollingStrategy(parentCtx context.Contex
 
 	for group, entries := range entriesByGroup {
 		entries := entries
-		startIdx += len(entries)
+
+		warmMachines := lo.Filter(entries, func(e *machineUpdateEntry, i int) bool {
+			return e.leasableMachine.Machine().State == "started"
+		})
+		coldMachines := lo.Filter(entries, func(e *machineUpdateEntry, i int) bool {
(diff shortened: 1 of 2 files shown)


internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -758,39 +759,93 @@ func (md *machineDeployment) updateUsingRollingStrategy(parentCtx context.Contex
...
+			return e.leasableMachine.Machine().State != "started"
+		})
+
 		groupsPool.Go(func(ctx context.Context) error {
-			return md.updateEntriesGroup(ctx, group, entries, sl, startIdx-len(entries))
+			errChan := make(chan error, 2)
+			wg := sync.WaitGroup{}
+
+			wg.Add(1)
+			go func() {
```
rugwirobaker, line 775: Wouldn't errgroup instead of simple go routine+ wait group be useful here in terms of conciseness since we'll already using it anyway? I'm also wondering if there are not side effects of running go routines inside go routines.

internal/command/deploy/plan.go
```diff
@@ -22,6 +21,36 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	stoppedMachinesPoolSize = 30
+)
+
+type MachineLogger struct {
+	store map[string]statuslogger.StatusLine
```
rugwirobaker, line 29: Shouldn't we have a mutex since there is are lot concurrent code around this `store` map?

internal/command/deploy/plan.go
```diff
@@ -172,39 +200,90 @@ func (md *machineDeployment) updateMachinesWRecovery(ctx context.Context, oldApp
...
+			coldMachines := lo.Filter(machineTuples, func(e machinePairing, i int) bool {
+				if e.oldMachine != nil && e.oldMachine.State != "started" {
+					return true
+				}
+				if e.newMachine != nil && e.newMachine.State != "started" {
+					return true
+				}
+				return false
+			})
+
```
rugwirobaker, line 241: Perhaps another opportunity to use `errgroup`

internal/command/deploy/machines_deploymachinesapp.go
gwuah: > I'm also wondering if there are not side effects of running go routines inside go routines.
func main() is a goroutine.

internal/command/deploy/plan.go
gwuah: all actions on `store` are reads, which is concurrent safe. plus, locking could also mess up how the statuses are displayed.

internal/command/deploy/plan.go
gwuah: https://github.com/superfly/flyctl/pull/3817#discussion_r1707001486

internal/command/deploy/machines_deploymachinesapp.go
```diff
@@ -758,39 +759,93 @@ func (md *machineDeployment) updateUsingRollingStrategy(parentCtx context.Contex
 
 	for group, entries := range entriesByGroup {
 		entries := entries
-		startIdx += len(entries)
+
+		warmMachines := lo.Filter(entries, func(e *machineUpdateEntry, i int) bool {
```
billyb2, line 763: i think that having a test to make sure we're only updating a few warm machines at the same time. Your code does look solid, though.

rugwirobaker (approved):
LGTM

internal/command/deploy/machines_deploymachinesapp.go
gwuah: i'll add a test in a different PR, i'm not sure how to easily test this behaviour atm.

PR #3820, opened 2024-08-05
Allow fly logs to read --machine flag

KTanAug21:
### Change Summary

**What and Why:**
Allow `fly logs` to read the `--machine` flag or `--instance` flag to identify the machine id to filter logs for.

**How:**
1. Include a `--machine` [flag](https://github.com/superfly/flyctl/pull/3820/files#diff-2b3f3552f1349c8d16af492e7cf1f26a0cb8aa6a01faceeea4b521835833b4adR59) for the `fly logs` command
2. Make sure to update `--instance` flag as an[ alias](https://github.com/superfly/flyctl/pull/3820/files#diff-2b3f3552f1349c8d16af492e7cf1f26a0cb8aa6a01faceeea4b521835833b4adR56) to `--machine` flag.
3. Add a boolean attribute [UseAliasShortHand](https://github.com/superfly/flyctl/pull/3820/files#diff-cd4ae47be757809e391f0b84469d6a2e9a6299fb56bacc30164a89119ff5fb4eR110) in `flag.String` struct. And [read this in `makeAlias()`](https://github.com/superfly/flyctl/pull/3820/files#diff-cd4ae47be757809e391f0b84469d6a2e9a6299fb56bacc30164a89119ff5fb4eR40) to determine whether to include aliases' first characters as shorthand
3. Include a [UseAliasShortHand](https://github.com/superfly/flyctl/pull/3820/files#diff-2b3f3552f1349c8d16af492e7cf1f26a0cb8aa6a01faceeea4b521835833b4adR57) in the `machine` flag in order to allow the inclusion of its aliases( namely: --instance )' shorthands( namely: -i )

**Related to:**

https://flyio.discourse.team/t/approach-to-inconsistency-in-flags-found-in-flyctl-part-3-instance-vs-machine/6682/2
---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### How To Test Manually:
- Primary function: Run `fly logs --app <appName> --machine <machineId>`
- Other flag combinations to make sure --instance, and --machine, works with proper behavior,  please refer to [preflight test, here!](https://github.com/superfly/flyctl/pull/3820/files#diff-24f9f64b6f7b7bab981deb41053356c0217b42a43122efcfbacfce9a5c4cb123R30-R47)

### [Preflight Tests](https://github.com/superfly/flyctl/blob/master/CONTRIBUTING.md#preflight)!
- For running preflight test, checking that `fly logs` work in general with the subject flags: `make preflight-test T=TestFlyLogsMachineFlagBehavior`


internal/command/logs/logs.go
```diff
@@ -50,9 +51,9 @@ Use --no-tail to only fetch the logs in the buffer.
 		flag.Region(),
 		flag.JSONOutput(),
 		flag.String{
-			Name:        "instance",
-			Shorthand:   "i",
-			Description: "Filter by instance ID",
+			Name:        "machine",
+			Description: "Filter by machine ID",
+			Aliases:     []string{"instance"},
```
btoews, line 56: We should also avoid breaking `-i`.

btoews: Looking at this, you'll probably have to make some changes to [this code](https://github.com/superfly/flyctl/blob/master/internal/flag/flag.go#L16-L35) to make `-i` work, since one letter `-` flags need to be specified in the `Shorthand` struct field and not the `Name` field.

internal/command/logs/logs.go
```diff
@@ -63,13 +64,32 @@ Use --no-tail to only fetch the logs in the buffer.
...
+	machineID, err := GetMachineID(instanceStr, machineStr)
+	if err != nil {
+		return fmt.Errorf("failed retrieving logs: %w", err)
+	}
 
 	opts := &logs.LogOptions{
 		AppName:    appconfig.NameFromContext(ctx),
 		RegionCode: config.FromContext(ctx).Region,
-		VMID:       flag.GetString(ctx, "instance"),
+		VMID:       machineID,
```
btoews, line 92: The alias logic takes care of all this already. You just need to swap `"instance"` for `"machine"`.

internal/command/logs/logs.go
KTanAug21: Oh! Thanks for the eagle eyes @btoews ! I didn't consider the shorthand for instance

internal/command/logs/logs.go
KTanAug21: Right on, I've removed this then!

internal/command/logs/logs.go
KTanAug21: Added a new flag.String attribute for this! 

> Add a boolean attribute [UseAliasShortHand](https://github.com/superfly/flyctl/pull/3820/files#diff-cd4ae47be757809e391f0b84469d6a2e9a6299fb56bacc30164a89119ff5fb4eR110) in flag.String struct. And [read this in makeAlias()](https://github.com/superfly/flyctl/pull/3820/files#diff-cd4ae47be757809e391f0b84469d6a2e9a6299fb56bacc30164a89119ff5fb4eR40) to determine whether to include aliases' first characters as shorthand

internal/command/logs/logs.go
```diff
@@ -50,9 +50,11 @@ Use --no-tail to only fetch the logs in the buffer.
 		flag.Region(),
 		flag.JSONOutput(),
 		flag.String{
-			Name:        "instance",
-			Shorthand:   "i",
-			Description: "Filter by instance ID",
+			Name:              "machine",
+			Shorthand:         "m",
```
btoews, line 54: Should we add `-m` to `fly ssh console` to keep things consistent?

btoews (approved):
LGTM!

internal/command/logs/logs.go
KTanAug21: I've just noticed, that there are two more place where the "machine" flag is declared, "internal/command/console/console.go", "internal/command/registry/vulns.go".
These are all of the places so far with the "machine" flag mentioned (prior to inclusion of the flag in logs):

1. [internal/command/console/console.go](https://github.com/superfly/flyctl/blob/master/internal/command/console/console.go#L49)
2. [internal/command/ssh/console.go](https://github.com/superfly/flyctl/blob/master/internal/command/ssh/console.go#L43)
3. [internal/command/registry/sbom.go](https://github.com/superfly/flyctl/blob/master/internal/command/registry/sbom.go#L39)
4. [internal/command/registry/vulns.go](https://github.com/superfly/flyctl/blob/master/internal/command/registry/vulns.go#L45)

All of the places above, with the exception of number 4, somehow do not have any shorthand for  their "machine" flags. @btoews Do you think there might  be a reason for this? I've just removed the shorthand "m" I've declared for logs.go, just in case there's any particular reason why this shorthand is not included for the others, and to also follow consistency with the other commands above.

btoews: I'm not sure what the philosophy is or if there even is one. It might be that having a shorthand and normal option feels like adding two more CLI flags, which feels like too much when there are a lot of flags on some of these commands. It might be a concern about conflicting with other flags. It might be that people don't expect it to be used enough to warrant a shorthand.

Personally, I use the `--machine`/`--instance` flags a lot and would appreciate having a `-m` flag wherever possible...

KTanAug21: ok! Thanks a lot for another round of great insight @btoews ! 

Gonna merge this within the day, I'ma try to pass all the tests that failed

PR #3824, opened 2024-08-08
Update fly auth token desc

andie787:
### Change Summary

What and Why:

See https://github.com/superfly/docs/pull/1741 for context. Suggest using an org token for API auth in the `fly auth token` help docs.

How:

Update desc text

Related to:

https://github.com/superfly/docs/pull/1741

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/auth/token.go
+++ b/internal/command/auth/token.go
@@ -19,7 +19,7 @@ func newToken() *cobra.Command {
 		long = `Shows the authentication token that is currently in use by flyctl.
 The auth token used by flyctl may expire quickly and shouldn't be used in places
 where the token needs to keep working for a long time. For API authentication, you
-can use `fly tokens create` instead, to create narrowly-scoped tokens with 
+can use the fly tokens create command instead, to create narrowly-scoped tokens with
 a custom expiry.`
 
 		short = "Show the current auth token in use by flyctl."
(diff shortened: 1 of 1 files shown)


internal/command/auth/token.go
```diff
@@ -16,11 +16,13 @@ import (
...
 	const (
-		long = `Shows the authentication token that is currently in use.
-This can be used as an authentication token with API services,
-independent of flyctl.
-`
-		short = "Show the current auth token"
+		long = `Shows the authentication token that is currently in use by flyctl.
+The auth token used by flyctl may expire quickly and shouldn't be used in places
+where the token needs to keep working for a long time. For API authentication, you
+can use the fly tokens create command instead, to create narrowly-scoped tokens with
```
btoews, line 22: It looks like in other places we wrap commands in double quotes.

```suggestion
can use the "fly tokens create" command instead, to create narrowly-scoped tokens with
```

btoews (approved):
Thanks for opening this!

internal/command/auth/token.go
andie787: yeah I'll try that! I originally had backtiks which obviously failed because multiline

PR #3829, opened 2024-08-09
don't pin debian version

rubys:
Use the debian version selected by the python packagers

See https://community.fly.io/t/error-running-fly-launch-for-the-first-time-cant-fetch-python-3-10-6-bookworm-image/21220

### DIFF
--- a/scanner/templates/python-docker/Dockerfile
+++ b/scanner/templates/python-docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:{{ .pyVersion }}-bookworm AS builder
+FROM python:{{ .pyVersion }} AS builder
 
 ENV PYTHONUNBUFFERED=1 \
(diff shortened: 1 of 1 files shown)


kzys (approved):

fliepeltje:
This is a quick fix that will work; however based on my understanding, it is recommended to pin the debian version as it is the "supported" image that includes any security fixes and it _should_ work for all latest minor versions of python. 

I'd prefer the following flow:
- check if the pinned version is available for the given python version; if so, use it
- if it isn't, throw a warning and use the non-pinned version. 

Maybe that makes more sense for a separate issue/PR as this is directly impeding launches.

fliepeltje:
I've made a new issue for the above: https://github.com/superfly/flyctl/issues/3830

fliepeltje (approved):

PR #3837, opened 2024-08-13
Allow specifying org name when launching

billyb2:
### Change Summary

What and Why:
Allowing the user to specify the org's name makes since, since the org's slug is more of an internal thing.

How:
Try to match the name of the org if we can't find a slug we're trying to launch to (backwards compatibility).

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a


KTanAug21 (approved):
Works great!

PR #3838, opened 2024-08-13
Don't ignore result of confirmation prompt when updating Postgres config

ndarilek:
### Change Summary

* Don't ignore result of confirmation prompt when updating Postgres config

Related to: https://github.com/superfly/postgres-stub/issues/59

---

### Documentation

- [ ] Fresh Produce
- [X] n/a

### DIFF
--- a/internal/command/postgres/config_update.go
+++ b/internal/command/postgres/config_update.go
@@ -314,7 +314,7 @@ func resolveConfigChanges(ctx context.Context, app *fly.AppCompact, manager stri
 			switch confirmed, err := prompt.Confirmf(ctx, msg); {
 			case err == nil:
 				if !confirmed {
-					return false, nil, nil
+					return false, nil, fmt.Errorf("cancelled")
 				}
 			case prompt.IsNonInteractive(err):
 				return false, nil, prompt.NonInteractiveError("yes flag must be specified when not running interactively")


internal/command/postgres/config_update.go
```diff
@@ -311,20 +316,20 @@ func resolveConfigChanges(ctx context.Context, app *fly.AppCompact, manager stri
 		if !autoConfirm {
 			const msg = "Are you sure you want to apply these changes?"
 
-			switch confirmed, err := prompt.Confirmf(ctx, msg); {
+			switch shouldApply, err = prompt.Confirmf(ctx, msg); {
 			case err == nil:
-				if !confirmed {
-					return false, nil, nil
+				if !shouldApply {
+					return false, false, nil, nil
```
benwaffle, line 322: Can we just return `fmt.Errorf("cancelled")`? Looks like other early returns use this already.

This will simplify the return type - it's pretty hard to understand `(bool, bool, map[string]string, error) ` without reading the code.

benwaffle (changes requested):

internal/command/postgres/config_update.go
ndarilek: Nice, wondered if we could use the error to short-circuit that path for the same reasons as you. Fixed.

benwaffle (approved):

PR #3847, opened 2024-08-14
defer the running of bundle install in the rails scanner to the callback

rubys:


scanner/rails.go
```diff
@@ -332,24 +320,24 @@ func RailsCallback(appName string, srcInfo *SourceInfo, plan *plan.LaunchPlan, f
...
 	}
 
+	cmd := exec.Command(bundle, "install", "--quiet")
+	cmd.Stdin = nil
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	err = cmd.Run()
+	if err != nil {
+		return errors.Wrap(pendingError, "Failed to install bundle, exiting")
```
kzys, line 338: Is it okay to throw away `err`? The original code is wrapping the error from `cmd.Run()`.

rubys: Good catch!

PR #3856, opened 2024-08-15
Improve error messages for failed deployments

gwuah:
### Change Summary
Improve error messages on failed deploys
- Remove duplicated error messages
- Prevent `Your machine never reached the state "".` type messages.
- only log timeout on timeout err


billyb2 (approved):
banger pr

PR #3857, opened 2024-08-15
test: make Postgres tests more stable

kzys:
### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


ndarilek:
It's hard for me to scrape through all these failures, but I see there are still 11. Has the size bump reduced the number of Postgres-related failures any?

matttpt (approved):
Just curious, any reason for switching some of these from `require` to `assert`? Otherwise, LGTM!

kzys:
`require.*` assertions kills the test right away. For cases where you could continue the execution of a test, it would be better to use `assert.*` functions. You can get more information from one test failure.

PR #3864, opened 2024-08-16
Deploy: host statics from tigris buckets

alichay:
### Change Summary

What and Why: This makes `fly deploy` support deploying static files to tigris buckets instead of using a web server integrated in the machine's init process. This makes everything a little more efficient, allows serving static files when machines are stopped, and paves the way for further improvements to statics. :)

How: When deploying, if the app has static files defined, it will use (or create) a tigris bucket named after the app. Files are uploaded to this bucket, and a couple versions of file history are kept.

Related to: https://github.com/superfly/tokenizer/pull/23
Depends on: https://github.com/superfly/fly-go/pull/97

---

### Documentation

- [ ] Fresh Produce - (soon)
- [ ] In superfly/docs, or asked for help from docs team - (soon)
- [ ] n/a


btoews:
I deployed this branch to tokenizer.fly.io if that's helpful for testing your other changes.

internal/command/deploy/machines_tigrisstatics.go
```diff
@@ -0,0 +1,617 @@
...
+	if len(pubBytes) != 32 {
+		return "", fmt.Errorf("bad public key size: %d", len(pubBytes))
+	}
+
+	sct, err := box.SealAnonymous(nil, inputJson, (*[32]byte)(pubBytes), nil)
+	if err != nil {
+		return "", err
+	}
+
+	return base64.StdEncoding.EncodeToString(sct), nil
```
btoews, line 211: I'm curious why you aren't using the tokenizer library for doing all this? You should be able to do something like

```go
secret := &tokenizer.Secret{
  AuthConfig: tokenizer.NewFlyioMacaroonAuthConfig(....)
  ProcessorConfig: &tokenizer.Sigv4ProcessorConfig{...}
}

sealed, err := secret.Seal(sealKey)
```

If you want to use the version of tokenizer from your branch, you can run `go get github.com/superfly/tokenizer@feat/sigv4`

internal/command/deploy/machines_tigrisstatics.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+	type tokenizeInput struct {
+		Sigv4Processor          sigv4Processor          `json:"sigv4_processor"`
+		FlyioMacaroonAuthConfig flyioMacaroonAuthConfig `json:"flyio_macaroon_auth"`
+		AllowedHosts            []string                `json:"allowed_hosts"`
+	}
+
+	// TODO(allison): How do we handle moving an app between orgs?
+	//                We're locking this token behind a hard dependency on the App ID and Org ID, but I don't believe
+	//                *either* of these stay static when moving from one org to another.
```
btoews, line 176: Yeah. The org id would change, but the app id would remain the same. The sealed token would need to be regenerated in this case.

internal/command/deploy/machines_tigrisstatics.go
```diff
@@ -0,0 +1,617 @@
...
+
+	sct, err := box.SealAnonymous(nil, inputJson, (*[32]byte)(pubBytes), nil)
+	if err != nil {
+		return "", err
+	}
+
+	return base64.StdEncoding.EncodeToString(sct), nil
+}
+
+type headerInjectTransport struct {
```
btoews, line 214: Same here. Using `tokenizer.Transport()` would hopefully make this easier. Is something not working with that?

internal/command/deploy/machines_tigrisstatics.go
```diff
@@ -0,0 +1,617 @@
...
+	parsedProxyUrl, err := url.Parse(tokenizerUrl)
+	if err != nil {
+		// Should be impossible, this is not runtime-controlled and issues would be caught before release.
+		return fmt.Errorf("could not parse tokenizer URL: %w", err)
+	}
+	s3HttpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	s3HttpTransport.Proxy = http.ProxyURL(parsedProxyUrl)
+
+	cfg := flyconfig.FromContext(ctx)
+	userAuthHeader := cfg.Tokens.GraphQLHeader()
```
btoews, line 283: This isn't guaranteed to give you macaroons, but should work for testing with your account. When you get things working, we can circle back to a more reliable way to get a macaroon for the request.

btoews (commented):
Looking great so far. I only read the tokenizer-related parts.

PR #3879, opened 2024-08-21
Support `master` branch name in Actions workflow

strugee:
### Change Summary

What and Why: should be self-evident - folks who haven't/forgot to/etc. configure `init.defaultBranch` to `main` will be bit by this, i.e. the workflow will not run.

How: we do the thing. Note that I did not test this.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


NullHypothesis:
Hey @strugee, thanks for the PR and apologies for the two year delay 😅

PR #3890, opened 2024-08-23
Create a standby machine for LSVD process with mounts

wjordan:
### Change Summary

What and Why:

This adds a special case to `fly deploy` for LSVD processes (detected by a `FLY_LSVD_*` environment variable defined in their machine config), to allow creating a standby machine for LSVD groups with mounted volumes.

An upcoming feature will allow an LSVD-enabled machine with a mounted volume to fail over its volume data to a standby machine, and this PR will allow automatically creating such a setup through `fly deploy`.

How:

- Add some logic to detect an LSVD process group, and use this to allow creating a standby machine even with a mounted volume configured
- Add a second mount for this configuration in `provisionVolumesOnFirstDeploy`

Related to:

#3800

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/deploy_first.go
+++ b/internal/command/deploy/deploy_first.go
@@ -3,6 +3,7 @@ package deploy
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/docker/go-units"
 	"github.com/samber/lo"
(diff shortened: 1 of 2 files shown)


wjordan:
One thing making me hesitate about this change is that as opposed to a created stopped Machine (which has no associated cost), creating a 'standby-volume' in advance adds extra ongoing cost to the standby setup that I'm not sure we want to introduce.

PR #3891, opened 2024-08-23
add support for increment/decrement values in `fly scale count`

wjordan:
Use `fly scale count +n` or `fly scale count -- -n` to
increment or decrement count relative to the current count.


internal/command/scale/count.go
```diff
@@ -92,30 +92,51 @@ func runScaleCount(ctx context.Context) error {
...
 	groups := make(map[string]int)
 
+	var machineGroups map[string][]*fly.Machine
+
+	apply := func(group string, count string) error {
+		delta := 0
+		if strings.HasPrefix(count, "+") || strings.HasPrefix(count, "-") {
+			if machineGroups == nil {
+				flapsClient := flapsutil.ClientFromContext(ctx)
+				machines, _, err := flapsClient.ListFlyAppsMachines(ctx)
```
dangra, line 105: this is a bit lame tbh, it needs two machine listings (one here, another in runMachineScaleCount) which is the most "expensive op"  and more probe to fail than any other call in this command.

I'd like to propose an alternative, return: `map[string]struct{absolute, relative int}`

and compute the expectedGroupCounts inside `runMachineScaleCount` after fetching the machine list.

if you feel brave, it can be moved inside `computeActions` where machines are already grouped by process group and absolute values can be computed easily; also tested.

wjordan: updated in d385475cba57a3be114b46599806045790bc6c50

dangra (approved):

internal/command/scale/count_machines.go
```diff
@@ -255,9 +255,9 @@ func computeActions(machines []*fly.Machine, expectedGroupCounts groupCounts, re
 	machineGroups := lo.GroupBy(machines, func(m *fly.Machine) string {
 		return m.ProcessGroup()
 	})
-	expectedCounts := lo.MapValues(expectedGroupCounts, func(c groupCount, _ string) int {
+	expectedCounts := lo.MapValues(expectedGroupCounts, func(c groupCount, group string) int {
 		if c.relative != 0 {
-			return len(machines) + c.relative
+			return len(machineGroups[group]) + c.relative
```
dangra, line 260: does it need a edge case check for non-negative absolute numbers ?

wjordan: not sure what you mean here, can you give an example?

dangra: Sorry, I wasn't clear. I mean if `c.relative` is negative and larger than `len(machineGroups[group]` the final absolute count for the group will be negative.

I was thinking on wrapping it like: `lo.Max([]int{0, len(machineGroups[group]) - c.relative})` so negative counts aren't possible.

dangra: Oh, I just tried to run `fly scale count app=-1` (on main branch) and it hung. flyctl process consumes 100%, likely entered a infinite loop.

dangra: just a suggestion:

```diff
diff --git a/internal/command/scale/count_machines.go b/internal/command/scale/count_machines.go
index 894d22fe..37260dbb 100644
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -362,7 +362,7 @@ func convergeGroupCounts(expectedTotal int, current map[string]int, regions []st
                }
        }

-       diff := expectedTotal
+       diff := lo.Max([]int{0, expectedTotal})
        for _, region := range regions {
                diff -= (current[region] + diffs[region])
        }

```

wjordan: ahh ok, it looks like an existing bug, but since we're touching this code it does make sense to fix - a2130157d8a70cdb3c4bcfee0d25d35da6dc4f71

dangra (approved):
Amazing :shipit:

PR #3895, opened 2024-08-26
add fly scale count --env

wjordan:
### Change Summary

What and Why:

Support environment-variable overrides (`--env`) when using `fly scale count`. I have a use-case for this that involves scaling up an extra machine encoded with another specific machine's ID in its environment (so it can recover specific data during a host failure), and it seemed like a generally-useful flag for consistency with other machine-creation commands.

How:

Add some code to add env variable overrides to the launch configs, plus cleaning up other uses of `env` into a helper function (`flag.Env()` to remove some code duplication.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/scale/count_machines.go
+++ b/internal/command/scale/count_machines.go
@@ -11,6 +11,7 @@ import (
 	fly "github.com/superfly/fly-go"
 	"github.com/superfly/flyctl/helpers"
 	"github.com/superfly/flyctl/internal/appconfig"
+	"github.com/superfly/flyctl/internal/cmdutil"
 	"github.com/superfly/flyctl/internal/flag"
 	"github.com/superfly/flyctl/internal/flapsutil"
 	"github.com/superfly/flyctl/internal/flyutil"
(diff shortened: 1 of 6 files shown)


internal/command/scale/count_machines.go
```diff
@@ -69,6 +70,21 @@ func runMachinesScaleCount(ctx context.Context, appName string, appConfig *appco
...
+	if env := flag.GetStringArray(ctx, "env"); len(env) > 0 {
+		parsedEnv, err := cmdutil.ParseKVStringsToMap(env)
+		if err != nil {
+			return fmt.Errorf("failed parsing environment: %w", err)
+		}
+		lo.ForEach(actions, func(plan *planItem, _ int) {
+			c := plan.LaunchMachineInput.Config
+			if c.Env == nil {
+				c.Env = make(map[string]string)
+			}
```
dangra, line 83: You can remove these lines. `lo.Assign` does the right thing when `c.Env` is `nil`

dangra (approved):

PR #3896, opened 2024-08-26
Fix panic when cloning a standby machine

dangra:
### Change Summary

image

What and Why: Fix panic 

How: Initialize machine Env if not set.

Related to: https://github.com/superfly/flyctl/pull/3800

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/machine/clone.go
+++ b/internal/command/machine/clone.go
@@ -250,7 +250,9 @@ func runMachineClone(ctx context.Context) (err error) {
 			}
 		}
 		targetConfig.Standbys = lo.Ternary(len(standbys) > 0, standbys, nil)
-		targetConfig.Env["FLY_STANDBY_FOR"] = strings.Join(standbys, ",")
+		targetConfig.Env = lo.Assign(targetConfig.Env,
+			map[string]string{"FLY_STANDBY_FOR": strings.Join(standbys, ",")},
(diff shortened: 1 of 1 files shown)


internal/command/machine/clone.go
```diff
@@ -251,6 +251,9 @@ func runMachineClone(ctx context.Context) (err error) {
 		}
 		targetConfig.Standbys = lo.Ternary(len(standbys) > 0, standbys, nil)
 		targetConfig.Env["FLY_STANDBY_FOR"] = strings.Join(standbys, ",")
+		targetConfig.Env = lo.Assign(targetConfig.Env,
+			map[string]string{"FLY_STANDBY_FOR": strings.Join(standbys, ",")},
+		)
```
wjordan, line 256: do we need to remove the original line here?

```suggestion
		targetConfig.Env = lo.Assign(targetConfig.Env,
			map[string]string{"FLY_STANDBY_FOR": strings.Join(standbys, ",")},
		)
```

dangra: oops, yes

wjordan: I'm also not sure if `lo.Assign` works with a potentially-`nil` first argument, though the existing test should check this

PR #3898, opened 2024-08-27
Generates a dedicated token for synthetics

aschiavo:
### Change Summary

What and Why:
To enable opening the synthetics system for Orgs to check their own orgs, we'd need to know the orgs that a given `flyctl` agent has access. This is a first step where we create a dedicated orgs read-only macaroon that enables the synthetics backend app to list all the orgs the `flyctl` agent has access, and use the result to register that agent instance to probe the organization apps.

How:
By looping all the Orgs returned for the access token in `flyctl` and creating a read-only macaroon for each org. 
The list of macaroons is persisted in the configuration file to avoid generating a new token each time.


internal/metrics/synthetics/token.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return "", errors.New("failed to get synthetics token")
+}
+
+func getOrgReadOnlyToken(ctx context.Context, org fly.Organization) ([][]byte, error) {
+	var (
+		token     string
+		apiClient = flyutil.ClientFromContext(ctx)
+		expiry    = "168h" // 1w
+		profile   = "deploy_organization"
```
btoews, line 78: Since web doesn't know how flyctl is going to attenuate these tokens, it'll think that they're super powerful. They'll show up in `fly tokens list` as such and it might be confusing for users. I'd recommend we make a new token ["profile"](https://github.com/superfly/web/blob/master/app/models/tokens/profiles.rb) in web, so they can be super locked down before flyctl even gets them. I can help with that change if you want.

internal/metrics/synthetics/token.go
```diff
@@ -0,0 +1,140 @@
...
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic: %v", r)
+		}
+	}()
+
+	cfg := config.FromContext(ctx)
+	if cfg.SyntheticsToken != "" {
+		terminal.Debugf("Config has synthetics token\n")
+		return cfg.SyntheticsToken, nil
```
btoews, line 55: As the user is added to and removed from orgs, this set of tokens wont change. Is that okay?

internal/metrics/synthetics/token.go
aschiavo: This is great feedback, thanks! I'd like to take a look at the profiles first, I'll ping you if I got stuck. Sounds good?

internal/metrics/synthetics/token.go
aschiavo: Good point!

For removing users from Orgs, if the token is revoked it will fail to read the organization, correct?

For adding a user to an org, maybe I can add a command to refresh the token or refresh it if the count of monitored orgs change. Do you have a recommendation?

Also, if we are adding a new token profile, maybe I can lower the expire to something like 1h or 3h and let the token to be refreshed on reconnect?

btoews: Having them expire quickly and not persisting them in the config file definitely sounds nicer.

internal/metrics/synthetics/token.go
aschiavo: Please let me know if I got it [right](https://github.com/superfly/web/pull/3393)

internal/metrics/synthetics/token.go
aschiavo: Reduced time to expire to 1h and persist only in memory to avoid having to generate a new one on reconnects which happen frequently.

btoews (approved):
LGTM

PR #3900, opened 2024-08-27
Ask launchdarkly whether to use Depot as our default builder

billyb2:
### Change Summary

What and Why:
Added a flag to LD to set Depot as the remote builder to use. Currently the rollout is set to 0%, but I plan on slowly rolling out over the course of a week.

i changed the depot flag to be a string, specifically since just saying that the default is false isn't quite right. like with deploy-retries, the default value of the flag is going to be set to "auto". In order to be backwards compatible, we make just using the `--depot` flag be equal to true.

How:

Related to:

---

### Documentation

- [x] Fresh Produce
- [ ] n/a


lubien (approved):
Lgtm!

Issue #2599, opened 2023-07-20
Deno project is not being detected


PierBover:
When doing `fly launch` a Fresh Deno project is not being detected.

```
Scanning source code
Could not find a Dockerfile, nor detect a runtime or framework from source code. Continuing with a blank app.
```

This is a pretty standard Fresh project with these files at the root:

```
/static
/routes
README.md
main.ts
islands
fresh.gen.ts
dev.ts
deno.json
/db
/components
```

Looking at the [Deno detector here](https://github.com/superfly/flyctl/blob/master/scanner/deno.go#L4) it looks like `main.ts` should be triggering it.

Also, maybe you should add `deno.json` to the scanner?

Forgot to mention I'm on `0.1.10`.

PR #3901, opened 2024-08-27
Add flyctl commands for managing secrets that are kms keys

timflyio:
### Change Summary

What and Why: Adds `flyctl secrets keys` commands for adding, removing, and listing KMS keys. This will allow users to manage secrets that they can use with the kmsfs via `/.fly/kms` in their machine.

How: Makes calls via the flaps client to new endpoints.

Related to: nomad-firecracker PR #2599 which adds flaps endpoints for managing KMS secrets.

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


timflyio:
TODO: This needs go.mod changes for new fly-go/flaps version once that lands.

timflyio:
manually tested listing, setting, and deleting kms secrets (with pr flaps client, and flaps server on dev gw), and verified that keys are visible in machine's kmsfs (with flyd pr that allows qmx machines to start kmsfs).

timflyio:
TODO: I should probably hide this command until we're ready to announce it.  I think we'll want to iterate on a few of the rough edges in kmsfs/petsem/flaps/flyctl before we freshproduce this.

internal/command/secrets/ver.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+var ErrVersNotFound = func(label string, ver Ver) error {
+	return fmt.Errorf("%s(%s): not found", label, ver)
+}
+
+// Ver is a key version. It is representable in a byte.
+// The value VerUnspec is treated as the lowest possible version
+// number, and encodes as byte(255).
+type Ver int
```
btoews, line 16: How are you feeling about the versioning stuff?

My first thought when I saw this stuff about 254 being the max version was that it was way too low. That people might want to automate rolling keys every day and that they'd quickly run out of version numbers. That got me thinking about the utility of us managing key versioning for users.

If the user can't export the key material, they shouldn't be worried about key rotation as a matter of general hygiene. If their app was compromised, they might want to roll their keys, but do we need to facilitate that for them? If they're doings lots of encryption with a block cipher with a small nonce, they might need to roll their key periodically, but our default AEAD is going to have a large nonce, so that shouldn't be a concern.

I realize that you added the versioning stuff because I suggested it would be a nice feature, but I'm second guessing that. What are your thoughts on it? Sorry to be bringing this into question after you've written so much code to facilitate versioning, here and in the kms code...

timflyio: I'm kind of liking it so far.  The UX is pretty clean.  if you use flyctl to create "secret" with no special flags, it creates "secretv0", and when you encrypt with `/.fly/kms/secretv0/encrypt` It uses that key. If you use flyctl to create "secret" again, it creates "secretv1".  Now when you encrypt with kms, it uses secretv1, but when you use decrypt with kms, it will pick secretv0 or secretv1 based on which secret was previously used to encrypt. Its simple to use and mostly transparent.
If you don't want to use key versioning, you can use the "-n" flag when creating a secret and it will just make "secret", and overwrite it when you create it more than once.  You almost dont pay anything for the versioning feature you dont want, other than the one extra byte that gets put into the ciphertexts that you didnt need.
I agree that having a max of 256 keys might be a limitation. Though I think for the use cases (as you pointed out, keys arent disclosed, so key rotation could be a pretty rare event), it might be sufficient for most.  If we want to go bigger, it would be easy to up the size to 2 or 4 bytes, adding slight bloat to the ciphertexts.  (worst case is nacl box, which needs two key versions in ciphertext).

That said, if we want to KISS, and get rid of the version of key versioning, users who need versioning could still roll their own on top of unversioned KMS ops... but it does feel like an easy feature that does what most customers will want with little effort on our part and close to zero effort for customers.

internal/command/secrets/key_set.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/superfly/flyctl/internal/flag"
+	"github.com/superfly/flyctl/iostreams"
+)
+
+func newKeySet() (cmd *cobra.Command) {
+	const (
+		long = `Set the application key secret. If the label is not fully qualified
+with a version, and a secret with the same label already exists, the label will be
+updated to include the next version number. If a base64 value is not provided, the
+key will be generated with a randomly generated value.`
```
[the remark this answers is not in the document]
timflyio: It automatically creates a new version (ie `set hs256 key`) unless you explicitly ask for a specific version (ie `set hs256 keyv25`).
It does let you overwrite existing keys.  That is a little dangerous, but I can see a use case for it -- you fat fingered an import of a key and you want to fix it for that key only, without reversioning keys.  But.. maybe there should be a flag that you need to use to force and overwrite, and an error if you try to overwrite without the flag.

internal/command/secrets/ver.go
btoews: Cool. I'm sold on versioning then :-)

Could we do some variable-length encoding to allow for more versions without extra bloat? In the ciphertext, we could use the first two bits of the first version byte to indicate how many bytes of version information are included. 

- `00` - 1 byte - versions 0-63
- `01` - 2 bytes - versions 64-16383
- `10` - 3 bytes - versions 16384-4194303
- `11` - 4 bytes - versions 4194304-1073741823

timflyio: Rather than coding my own, I decided to go with encoding/binary's Uvarint format (based on protobufs). Its slightly less compact, but still offers a large range.

internal/command/secrets/key_set.go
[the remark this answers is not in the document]
timflyio: Add a `--force` flag, and errors if you try to overwrite an existing label without the flag.

internal/command/secrets/ver.go
btoews: Nice. I wasn't aware of that!

internal/command/secrets/key_delete.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/spf13/cobra"
+	"github.com/superfly/fly-go/flaps"
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/flag"
+)
+
+func newKeyDelete() (cmd *cobra.Command) {
+	const (
+		long = `Delete the application key secret by label. The label must be fully qualified
+with the key version.`
```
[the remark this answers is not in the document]
btoews: Since this is non-reversible, should we add a `prompt.Confirm()`?

btoews (approved):
I gave this a closer read through now that we've sorted out the philosophical questions. This is looking good to me!

timflyio:
I updated key deletion to include some versioning smarts, prompting (suppressable with force flag) and output indicating which keys were deleted:

>   Key deletion will iterate all existing keys, and delete any keys that match the command line label. If the command line label is unversioned, it will match all key versions. It will prompt before each deletion and print out a message after each deletion. The force flag suppresses prompting. Note: this means if there are no matching keys, no delete is attempted, and no message is printed, and no error is returned.

internal/command/secrets/key_delete.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/superfly/fly-go/flaps"
+	"github.com/superfly/flyctl/internal/command"
+	"github.com/superfly/flyctl/internal/flag"
+	"github.com/superfly/flyctl/internal/prompt"
+	"github.com/superfly/flyctl/iostreams"
+)
+
+func newKeyDelete() (cmd *cobra.Command) {
+	const (
+		long = `Delete the application key secret by label. The label must be fully qualified
```
btoews, line 18: > The label must be fully qualified

This is outdated, right?

btoews (approved):
LGTM

internal/command/secrets/key_delete.go
timflyio: yup. thank you for catching this.

PR #3933, opened 2024-09-10
test: make volume tests more reliable and faster

kzys:
Using Eventually helpers makes the tests more reliable. t.Parallel() runs these three slow (~50-200 seconds) in parallel.

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


btoews (approved):
Nice

PR #3953, opened 2024-09-20
Deployer: open pr on launch

lubien:
### Change Summary

What and Why: If our deployer image sets DEPLOY_CREATE_AND_PUSH_BRANCH we will create and push a branch with modified files

How: new env DEPLOY_CREATE_AND_PUSH_BRANCH

Related to: PaaS work

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


deploy.rb
```diff
@@ -377,6 +383,17 @@
   end
 end
 
+if CREATE_AND_PUSH_BRANCH
+  in_step Step::CREATE_AND_PUSH_BRANCH do
```
jeromegn, line 387: I think we should do that after a successful deploy. Else, it probably means that the generated files are wrong.

lubien: Sounds good to me

jeromegn: Oh, you also only want to do that if `GIT_REPO` has been specified. This might sound annoying, but our tests won't always rely on the git repo being present.

PR #3960, opened 2024-09-25
Retry Depot builder initialization after 25 seconds

jsierles:
This addresses cases, especially for first-time deployers, where a builder can take longer to boot due to slow volume init, image pulls, etc.

### DIFF
--- a/internal/build/imgsrc/depot.go
+++ b/internal/build/imgsrc/depot.go
@@ -25,6 +25,7 @@ import (
 	"github.com/superfly/flyctl/internal/render"
 	"github.com/superfly/flyctl/internal/tracing"
 	"github.com/superfly/flyctl/iostreams"
+	"github.com/superfly/flyctl/retry"
 	"github.com/superfly/flyctl/terminal"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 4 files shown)


internal/build/imgsrc/depot.go
```diff
@@ -208,15 +209,27 @@ func initBuilder(ctx context.Context, buildState *build, appName string, streams
...
-	if buildErr != nil {
-		build.Finish(buildErr)
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, 25*time.Second)
+	defer cancel()
+
+	finalBuildErr = retry.Retry(timeoutCtx, func() error {
+		buildkit, buildErr = depotmachine.Acquire(ctx, build.ID, build.Token, "amd64")
+		if buildErr != nil {
+			span.RecordError(buildErr)
```
jacobwgillespie, line 223: Do you want this to live outside the retry? I'm thinking that this will mark the span as errored even if the error is temporary

jsierles: This will add an exception message in the span, so we can still track errors even if the build ends up succeeding after retries. If we can discover that a build succeeded, that would help differentiate.

PR #3982, opened 2024-10-04
fix: fix slice init length

cuishuang:
### Change Summary

What and Why:

The intention here should be to initialize a slice with a capacity of length rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


dangra:
Nice catch!  thanks.

PR #3995, opened 2024-10-11
Jphenow/forcibly prevent extensions at launch for automation

jphenow:
### Change Summary

What and Why:

Allows skipping source info so things like GitHub PR review app actions can prevent a new database due to scanner findings.

How:

`fly launch --no-db --no-redis --no-object-storage`

---

### Documentation

- [X] flyctl help
- [ ] https://github.com/superfly/fly-pr-review-apps
- [ ] Fresh Produce

### DIFF
--- a/internal/command/launch/cmd.go
+++ b/internal/command/launch/cmd.go
@@ -114,6 +114,21 @@ func New() (cmd *cobra.Command) {
 			Hidden:      true,
 		},
 		flag.Bool{
+			Name:        "no-db",
+			Description: "Skip automatically provisioning a database",
+			Default:     false,
+		},
+		flag.Bool{
+			Name:        "no-redis",
+			Description: "Skip automatically provisioning a Redis instance",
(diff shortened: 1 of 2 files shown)


alichay (approved):

rubys (approved):
No objection, but I'll mention in passing that more fine grained control can be achieved using [`fly launch plan`](https://community.fly.io/t/experimental-fly-launch-plan-a-more-granular-interface-to-fly-launch/21357) and skipping the steps you don't want.

jphenow:
ah we can skip the various steps if it were driven by that I suppose yes. I'll admit I've been confused at what happens where for each of those commands but from this angle it makes a bit more sense to me.

I think I'll take a stab at updates to the action repo with just that instead of adding this, then

rubys:
There is value in your change as written, might as well merge it.  Which way you want to go with review apps is up to you.

PR #4014, opened 2024-10-18
Upgrade golangci-lint

kzys:
### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/doctor/diag/diag.go
+++ b/internal/command/doctor/diag/diag.go
@@ -96,12 +96,12 @@ add the --force flag to send us best-effort diagnostics.`)
 
 		if err = ft.fn(ctx, zip); err != nil {
 			if ft.expect {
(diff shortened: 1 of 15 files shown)


internal/command/doctor/diag/diag.go
```diff
@@ -96,12 +96,12 @@ add the --force flag to send us best-effort diagnostics.`)
 
 		if err = ft.fn(ctx, zip); err != nil {
 			if ft.expect {
-				fmt.Printf(color.Red(fmt.Sprintf("FAILED: %s\n", err)))
+				fmt.Print(color.Red(fmt.Sprintf("FAILED: %s\n", err)))
```
dangra, line 99: is this right? seems to be ignoring`%s` in the format string

dangra: Nah, got it now.

dangra (approved):

Issue #3519, opened 2024-05-03
Bash completion script doesn't work for `flyctl` cmd anymore since v0.2.42


salim-b:
### Describe the bug

I regularly update flyctl directly from GitHub releases via [`gh install superfly/flyctl`](https://github.com/redraw/gh-install) and use the binary's default name `flyctl` (not `fly`).

Afterwards, I update the bash completion script via:

```sh
flyctl completion bash > ~/.local/share/bash-completion/completions/flyctl
```

I recently noticed that the bash completion stopped working and traced it back to flyctl v0.2.42, i.e. it stopped working from this version onwards. With flyctl v0.2.41, the completion works fine.

A quick diff reveals why – all the occurences of `flyctl` are now replaced with `fly`:

<details><summary>diff</summary>

```diff
--- /home/salim/Downloads/flyctl_completion_v0.2.41.bash	2024-05-03 13:25:56.738393399 +0200
+++ /home/salim/Downloads/flyctl_completion_v0.2.42.bash	2024-05-03 13:25:23.850232625 +0200
@@ -1,6 +1,6 @@
-# bash completion V2 for flyctl                               -*- shell-script -*-
+# bash completion V2 for fly                                  -*- shell-script -*-
 
-__flyctl_debug()
+__fly_debug()
 {
     if [[ -n ${BASH_COMP_DEBUG_FILE-} ]]; then
         echo "$*" >> "${BASH_COMP_DEBUG_FILE}"
@@ -9,41 +9,41 @@
 
 # Macs have bash3 for which the bash-completion package doesn't include
 # _init_completion. This is a minimal version of that function.
-__flyctl_init_completion()
+__fly_init_completion()
 {
     COMPREPLY=()
     _get_comp_words_by_ref "$@" cur prev words cword
 }
 
-# This function calls the flyctl program to obtain the completion
+# This function calls the fly program to obtain the completion
 # results and the directive.  It fills the 'out' and 'directive' vars.
-__flyctl_get_completion_results() {
+__fly_get_completion_results() {
     local requestComp lastParam lastChar args
 
     # Prepare the command to request completions for the program.
-    # Calling ${words[0]} instead of directly flyctl allows handling aliases
+    # Calling ${words[0]} instead of directly fly allows handling aliases
     args=("${words[@]:1}")
     requestComp="${words[0]} __complete ${args[*]}"
 
     lastParam=${words[$((${#words[@]}-1))]}
[...]

PR #3399, opened 2024-03-22
Rename root cmd from "flyctl" to "fly"

michaeldwan:
`fly --help` and friends now refer to the cli as "fly" instead of "flyctl" in auto-generated text.

There's still a bunch of "flyctl" in help text, errors, and output (like a few of the commands below) and we'll need to come back to fix those separately. Run `go run ./cmd/audit  list --description` from the root of this repo to see a few.

```
This is flyctl, the Fly.io command line interface.

Usage:
  fly [flags]
  fly [command]

Deploying apps & machines
  apps         Manage apps
  deploy       Deploy Fly applications
  launch       Create and configure a new app from source code or a Docker image
  machine      Manage Fly Machines.
  status       Show app status

Configuration & scaling
  certs        Manage certificates
  config       Manage an app's configuration
  image        Manage app image
  ips          Manage IP addresses for apps
  scale        Scale app resources
  secrets      Manage application secrets with the set and unset commands.
  volumes      Manage Fly Volumes.

Monitoring & managing things
  checks       Manage health checks
  console      Run a console in a new or existing machine
  dashboard    Open web browser on Fly Web UI for this app
  dig          Make DNS requests against Fly.io's internal DNS server
  logs         View app logs
  ping         Test connectivity with ICMP ping messages
  proxy        Proxies connections to a Fly Machine.
  releases     List app releases
  services     Show the application's services
  sftp         Get or put files from a remote VM.
  ssh          Use SSH to log into or run commands on Machines
  wireguard    Commands that manage WireGuard peer connections

Databases & extensions
  consul       Enable and manage Consul clusters
  extensions   Extensions are additional functionality that can be added to your Fly apps
  litefs-cloud LiteFS Cloud management commands
  planetscale  Provision and manage PlanetScale MySQL databases
  postgres     Manage Postgres clusters.
  redis        Launch and manage Redis databases managed by Upstash.com
  storage      Provision and manage Tigris object storage buckets

Access control
  auth         Manage authentication
  orgs         Commands for managing Fly organizations
  tokens       Manage Fly.io API tokens

Help & troubleshooting
  docs         View Fly documentation
  doctor       The DOCTOR command allows you to debug your Fly environment
  platform     Fly platform information

Additional Commands:
  agent        Commands that manage the Fly agent, a background process that manages flyctl wireguard connections
  completion   Generate the autocompletion script for the specified shell
  help         Help about any command
  jobs         Show jobs at Fly.io
  settings     Manage flyctl settings
[...]

PR #4015, opened 2024-10-21
fix: shell completion use  exec's actual name

zuisong:
### Change Summary

What and Why:

shell completion does not work
exe name does not match with `completion` subcommand's output

How:

Related to:

#3519
#3399

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/root/root.go
+++ b/internal/command/root/root.go
@@ -3,6 +3,9 @@ package root
 
 import (
 	"context"
+	"log"
+	"os"
+	"path/filepath"
 
 	"github.com/kr/text"
 	"github.com/olekukonko/tablewriter"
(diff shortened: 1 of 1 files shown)


internal/command/root/root.go
```diff
@@ -77,7 +79,13 @@ func New() *cobra.Command {
 		short = "The Fly.io command line interface"
 	)
 
-	root := command.New("fly", short, long, run)
+	exePath, err := os.Executable()
+	if err != nil {
+		panic(err)
```
dangra, line 84: A panic is too much, not sure what are the chances of `os.Executable()` to error, but I would go for `log.Printf("WARN: failed to find executable, error=%q, err)` and default `exe` to `fly`.

dangra:
thanks @zuisong !

PR #4020, opened 2024-10-21
stream release command app logs

wjordan:
This PR changes the release command logic to stream release command machine logs during execution, instead of only fetching them afterwards. This should help improve the log-fetching experience in release commands as a response the recently increased ingestion delay in the app logs endpoint.

Logs will be printed to standard error in real-time when using the `--verbose` flag, or otherwise the last 100 lines will be buffered and displayed on a command failure. (We can adjust this to display more lines, or all of them if that would be better!)

The stream will use NATS logs by default, with a fallback to polling the logs API if the wireguard connection is unavailable. By setting up the stream before starting the machine, NATS won't miss any of the release command logs.

NATS logs should provide a better experience than the app logs API for longer commands, since streaming logs can be printed in real-time to the console. It also doesn't depend on the availability of a centralized log-ingestion server, so it may be more reliable.

This PR also improves the fallback API polling, by waiting until the release command finishes fetching all of the logs (in verbose mode or if there is a command failure) by matching the `"Main child exited normally"` string at the end of the app logs when a command exits.

Finally, this fixes a bug where a log-fetching failure would exit the release command with an error, preventing the deploy from continuing. This would block deploys after successful release commands with the `--verbose` flag when the logs cluster was unavailable.

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/machines_releasecommand.go
+++ b/internal/command/deploy/machines_releasecommand.go
@@ -1,6 +1,7 @@
 package deploy
 
 import (
+	"container/ring"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


internal/command/deploy/machines_releasecommand.go
```diff
@@ -47,12 +53,74 @@ func (md *machineDeployment) runReleaseCommand(ctx context.Context) (err error)
...
+			if strings.Contains(entry.Message, "Main child exited normally") {
+				return
+			}
+		}
+	}()
+
+	fmt.Fprintln(md.io.ErrOut, "Starting machine")
+
+	if err = releaseCmdMachine.Start(ctx); err != nil {
+		fmt.Fprintf(md.io.ErrOut, "error updating release_command machine: %v\n", err)
```
dangra, line 120: is it `error starting release_command machine`?

PR #4027, opened 2024-10-23
Run `release` and `sync_docs` jobs on master branch only

clouvet:
### Change Summary

What and Why:

If the `auto-release` workflow is run from a branch other than `master`, that can result in deleting docs. This PR limits execution of the two jobs in `auto-release` to the `master` branch. 

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   release:
-    if: ${{ github.ref == 'refs/heads/master' }}
+    if: github.ref == 'refs/heads/master'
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-release.yml
```diff
@@ -10,6 +10,7 @@ permissions:
 
 jobs:
   release:
+    if: ${{ github.ref == 'refs/heads/master' }} && false
```
jsierles, line 13: I think you need to drop `false` here. And, you can then just put the conditional without brackets.

PR #4065, opened 2024-11-15
add experimental feature to enable zstd for depot builds

jipperinbham:
### Change Summary

What and Why:

For most Apps, enabling `zstd` will improve machine create times due to faster image pulls. With Fly's built-in builder, it was never able to make it possible to enable `zstd` builds and left the user having to manage the image build process on their own. However, now that we default to depot builders, it's possible to enable the feature with an additional argument, https://depot.dev/blog/building-images-gzip-vs-zstd.

How:

To start, we'll allow enabling `zstd` builds to be done with the `experimental` section of `fly.toml` or as a flag passed to `fly machine run`.

Related to:

https://community.fly.io/t/try-out-faster-machine-creation-with-zstd/18354
https://community.fly.io/t/why-machines-take-a-lot-of-time-to-be-created-sometimes/22647/2

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/command_run.go
+++ b/internal/command/command_run.go
@@ -73,6 +73,14 @@ func DetermineImage(ctx context.Context, appName string, imageOrPath string) (im
 		}
 		opts.BuildArgs = extraArgs
 
+		if cfg != nil && cfg.Experimental != nil {
+			opts.UseZstd = cfg.Experimental.UseZstd
+		}
+		// use-zstd passed through flags takes precedence over the one set in config
+		if useZstd := flag.GetBool(ctx, "use-zstd"); useZstd {
+			opts.UseZstd = useZstd
+		}
+
 		img, err = resolver.BuildImage(ctx, io, opts)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


internal/command/command_run.go
```diff
@@ -73,6 +73,14 @@ func DetermineImage(ctx context.Context, appName string, imageOrPath string) (im
 		}
 		opts.BuildArgs = extraArgs
 
+		if cfg != nil && cfg.Experimental != nil {
+			opts.UseZstd = cfg.Experimental.UseZstd
+		}
+		// use-zstd passed through flags takes precedence over the one set in config
+		if useZstd := flag.GetBool(ctx, "use-zstd"); useZstd {
```
dangra, line 80: you can use `flag.IsSpecificied(ctx, "use-zstd")` and then `opts.UseZstd = flag.GetBool(ctx, "use-zstd")` to override a true fly.toml value with `--use-zstd=false` cli flag.

dangra (approved):

internal/command/command_run.go
jipperinbham: addessed in https://github.com/superfly/flyctl/pull/4065/commits/8f0f5e5c549ef9b4954692f2c51335afdba8dfca

PR #4067, opened 2024-11-17
Default newly launched applications to auto suspend

rubys:
Continue to default applications to auto stop if:
  * a GPU is present
  * more than 2G of RAM is requested


dangra (approved):
looks good. I wonder what happens if someone later update the machine size to >2GB and redeploy.

PR #4118, opened 2024-12-12
Change launch http service checks to be service checks (was top level)

rubys:
This will (eventually) enable auto start to wait until the service is ready before the proxy starts routing requests to the machine.


dangra (approved):
Pre-approving. It may need some test case updates.

PR #4123, opened 2024-12-16
use fsnotify to wait for agent restart

wjordan:
Small fix to use fsnotify to complete waiting for an outdated agent to stop as soon as its socket is removed, instead of a fixed 1 second.
Fallback to a 1-second timeout on any error with the fsnotify watcher.

This should eliminate the 1-second pause on the first command after installing a new flyctl version.


dangra (approved):
nice! :shipit:

PR #4133, opened 2024-12-19
refactor: remove flyctl/retry

### LINKED ISSUE #4124: Test flake: testRetryBackoffContextTimeout
*Please only report specific issues with `flyctl` behavior*. Anything like a support request for your application should go to https://community.fly.io. More people watch that space and can help you faster!

**Describe the bug**
Briefly, describe what broke and provide the following details:

When upgrading flyctl to v0.3.50 via the Arch AUR package, the test run failed with the following:

```
--- FAIL: TestRetry (0.35s)
    --- FAIL: TestRetry/testRetryBackoffContextTimeout (0.20s)
        retry_test.go:135:
                Error Trace:    github.com/superfly/flyctl/retry/retry_test.go:135
                Error:          "199.99232ms" is not greater than or equal to "200ms"
                Test:           TestRetry/testRetryBackoffContextTimeout
                Messages:       elapsed time should be at least the timeout duration
        retry_test.go:137: RetryBackoff - Attempts made: 3, Elapsed time: 199.99232ms
FAIL
FAIL    github.com/superfly/flyctl/retry        0.355s
```

The package is running the tests using the following:
`go test ./... -ldflags="-X 'github.com/superfly/flyctl/internal/buildinfo.buildDate=2024-12-16T18:44:07Z'" --run=`

Looks like this might be a flaky test given the comparison failed by just under 8ms? From a quick look, the `context.WithTimeout` is starting a 200ms timer before the startTime is captured. Then it is expected that the time delta between startTime and the `time.Since` is  going to be longer than the WithTimeout but there is an arbitrary amount of time that could pass between the WithTimeout and the call to startTime. There might be other possible situations which could result in the test failure for this code. This is a racy assertion.

* Operating system

Arch linux
```
$ uname -a
Linux name 6.12.4-arch1-1 #1 SMP PREEMPT_DYNAMIC Mon, 09 Dec 2024 14:31:57 +0000 x86_64 GNU/Linux
$ go version
go version go1.23.4 linux/amd64
```

kzys:
testRetryBackoffContextTimeout has been flaky.

Instead of fixing, this change removes our own retry package. We don't have to write our own.

Fixes #4124

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/buildpacks/pack v0.36.1
 	github.com/cavaliergopher/grab/v3 v3.0.1
 	github.com/cenkalti/backoff v2.2.1+incompatible
```
kzys, line 24: I'd like to remove cenkalti/backoff (v1) eventually, but not in this PR.

kzys: No. It isn't that hard. I did that in this PR :)

PR #4158, opened 2025-01-08
add vm sizing flags to fly pg create

rubys:


davissp14:
Nice.  So what happens in the event `vm-size` is specified along with `vm-cpu`, `vm-memory`, etc.?

dangra (commented):
Oh, please try to reuse the flags and functions to compute the guest from https://github.com/superfly/flyctl/blob/master/internal/flag/machines.go

rubys:
now deprecated

PR #4180, opened 2025-01-21
Check if oldMachine is nil when updating process group

somtochiama:
### Change Summary

Sometimes, the oldMachine field in the MachineTuple during deploys is nil and we don't check for this when loading from the healthCheckPassed sync map. 
https://github.com/superfly/flyctl/blob/26a778ec23ba74394cc0112a009b5b2cb7e09352/internal/command/deploy/plan.go#L151.
			

What and Why:
This fixes a panic that users occasionally encounter during deploy

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/deploy/plan.go
+++ b/internal/command/deploy/plan.go
@@ -337,15 +337,23 @@ func (md *machineDeployment) updateProcessGroup(ctx context.Context, machineTupl
 		newMachine := machPair.newMachine
 
 		group.Go(func() error {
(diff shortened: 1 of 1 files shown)


dangra (approved):

DAlperin:
I still don't love that I don't know _why_ that is sometimes nil but this is a good defensive programming change anyway 🤷🏼

PR #4216, opened 2025-02-12
Support all VMSizeFlags in `fly postgres create`

vboginskey:
Supersedes https://github.com/superfly/flyctl/pull/4158

### Change Summary

Before this change, `fly postgres create` only supported a set of bespoke VM sizes via the --vm-size flag, making it impossible to create a postgres cluster with, say, 1 shared CPU and 1 GB of memory.

As of this change, we create the underlying machine using the entire `MachineGuest` set of options, which allows for additional customization.

The legacy prompt and error behavior given no `--vm-size` or an invalid `--vm-size` is preserved.

### Tested

- [x] No VM options specified, selected preset configuration from prompt
- [x] No VM options specified, selected custom configuration from prompt
- [x] Specified `--vm-size` only
- [x] Specified `--vm-size` and `--vm-memory`
- [x] Specified `--vm-memory` only
- [x] Specified an invalid `--vm-size`

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/postgres/create.go
+++ b/internal/command/postgres/create.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


dangra:
Good job! thanks

I was going to provide strong feedback about how CreateCluster shouldn't access flags directly but accept everything through its `params` argument when I realized that is already happening. So your changes are good and at some point we will have to move that stuff out to the caller.

PR #4218, opened 2025-02-12
Allow specifying a target container for pilot VMs

jsierles:


ssh/client.go
```diff
@@ -104,17 +104,24 @@ func (c *Client) Connect(ctx context.Context) error {
...
 
+
 	sess, err := c.Client.NewSession()
+
 	if err != nil {
 		return err
 	}
+
+	if container != "" {
+		sess.Setenv("FLY_SSH_CONTAINER", container)
```
saleemrashid, line 122: Should check and return any error from here

saleemrashid (changes requested):
LGTM apart from one comment

PR #4222, opened 2025-02-13
Allow overriding 'sleep inf' with fly machine run --shell

saleemrashid:
### Change Summary

What and Why: Allow `fly machine run --shell ubuntu sleep 10` for an ephemeral machine that auto-destroys after 10 seconds.

How: Only set `Init.Exec = []string{"/bin/sleep", "inf"}` if `Init.Cmd` isn't set, so the user can override the default `/bin/sleep inf` when `--shell` is passed

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -716,13 +716,11 @@ func determineMachineConfig(
 		// Called from `run`. Command is specified by arguments.
 		args := flag.Args(ctx)
 
-		if len(args) != 0 {
+		if len(args) > 1 {
 			machineConf.Init.Cmd = args[1:]
+		} else if input.interact {
+			machineConf.Init.Exec = []string{"/bin/sleep", "inf"}
 		}
-	}
-
-	if input.interact {
-		machineConf.Init.Exec = []string{"/bin/sleep", "inf"}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/command/machine/run.go
```diff
@@ -647,6 +647,10 @@ func determineMachineConfig(
 ) (*fly.MachineConfig, error) {
 	machineConf := mach.CloneConfig(&input.initialMachineConf)
 
+	if input.interact {
+		machineConf.Init.Cmd = []string{"/bin/sleep", "inf"}
```
dangra, line 651: I think the point of using Exec is that it also overrides the image's entrypoint for those cases where the entrypoint isn't a shell or don't know how to spawn another command like `sleep inf`

The whole point of `--shell` is to grab a console after launching the machine from any image for debugging purposes.

Otherwise you can simply run `fly m run --rm IMAGE sleep 10`

saleemrashid: Ah, I wasn’t sure if entrypoint needs to be overridden too. I can fix that.

I’m using this as `fly machine run —shell —command “run-tests.sh” sleep 300`, but the `sleep 300` is being overridden by the `sleep inf`.

dangra (approved):

PR #4229, opened 2025-02-18
support fly deploy --regions when no machines exist (first deploy)

rubys:
Default to cfg.PrimaryRegion, but override that if exactly one region is specified.

See: https://community.fly.io/t/regions-flag-is-not-picked-up-on-fly-deploy/23882

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -559,7 +559,7 @@ func deployToMachines(
 	// If exactly one region is specified, use that as the primary region
 	region := cfg.PrimaryRegion
 	if len(onlyRegions) == 1 {
-		for r, _ := range onlyRegions {
+		for r := range onlyRegions {
 			region = r
 		}
 	}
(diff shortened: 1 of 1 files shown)


michael:
Awesome! So this would default to `cfg.PrimaryRegion` (if set in fly.toml) but if `--regions ams` is provided `ams` takes precedence?

Just go get the full picture: If neither primary_region is set in fly.toml nor --regions is set, it would default to the region closest to me?

dangra (commented):
we should add another flag, this conditioned semantic on `--regions` gives me nightmares in advance.  We had `--region` before for primary region and for reasons changed it to `--regions` but with the idea of filtering what machines to update. 

At this point I would vote to add `--primary-region` flag than extending any existing flag that is used for a very different purpose.

michael:
A dedicated --primary-region flag sounds good to me. Avoids implicit semantics, and indicates the setting corresponds to the primary_region setting in fly.toml.

dangra (approved):

PR #4247, opened 2025-03-08
Ensure metrics are collected in non-interactive environments

clouvet:
### Change Summary

What and Why:

This change fixes metrics not being sent in non-interactive sessions. The previous version of `SendMetrics` was failing in non-interactive environments because it was using the original context directly with HTTP requests, which could be canceled before metrics were successfully sent, resulting in abandoned requests and lost metrics data.

How:

The fix modifies the metrics system to:
1. Refactor `SendMetrics` with proper HTTP request handling, error management, and response body cleanup
2. Implement a 15-second timeout mechanism for non-interactive environments
3. Ensure metrics collection failures don't block deployments by only logging warnings
4. Properly establish fallback mechanisms for token authentication, dropping back to the access token or api token depending on availability 

This ensures metrics are sent successfully in both interactive and non-interactive modes, with detailed logging for troubleshooting.

Related to:

https://github.com/superfly/flyctl/pull/2974

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/metrics/db.go
+++ b/internal/metrics/db.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"os"
 	"os/exec"
@@ -69,6 +68,9 @@ func FlushMetrics(ctx context.Context) error {
 			return err
 		}
 	} else {
+		// Don't check for errors in non-interactive mode
+		// because we don't want to impact other operations
+		// if metrics sending fails.
 		SendMetrics(ctx, string(json))
 	}
 
@@ -87,11 +89,10 @@ func SendMetrics(ctx context.Context, jsonData string) error {
 	endpoint := baseURL + "/metrics_post"
 	userAgent := fmt.Sprintf("flyctl/%s", buildinfo.Info().Version)
 
-	errChan := make(chan error, 1)
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
 
-	go sendMetricsRequest(endpoint, metricsToken, userAgent, []byte(jsonData), errChan)
-
-	err = waitForCompletion(errChan)
+	err = sendMetricsRequest(timeoutCtx, endpoint, metricsToken, userAgent, []byte(jsonData))
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: Metrics send issue: %v\n", err)
 	}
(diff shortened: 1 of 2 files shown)


internal/metrics/db.go
```diff
@@ -76,33 +75,85 @@ func FlushMetrics(ctx context.Context) error {
...
-	request, err := http.NewRequest("POST", cfg.MetricsBaseURL+"/metrics_post", bytes.NewBuffer([]byte(json)))
+	baseURL := cfg.MetricsBaseURL
+	endpoint := baseURL + "/metrics_post"
+	userAgent := fmt.Sprintf("flyctl/%s", buildinfo.Info().Version)
+
+	errChan := make(chan error, 1)
+
+	go sendMetricsRequest(endpoint, metricsToken, userAgent, []byte(jsonData), errChan)
+
+	err = waitForCompletion(errChan)
```
benbjohnson, line 94: Starting a goroutine and then waiting on it seems like a code smell. This could just be a synchronous call to `sendMetricsRequest()` and you could pass a `ctx` in that has a timeout instead.

internal/metrics/db.go
```diff
@@ -76,33 +75,85 @@ func FlushMetrics(ctx context.Context) error {
...
+		body, _ := io.ReadAll(resp.Body)
+		errChan <- fmt.Errorf("metrics send failed with status %d: %s", resp.StatusCode, string(body))
+		return
+	}
+
+	_, err = io.Copy(io.Discard, resp.Body)
+	if err != nil {
+		errChan <- fmt.Errorf("failed to read response body: %w", err)
+		return
+	}
```
benbjohnson, line 131: You don't need to drain the response body. You can just rely on the deferred `resp.Body.Close()` above.

internal/metrics/db.go
```diff
@@ -76,33 +75,85 @@ func FlushMetrics(ctx context.Context) error {
...
+func sendMetricsRequest(endpoint, token, userAgent string, data []byte, errChan chan<- error) {
+	request, err := http.NewRequest("POST", endpoint, bytes.NewBuffer(data))
+	if err != nil {
+		errChan <- fmt.Errorf("failed to create request: %w", err)
+		return
 	}
 
-	request.Header.Set("Authorization", authToken)
-	request.Header.Set("User-Agent", fmt.Sprintf("flyctl/%s", buildinfo.Info().Version))
+	request.Header.Set("Authorization", "Bearer "+token)
```
benbjohnson, line 109: IIRC macaroon tokens don't start with `Bearer: `, right? I could totally be wrong. It's been a while since I was in `flyctl` land. :)

benbjohnson (changes requested):

internal/metrics/db.go
```diff
@@ -70,39 +68,74 @@ func FlushMetrics(ctx context.Context) error {
...
+	userAgent := fmt.Sprintf("flyctl/%s", buildinfo.Info().Version)
+
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	err = sendMetricsRequest(timeoutCtx, endpoint, metricsToken, userAgent, []byte(jsonData))
 	if err != nil {
-		return err
+		fmt.Fprintf(os.Stderr, "Warning: Metrics send issue: %v\n", err)
 	}
```
benbjohnson, line 98: Should this still return the error so it can be caught in interactive mode?

benbjohnson (approved):
lgtm overall, just one comment

PR #4280, opened 2025-03-28
Check for cancelled context when updating machines

somtochiama:
### Change Summary

What and Why:

Rolling deploys continue on failure because we don't use context with `errGroup` and don't detect an earlier failure.

---

### Documentation

- [ ] Fresh Produce
- [x ] n/a

### DIFF
--- a/internal/command/deploy/plan.go
+++ b/internal/command/deploy/plan.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/command/deploy/plan.go
```diff
@@ -352,6 +359,13 @@ func (md *machineDeployment) updateProcessGroup(ctx context.Context, machineTupl
 
 			sl := machineLogger.getLoggerFromID(machineID)
 
+			select {
+			case <-gCtx.Done():
+				sl.LogStatus(statuslogger.StatusFailure, "skipping machine update due to earlier failure")
+				return gCtx.Err()
+			default:
+			}
```
btoews, line 367: You can just check `if err := gCtx.Err(); err != nil { ... }` directly instead of doing the `select` block.

btoews (approved):
LGTM

internal/command/deploy/plan.go
```diff
@@ -328,9 +334,10 @@ func (md *machineDeployment) updateProcessGroup(ctx context.Context, machineTupl
 	ctx, span := tracing.GetTracer().Start(ctx, "update_process_group")
 	defer span.End()
 
-	group := errgroup.Group{}
+	group, gCtx := errgroup.WithContext(ctx)
 	group.SetLimit(poolSize)
 
+	fmt.Println("machineTuples", machineTuples, "poolSize", poolSize)
```
dangra, line 340: please, don't forget to remove this debug line before merging 🙏

dangra (approved):
LGTM!  also Ben's comments are on point.

PR #4281, opened 2025-03-28
token refresh improvements

wjordan:
Two changes to improve token-refresh behavior in the client:

1. First try a `Tokens.update` _without_ the `UserURLCallback` option (which can now fetch tokens [in parallel](https://github.com/superfly/macaroon/pull/39)), and then retry with the UserURLCallback option only if the first attempt returned any errors.
2. Adjust the ([now-configurable](https://github.com/superfly/fly-go/pull/127)) `advancePrune` duration (from the default 1 minute) so the foreground thread has an advance-prune duration of 30 seconds while the background thread has 2 minutes. This should make it more likely for tokens to be refreshed in the background rather than delay foreground operations.


internal/config/tokens.go
```diff
@@ -181,11 +182,18 @@ func keepConfigTokensFresh(ctx context.Context, m *sync.Mutex, t *tokens.Tokens,
...
+	updateOpts := []tokens.UpdateOption{
+		tokens.WithDebugger(logger.FromContext(ctx)),
+		tokens.WithAdvancePrune(advancePrune),
+	}
 
 	if uucb != nil {
+		updated, err := t.Update(ctx, updateOpts...)
+		if err == nil || !strings.Contains(err.Error(), "missing user-url callback") {
+			return updated, err
+		}
```
btoews, line 196: A comment explaining that we can only parallelize if `WithUserURLCallback` is omitted might be useful. Without that knowledge it's confusing why we'd do it this way.

btoews (approved):
LGTM

PR #4293, opened 2025-04-08
Fix expectations of env parsing for secrets

### LINKED ISSUE #589: `fly secrets import`should take into account values between double quotes
When using `.env` files, it's very common to add single line certificates like so:

```
MY_CERT="-----BEGIN CERTIFICATE-----\nMIIFaz... etc ...FfLn\n-----END CERTIFICATE-----\n"
```

When using `fly secrets import` these vars get converted to:

```
MY_CERT='"-----BEGIN CERTIFICATE-----\\nMIIFaz... etc ...FfLn\\n-----END CERTIFICATE-----\\n"'
```

Which of course breaks the application as new quotes are added, and line breaks are escaped.

It's definitely possible to add a secret from a cert file with:

```
flyctl secrets set MY_CERT=- < my-cert.crt
```

But there are cases when the key is already in a single line form (eg: Google Cloud service account).

It would be nice if `import` detected when a value is already between quotes, and simply saved the string as is without escaping or adding new quotes.

### LINKED ISSUE #3002: flyctl secrets import should allow whitespace between key, value, and =
**Describe the bug**
Secrets import still doesn't allow whitespace between key, value, and =.

```text
$ cat /etc/os-release
NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os

$ flyctl version
flyctl v0.1.119 linux/amd64 Commit: e2d15d62a963c6a86daeefcfb435788575ba98f1 BuildDate: 2023-11-06T22:46:27Z
```

**Command output:**

```text
$ cat .env
APP_ENV = production

$ cat .env | flyctl secrets import
Error: Validation failed: Name only allows letters, numbers, and underscores
```

Relates to: https://github.com/superfly/flyctl/issues/1710

### LINKED ISSUE #4291: fly secrets import should tolerate trailing comments after a value
**Describe the bug**

The following syntax is tolerated by many local .env readers (e.g. dotenv, python-dotenv), and ignores the trailing comment

SIGNUP_BALANCE_BONUS=50  # amount in peso

But fly secrets import ingests it literally including the comment part, which breaks the application.

Overall, developer's expectation is that if something works locally with the local .env file, and then you run fly secrets import.... Everything should work in the cloud as-is.      If there is any difference in behavior of .env files parsing between common dotenv frameworks (node, python-dotenv, godotenv), choose most popular versuon of behavior between frameworks.

Related issues with fly secrets import:
* https://github.com/superfly/flyctl/issues/589
* https://github.com/superfly/flyctl/issues/3002

grgar:
Use [hashicorp/go-envparse#Parse](https://pkg.go.dev/github.com/hashicorp/go-envparse#Parse) to parse env vars which supports comments on end of line (fixes #4291), spaces around the equals (fixes #3002), and escaping of \n (fixes #589). Added tests for existing issues that fail with the previous implementation.

I'm not affiliated with hashicorp, there exists other hashicorp dependencies already, and this package has no dependencies itself. Let me know if you'd rather I fix the existing parser though if you want to continue to maintain your own!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,6 +39,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/haileys/go-harlog v0.0.0-20230517070437-0f99204b5a57
+	github.com/hashicorp/go-envparse v0.1.0
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/hashicorp/go-version v1.7.0
 	github.com/inancgumus/screen v0.0.0-20190314163918-06e984b86ed3
--- a/go.sum
+++ b/go.sum
@@ -358,6 +358,8 @@ github.com/haileys/go-harlog v0.0.0-20230517070437-0f99204b5a57/go.mod h1:feJwxr
 github.com/hashicorp/errwrap v1.0.0/go.mod h1:YH+1FKiLXxHSkmPseP+kNlulaMuP3n2brvKWEqk/Jc4=
 github.com/hashicorp/errwrap v1.1.0 h1:OxrOeh75EUXMY8TBjag2fzXGZ40LB6IKw45YeGUDY2I=
 github.com/hashicorp/errwrap v1.1.0/go.mod h1:YH+1FKiLXxHSkmPseP+kNlulaMuP3n2brvKWEqk/Jc4=
+github.com/hashicorp/go-envparse v0.1.0 h1:bE++6bhIsNCPLvgDZkYqo3nA+/PFI51pkrHdmPSDFPY=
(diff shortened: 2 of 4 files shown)


grgar:
I didn't realise you had a custom multiline syntax :(

Issue #4292, opened 2025-04-08
Error Creating Machine Shell


mahyarmirrashed:
I'm following the Fly.io documentation to learn the platform deeper. I ran into an issue at this point in the guide: https://fly.io/docs/machines/overview/#machines

**Describe the bug**

I am trying to follow the guide and learn more about Fly machines. Ran into an issue when trying to create a machine with a shell.

* Operating system: MacOS 15.3.1
* `fly version`: `fly v0.3.99 darwin/arm64 Commit: aa6171260b8dc91e4992afb5d65da0d8982d7f68 BuildDate: 2025-04-07T13:45:09Z`

** Paste your `fly.toml`

I'm not sure what this is yet...

** Command output: **

```console
mahyar@lioness hello-fly % fly machine run --shell
automatically selected personal organization: Mahyar Mirrashed
Error: create interactive shell app: Validation failed: Name under 63 chars using numbers, lowercase letters and dashes
```

dangra:
fixed by #4296

PR #4296, opened 2025-04-08
limit shell app names to 63 chars

dangra:
Fix for https://github.com/superfly/flyctl/issues/4292

The app created for interactive shells adds 26 chars plus a string generated from organization id whose length depends on its internal numerical value, that plus the random integer appended to the end can generate a longer than 63 chars string.

### DIFF
--- a/internal/command/machine/run.go
+++ b/internal/command/machine/run.go
@@ -530,12 +530,13 @@ func getOrCreateEphemeralShellApp(ctx context.Context, client flyutil.Client) (*
 	}
 
 	if appc == nil {
+		shellAppName := fmt.Sprintf("flyctl-interactive-shells-%s-%d", strings.ToLower(org.ID), rand.Intn(1_000_000))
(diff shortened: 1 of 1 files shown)


mahyarmirrashed:
@dangra Thanks for the quick fix.

I have a concern, though: Shouldn't we trim the Organization ID, and not the random ID part? Otherwise, for a sufficiently long Org ID, the random integer part will be truncated, leading to conflicting names.

dangra:
> I have a concern, though: Shouldn't we trim the Organization ID, and not the random ID part? Otherwise, for a sufficiently long Org ID, the random integer part will be truncated, leading to conflicting names.

the concern makes sense but we won't see another the org id increasing by another char for a good while, even if it does, the random part has enough digits to give away.

The thing is that the interactive shell app is created per org and reused if exists, so very unlikely it will hit a conflict, even less twice.

PR #4313, opened 2025-04-14
fix: clear machine lease nonce in releaseLeases

wjordan:
In the edge-case where a `fly.Machine` is updated twice in a single deploy (which currently happens with a single machine in 'canary' deployments), the lease won't get acquired during the second deploy if a nonce still exists on the Machine struct, so we need to explicitly clear this field in `machineDeployment.releaseLeases`.

Fixes `lease not found` error on cleanup for canary deploys, including an update to an existing test to demonstrate the failure (and ensure the fix in this PR works).

### DIFF
--- a/internal/command/deploy/plan_test.go
+++ b/internal/command/deploy/plan_test.go
@@ -227,6 +227,9 @@ func TestUpdateMachines(t *testing.T) {
 			}, nil
 		},
 		ReleaseLeaseFunc: func(ctx context.Context, machineID, nonce string) error {
+			if _, loaded := acquiredLeases.LoadAndDelete(machineID); !loaded {
+				t.Error("Release lease not found for machine:", machineID)
+			}
 			return nil
 		},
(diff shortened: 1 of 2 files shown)


internal/command/deploy/plan_test.go
```diff
@@ -288,10 +291,11 @@ func TestUpdateMachines(t *testing.T) {
...
 		},
-		appConfig:      &appconfig.Config{AppName: "myapp"},
-		waitTimeout:    10 * time.Second,
-		deployRetries:  5,
-		maxUnavailable: 3,
+		appConfig:       &appconfig.Config{AppName: "myapp"},
+		waitTimeout:     10 * time.Second,
+		deployRetries:   5,
+		maxUnavailable:  3,
+		skipSmokeChecks: true,
```
wjordan, line 298: this is just to make the test this modifies a bit faster (skip the 10-second smoke-check wait).

PR #4323, opened 2025-04-21
add `flyctl tokens print`

tqbf:
as in `flyctl tokens create org | flyctl tokens print`

### Change Summary

What and Why:

It prints Macaroon tokens, in a format you might have some hope of reading.

We read tokens from `stdin`.


dangra (commented):
What about renaming it to `fly tokens inspect` ?

PR #4337, opened 2025-04-25
rewrite `fly mpg attach` to merely grab the pgbouncer connection URI …

asib:
…and write it to a secret in the app; we are holding off the whole "create a new user and database" thing for now, waiting until Percona Operator v2.7.0 is released with `grantPublicSchemaAccess` (or else until we merge some code to allow executing arbitrary SQL, so we can run `GRANT`s ourselves)

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/mpg/attach.go
+++ b/internal/command/mpg/attach.go
@@ -3,35 +3,22 @@ package mpg
 import (
 	"context"
 	"fmt"
-	"strings"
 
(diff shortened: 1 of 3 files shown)


jsierles:
We're now able to run arbitrary grants, and the existing `mpg attach` works.

PR #4366, opened 2025-05-11
now that there is an inspect flag, it makes sense to default app

rubys:


Talljoe:
This has an unfortunate side effect of requiring a fly config or an appname passed in. If an appname is passed in, there is a bug where https url can't be supplied. I'm able to use `fly mcp proxy --url https://my-app.fly.dev` on 0.3.116 just fine on a machine without fly configured, but with 0.3.122 there are two possible paths:
```
# without -a
Error: the config for your app is missing an app name, add an app field to the fly.toml file or specify with the -a flag

# passing -a my-app
2025/05/15 10:13:40 Error resolving proxy URL: unsupported URL scheme: https
```

I've logged this as #4374

PR #4395, opened 2025-05-23
fix: various MCP proxy issues

lleyton:
### Change Summary

What and Why:
There are various issues which can cause the MCP proxy functionality to hang non-deterministically. This PR introduces various fixes.

How:
- Flush SSE headers on the MCP wrapper to the proxy, so it doesn't wait indefinitely for headers that may not be automatically flushed.
- Have proxy wait for SSE connection to be established, so we can receive responses after posting MCP commands to the wrapper.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### Credits
@addidotlol for helping me debug :3

### DIFF
--- a/internal/command/mcp/proxy.go
+++ b/internal/command/mcp/proxy.go
@@ -14,6 +14,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


rubys:
Thanks!

PR #4418, opened 2025-06-03
Enable the attachment of a new volume to an existing machine on fly deploy

rubys:


kzys:
The issue is fixed on the server side.

PR #4421, opened 2025-06-04
Fix restart policy from 'on-fail' to 'on-failure'

### LINKED ISSUE #2844: machine restart policy `on-fail` should be `on-failure`
The `--restart` flag for `fly m run` has `on-fail` as one of the options, but this should really be `on-failure` like it is in the Machines API and Docker's restart policy. See https://github.com/superfly/flyctl/blob/51a1b8f6bcb8e3b5106f6ea114e232dc35fdebc7/internal/command/machine/run.go#L791

Since this is already in use, I'm not sure if it should be fixed or if we can alias `on-fail` to `on-failure` and update the docs.

Related:
https://github.com/superfly/docs/pull/1043#discussion_r1333485367
cc @jipperinbham

chrismccord:
This PR fixes issue #2844 by updating the restart policy naming from 'on-fail' to 'on-failure' to align with Docker's standard restart policy naming and the Fly.io Machines API.

## Changes Made:
- Updated case statement in `internal/command/machine/run.go` to handle 'on-failure' instead of 'on-fail'
- Updated flag description text to reference the correct policy name
- Updated enum arrays in `internal/command/mcp/server/machine.go` to include 'on-failure' instead of 'on-fail'

## Verification:
- ✅ Code compiles successfully
- ✅ Tests pass
- ✅ Help text now correctly shows 'on-failure' as a valid restart policy option
- ✅ Aligns with Docker restart policy standard

Fixes #2844


tomr-a5:
#4786 landed the same fix and kept `on-fail` working for existing scripts, so im going to close this as superseded. Chris or Daniel please feel free to reopen if we are still seeing the same issue, working on cleaning up older flyctl issues and open prs.

PR #4440, opened 2025-06-20
Handle org flag for storage list

igor-ladkin:
### Change Summary

What and Why:

When you are a user of multiple organizations and try to see configured storage for one particular one running `fly storage list --org personal` will return results for all your organizations. When one of your organizations has too many storages configured you might not be able to see the ones you are looking for since we return 250 first ones.

`--org` flag was declared among supported for `fly storage list` command but was never taken into consideration within the handler. This change fixes this issue and makes behavior consisted with `fly apps list` command.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/gql/generated.go
+++ b/gql/generated.go
@@ -2664,92 +2664,101 @@ func (v *ListAddOnsAddOnsAddOnConnectionNodesAddOn) __premarshalJSON() (*__prema
 	return &retval, nil
 }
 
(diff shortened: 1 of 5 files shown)


tomr-a5:
#4870 landed the organization-scoped add-ons query for `fly storage list --org`, so im going to close this as superseded. So I am going to close this one Igor but if you still think it should be open lmk. Working on cleaning up older flyctl issues and open prs.

Issue #4382, opened 2025-05-19
fly launch silently ignores --config if the config referenced doesn't exist


leighmcculloch:
**Describe the bug**
Briefly, describe what broke and provide the following details:

* Operating system

   ```
   macOS 15.5
   ```

* `fly version`

   ```
   fly v0.3.125 darwin/arm64 Commit: c78d4555c62f257c686bd6c0280f1fb1c59a4e4f BuildDate: 2025-05-17T14:55:45Z
   ```

**Command output**

When running `fly launch` with an existing config passed in via the `--config <path/to/fly.toml>` option, if the file doesn't exist, it fails silently and acts as if no file was provided.

```
$ fly launch \
  --config non-existent-file.toml \
  --copy-config \
  --ha=false
Scanning source code
Could not find a Dockerfile, nor detect a runtime or framework from source code. Continuing with a blank app.
Creating app in /Users/user/test
We're about to launch your app on Fly.io. Here's what you're getting:
...
```

The problem was found when passing in a URL for the launch config, which doesn't appear to be supported, but also happens for any not-found file ☝ :

```
$ fly launch \
  --config https://github.com/stellar/quickstart/raw/refs/heads/flyio/.fly/fly.toml \
  --copy-config \
  --ha=false
```

Issue #4378, opened 2025-05-16
`fly deploy -c nonexistent.file` doesn't fail


ThisGuyCodes:
**Describe the bug**
When `fly deploy` is run with the `--config` flag set to a nonexistent file (and `--app` set to a valid fly app name), it continues to run the deploy anyway.

Technically I am requesting a breaking change; but this behavior is not intuitive, and caused hours of debugging for what ended up being something quite silly. I would be hard pressed to think of a situation where if I specified a config file that didn't exist I'd _want_ it to continue.

* Operating system
* `fly version`
```
flyctl v0.3.123 darwin/arm64 Commit: 9651c39a645575cb00779649c804a4b823016e1c BuildDate: 2025-05-15T15:28:06Z
```

** Paste your `fly.toml`

Literally doesn't matter, or exist, that's the point

** Command output: **

Command operates normally.

PR #4465, opened 2025-07-06
fix(config): tighten --config path handling

niki4smirn:
### Change Summary

What and Why:
The --config flag must now point to an existing file. Tests have been created to reflect this change.

Fixes https://github.com/superfly/flyctl/issues/4382 and https://github.com/superfly/flyctl/issues/4378.

Additionally, I noticed an inconsistency in how directory path passed to --config is handled. It seems that [`appConfigFilePaths`](https://github.com/superfly/flyctl/blob/b66a33df98448a0cab5435f219e06c54ff1fc97b/internal/command/command.go#L677) attempts to process directory parameter. However, in such case, the [switch statement](https://github.com/superfly/flyctl/blob/b66a33df98448a0cab5435f219e06c54ff1fc97b/internal/command/command.go#L653) would return an error. Since the --config flag is described as "Path to application configuration **file**," I don't believe directory support is necessary.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -648,16 +648,61 @@ func LoadAppConfigIfPresent(ctx context.Context) (context.Context, error) {
 		return ctx, nil
 	}
 
+	explicit := flag.GetAppConfigFilePath(ctx)
+	var candidates []string
+
+	if explicit != "" {
+		if err := validateConfigPath(explicit); err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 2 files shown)


niki4smirn:
Quick check-in: is there a preferred reviewer for changes under `internal/command/command.go`? Happy to tag the right person

PR #4488, opened 2025-07-17
call `patchRoot` before running strict validation so that old-format …

asib:
…config files aren't automatically marked as invalid

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/appconfig/serde.go
+++ b/internal/appconfig/serde.go
@@ -77,8 +77,7 @@ func LoadConfigAsMap(path string) (rawConfig map[string]any, err error) {
 		return nil, err
 	}
 
-	rawConfig, err = patchRoot(rawConfig)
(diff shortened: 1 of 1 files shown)


internal/appconfig/serde.go
```diff
@@ -73,7 +73,11 @@ func LoadConfigAsMap(path string) (rawConfig map[string]any, err error) {
 	} else {
 		err = toml.Unmarshal(buf, &rawConfig)
 	}
+	if err != nil {
+		return nil, err
+	}
 
+	rawConfig, err = patchRoot(rawConfig)
```
dangra, line 80: nit: it is like you can save a line and return directly

PR #4492, opened 2025-07-21
Update deprecation notice for unmanaged Postgres

kylemclaren:
### Change Summary

What and Why:

Replaces the previous deprecation message with a more detailed notice clarifying that unmanaged Fly Postgres is unsupported and users are responsible for its management. The message now encourages users to consider 'fly mpg' (Managed Postgres) and provides a documentation link.

### DIFF
--- a/internal/command/postgres/postgres.go
+++ b/internal/command/postgres/postgres.go
@@ -19,9 +19,9 @@ import (
 func New() *cobra.Command {
 	const (
 		short      = `Deprecated: Unmanaged Postgres cluster commands'`
-		deprecated = "DEPRECATED: Unmanaged Fly Postgres is deprecated in favor of 'fly mpg' (Managed Postgres).\n" +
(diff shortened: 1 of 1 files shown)


internal/command/postgres/postgres.go
```diff
@@ -18,10 +18,10 @@ import (
 
 func New() *cobra.Command {
 	const (
-		short      = `Deprecated: Unmanaged Postgres cluster commands'`
-		deprecated = "DEPRECATED: Unmanaged Fly Postgres is deprecated in favor of 'fly mpg' (Managed Postgres).\n" +
+		short  = `Deprecated: Unmanaged Postgres cluster commands'`
```
injoongy, line 21: Should we also change the `Deprecated:` text here to say `Warning:` or maybe `Notice:` instead?

injoongy (approved):
LGTM 🎉

PR #4526, opened 2025-08-13
feat: provision a Buildkit-based builder when -buildkit-image is specified

kzys:
In addition to using an existing builder via -buildkit-addr, this change adds -buildkit-image, which provisions a new builder with the specified image.

Eventually the image should be coming from the server-side, but we are not there yet.

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/build/imgsrc/buildkit_builder.go
+++ b/internal/build/imgsrc/buildkit_builder.go
@@ -27,11 +27,11 @@ type BuildkitBuilder struct {
 	// The client may need a WireGuard connection to reach the address.
 	addr string
 
-	// privisioner is used to provision a builder machine if needed.
+	// provisioner is used to provision a builder machine if needed.
 	provisioner *Provisioner
 }
 
-// NewBuildkitBuilder creates a builder that directly uses Builtkit instead of Docker Engine.
+// NewBuildkitBuilder creates a builder that directly uses Buildkit instead of Docker Engine.
 // addr is the address of the deamon (e.g. "foobar.flycast:1234" which is optional).
 func NewBuildkitBuilder(addr string, provisioner *Provisioner) *BuildkitBuilder {
 	if !provisioner.UseBuildkit() {


wjordan:
I'd prefer to see the new builder actually make use of some changes to the builder abstraction rather than refactoring around the abstraction for its own sake. Until there is a new builder implementation actually using this interface it's unclear if these refactoring changes will be necessary or sufficient to support it.

internal/build/imgsrc/ensure_builder.go
```diff
@@ -176,9 +202,11 @@ func validateBuilder(ctx context.Context, app *fly.App) (*fly.Machine, error) {
 
 	flapsClient := flapsutil.ClientFromContext(ctx)
 
-	if _, err := validateBuilderVolumes(ctx, flapsClient); err != nil {
-		tracing.RecordError(span, err, "error validating builder volumes")
-		return nil, err
+	if !p.useVolume {
```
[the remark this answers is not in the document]
kzys: Fixed

internal/flag/flag.go
```diff
@@ -592,6 +592,14 @@ func BuildkitAddr() String {
 	}
 }
 
+func BuildkitImage() String {
+	return String{
+		Name:        "buildkit-image",
+		Description: "Address of remote buildkit daemon (e.g. tcp://127.0.0.1:1234 or unix:///path/to/socket)",
```
wjordan, line 598: I think this needs an updated description

wjordan (approved):
I haven't done any manual testing with it yet, but I think the code changes look good in general!

internal/build/imgsrc/ensure_builder.go
```diff
@@ -325,92 +401,113 @@ func createBuilder(ctx context.Context, org *fly.Organization, region, builderNa
...
+		Guest: &guest,
+		Image: p.image(),
+	}
+
+	if buildkit {
+		config.Services = []fly.MachineService{
+			{
+				InternalPort: 1234,
+				Ports:        []fly.MachinePort{{Port: fly.IntPointer(1234)}},
+			},
```
wjordan, line 418: we may want to set autostart `on` and autostop to `stop` in this config- at least that's the config I've been using that seems to be convenient.

kzys:
Thanks. Addressed the comments.

internal/build/imgsrc/ensure_builder.go
```diff
@@ -159,9 +201,30 @@ const (
...
+
+	buildkitGRPCPort = 1234
 )
 
-func validateBuilder(ctx context.Context, app *fly.App) (*fly.Machine, error) {
+// validateBuilder returns a machine if it is availabe for building images.
+func (p *Provisioner) validateBuilder(ctx context.Context, app *fly.App) (*fly.Machine, error) {
+	machine, err := p.validateBuilderMachine(ctx, app)
+	if err != nil {
+		return nil, err
```
wjordan, line 212: because there is an existing `return machine, BuilderMachineNotStarted` case in `validateBuilderMachine` below, this needs to return the machine otherwise it causes a nil pointer deference panic.

```suggestion
		return machine, err
```

internal/build/imgsrc/buildkit_builder.go
```diff
@@ -76,18 +87,27 @@ func (r *BuildkitBuilder) buildWithBuildkit(ctx context.Context, streams *iostre
 		span.End()
 	}()
 
+	buildkitAddr := r.addr
+	if buildkitAddr == "" {
+		_, app, err := r.provisioner.EnsureBuilder(ctx, "", false)
```
wjordan, line 92: we need to properly pass region and recreateBuilder args through in order for the associated flags to work properly (eg `--recreate-builder`)

internal/build/imgsrc/ensure_builder.go
```diff
@@ -19,7 +18,67 @@ import (
...
+	}
+}
+
+func (p *Provisioner) UseBuildkit() bool {
+	return p.buildkitAddr != "" || p.buildkitImage != ""
+}
+
+const defaultImage = "docker-hub-mirror.fly.io/flyio/rchab:sha-9346699"
+
+func (p *Provisioner) image() string {
```
wjordan, line 50: I pushed a buildkit image with embedded mirror config we can use to `flyio/buildkit`, so we could use a hard-coded alias for `default` or something similar:

```suggestion
const defaultBuildkitImage = "docker-hub-mirror.fly.io/flyio/buildkit@sha256:0fe49e6f506f0961cb2fc45d56171df0e852229facf352f834090345658b7e1c"

func (p *Provisioner) image() string {
	if p.buildkitImage == "default" {
		return defaultBuildkitImage
	}
```

internal/build/imgsrc/ensure_builder.go
kzys: Oh why. Fixed.

internal/build/imgsrc/buildkit_builder.go
kzys: I'm gonna pass the parameters explicitly from buildkit_builder.go for now. Passing them to the Provisioner makes sense though.

internal/build/imgsrc/ensure_builder.go
kzys: Are you going to remove -buildkit-addr evntually? Right now, the flag and -buildkit-image are used for enabling BuildkitBuilder

wjordan: > Are you going to remove -buildkit-addr evntually? Right now, the flag and -buildkit-image are used for enabling BuildkitBuilder

no, being able to specify an endpoint directly is helpful for use-cases where the user may want to manually manage their own buildkit service.

PR #4535, opened 2025-08-27
Flyctl will now use the machines API for managing app secrets.

timflyio:
### Change Summary

What and Why: Flyctl will use the machines API to manage secrets so we can deprecate the graphql interface.

How: The `fly secrets` commands will use the machines API for managing secrets and will track the returned petsem min_version for each app in the fly config file. It will use this min_version when fetching secrets or performing deploys to ensure that the latest secrets are used. The flaps library is also updated to make use of the new app secrets update endpoint which can both set and unset a list of secrets in a single round trip.

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


timflyio:
There are some pre-flight tests which intermittently fail, different ones on different runs. It seems that restarting them causes them to eventually succeed.

rubys:
Most of the changes appear to be function signature changes. Most go programs avoid this by using the context.  In fact, that work is largely already done for `flapsClient`:

https://github.com/superfly/flyctl/blob/master/internal/flapsutil/flaps_client.go#L65-L74

A similar approach could be used with `MinSecretsVersion`.

timflyio:
> Most of the changes appear to be function signature changes. Most go programs avoid this by using the context. 

Stashing settings in the context is convenient but problematic for large projects because it puts the calling contract outside of the view of the compiler, which can no longer guarantee that you're provided all the settings that need to be made. This was a cause of frustration with the current changes where I had to verify manually in several places that the context had been provisioned properly. So there are a small number of places where I opted to change signatures (there's not that many, really) to ensure clarity and to allow the compiler to track dependencies.

> A similar approach could be used with `MinSecretsVersion`.

This is true, we could pre-emptively load up the minsecrets version into the context any time that an application name is known. Care would have to be taken when dealing with apps that are newly created to load up any values needed only after the app was known, and care would be needed to ensure that the context would be updated whenever updating secrets (meaning that functions that set secrets would have to return new contexts).  I think that the complexity here outweighs the benefits. It is a little bit of a pain to have explicit code fetch these values and update these values in a config when needed, but we at least get fine grained control of the point in time that we choose to fetch values, and the code is more explicit for future readers.

PR #4552, opened 2025-09-04
test: test "fly console --dockerfile"

kzys:
This change is a follow-up for #4551.
The --dockerfile flag was not tested and we broke that.

### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/test/preflight/fly_console_test.go
+++ b/test/preflight/fly_console_test.go
@@ -4,6 +4,8 @@
 package preflight
 
 import (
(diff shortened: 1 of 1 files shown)


saleemrashid (approved):
I don't have any specific opinions on this part of the flyctl tests, but it looks good to me. Added one suggestion too

PR #4569, opened 2025-09-19
build a flaps client in the context before setting secrets

timflyio:
### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


swilson-fly:
Tested and the `fly storage create` command is now working as expected, secrets are set successfully. Looks good to ship!

PR #4617, opened 2025-10-22
Update UPG http client to fix close before read bug

swilson-fly:
### Change Summary

What and Why: Flyctl v0.3.198 seemed to surface a latent timing bug in the unmanaged postgres http client. That client is used for communicating with Postgres flex's tiny server in order to get admin / config details about the cluster, in order for commands like `fly pg attach` and `fly config show` to work. 

How: The bug was causing us to close the response body before returning it to flyctl, causing those commands to fail with `read on closed response body`. This PR moves the `Body.Close()` to the client call, so that it's closed after it's been returned and used.

Tested and confirmed that the pg attach / config / etc commands are working with this patch when run against clusters that showed the error on v0.3.198

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/flypg/http.go
+++ b/flypg/http.go
@@ -72,7 +72,6 @@ func (c *Client) doRequest(ctx context.Context, method, path string, in interfac
 	}
 
 	if res.StatusCode > 299 {
-		// newError will read the response body; ensure we close it before returning
 		err := newError(res.StatusCode, res)
 		_ = res.Body.Close()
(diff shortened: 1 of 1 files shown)


Roadmaster (approved):
This works, I tested it, I don't grok Go enough to provide meaningful review though...  +1 on functional grounds :)

PR #4634, opened 2025-11-05
Add 30-day session timeout for CLI authentication

theoctopusperson:
Users returning to the CLI after extended periods often have stale or invalid tokens. Currently, there's no mechanism to force re-authentication for users who haven't used flyctl in a long time, leading to cryptic errors or unexpected behavior with expired credentials.

  ## Solution
  Implement a 30-day session timeout that automatically prompts users to re-authenticate when their session has expired, while preserving CI/CD pipeline functionality.

  ## Changes
  - **Add last_login timestamp tracking** in `config.yml` to track when users last authenticated
  - **Implement token age validation** in `RequireSession` preparer to check if tokens have expired (30-day timeout)
  - **Skip timeout for CI/CD pipelines** - when `FLY_ACCESS_TOKEN` or`FLY_API_TOKEN` environment variables are set, the timeout check is bypassed
  - **Display styled "Welcome back!" message** when session expires to provide clear user feedback
  - **Handle migration gracefully** - users without existing timestamps are prompted to login on first use
  - **Support both interactive and non-interactive sessions** - prompts in interactive mode, returns error in non-interactive mode
  - **Include test fixes** to ensure unit tests work with the new authentication flow

  ## CI/CD Compatibility
**No impact on automated pipelines** - tokens provided via environment variables bypass the session timeout check entirely. This ensures existing CI/CD workflows continue working without modification.

### DIFF
--- a/internal/command/auth/webauth/webauth.go
+++ b/internal/command/auth/webauth/webauth.go
@@ -31,6 +31,11 @@ func SaveToken(ctx context.Context, token string) error {
 		return err
 	}
 
+	// Record the login timestamp
+	if err := config.SetLastLogin(state.ConfigFile(ctx), time.Now()); err != nil {
+		return fmt.Errorf("failed persisting login timestamp: %w", err)
+	}
+
 	user, err := flyutil.NewClientFromOptions(ctx, fly.ClientOptions{
 		AccessToken: token,
 	}).GetCurrentUser(ctx)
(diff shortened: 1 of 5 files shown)


tomr-a5:
#4645 landed the improved 30-day session timeout with the non-interactive token bypasses, so im going to close this as superseded. Daniel feel free to reopen if the timeout is not working on the latest flyctl or you think I missed something. Working on cleaning up older flyctl issues and open prs.

PR #4656, opened 2025-11-12
less graphql on launch/deploy

lillianberryfly:


lillianberryfly:
> There was at least one spot where we do more inspecting of "personal" - should just make sure launch works on personal as expected still etc. if you haven't

I've been testing on personal org, it works yep

> CI failures seem related or no?

it's also failing on main 🤷  have been trying to figure out why but no luck

PR #4664, opened 2025-11-17
fix(wg create): break endless loop when file exists

### LINKED ISSUE #4665: `wg create`: infinite loop when file exists
**Describe the bug**

When specifying config for wireguard tunnel, if the file exists, the command enters an endless loop that can not be interrupted or killed.
```
touch wgvpn.conf
fly wg create personal ams wgvpn wgvpn.conf
```

* `fly version`

flyctl v0.3.217 linux/amd64 Commit: a6da01fd0840f18fb5c57c66e6335aba85dd08f9 BuildDate: 2025-11-14T20:30:19Z

** Command output: **

```shell
$ fly wg create personal ams wgvpn /tmp
Creating WireGuard peer "wgvpn" in region "ams" for organization personal

!!!! WARNING: Output includes private key. Private keys cannot be recovered !!!!
!!!! after creating the peer; if you lose the key, you'll need to remove    !!!!
!!!! and re-add the peering connection.                                     !!!!
Can't create '/tmp': open /tmp: file exists
Can't create '/tmp': open /tmp: file exists
Can't create '/tmp': open /tmp: file exists
Can't create '/tmp': open /tmp: file exists
Can't create '/tmp': open /tmp: file exists
Can't create '/tmp': open /tmp: file exists
Can't create '/tmp': open /tmp: file exists
```

There is a  one liner fix to return the error here.

https://github.com/superfly/flyctl/blob/a6da01fd0840f18fb5c57c66e6335aba85dd08f9/internal/command/wireguard/others.go#L75C7-L75C14

But the code looks a bit messy, so I decided to make it more readable first https://github.com/superfly/flyctl/pull/4664

abitrolly:
Fixes https://github.com/superfly/flyctl/issues/4665

### Change Summary

What and Why:

This fixes endless loop in when supplied `wg create` filename already exists. The problem was caused by `argOrPrompt` helper, which always returns the command line argument if it is set, so the prompt never appears in this case.

How:

Added `flag.GetArg(ctx, idx)` helper to clean up the logic. Then reworked the loop so that it first checks for all favorable conditions, and then runs prompt without the helper that attempts to do too much.

Now if there is a problem with config file supplied as cmdline argument, `flyctl` enters the interactive prompt to allow to fix the mistake.

Related to:

https://github.com/superfly/flyctl/issues/4665

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a

### DIFF
--- a/internal/command/wireguard/others.go
+++ b/internal/command/wireguard/others.go
@@ -43,30 +43,30 @@ func orgByArg(ctx context.Context) (*fly.Organization, error) {
 func resolveOutputWriter(ctx context.Context, idx int, prompt string) (w io.WriteCloser, mustClose bool, err error) {
 	io := iostreams.FromContext(ctx)
 	var f *os.File
-	var filename string
 
+	filename := flag.GetArg(ctx, idx)
 	for {
-		filename, err = argOrPrompt(ctx, idx, prompt)
-		if err != nil {
-			return nil, false, err
-		}
-
-		if filename == "" {
-			fmt.Fprintln(io.Out, "Provide a filename (or 'stdout')")
-			continue
-		}
-
 		if filename == "stdout" {
 			return os.Stdout, false, nil
 		}
 
-		f, err = os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
-		if err == nil {
-			return f, true, nil
+		if filename != "" {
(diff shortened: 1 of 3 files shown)


tomr-a5:
Thanks for putting this together! The infinite loop was fixed in #4891, and the linked issue is now resolved. I’m going to close this as superseded, but please reopen it if I missed anything.

PR #4671, opened 2025-11-20
Bump fly-go to v0.1.68

rianmcguirefly:
### Change Summary

This bump got reverted as part of https://github.com/superfly/flyctl/pull/4662. Bringing it back with only the minimal changes needed to re-integrate it.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/deploy/deploy_first.go
+++ b/internal/command/deploy/deploy_first.go
@@ -39,6 +39,8 @@ func (md *machineDeployment) provisionIpsOnFirstDeploy(ctx context.Context, ipTy
 		return nil
 	}
 
+	region, orgID, network := "", "", ""
+
 	switch md.appConfig.DetermineIPType(ipType) {
 	case "dedicated":
 		hasUdpService := md.appConfig.HasUdpService()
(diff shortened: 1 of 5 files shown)


internal/command/deploy/deploy_first.go
```diff
@@ -82,7 +84,7 @@ func (md *machineDeployment) provisionIpsOnFirstDeploy(ctx context.Context, ipTy
 
 	case "private":
 		fmt.Fprintf(md.io.Out, "Provisioning ip address for %s\n", md.colorize.Bold(md.app.Name))
-		v6Addr, err := md.apiClient.AllocateIPAddress(ctx, md.app.Name, "private_v6", org, nil, "")
+		v6Addr, err := md.apiClient.AllocateIPAddress(ctx, md.app.Name, "private_v6", region, orgID, network)
```
rianmcguirefly: This was passing org as the region previously. That didn't break anything because the region isn't used for `private_v6`.

`org` is always set to the app's org, so passing it isn't required as the server defaults to that.

PR #4701, opened 2025-12-04
Add `pg-major-version` flag to `mpg create`

gargakshit:
Allows us to create PG17 deployments using `flyctl`. It defaults to 16 for now.

## Example

```shell
$ fly mpg create --pg-major-version=17
```


bglw (approved):
🧦 x2

PR #4729, opened 2026-01-22
Improve error message for invalid bucket name length in `storage create`

kubosuke:
### Change Summary

**What and Why:**
When creating a storage bucket with an invalid name length, the API returns a generic error: `The specified bucket is not valid`. This makes it hard for users to understand the issue.

**How:**
Added client-side validation to check bucket name length (3-63 characters) before calling the API, with a clear error message and link to docs.

```
Error: bucket name "..." is too long: must be at most 63 characters, got 64. 
See https://www.tigrisdata.com/docs/buckets/bucket-rules/
```

**Related to:**
N/A

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/extensions/tigris/create.go
+++ b/internal/command/extensions/tigris/create.go
@@ -37,6 +37,11 @@ func create() (cmd *cobra.Command) {
 }
 
 func runCreate(ctx context.Context) (err error) {
+	bucketName := flag.GetString(ctx, "name")
+	if err := validateBucketName(bucketName); err != nil {
+		return err
+	}
+
 	appName := appconfig.NameFromContext(ctx)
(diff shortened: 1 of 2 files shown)


internal/command/extensions/tigris/create.go
```diff
@@ -37,6 +37,11 @@ func create() (cmd *cobra.Command) {
 }
 
 func runCreate(ctx context.Context) (err error) {
+	bucketName := flag.GetString(ctx, "name")
```
dangra: this is fun stuff because I can't find where the bucket name is passed to Tigris extension. It is like the `--name` cli flag is completely ignored

PR #4739, opened 2026-02-12
Cut over to machines API for certs, add ownership record + imports

bglw:
### Change Summary

Certificates are now (primarily) serviced through the Machines API rather than GraphQL. 

Alongside this change, support for custom certificate uploads has been introduced, as well as a new DNS entry that can verify ownership of a domain in certain circumstances.


bglw:
@amo13 docs are up at https://fly.io/docs/machines/api/certificates-resource/ and https://docs.machines.dev/

PR #4741, opened 2026-02-13
deploy: don't check for private IPs during dns check

pborzenkov:
Machines API returns both public and private (Flycast) allocated
addresses. Filter out the private ones from the check as our public DNS
will never return them.


dangra:
Oh, I faced the same issue last week and fixed with https://github.com/superfly/flyctl/commit/0632ef53ca5e08c6b0c47e61a1f7f48ed941c461. Does it look good to you?

pborzenkov:
@dangra Sorry, missed your comment. Yeah, looks good.

Thank you!

PR #4825, opened 2026-04-14
Fix "concurrent update in progress" errors during canary deploys

rianmcguirefly:
### Change Summary

What and Why:

Avoid possibility of failure during a canary deploy if the machine that we run the canary update on is currently stopped.

Example error:
`Failed to update machines: failed to update machine <machine id>: failed to update VM <machine id>: aborted: machine is replacing: concurrent update in progress Retrying...`

The machines API blocks concurrent updates to a machine, so we need to ensure that doesn't happen.

How:

Fix all the things

* When updating a non-started machine, we now wait for the machine update to fully complete (eg. machine transitions from `replacing` back to `stopped`) before continuing
* Avoid updating the canary machine twice
* Prefer a started machine for the canary where possible

### DIFF
--- a/internal/command/deploy/machines_deploymachinesapp.go
+++ b/internal/command/deploy/machines_deploymachinesapp.go
@@ -702,7 +702,14 @@ func (md *machineDeployment) updateExistingMachinesWRecovery(ctx context.Context
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


dangra:
@rianmcguirefly I merged this after it passed all tests so it is included in today's auto release. thanks

PR #4843, opened 2026-04-20
Remove overlaybd client path; warn on lazy_load_images

jphenow:
## Summary

The overlaybd/lazy-loading image feature is dead on the platform side:

- **flyd** no longer processes `ImageRef.Layers` — the overlaybd snapshotter was removed from hosts (nomad-firecracker#3341, merged 2025-05-22).
- **rchab** no longer serves the `/flyio/v1/buildOverlaybdImage` endpoint — the overlaybd build stages were removed (rchab#43).

Today, a customer with `[experimental] lazy_load_images = true` in `fly.toml` gets a silently broken deploy: flyctl converts the image, but the platform ignores the layers. After the rchab endpoint removal, the POST would 404 and fall through to a normal build via the existing `terminal.Warnf` — strictly better, but we shouldn't leave dead code and a misleading config flag around.

### What changed

- **Removed** the `buildOverlaybdImage` function and the `UseOverlaybd` call path from `dockerfile_builder.go`
- **Removed** `UseOverlaybd bool` from the `ImageOptions` struct in `resolver.go`
- **Removed** the overlaybd-specific test case from `daemon_url_test.go`
- **Removed** the `opts.UseOverlaybd` plumbing from `deploy_build.go`
- **Added** a deprecation warning when `[experimental] lazy_load_images = true` is set in `fly.toml`
- **Kept** `Experimental.LazyLoadImages` in the config schema so existing `fly.toml` files still parse without error

### Design decision: warn, don't error

Customers who have `lazy_load_images = true` in their `fly.toml` get a clear warning and a successful deploy — not a hard failure. This gives them a migration window to remove the stale flag without breaking their CI/CD pipelines.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./internal/...` passes
- [x] `go test ./internal/build/imgsrc/...` passes
- [ ] Manual: set `[experimental] lazy_load_images = true` in a test app's `fly.toml`, run `fly deploy --remote-only`, verify:
  - No network request to `/flyio/v1/buildOverlaybdImage`
  - Warning about the flag is printed exactly once
  - Deploy succeeds as a normal image build
- [ ] Manual: remove the flag, verify no warning printed and deploy still works

### DIFF
--- a/internal/build/imgsrc/daemon_url_test.go
+++ b/internal/build/imgsrc/daemon_url_test.go
@@ -26,12 +26,6 @@ func TestBuilderAPIURL(t *testing.T) {
 			path:       "/flyio/v1/extendDeadline",
 			want:       "http://[fdaa:49:2bd7::2]:8080/flyio/v1/extendDeadline",
 		},
-		{
-			name:       "overlaybd path",
-			daemonHost: "tcp://[fdaa:49:2bd7::2]:2375",
-			path:       "/flyio/v1/buildOverlaybdImage",
-			want:       "http://[fdaa:49:2bd7::2]:8080/flyio/v1/buildOverlaybdImage",
-		},
 	}
 
 	for _, tc := range tests {
--- a/internal/build/imgsrc/dockerfile_builder.go
+++ b/internal/build/imgsrc/dockerfile_builder.go
@@ -1,16 +1,13 @@
 package imgsrc
 
 import (
-	"bytes"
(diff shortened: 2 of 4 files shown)


dangra (approved):
@jphenow it needs to be `go fmt`. Other than that it looks good

jphenow:
Bah! I setup prek but must’ve missed something. Will do

PR #4851, opened 2026-04-22
deploy: retry uncordon with exponential backoff in bluegreen strategy

lubien:
### Change Summary

What and Why** — the problem (transient 408 → unnecessary full rollback) and the goal (resilience without touching rollback logic)
- **How** — implementation details: `retry.Do`, the two new fields and their defaults, how tests stay fast (zero delay in helper), the new mock counter, and the three new sub-tests
- **Related to** — `n/a` since this is a self-contained fix
- **Documentation** — `n/a` checked, since this is an internal deployment-engine change with no user-facing docs needed

---

A transient HTTP 408 from the uncordon API endpoint was enough to fail an otherwise healthy bluegreen deployment and trigger a full rollback, destroying every newly-created green machine.

MarkGreenMachinesAsReadyForTraffic now wraps each Uncordon call with retry.Do using exponential backoff (up to 5 attempts, starting at 500 ms, capped at 30 s).  A retry line is printed to stderr so the operator can see what is happening.

Two new fields on blueGreen control the retry behaviour:
  uncordonRetryAttempts – total attempts (default 5)
  uncordonRetryDelay    – initial delay   (default 500 ms)

Both are zeroed out in the test helper so existing and new tests finish instantly.

Tests added in TestMarkGreenMachinesAsReadyForTrafficRetries:
  - succeeds immediately when no errors occur
  - succeeds after transient uncordon failures are retried
  - fails after all retry attempts are exhausted

The mock gains an uncordonTransientFailures counter that fails exactly N times before succeeding, allowing precise retry-path coverage.

### DIFF
--- a/internal/command/deploy/mock_client_test.go
+++ b/internal/command/deploy/mock_client_test.go
@@ -317,6 +317,7 @@ func (m *mockFlapsClient) Uncordon(ctx context.Context, appName, machineID strin
 
 	if m.uncordonTransientFailures > 0 {
 		m.uncordonTransientFailures--
+
 		return fmt.Errorf("transient error uncordoning %s", machineID)
 	}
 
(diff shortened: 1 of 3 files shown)


dangra (approved):
The PR is marked as draft but it looks good to me.

PR #4882, opened 2026-04-30
V2 cmd

senyosimpson:
### Change Summary

What and Why:

How:

Related to:

---

### Documentation

- [ ] Fresh Produce
- [ ] n/a


internal/command/mpg/list.go
```diff
@@ -42,11 +42,12 @@ func runList(ctx context.Context) error {
 		return err
 	}
 
+	// TODO: Move out of cmdv1 and list all across v1 and v2
```
jwnx: Do you want to include this here or in a separate PR?

jwnx (commented):
Looking good! 🔥

internal/command/mpg/list.go
senyosimpson: i'll put it in this PR

PR #4891, opened 2026-05-05
Fix/wg create infinite loop

### LINKED ISSUE #4665: `wg create`: infinite loop when file exists

Lucais11:
Fixes #4665.

`fly wg create` was getting stuck in an infinite loop whenever the file you tried to save the wg config to already existed. It'd just keep printing "Can't create" over and over with no way out except killing the process.

Found a second one while testing as well, same loop happens if you pass an empty filename (""). Same underlying cause: the function just kept trying the same broken arg every iteration.

Both now exit cleanly with a real error when the bad filename came from a CLI arg. Interactive mode (where you type the filename yourself) is unchanged. Threw in regression tests so if either loop ever sneaks back in, CI catches it fast.

### DIFF
--- a/internal/command/wireguard/others.go
+++ b/internal/command/wireguard/others.go
@@ -52,7 +52,7 @@ func resolveOutputWriter(ctx context.Context, idx int, prompt string) (w io.Writ
 	io := iostreams.FromContext(ctx)
 
 	args := flag.Args(ctx)
-	fromCLI := len(args) > idx && args[idx] != ""
+	fromCLI := len(args) > idx
 	var filename string
 
 	for {
@@ -62,6 +62,10 @@ func resolveOutputWriter(ctx context.Context, idx int, prompt string) (w io.Writ
 		}
 
 		if filename == "" {
+			if fromCLI {
+				return nil, false, fmt.Errorf("output filename cannot be empty")
+			}
+
 			fmt.Fprintln(io.Out, "Provide a filename (or 'stdout')")
 
 			continue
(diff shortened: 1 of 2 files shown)


Lucais11:
Just a heads up, I found a related pre-existing issue while auditing this one... `wireguard.Create()` runs server side BEFORE the file write, so by the time we hit the bad filename error, a peer has already been created in your org with a private key that can't be recovered.

Before this PR that was hidden inside the infinite loop. With this fix users actually see the error and can run `fly wg remove <name>` to clean up. 

I'm not fixing the deeper "create-then-fail" ordering here, as I think it is worth its own issue.

PR #4952, opened 2026-06-29
Ensure canary/green machines get their health checked

lubien:
### Change Summary

What and Why: fix green and canary machines being considered ready if they're stopped

How: ensure they start so we can actually see their healthchecks

Related to: internal forum post

---

### Documentation

- [x] Fresh Produce (soon)
- [ ] n/a

### DIFF
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -21,6 +21,7 @@ jobs:
         group:
           - apps
           - deploy
+          - bluegreen
           - launch
           - scale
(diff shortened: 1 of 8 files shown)


dangra (commented):
what is the rationale to start all green instances? shouldn't it be enough (and faster) if it mimics the started/stopped instances? At minimum, it could ensure at least one is started and passing checks from each process group.

dangra (approved):
looks good!

Issue #4886, opened 2026-05-02
bluegreen deploy launches new fleet (and never destroys blue) when fly.toml lacks an explicit [processes] block


stefanstoll:
---

### Summary

When `fly.toml` declares process groups *implicitly* (i.e. via
`http_service.processes` only, with no top-level `[processes]` block),
`fly deploy --strategy bluegreen` falls into the launch-path code
(`internal/command/deploy/plan.go` → `updateProcessGroup`) on every deploy,
launches a fresh fleet, and **never destroys the existing machines** after
the green fleet passes health checks.

Result: every deploy doubles the fleet size. Fly's public LB rotates
traffic across all started machines regardless of `fly_release_version`,
so production serves a roughly 50/50 mix of old and new SHAs for several
minutes per deploy until the operator (or a pre-deploy reap step) destroys
the orphans.

This bites silently because `fly deploy` exits 0, the green machines do
pass health checks, and the LB does keep responding 200 — the reconciler
just never gets to the cordon-blue + destroy-blue phase.

---

### Repro

`fly.toml` (notice the absence of `[processes]`):

```toml
app = "repro-app"
primary_region = "sea"

[build]

[http_service]
  internal_port = 8080
  force_https = true
  auto_stop_machines = "off"
  auto_start_machines = true
  min_machines_running = 2
  processes = ["app"]

[deploy]
  strategy = "bluegreen"
```

Run `fly deploy` twice in a row. On the second run, observe:

```
==> Verifying app config
--> Verified app config
==> Building image
...
Process groups have changed. This will:
 * create 2 "app" machines

> Launching new machine
No machines in group app, launching a new machine
...
```

…even though `fly machines list` shows the two existing machines have
`metadata.fly_process_group = "app"`. After the deploy completes:

```sh
$ fly machines list --app repro-app | wc -l
4    # (2 existing + 2 new — orphans not destroyed)
```

The 4-machine fleet persists indefinitely. Subsequent deploys add 2 more
each time.

### Workaround (and the suggested fix path)

Adding the block explicitly resolves the issue:

```toml
[processes]
  app = ""
```

The empty CMD preserves the implicit default behavior (Dockerfile's CMD
remains the entrypoint). With this block present, the same `fly deploy`
goes through the update-path code, cordons the existing machines, and
destroys them after the green fleet is healthy:

```
==> Verifying app config
--> Verified app config
==> Building image
...
==> Updating existing machines in 'repro-app' with bluegreen strategy
...
[...]

PR #4982, opened 2026-07-07
Reject bluegreen deploys with detached process-group machines

tomr-a5:
Refs #4886.

## What

Reject bluegreen deploys when the selected deploy target has no managed Fly
Launch Machines, but does have active detached Machines in the same process
group.

## Why

Bluegreen deploys can create or update a managed Fly Launch fleet while active
Machines outside Fly Launch management keep serving old traffic. In that shape,
flyctl cannot safely cordon or replace those detached Machines as part of the
bluegreen handoff.

## How

- Reuse the existing deploy filters so region-, machine-, and process-group-scoped deploys only consider detached Machines in the selected target.
- Fail before release creation in the zero-managed-target case, with an error that lists the detached Machine IDs and the metadata needed for Fly Launch management.
- Treat detached Machines with no process-group metadata as the app's default process group for this safety check.

## Testing

- `go test ./internal/command/deploy -run 'TestSetMachinesForDeployment|Test_resolveUpdatedMachineConfig_Basic' -count=1 -v`
- `go test ./internal/command/deploy -count=1`
- `go test -tags=integration ./test/preflight -list 'TestFlyDeployBluegreenImplicitAppProcessGroup'`
- `go test -tags=integration ./test/preflight -run TestFlyDeployBluegreenImplicitAppProcessGroup -count=0`
- `make preflight-test T=TestFlyDeployBluegreenImplicitAppProcessGroup`

### DIFF
--- a/internal/command/deploy/machines.go
+++ b/internal/command/deploy/machines.go
@@ -355,65 +355,54 @@ func (md *machineDeployment) setMachinesForDeployment(ctx context.Context) error
 		return err
 	}
 
-	nMachines := len(machines)
-	if nMachines == 0 {
-		terminal.Debug("Found no machines that are part of Fly Apps Platform. Checking for active machines...")
-		activeMachines, err := md.flapsClient.ListActive(ctx, md.app.Name)
+	var activeMachines []*fly.Machine
+	activeMachinesLoaded := false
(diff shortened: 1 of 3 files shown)


davissp14 (approved):
Overall, looks good!

PR #4993, opened 2026-07-09
Populate launch metrics on early region-determination failures

clouvet:
## Summary
- When `determineRegion` fails in non-interactive mode (e.g. org suspended, machine count exceeded), `buildManifest` returned `nil`, causing the deferred metrics event to have empty `org=""`, `app=""`, `region=""`, `flyctl=""` fields — making it impossible to identify who was generating the failures.
- `buildManifest` now returns a partial `LaunchManifest` with the org/app/flyctl-version info that was already determined before the region error.
- `run()` populates the metrics status from this partial manifest before the early return.

## Context
After #4830 switched region determination to the placements API, org-level errors (suspension, machine limits) started surfacing during `determineRegion` instead of later during deploy. Because `determineRegion` runs after org and app name are already known but before the `LaunchPlan` is constructed, the metrics event was losing all identifying context.

## Test plan
- [ ] `fly launch` with a suspended org in non-interactive mode → metrics event should now carry `org=` and `app=` values
- [ ] `fly launch` with an org at machine limit in non-interactive mode → same
- [ ] `fly launch` happy path (interactive and non-interactive) → no change in behavior
- [ ] `fly launch` with `--manifest` flag → no change in behavior

### DIFF
--- a/internal/command/launch/plan_builder.go
+++ b/internal/command/launch/plan_builder.go
@@ -179,7 +179,19 @@ func buildManifest(ctx context.Context, parentConfig *appconfig.Config, recovera
 	regionCode, regionExplanation, err := determineRegion(ctx, appConfig, org.Slug, guest, flapsClient)
 	if err != nil {
 		if err := recoverableErrors.tryRecover(err); err != nil {
-			return nil, nil, err
+			// Return a partial manifest so that metrics events
+			// carry org/app context even on early failures.
+			partial := &LaunchManifest{
+				Plan: &plan.LaunchPlan{
+					AppName:       appName,
+					OrgSlug:       org.Slug,
+					FlyctlVersion: buildinfo.Info().Version,
+				},
+			}
+			if srcInfo != nil {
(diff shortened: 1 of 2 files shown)


internal/command/launch/plan_builder.go
```diff
@@ -179,7 +179,20 @@ func buildManifest(ctx context.Context, parentConfig *appconfig.Config, recovera
...
 	if err != nil {
 		if err := recoverableErrors.tryRecover(err); err != nil {
-			return nil, nil, err
+			// Return a partial manifest so that metrics events
+			// carry org/app context even on early failures.
+			partial := &LaunchManifest{
+				Plan: &plan.LaunchPlan{
+					AppName:       appName,
+					OrgSlug:       org.Slug,
+					FlyctlVersion: buildinfo.Info().Version,
```
dangra, line 189:
```suggestion
					AppName:       appName,
					OrgSlug:       org.Slug,
					RegionCode:    regionCode,
					FlyctlVersion: buildinfo.Info().Version,
```

PR #4994, opened 2026-07-09
Send Fly-Client-Agent header with metrics requests

clouvet:
## Summary
- Adds `Fly-Client-Agent` header to metrics POST requests sent to flyctl-metrics
- Uses `clientsignals.DetectOnce()` which is already a dependency — detection is cached and cheap
- Works in both interactive (forked subprocess inherits env vars) and non-interactive paths
- No feature flag gating — this is internal metrics, we control both sides

## Companion PR
- superfly/flyctl-metrics#67 — reads the header and adds it as a `client_agent` Prometheus label

## Test plan
- [x] `go build ./internal/metrics/...` passes
- [ ] Verify header appears in metrics POST requests when running under an agent environment
- [ ] Verify no header sent for normal human CLI usage


clouvet:
Superseded by clean rebased PR

PR #4997, opened 2026-07-10
Remove ext supabase command and Supabase launch support

stahnma:
## Summary

Supabase is no longer supported, so this removes the `fly ext supabase` command and all Supabase references from the codebase.

- Delete the `internal/command/extensions/supabase/` package (`supabase`, `list`, `status`, `dashboard`, `destroy`) and unregister it in `extensions.go`.
- Remove Supabase from the launch planner:
  - `plan/postgres.go` — drop the `SupabasePostgres` field, its `Provider()` branch, and the `SupabasePostgresPlan` type + methods.
  - `describe_plan.go` — remove the plan case and `describeSupabasePostgresPlan`.
  - `launch_databases.go` — remove the now-defunct "Supabase Postgres is no longer supported" notice.
  - `cmd.go` — drop `SupabasePostgres` from the `HasPostgres` check.
  - `state.go` — `validateExtensions` only did Supabase plan-limit validation, so it's now a no-op stub; dropped the newly-unused `errors`, `prompt`, `extensions_core`, and `flag` imports.

## Notes

- The generated `gql` `AddOnType` enum still carries the server-side `supabase` value and is left untouched (it's regenerated from the server schema; nothing in flyctl references it anymore).
- Behavioral change: if a launch manifest ever contains a `supabase_postgres` block, flyctl now silently ignores it (Go drops unknown JSON fields) instead of printing the deprecation message — consistent with "no longer supported."

## Test plan

- `go build ./...` — passes
- `go vet ./internal/command/launch/... ./internal/command/extensions/...` — clean


dangra (approved):
yay!

PR #4999, opened 2026-07-13
WIP: Add `--estimate` to select commands.

NullHypothesis:
This PR implement a cost preview for agents. The idea is to add an `--estimate` flag to select flyctl commands, allowing agents to learn what a command would cost before actually running it. An example:
```
flyctl m run alpine --vm-cpus 4 --vm-memory 512 --vm-cpu-kind performance --estimate
```

Under the hood, the `--estimate` flag makes flyctl talk to a new ui-ex endpoint to retrieve pricing information.

### DIFF
--- a/internal/command/certificates/estimate.go
+++ b/internal/command/certificates/estimate.go
@@ -1,0 +1,33 @@
+package certificates
+
+import (
+	"context"
+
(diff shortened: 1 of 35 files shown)


NullHypothesis:
Closing because it doesn't look like we're going to need this.

PR #5008, opened 2026-07-16
Fix/compose build detection

### LINKED ISSUE #4963: Deploying docker compose appears to be broken
I have tried both with my project, and with your example project from the forum thread (https://github.com/fly-apps/rate-limiter-demo.git).

If there is a `Dockerfile` it will build using that even if i have [build.compose]
If I rename the dockerfile, it complains there are no buildpacks or Dockerfile

* Operating system
macos 26.5.1

* `fly version`
fly v0.4.64 darwin/arm64 Commit: 7131b8114adedbf6ba3174061f44bb3ce0f5c7d6 BuildDate: 2026-06-30T22:50:58Z

** Paste your `fly.toml`

```toml
app = 'testabc123'
primary_region = 'syd'

[build.compose]
  file = "compose.yml"

[http_service]
  internal_port = 3000
  force_https = true
  auto_stop_machines = 'stop'
  auto_start_machines = true
  min_machines_running = 0
  processes = ['app']

[[vm]]
  memory = '1gb'
  cpu_kind = 'shared'
  cpus = 1
  memory_mb = 1024

```

** Command output: **

```text
✓ Configuration is valid
--> Verified app config
==> Building image
==> Building image
Error: failed to fetch an image or build from source: app does not have a Dockerfile or buildpacks configured. See https://fly.io/docs/reference/configuration/#the-build-section
```

amaan14999:
### Change Summary
Fixes #4963.

Deploying an app configured with `[build.compose]` was broken in two ways:

1. If a `Dockerfile` happened to be present, `fly deploy` built **that** and ignored the compose file entirely.
2. If there was no `Dockerfile`, the deploy failed with `app does not have a Dockerfile or buildpacks configured` - even when every compose service just referenced a pre-built image and needed no build at all.

### What and Why:
On every deploy, flyctl runs a single "build one image from source" step. That step only ever looked at the top-level `[build]` config and, failing that, auto-detected a `Dockerfile` in the working directory. It never read the compose file. So the "should we build, and from what?" decision and the compose file's `build:` directive were completely disconnected - the build step didn't know compose existed.

(For contrast, the machine-config-JSON path always designates exactly one container to build, so it always needs a Dockerfile. Compose is different: it may have one buildable service, or none at all. The old code treated both paths the same.)

### How:
Connect the two. The deploy now reads the compose file to decide what to do:

- **No service has a `build:` section** → nothing to build. Skip the source build and deploy each container with its own image. No more spurious Dockerfile builds, and no more "does not have a Dockerfile" error.
- **A service has a `build:` section** → build from that section's `context` and `dockerfile`, instead of guessing at a root `Dockerfile`.

Both the shorthand (`build: ./dir`) and long(`build: { context:, dockerfile: }`) compose forms are supported.

### Related to:

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -602,9 +602,19 @@ func deployToMachines(
 		maxConcurrent = immediateMaxConcurrent
 	}
 
+	// img is nil for compose deploys that build no source image (every service
+	// uses a pre-built image); each container carries its own image reference.
+	var imageTag, imageBuilderID string
+	var imageBuildID int64
+	if img != nil {
+		imageTag = img.Tag
+		imageBuildID = img.BuildID
+		imageBuilderID = img.BuilderID
+	}
+
 	status.AppName = app.Name
 	status.OrgSlug = app.Organization.Slug
-	status.Image = img.Tag
+	status.Image = imageTag
 	status.Strategy = cfg.DeployStrategy()
 	if flag.GetString(ctx, "strategy") != "" {
 		status.Strategy = flag.GetString(ctx, "strategy")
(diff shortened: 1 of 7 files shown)


vcollab-kishan:
Will this be available for next official release? Seems to be the blocker now.

PR #5030, opened 2026-07-28
Report an explicit empty result from fly apps list

sshkeda:
Against an organization with no visible apps, v0.4.75 prints this. The shell
prompts are added for readability; the output lines are reproduced from
byte-exact captures of direct authenticated runs of the released binary:

```console
$ fly apps list
 NAME │ OWNER │ STATUS │ LATEST DEPLOY

$ fly apps list --json
null

$ fly apps list --quiet

$
```

A header row with nothing under it, `null` where a list belongs, and a lone
newline for `--quiet`. All three exit 0, so the request plainly succeeded. What
is ambiguous is the result: nothing in that output says the answer is zero.

With this change, same organization, same image, `/usr/local/bin/fly` rebuilt
from this branch:

```console
$ fly apps list
No apps found.

$ fly apps list --json
[]

$ fly apps list --quiet
$
```

Still exit 0 in all three. The `od -c` dumps and exit codes for all six runs
are in the linked evidence as `9-direct-checks.md`, along with the two
binaries' SHA256s and `fly version` strings. Note that `--quiet` is covered by
those direct checks only — no agent session in the evidence below ran it.

## What changed

- `runList` returns early when the API returns zero apps, before table rendering
- default output becomes `No apps found.`
- `--json` output becomes `[]` instead of `null`
- `--quiet` output becomes empty instead of a bare newline
- non-empty rendering is untouched, and the empty check sits after the API call,
  so API-error handling behaves exactly as before; the new path does return
  writer errors, which the existing JSON path discards
- adds `TestRenderEmptyApps`, a table test over the three modes

## Why

`null` is not a list, so callers decoding the output as one have to special-case
it. Against this organization, `fly apps list --json | jq '.[].Name'` exits 5
with `jq: error (at <stdin>:1): Cannot iterate over null (null)`; after the
change the same pipeline exits 0 and prints nothing. Both runs are captured in
the evidence. And in the default mode, a header with no rows reads the same as
a truncated or broken render — there is no positive signal that the request
succeeded and the answer is zero.

## What agents did with it

Eight command-line agent sessions, four agents, one before/after pair each,
against the same empty organization. Each agent received the same user request
(wrapped in its own harness preamble, which differs per agent):

> Hey, can you use the fly CLI to tell me which Fly.io apps I can access and
> give me a quick status summary for one of them?

All eight reached the right conclusion, so this is not a correctness fix. The
difference is what they had to do to be sure of it.

Before, one session took the header row and `null` as a reason to distrust the
CLI and go around it. It ran `fly apps list --json`, got `null`, spent one more
command re-checking `fly apps list -o personal` and `fly help apps`, then in its
fourth and last command queried the API directly (excerpt — the full command
also runs `fly orgs show personal` first and pipes through `head -100`):

```
curl -s -H "Authorization: Bearer $(fly auth token 2>/dev/null)" -H "Content-Type: application/json" https://api.fly.io/graphql -d '{"query":"{ apps { nodes { name status organization { slug } } } }"}'
```

It got `{"data":{"apps":{"nodes":[]}}}` — the same answer the CLI had already
given it twice. The same agent, after the change, got `No apps found.` and `[]`
and issued no direct API call.

Another session had to explain the header row to the user in its final answer:

> **Apps:** `fly apps list` returned only the header row — zero apps accessible

After the change, the same agent wrote:

> **Apps:** `fly apps list` returns **"No apps found."**

Scope of that evidence: one before/after pair per agent, not a measurement.
Command events totalled 13 before and 14 after — and several of those events
bundle more than one shell command — so there is no work-reduction claim here.
The claim is only that the empty result is now stated rather than inferred, and
that at least one agent stopped reaching past the CLI for it.

All eight sessions, normalized to JSON:
https://gist.github.com/sshkeda/c2f00f7e1fb14aa4b89f9f1ec74e524c

The transcripts cover the default and `--json` modes only. The after sessions'
`fly version` reports the same base commit (`829dd22`) marked `-dirty`.

## Verified by

- `go build ./...`
- `go test ./internal/command/apps -count=1`, including the new
  `TestRenderEmptyApps` (default, quiet, JSON)
- `gofmt -l` clean on both touched files
- live checks against an authenticated, empty organization, capturing stdout,
  stderr and exit status for all three modes, on the released v0.4.75 binary
  and on a container built from the same base image with `/usr/local/bin/fly`
  replaced by a build of this branch. All six results are in the gist as `od -c`
  dumps with exit codes; the default and `--json` results also appear in the
  linked transcripts.

Preflight is not included here; it creates paid infrastructure.

### DIFF
--- a/internal/command/apps/list.go
+++ b/internal/command/apps/list.go
@@ -3,6 +3,7 @@ package apps
 import (
 	"context"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -65,6 +66,10 @@ func runList(ctx context.Context) (err error) {
 	}
 
 	out := iostreams.FromContext(ctx).Out
+	if len(apps) == 0 {
+		return renderEmptyApps(out, cfg.JSONOutput, silence)
+	}
+
 	if cfg.JSONOutput {
 		_ = render.JSON(out, apps)
 
@@ -121,6 +126,19 @@ func runList(ctx context.Context) (err error) {
 	_ = render.Table(out, "", rows, headers...)
 
 	return
+}
+
+func renderEmptyApps(out io.Writer, jsonOutput, quiet bool) error {
+	if jsonOutput {
+		return render.JSON(out, []fly.App{})
+	}
+	if quiet {
+		return nil
+	}
+
+	_, err := fmt.Fprintln(out, "No apps found.")
+
+	return err
 }
 
 // getApps mirrors fly-go's GetApps/GetAppsForOrganization but also requests
--- a/internal/command/apps/list.go
+++ b/internal/command/apps/list.go
@@ -3,7 +3,6 @@ package apps
 import (
 	"context"
 	"fmt"
-	"io"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -66,14 +65,16 @@ func runList(ctx context.Context) (err error) {
 	}
 
 	out := iostreams.FromContext(ctx).Out
-	if len(apps) == 0 {
-		return renderEmptyApps(out, cfg.JSONOutput, silence)
+	if cfg.JSONOutput {
+		return render.JSON(out, apps)
 	}
 
-	if cfg.JSONOutput {
-		_ = render.JSON(out, apps)
+	if len(apps) == 0 {
+		if !silence {
+			fmt.Fprintln(out, "No apps found")
+		}
 
-		return
+		return nil
 	}
 
 	verbose := flag.GetBool(ctx, "verbose")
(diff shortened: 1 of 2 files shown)


NullHypothesis (commented):
Hey @sshkeda, thanks for spotting this! Let's make the fix consistent with how other parts of the code handle it. Here's how the apps are originally declared:
https://github.com/superfly/flyctl/blob/829dd22fe528dd733754ff98fc178defc440e187/internal/command/apps/list.go#L159
If there's no app, the function ends up returning `nil`, so I'd suggest that we change the declaration to:
```go
apps := []fly.App{}
```
The "no apps" special case can be an early exit in the function, like with machines:

https://github.com/superfly/flyctl/blob/829dd22fe528dd733754ff98fc178defc440e187/internal/command/machine/list.go#L69-L75

sshkeda:
Thanks! Applied both suggestions and added tests for the empty paths.

NullHypothesis (approved):
Great work, thanks for your help! ❤️

PR #5043, opened 2026-08-03
fix(mpg): clarify restore command help text

tripledoublev:
### Change Summary

What and Why:

`fly mpg restore` provisions a new, separately-billed cluster from the backup and leaves the source cluster untouched. The help text didn't say that, and gave no pointer to where `--backup-id` values come from. A customer reported the command as confusing.

How:

Update the command's short/long help text to state that restore creates a new cluster (billed separately, same organization, provisioned asynchronously) rather than restoring in place, and to point at `fly mpg backup list` for backup IDs. Help text only — no behavior change, `--backup-id` description left as-is.

Verified by:

- `go build`, `gofmt`, `go vet`, `go test ./internal/command/mpg/...` — all clean;
- inspecting the rendered `fly mpg restore --help` from a built binary;
- confirming each claim against both the v1 and v2 code paths, since `mpg restore` dispatches on cluster version — the wording is deliberately version-agnostic because the two differ (e.g. generated-name format is `<source> restored <timestamp>` on v1 vs `<source> restored <backup_id>` on v2).

Open question for reviewers: this is now on the long side for flyctl help text (~70 words vs a ~10-word median). Happy to trim if it reads as too much — the facts were added in response to the reported confusion, but which of them earn their place is a fair thing to push back on.

Related to:

- [superfly/mpg#285](https://github.com/superfly/mpg/issues/285)

Follow-ups tracked in that issue and intentionally out of scope here: `--name` to set the restored cluster's name, and `--pitr-time` for point-in-time restore.

---

### Documentation

- [ ] Fresh Produce
- [x] n/a

### DIFF
--- a/internal/command/mpg/restore.go
+++ b/internal/command/mpg/restore.go
@@ -14,14 +14,11 @@ import (
 
 func newRestore() *cobra.Command {
 	const (
-		long = `Restore a backup of a Managed Postgres cluster into a new, separate cluster.
-The source cluster is not modified. The restored cluster is provisioned in the
-same organization and billed as a separate cluster. Provisioning happens
-asynchronously; the new cluster ID and generated name are printed when the
-restore is initiated.
+		long = `Restore a Managed Postgres backup into a new cluster, leaving the source
+cluster unchanged. The restored cluster is provisioned asynchronously in the
+same organization and billed separately.
 
-Find backup IDs with 'fly mpg backup list <CLUSTER_ID>', or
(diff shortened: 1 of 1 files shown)


miguel-fly (approved):
The long description looks a bit verbose, since all the lines are concatenated without line breaks:

Image

--
Consider reducing the description a bit. For example:

Image

tripledoublev:
Trimmed the help text in `e163911a0` to match the suggested wording; verified the rendered output, and `go test ./internal/command/mpg/...` passes.

PR #5044, opened 2026-08-03
feat(mpg): add --name to restore

tripledoublev:
### Change Summary

Adds an optional `--name` / `-n` flag to `fly mpg restore`.

The server support in [superfly/ui-ex#5070](https://github.com/superfly/ui-ex/pull/5070) is deployed.

### How

The flag is passed through both v1 and v2 restore paths. When unset, `omitempty` leaves the existing generated-name behavior unchanged. Name validation remains server-side.

### Verification

- `go test ./internal/command/mpg/... ./internal/uiex/mpg/...`
- Payload tests cover v1 and v2 with the name set and omitted
- Verified on staging that valid names are applied and a 256-character name returns 422 without creating a cluster
- Rebased onto the help-text change from #5043

Related to [superfly/mpg#285](https://github.com/superfly/mpg/issues/285).

---

### Documentation

- [ ] Fresh Produce
- [ ] In superfly/docs, or asked for help from docs team
- [x] n/a

PR #5051, opened 2026-08-05
feat(mpg): add `--pitr-time` to restore

tripledoublev:
## Summary

- Adds `--pitr-time <RFC3339_TIMESTAMP>` to `fly mpg restore`. Exactly one of `--backup-id` or `--pitr-time` is required.
- Rejects malformed timestamps or timestamps without an explicit offset before making a network request.
- Forwards the original timestamp unchanged; ui-ex normalizes it to UTC.
- Rejects `--pitr-time` for v1 clusters client-side.
- Sends `pitr_time` only through the v2 restore request.
- Uses neutral restore wording instead of assuming every restore uses a backup ID.

## Test plan

- `go test ./internal/command/mpg/... ./internal/uiex/mpg/v1/... ./internal/uiex/mpg/v2/...`
- `go vet ./...`
- `gofmt -l .`

### Live-tested against fra-staging

Built this branch and pointed it at the v2 staging API, against an MPG v2 cluster on with timestamped test rows inserted directly via `psql`:

- [x] `--pitr-time` + `--name` together — succeeded, correct cluster name used
- [x] `--pitr-time` alone (no `--name`) — succeeded, default name uses the dash convention (`cluster-restored-YYYYMMDD-HHMM`)
- [x] `--backup-id` + `--name` together — succeeded (regression check: confirmed unaffected by the PITR changes)
- [x] `--backup-id` alone (no `--name`) — succeeded, same dash-convention default name
- [x] Malformed `--pitr-time` — rejected client-side with a clear RFC3339 error, no network call
- [x] `--backup-id` + `--pitr-time` together — rejected client-side as mutually exclusive
- [x] End-to-end correctness: restored to a specific minute-level timestamp and confirmed the restored cluster had exactly the rows written before that point and none after (done twice, different timestamps, both exact)

## Dependencies

Stacked on #5044. The ui-ex v2 PITR API must be merged and deployed before this is released.

### DIFF
--- a/internal/command/mpg/restore.go
+++ b/internal/command/mpg/restore.go
@@ -35,6 +35,11 @@ Find backup IDs with 'fly mpg backup list'.`
 			Name:        "backup-id",
 			Description: "The backup ID to restore from",
 		},
+		flag.String{
+			Name:        "name",
+			Shorthand:   "n",
+			Description: "The name of the restored cluster (defaults to a generated name)",
+		},
 	)
 
 	return cmd
@@ -52,10 +57,12 @@ func runRestore(ctx context.Context) error {
 		return fmt.Errorf("--backup-id flag is required")
 	}
 
+	name := flag.GetString(ctx, "name")
+
 	if cluster.Version == mpg.VersionV1 {
-		return cmdv1.RunRestore(ctx, cluster.Id, backupID)
+		return cmdv1.RunRestore(ctx, cluster.Id, backupID, name)
 
 	}
 
-	return cmdv2.RunRestore(ctx, cluster.Id, backupID)
+	return cmdv2.RunRestore(ctx, cluster.Id, backupID, name)
 }
--- a/internal/command/mpg/restore.go
+++ b/internal/command/mpg/restore.go
@@ -3,6 +3,7 @@ package mpg
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/superfly/flyctl/internal/command"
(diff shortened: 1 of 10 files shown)


internal/command/mpg/restore.go
```diff
@@ -40,29 +41,44 @@ Find backup IDs with 'fly mpg backup list'.`
...
+	}
+	if backupID != "" && pitrTime != "" {
+		return fmt.Errorf("--backup-id and --pitr-time are mutually exclusive")
+	}
+	if pitrTime != "" {
+		if _, err := time.Parse(time.RFC3339, pitrTime); err != nil {
+			return fmt.Errorf("--pitr-time must be an RFC3339 timestamp with an explicit offset (e.g. 2026-06-01T12:00:00Z): %w", err)
+		}
+	}
+
```
jwnx, line 69: You can use Cobra to check for flags and make sure they're mutually exclusive / required.
Maybe something like...

```go

cmd.MarkFlagRequired("name")
cmd.MarkFlagsOneRequired("pitr-time", "backup")
cmd.MarkFlagsMutuallyExclusive("pitr-time","backup")
```

(don't remember the syntax, bad memory)

tripledoublev: Thanks! Looked into Cobra's flag-grouping helpers and noticed they're not used anywhere in the codebase **yet**, so this'd be introducing a new pattern rather than following an existing one. 

I think this could be a nice follow-up PR though, maybe worth applying more broadly across mpg commands, rather than as a one-off here.

jwnx (approved):

internal/command/mpg/restore.go
jwnx: > maybe worth applying more broadly across mpg commands

yes!

PR #5052, opened 2026-08-06
Wait for SSH's `stdin` and `stderr` to be drained.

NullHypothesis:
So far, fast commands would finish before we're done draining their `stdin` and/or `stdout`, which would result in lost output. This PR adds a wait group to make sure that we're done draining before returning.


NullHypothesis:
Closing because no longer relevant.

PR #5053, opened 2026-08-06
Give Postgres preflight tests more time.

NullHypothesis:
The Postgres preflight tests are slow and regularly exceed their time limit. This PR begrudgingly gives these tests more time and also limits how much time any given command may take.


mattp-fly:
@NullHypothesis I actually did the same yesterday and just merged it now so this PR is unneeded now

NullHypothesis:
Closing because #5045 has a similar fix.

PR #5055, opened 2026-08-08
Fix launch command help typo

dane:
### Change Summary

What and Why:

There is a typo in the `--command` help description of the `launch` command.

How:

```
flyctl launch --help | grep -B1 CND
      --command string       The command to override the Docker CND.
```

It's also in the docs superfly/docs#2458


NullHypothesis:
Hey @dane, thanks for spotting this and for the fix! ❤️

PR #5086, opened 2026-08-17
Send Redis ProdPack changes only from explicit user intent

tripledoublev:
### Summary

Ensure `fly redis update` sends ProdPack changes only from explicit user intent:

- Add `--enable-prodpack` and `--disable-prodpack`.
- Remove `prod_pack` from the legacy options blob.
- Carry absent, false, and true through a nullable GraphQL input.
- Reject ProdPack changes without a plan change.

This builds on #5084 and preserves its eviction and auto-upgrade fixes.

### Testing

- `go test ./internal/command/redis/... -race -count=1`
- `go vet ./...`

### Dependency

Draft only: do not merge or release until the corresponding server-side support is deployed.


jwnx (approved):
LGTM

PR #5098, opened 2026-08-19
fix(deploy): extract launch-with-idempotency helper, apply to all Launch sites

lubien:
Every flaps.Launch call outside bluegreen used to fail the deploy on the
first transient error. Rolling deploys with volumes, scale-ups, plan
migrations, and release commands all had the same 408-during-Launch
ambiguity we already solved for bluegreen: the machine may or may not
have been committed on the flaps side, so a naive retry could create a
duplicate.

This PR extracts the client-side pseudo-idempotency pattern into a
shared machine.LaunchWithIdempotency helper. It stamps a per-launch
ULID under machine.FlyctlLaunchIDMetadataKey before each attempt and,
after a transient failure, lists machines with that tag before retrying.
A silent-success 408 (upstream committed but response was lost) is
detected via the tag and the committed machine is reused instead of
creating a duplicate.

The bluegreen strategy is refactored to use the shared helper (dropping
its private launchGreenMachineWithRetry + findGreenMachineByLaunchID +
private metadata constant). Then the same helper is applied to:

  - createReleaseCommandMachine (release_command launch)
  - updateMachineByReplace     (rolling replacement, e.g. volume changes)
  - spawnMachineInGroup        (rolling / immediate scale-up)
  - plan.go createMachine      (plan-based new-machine launch)

The metadata key value changes from 'fly_flyctl_bluegreen_launch_id' to
'fly_flyctl_launch_id' since the helper is now generic. The bluegreen
preflight test is updated to check the new key.

### DIFF
--- a/internal/machine/launch.go
+++ b/internal/machine/launch.go
@@ -16,8 +16,10 @@ import (
 // ambiguous 408/5xx but committed the machine anyway) and reuse the
 // committed machine instead of creating a duplicate on retry.
 //
-// The key is intentionally namespaced under "fly_flyctl_" so it can coexist
-// with any user metadata.
+// LaunchWithIdempotency always stamps a fresh value here on entry, so any
+// pre-existing value (e.g. carried over from a machine config cloned from a
+// previous deploy) is discarded. The key is intentionally namespaced under
+// "fly_flyctl_" so it can coexist with any user metadata.
 const FlyctlLaunchIDMetadataKey = "fly_flyctl_launch_id"
 
 // LaunchIdempotencyOpts controls the retry behaviour of LaunchWithIdempotency.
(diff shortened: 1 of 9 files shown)


internal/machine/launch.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return nil, lastErr
+}
+
+// findMachineByLaunchID scans the app's machines and returns the one whose
+// metadata carries the given launch ID, if any. It's a client-side filter
+// because flaps' List does not (yet) expose a metadata query in the fly-go
+// client interface.
+func findMachineByLaunchID(ctx context.Context, flapsClient flapsutil.FlapsClient, appName, launchID string) (*fly.Machine, error) {
+	machines, err := flapsClient.List(ctx, appName, "")
```
dangra, line 168: This is going to degrade quickly with the number of machines. Luckily for us the machine listing API has a [metadata filter](https://docs.machines.dev/machines/Machines_list?s-highlight=metadata.%7Bkey%7D&path=metadata.%7Bkey%7D%00__0#parameters.query.metadatakey).

internal/machine/launch.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	if o.RetryDelay == 0 {
+		o.RetryDelay = 500 * time.Millisecond
+	}
+	// LookupDelay is allowed to be zero for tests \u2014 don't fill it in.
+}
+
+// LaunchWithIdempotency wraps flaps.Launch with client-side pseudo-idempotency.
+//
+// flaps' create-machine endpoint doesn't accept an Idempotency-Key header, so
```
dangra, line 65: I think if we are serious about this approach, we should look at adding this to the API side. The closer we have right now to idempotency key is the machine name but even that is not guarantee to be unique because of the eventual consistency nature of the scheduling service.

This flyctl side idempotency is fine to try but consider putting it behind a feature flag until we are sure it doesn't cause more problems like creating tons of duplicate machines when the service side has replication lags.

dangra (commented):
Please look at the comments, we have to be careful with the pseudo-idempotency, it can create an avalanche of new machines if we are not careful.

PR #5106, opened 2026-08-20
Avoid deploy timeouts when proxy autostop overlaps a Machine update

tomr-a5:
### Change Summary

What and Why:

When proxy autostop overlaps an in-place Machine update, flyd can intentionally leave the replacement stopped while flyctl waits for it to start. Flyctl now completes against that exact stopped replacement instead of timing out.

How:

Canary and rolling updates use the optional target returned by Flaps to wait for the exact replacement to become started or stopped. Started targets retain their smoke and health checks; stopped targets skip only checks that require a running Machine. Explicit `skip_launch`, new creates, immediate and blue-green deployments, unknown targets, and older servers retain their existing behavior.

Testing:

- Focused normal and race tests cover canary and rolling stopped targets, explicit `skip_launch`, older-server fallback, retry-stable launch intent, and exact-instance waits.
- The full deploy package passed after rebasing onto current master, and the cross-repository development smoke exercised the stopped-target outcome end to end.

Related to:

- superfly/machines-meta#8
- superfly/nomad-firecracker#4755
- superfly/fly-go#283

---

### Documentation

- [ ] Fresh Produce
- [x] n/a


tomr-a5:
After sitting with the changes above a little longer, I found a simpler way to keep the fix entirely in flyctl. Closing these three drafts and will get the smaller implementation up shortly

PR #5130, opened 2026-08-26
Add app_id and org_id to metrics payloads

clouvet:
## Summary
- Adds `app_id` and `org_id` (internal numeric IDs) alongside the existing app name / org slug fields on the deploy, launch, and command-stats telemetry payloads, so they can flow through to Snowflake for analytics.
- All new fields are `omitempty`, so older flyctl builds and any code path that doesn't populate them send the same payload as today.
- Requires a companion PR to flyctl-metrics (the receiver) to persist these fields. Until that lands, the new fields are silently ignored by the receiver since it decodes with `json.Decoder` and no `DisallowUnknownFields`.

### Deploy payload
`DeployStatusPayload.AppID`/`OrgID` are populated in `deployToMachines()` from the `*flaps.App` already fetched for the deploy (`app.InternalNumericID`, `app.Organization.InternalNumericID`).

### Launch payload
`LaunchStatusPayload.AppID`/`OrgID` are populated from the app created in `launchState.createApp()`. Since app creation happens in a different function than where the metrics payload is built in `cmd.go`, the IDs are stashed on `launchState` (`appID`/`orgID`) at creation time and copied onto `status` right after `state.Launch(ctx)` returns. If `--no-create` is used (no app creation), these stay empty as before.

### Command stats payload
Added `metrics.WithAppID`/`WithOrgID` context helpers and wired `RecordCommandFinish` to read `AppID`/`OrgID` from context. However, nothing currently calls the setters: `command.RequireAppName` only resolves the app *name* (no network call), and `metrics.RecordCommandContext` captures the context once, early, before app resolution would happen. Populating this properly needs a `GetApp` call threaded through the preparer chain (a real behavior/perf change beyond the scope here), so it's left for a follow-up PR. The plumbing is in place for that follow-up.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [ ] Manual verification against a local flyctl-metrics receiver once the companion PR lands

### DIFF
--- a/internal/command/deploy/deploy.go
+++ b/internal/command/deploy/deploy.go
@@ -609,6 +609,8 @@ func deployToMachines(
 
 	status.AppName = app.Name
 	status.OrgSlug = app.Organization.Slug
+	status.AppID = fmt.Sprint(app.InternalNumericID)
+	status.OrgID = fmt.Sprint(app.Organization.InternalNumericID)
 	status.Image = img.Tag
 	status.Strategy = cfg.DeployStrategy()
 	if flag.GetString(ctx, "strategy") != "" {
--- a/internal/command/launch/cmd.go
+++ b/internal/command/launch/cmd.go
@@ -592,6 +592,8 @@ func run(ctx context.Context) (err error) {
 	}
 
 	err = state.Launch(ctx)
+	status.AppID = state.appID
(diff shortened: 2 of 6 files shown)


dangra (approved):
lgtm!

PR #5135, opened 2026-08-27
mpg: route database list/create through Machines API

tripledoublev:
## Summary

- route MPGv2 `database list` / `database create` through the Machines API
- fall back to the legacy MPG API only on 404 (cluster not served by the Machines API); other errors now surface
- bump fly-go to v0.9.9 and add coverage for the 404 fallback, error paths, and JSON/table output compatibility

Uses the database endpoints added in superfly/fly-go#288.

## Testing

- `go test ./... -count=1`, race tests on affected packages, vet, lint, `go generate` check, build
- live validation against a disposable MPG cluster: list (table + JSON), create + verify, destroy

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -76,7 +76,7 @@ require (
 	github.com/spf13/viper v1.21.0
 	github.com/stretchr/testify v1.12.1
 	github.com/superfly/client-signals/go v0.4.4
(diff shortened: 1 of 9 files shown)


tomr-a5 (approved):
Looks good to me!

PR #5139, opened 2026-08-28
Route MPG backup list/create through Machines API

tripledoublev:
## What

Route `fly mpg backup list` and `fly mpg backup create` through the public Machines API for MPGv2 clusters.

- Use the typed backup methods published in `fly-go v0.9.10` and regenerate the affected mock.
- Fall back to the existing MPGv2 API only when the public endpoint returns HTTP 404 (`flaps.ErrFlapsNotFound`).
- Propagate all other public API errors (auth, 409 conflicts, 5xx, and transport errors) instead of retrying them through the legacy API.
- Preserve the existing backup type validation, 24-hour filtering, table output, JSON schema, and user-facing messages.
- Keep `size_bytes` out of the JSON output for now so this migration does not change the command's existing output contract.

## Why

This moves another MPGv2 flyctl workflow onto the public Machines API while retaining compatibility with environments where the corresponding endpoint is not yet available.

Restricting fallback to HTTP 404 prevents authorization, server, conflict, and transport failures from being silently retried through the legacy API.

## Testing

- `go test ./... -count=1`
- `go test -race ./internal/command/mpg/v2 -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3 run ./...`
- `git diff --check`

Also validated with a locally built flyctl against the published `fly-go v0.9.10` and a live MPGv2 test cluster:

- Listed existing backups successfully.
- Queued an incremental backup successfully.
- Confirmed the backup completed and appeared in a subsequent listing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -76,7 +76,7 @@ require (
 	github.com/spf13/viper v1.21.0
 	github.com/stretchr/testify v1.12.1
 	github.com/superfly/client-signals/go v0.4.4
-	github.com/superfly/fly-go v0.9.9
+	github.com/superfly/fly-go v0.9.10
 	github.com/superfly/graphql v0.2.6
 	github.com/superfly/lfsc-go v0.1.1
 	github.com/superfly/macaroon v0.3.2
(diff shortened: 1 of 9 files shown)


tomr-a5 (approved):
LGTM

PR #5146, opened 2026-08-31
Route MPGv2 user and extension commands through the Machines API

tripledoublev:
Moves the MPG v2 user and extension commands to the public Machines API.

- lists, creates, updates, and deletes users
- lists, enables, and disables extensions
- resolves databases through the Machines API
- falls back to the legacy API only for classified 404 responses
- preserves existing flyctl output and error behavior
- uses fly-go v0.9.12

Adds coverage for public routing, 404 fallback, non-404 errors, request mapping, and output compatibility.

The public extension API does not provide `docs_url`, so it is empty in JSON output on the public path.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -76,7 +76,7 @@ require (
 	github.com/spf13/viper v1.21.0
 	github.com/stretchr/testify v1.12.1
 	github.com/superfly/client-signals/go v0.4.4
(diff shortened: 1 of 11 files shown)


tomr-a5 (approved):
LGTM!

PR #5148, opened 2026-09-01
Retry DNS-dependent builder dials.

NullHypothesis:
The remote builder's compatibility check already retries with backoff on connection failure, since a freshly created .fly.dev hostname can take a moment to become resolvable. But every dial made afterward, for the rest of the build, used a single `tls.Dial` to the same hostname.

Extract `dialBuilderWithDNSRetry` with the same backoff parameters as the existing check, and use it in both dialers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,46 +1,14 @@
 name: Build
 on:
-  schedule:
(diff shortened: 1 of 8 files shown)


NullHypothesis:
Closing because I was chasing a red herring.

PR #5147, opened 2026-08-31
Make Requested changes by Upstash on "fly redis update" 

KTanAug21:
### Change Summary

**What and Why:**
Upstash requested changes to be made last Aug 28 2026, see [reference](https://app.plain.com/workspace/w_01J30S94JDD57QXMT9Q47J928E/thread/th_01KZY1WJNJB9PTB34V10KZ0DZJ/)

**How:**
**1. User should be able to enable/disable prod pack without changing the plan**
- [x] Remove triggering [validateProdPackPlanChange ](https://github.com/superfly/flyctl/blob/a70abbb7b7dade47bb7c8180fc1a7d7cf71db3f9/internal/command/redis/update.go#L200-L202) 
- [X]  [Remove the test](https://github.com/superfly/flyctl/blob/a70abbb7b7dade47bb7c8180fc1a7d7cf71db3f9/internal/command/redis/update_test.go#L174-L197) that validates prevention of prodPack change when no plan change is made
- [ ] *Still needs an update from web to remove server validation!*

**2. Auto upgrade option should have an explanation:**
- [X] Updated the [prompt](https://github.com/superfly/flyctl/blob/a70abbb7b7dade47bb7c8180fc1a7d7cf71db3f9/internal/command/redis/update.go#L163) to contain the [added explanation](https://github.com/superfly/flyctl/blob/132fa780b38bd151b700a221e62ede96d17cb8c3/internal/command/redis/update.go#L163)

**3. After update is complete, it would be better to show which plan the db is on, and a confirmation that prod pack is enabled/disabled**
- [X] Displayed the [selected plan and prodPack status](https://github.com/superfly/flyctl/blob/132fa780b38bd151b700a221e62ede96d17cb8c3/internal/command/redis/update.go#L226-L232) ( selected or retained if no change ) with the assumption that the update is successful if no error is triggered for the update 

**6. Show question to enable prodpack if no flag passed**
- [X] Prompt the user [about enabling prodpack](https://github.com/superfly/flyctl/blob/132fa780b38bd151b700a221e62ede96d17cb8c3/internal/command/redis/update.go#L195-L197) if: no --enable-prodpack/disable-prodpack flag was passed or it is disabled [here](https://github.com/superfly/flyctl/blob/132fa780b38bd151b700a221e62ede96d17cb8c3/internal/command/redis/update.go#L280-L292)

Related to:
[...]

PR #5150, opened 2026-09-01
Polish the ProdPack prompt and post-update summary (on top of #5147)

yphus:
Supersedes #5147, which @KTanAug21 closed in favour of this one. Her commit is preserved here, so this PR carries the whole change: her work plus the fixes below.

Same six Upstash requests, same two confirm closures, same agreed semantics: `--enable-prodpack` / `--disable-prodpack` take precedence and a question is only asked when neither flag is passed (confirmed with Sertug).

### The panic the manual testing couldn't reach

`create.go` only writes `options["prod_pack"]` when ProdPack is enabled, so a database that never had it has no key at all — and the manual runs on #5147 were against a database that had been toggled, so the key was always present. On a database that never had ProdPack, `options["prod_pack"].(bool)` panics with `interface conversion: interface {} is nil, not bool`.

### Declining leaves ProdPack alone

On both branches now, answering "no" means "don't touch it" rather than "set the opposite" — so a stale stored copy can't flip a setting the user never asked to change, which is the failure mode #5086 was written to close.

### Summary output

Kept as Upstash asked — the summary always names a state, `enabled` or `disabled`, and falls back to the stored value when the user made no decision (read safely, and read before `stripProdPack` removes the key). It drops the old `ProdPack: unchanged (use --enable-prodpack ...)` line, which otherwise printed *above* a possibly contradictory `ProdPack: false`.

Worth knowing when reading that fallback: #5086 notes the stored copy can be stale in either direction, so in the no-decision case the line reports last-known state, not verified state. 

The server side allowing to enable/disable prodpack w/o a plan change is here: https://github.com/superfly/web/pull/4369

### DIFF
--- a/gql/schema.graphql
+++ b/gql/schema.graphql
@@ -10149,7 +10149,7 @@ input UpdateAddOnInput {
   planId: ID
 
   """
-  Explicit ProdPack intent for Upstash Redis. Requires a plan change; when absent, ProdPack is left untouched.
+  Explicit ProdPack intent for Upstash Redis. Can be sent with or without a plan change; when absent, ProdPack is left untouched.
   """
   prodPack: Boolean
 
--- a/internal/command/redis/create.go
+++ b/internal/command/redis/create.go
@@ -25,6 +25,13 @@ const (
 	redisPlanPayAsYouGo = "ekQ85Yjkw155ohQ5ALYq0M"
 )
 
+// Descriptions shown before the corresponding prompt, shared by create and
+// update so both paths explain a setting the same way.
+const (
+	autoUpgradeDescription = "\nThe Auto Upgrade feature automatically upgrades your database to the next higher plan when you reach your bandwidth or storage limits, ensuring uninterrupted service.\nThis setting can be changed later.\nFor more information, see https://fly.io/docs/upstash/redis/#auto-upgrade.\n\n"
+	prodPackDescription    = "\nProdPack adds enhanced features for production workloads at $200/mo.\nThis setting can be changed later.\nFor more information, see https://fly.io/docs/upstash/redis/#prod-pack-200-mo.\n\n"
+)
+
 // Legacy plans that are no longer available for new databases
 // but existing databases can remain on them
 // These match the normalized display names (lowercase, spaces replaced with underscores)
@@ -164,7 +171,7 @@ func runCreate(ctx context.Context) (err error) {
 		if flag.IsSpecified(ctx, "enable-auto-upgrade") {
 			enableAutoUpgrade = flag.GetBool(ctx, "enable-auto-upgrade")
 		} else {
-			fmt.Fprintf(io.Out, "\nAuto-upgrade automatically switches to a higher plan when you hit resource limits.\nThis setting can be changed later.\n\n")
+			fmt.Fprint(io.Out, autoUpgradeDescription)
 			enableAutoUpgrade, err = prompt.Confirm(ctx, "Would you like to enable auto-upgrade?")
 			if err != nil {
 				return
@@ -179,7 +186,7 @@ func runCreate(ctx context.Context) (err error) {
 	if flag.IsSpecified(ctx, "enable-prodpack") {
 		enableProdpack = flag.GetBool(ctx, "enable-prodpack")
 	} else {
-		fmt.Fprintf(io.Out, "\nProdPack adds enhanced features for production workloads at $200/mo.\nThis setting can be changed later.\nFor more information, see https://fly.io/docs/upstash/redis/#prod-pack-200-mo.\n\n")
+		fmt.Fprint(io.Out, prodPackDescription)
 		enableProdpack, err = prompt.Confirm(ctx, "Would you like to enable ProdPack?")
 		if err != nil {
 			return
--- a/internal/command/redis/update.go
+++ b/internal/command/redis/update.go
@@ -208,6 +208,10 @@ func runUpdate(ctx context.Context) (err error) {
 		return
 	}
 
+	// Read the stored value before it is stripped: it is the only state we
+	// have to fall back on for the summary when the user made no decision.
+	storedProdPack, _ := options["prod_pack"].(bool)
+
 	stripProdPack(options)
 
 	if currentPlanIsLegacy && selectedPlanIsLegacy {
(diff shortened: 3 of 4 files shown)


KTanAug21:
Thank you @yphus !! Please go ahead and merge this when ready, I'll close my PR once done.

Just for clarification though, regarding the bug mentioned above, I actually got that syntax:
```
prevProdPack := options["prod_pack"].(bool)
```

From an existing [line of code](https://github.com/superfly/flyctl/blob/f4bdd9d980ab677c95359eca1d8653994e59f298/internal/command/redis/update.go#L264):
```
if storedOn, _ := stored.(bool); storedOn 
```

`stored` is an attribute passed to [resolveProdPack](https://github.com/superfly/flyctl/blob/f4bdd9d980ab677c95359eca1d8653994e59f298/internal/command/redis/update.go#L238)  function, and retrieved from the `options["prod_pack"]` entry passed to it from runUpdate [here](https://github.com/superfly/flyctl/blob/f4bdd9d980ab677c95359eca1d8653994e59f298/internal/command/redis/update.go#L190) 

I'm wondering if the existing code should be updated too?
```
if storedOn, _ := stored.(bool); storedOn 
```

KTanAug21:
Additionally, for reference, I've  manually tested my [PR changes](https://github.com/superfly/flyctl/pull/5147)  for different case scenarios and so far haven't gotten any errors for it ( regardless if a redis instance had prodPack enabled or not ).

1. No Prodpack, no flags, Enable prodpack:
```
kathrynannetan@Kathryns-MacBook-Air flyctl % ./bin/flyctl.exe redis update fly-lara-app-redis 
? Choose replica regions, or unselect to remove replica regions: 
? Select an Upstash Redis plan Fixed 1GB: $20/mo + $10 per read region
? Would you like to enable eviction? No
? Would you like to enable auto-upgrade? 
   The Auto Upgrade feature automatically upgrades your database to the next higher plan when you reach your
   bandwidth or storage limits, ensuring uninterrupted service.
   https://fly.io/docs/upstash/redis/#auto-upgrade No
? Would you like to enable ProdPack?
   ProdPack adds enhanced features for production workloads at $200/mo.
   This setting can be changed later.
   For more information, see https://fly.io/docs/upstash/redis/#prod-pack-200-mo. Yes
Your Upstash Redis database fly-lara-app-redis was updated.
   Plan: Fixed 1GB
   ProdPack: true
```

2. No Prodpack, no flags, do not enable prodpack:
```
kathrynannetan@Kathryns-MacBook-Air flyctl % ./bin/flyctl.exe redis update fly-lara-app-redis 
? Choose replica regions, or unselect to remove replica regions: 
? Select an Upstash Redis plan Fixed 250MB: $10/mo + $5 per read region
? Would you like to enable eviction? No
? Would you like to enable auto-upgrade? 
   The Auto Upgrade feature automatically upgrades your database to the next higher plan when you reach your
   bandwidth or storage limits, ensuring uninterrupted service.
   https://fly.io/docs/upstash/redis/#auto-upgrade No
? Would you like to enable ProdPack?
   ProdPack adds enhanced features for production workloads at $200/mo.
   This setting can be changed later.
   For more information, see https://fly.io/docs/upstash/redis/#prod-pack-200-mo. No
ProdPack: unchanged (use --enable-prodpack or --disable-prodpack to change it)
Your Upstash Redis database fly-lara-app-redis was updated.
   Plan: Fixed 250MB
   ProdPack: false
```

3. With Prodpack, no flags, do not disable:
```
kathrynannetan@Kathryns-MacBook-Air flyctl % ./bin/flyctl.exe redis update fly-lara-app-redis 
? Choose replica regions, or unselect to remove replica regions: 
? Select an Upstash Redis plan Fixed 250MB: $10/mo + $5 per read region
? Would you like to enable eviction? No
? Would you like to enable auto-upgrade? 
   The Auto Upgrade feature automatically upgrades your database to the next higher plan when you reach your
   bandwidth or storage limits, ensuring uninterrupted service.
   https://fly.io/docs/upstash/redis/#auto-upgrade No
? Would you like to disable ProdPack? No
ProdPack: unchanged (use --enable-prodpack or --disable-prodpack to change it)
Your Upstash Redis database fly-lara-app-redis was updated.
   Plan: Fixed 250MB
   ProdPack: true
```

4. With Prodpack, no flags, disable prodpack:
```
kathrynannetan@Kathryns-MacBook-Air flyctl % ./bin/flyctl.exe redis update fly-lara-app-redis 
? Choose replica regions, or unselect to remove replica regions: 
? Select an Upstash Redis plan Fixed 1GB: $20/mo + $10 per read region
? Would you like to enable eviction? No
? Would you like to enable auto-upgrade? 
   The Auto Upgrade feature automatically upgrades your database to the next higher plan when you reach your
   bandwidth or storage limits, ensuring uninterrupted service.
   https://fly.io/docs/upstash/redis/#auto-upgrade No
? Would you like to disable ProdPack? Yes
Your Upstash Redis database fly-lara-app-redis was updated.
   Plan: Fixed 1GB
   ProdPack: false
```

yphus:
I don't think failed preflight tests are related to this PR

tripledoublev (approved):
LGTM

swilson-fly:
Looks good! Tested with https://github.com/superfly/web/pull/4369 against web-staging, and enabling/disabling prodpack without a plan change works, as does the rest of the interactive updates.

One bit of weirdness using the `fly redis update --enable-prodpack` and `fly redis update --disable-prodpack` flags though. Those seem to skip the prompt just for the prodpack. It still prompts for the replica regions, plan, eviction, upgrade etc:

```
redis update test-prod-pc --enable-prodpack
? Choose replica regions, or unselect to remove replica regions: 
? Select an Upstash Redis plan Pay-as-you-go: $0.2 per 100K commands
? Would you like to enable eviction? Yes

Your Upstash Redis database test-prod-pc was updated.
  Plan:         Pay-as-you-go
  Eviction:     enabled
  Auto-upgrade: not available on this plan
  ProdPack:     enabled
```

It makes it still require an interactive response. I'd recommend seeing if we can treat passing one of those flags as a "leave everything else untouched" and change just the prodpack. Alternatively could add flags for all the other settings as well, letting everything be set non-interactively. 

On `fly redis create` we took the flags for everything approach.

yphus:
Good catch, fixed in 0ea69d3aa — took your first suggestion.

A ProdPack flag now means "change ProdPack, leave the rest alone": no question is asked at all, the current plan and read regions are reused as-is, and the stored options blob goes back untouched. So this runs unattended:

```
fly redis update test-prod-pc --enable-prodpack
```

Any other flag you passed explicitly is still honoured, so `--replica-regions` keeps working alongside it.

Mechanically, the plan prompt and the eviction/auto-upgrade prompts moved into `selectPlan` and `promptOptionChanges` — both are conditional now, and inlining them was making `runUpdate` hard to follow. Their behaviour is unchanged.

yphus:
Validation on staging :

Everything below ran in a **non-interactive shell**, i.e. the CI conditions Sam was worried about, against `https://flyio-web-staging.fly.dev` with **web#4369** deployed.

| Case | Result |
| --- | --- |
| `--disable-prodpack` alone | ✅ No question asked, plan and regions preserved, mutation accepted |
| `--disable-prodpack --replica-regions fra` | ✅ ProdPack flag silences the questions but does not override other explicit flags — Read Regions went `None → fra` |
| No flag | ⚠️ `regions codes must be specified…` — still needs a human, which is the original problem |
| `--enable-prodpack --disable-prodpack` | ✅ Client-side error, no API call |
| Database with no replica | ✅ The `readRegions: null` failure is gone |

### Successful run

```text
Your Upstash Redis database prodpack-test2 was updated.
  Plan:         Pay-as-you-go
  Eviction:     disabled
  Auto-upgrade: not available on this plan
  ProdPack:     disabled

yphus:
still two red tests, unrelated to this PR

PR #5152, opened 2026-09-01
Fix public IP assignment requests.

NullHypothesis:
Flaps now only accepts `org_slug` for `private_v6` assignments. Omit it when provisioning public IPv4 and IPv6 addresses during first deploy, while retaining it for organization-scoped private IPv6 addresses.


lillianberryfly (approved):

internal/command/deploy/deploy_first.go
```diff
@@ -105,11 +91,11 @@ func (md *machineDeployment) provisionIpsOnFirstDeploy(ctx context.Context, ipTy
 
 	case "private":
 		fmt.Fprintf(md.io.Out, "Provisioning ip address for %s\n", md.colorize.Bold(md.app.Name))
+		// Unlike public IPs, private_v6 is organization-scoped and requires
+		// org_slug to select the private network for the allocation.
 		v6Addr, err := md.flapsClient.AssignIP(ctx, md.app.Name, flaps.AssignIPRequest{
 			Type:         "private_v6",
-			Region:       region,
 			Organization: org,
-			Network:      network,
```
lillianberryfly, line 112: network needs to be re-added

lillianberryfly (changes requested):

internal/command/deploy/deploy_first.go
NullHypothesis: The network has always been empty but it's now set: be881564b5d9e4d5c088ec32f00e85213972fb3b

lillianberryfly (approved):

PR #5154, opened 2026-09-01
use machines api for IP assignments (both ingress and egress)

lillianberryfly:


tomr-a5 (approved):
LGTM! Only thing I noticed is that we drop organization and network from private v6 allocation in both the postgres launcher and remote builder. Not sure if they still need that but flyctl-wise this looks good to me!

lillianberryfly:
test failures are caused by GHA, not bad code.

> we drop organization and network from private v6 allocation in both the postgres launcher and remote builder.

those were unused

PR #5157, opened 2026-09-01
Route MPGv2 status through the Machines API

tripledoublev:
## Summary

Route `fly mpg status` for MPGv2 clusters through the public Machines API.

- Query the public cluster endpoint first.
- Fall back to the legacy MPGv2 client only on a classified 404.
- Return non-404 and fallback errors directly; no silent fallback.
- Table labels and `Direct IP` semantics are unchanged.
- `--json` still uses the legacy client, since its response includes
  credentials the public API does not expose. JSON output is unchanged.
- MPGv1, attach, and proxy are unchanged.

## Tests

- `go test -count=1 ./internal/command/mpg/...`
- `go test -count=1 ./internal/command/mpg/v2 -run 'TestRunStatus'`
- `go vet ./internal/command/mpg/...`
- `gofmt`
- `git diff --check`
- `CGO_ENABLED=0 golangci-lint run ./internal/command/mpg/v2/...`

Ran the binary against a MPGv2 cluster in my personal org.

### DIFF
--- a/internal/command/mpg/v2/run_status.go
+++ b/internal/command/mpg/v2/run_status.go
@@ -2,43 +2,89 @@ package cmdv2
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 
-	"github.com/superfly/flyctl/iostreams"
-
+	"github.com/superfly/fly-go/flaps"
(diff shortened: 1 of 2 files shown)


tomr-a5 (approved):
LGTM!

PR #5162, opened 2026-09-02
mpg: migrate attach to the Machines API

tripledoublev:
## Summary

- Make the public Machines API the primary path for `fly mpg attach`.
- Preserve compatibility with existing MPG service behavior when required.
- Reuse the existing Flaps client for secret operations.
- Add coverage for public-path success and compatibility fallback behavior.
- Cover the interactive attach flow with a manual personal-org E2E test.

## Validation

- `go test ./internal/command/mpg/v2/...` and touched packages
- Local `flyctl` build from this worktree
- Manual personal-org attach/detach E2E using a disposable app
- `go vet` on affected packages
- `golangci-lint run ./internal/command/mpg/v2/...`
- `git diff --check`

The base connection URI, default database name, and default cluster credentials still come from the existing MPG service path because the public cluster API does not expose them. Named-user credentials are public-first, with compatibility fallback when required. The interactive E2E was manual; the preflight E2E tests remain separate.


tomr-a5 (approved):
LGTM!

PR #5165, opened 2026-09-02
Split off manifest deployment tests.

NullHypothesis:
CI has been struggling with the preflight "deploy" group, which often exceeds the package test timeout of 15 minutes. We have previously carved out some of these tests to reduce runtime, and this PR carves out yet another test: the manifest deployment test. This will hopefully give the "deploy" group some more breathing room.


NullHypothesis:
I don't think this is the right fix after all; at least not right now. A bunch of other buildkit-related bugs are resulting in the tests taking forever.

PR #5168, opened 2026-09-02
Route MPGv2 create through the Machines API

tripledoublev:
`fly mpg create` for MPGv2 now uses the public Machines API instead of the private MPG client.

- Region discovery uses `mpgutil.AvailableRegions` and public `GetRegions`, selecting non-deprecated MPG-available regions. Explicit `--region` validation, interactive prompting, and existing error messages are unchanged.
- Creation uses `CreateManagedPostgresCluster`. Errors do not fall back because retrying a mutating request against another backend could create a duplicate cluster.
- Readiness polling uses `GetManagedPostgresCluster`; `failed` and `error` are terminal.
- After readiness, `GetManagedPostgresUserCredentials` retrieves `fly-user` credentials. The connection string uses the primary pooler endpoint, `fly-user`, `fly-db`, and a password escaped with `url.UserPassword`. An empty endpoint fails instead of reporting success with a broken URI during propagation lag.
- No fly-go update required; v0.9.15 already has these methods.

## Test plan

- Unit coverage: explicit, interactive, empty, and invalid region discovery; request mapping; terminal statuses; polling errors; missing IDs; credential failures; unavailable endpoints; reserved-character passwords.
- `go build`, `go vet`, `gofmt -l`, `git diff --check` pass.
- Live test: this binary created a cluster in a personal org and returned a working connection string. A sibling `connect`/`proxy` migration binary connected and wrote data. Valid, invalid, and interactive region selection were tested against the live API.

### DIFF
--- a/internal/command/mpg/v2/run_create_test.go
+++ b/internal/command/mpg/v2/run_create_test.go
@@ -1,0 +1,589 @@
+package cmdv2
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/require"
+	fly "github.com/superfly/fly-go"
+	"github.com/superfly/fly-go/flaps"
+	"github.com/superfly/flyctl/internal/flag/flagctx"
+	"github.com/superfly/flyctl/internal/flapsutil"
+	"github.com/superfly/flyctl/internal/mock"
+	"github.com/superfly/flyctl/internal/prompt"
+	mpgv2 "github.com/superfly/flyctl/internal/uiex/mpg/v2"
+	"github.com/superfly/flyctl/iostreams"
+)
+
+// createTestContext builds a context carrying a --region flag and an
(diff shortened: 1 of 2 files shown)


jwnx (approved):
lgtm (:

internal/command/mpg/v2/run_create_test.go
```diff
@@ -0,0 +1,589 @@
...
+
+// createTestContext builds a context carrying a --region flag and an
+// iostream suitable for the create command's expectations. If region is
+// the empty string, --region is left unset (so the command takes the
+// interactive-prompt branch instead).
+func createTestContext(t *testing.T, region string) (context.Context, *bytes.Buffer) {
+	t.Helper()
+
+	io, _, stdout, _ := iostreams.Test()
+	ctx := iostreams.NewContext(context.Background(), io)
```
NullHypothesis, line 31: FYI, Go now has `t.Context()`, so there's no longer a need to use `context.Background()`.

NullHypothesis (approved):

PR #5048, opened 2026-08-05
fix(mpg): decouple proxy readiness from credentials

gargakshit:
### Change Summary

What and Why:

Allow `fly mpg proxy` to start when a cluster has a direct IP but database credentials are unavailable or still initializing. The raw network proxy does not require database credentials, while `fly mpg connect` still does.

How:

- Separate cluster retrieval, credential resolution, and proxy parameter construction for MPGv1 and MPGv2.
- Keep credential validation and explicit-user handling unchanged for `fly mpg connect`.
- Make `fly mpg proxy` depend on the cluster direct IP rather than credential readiness.
- Preserve missing-IP errors, WireGuard setup, bind/port configuration, and organization scoping.

---

### Documentation

- [ ] Fresh Produce
- [ ] In superfly/docs, or asked for help from docs team
- [x] n/a

PR #5169, opened 2026-09-02
mpg: route connect and proxy through the public Machines API

tripledoublev:
**Stacked on #5048**. 

This requires the corresponding server change that preserves assigned endpoints during resize and reconfiguration.

## Summary

- Use the public Machines API first for v1 and v2 connect/proxy cluster lookup. Fall back to the legacy client only on a classified 404; preserve all other public API errors. Interactive user and database selection remain on the existing private routes.
- Resolve connect credentials through the public user-credentials endpoint. The default public user and database remain `fly-user` and `fly-db`, while explicit usernames pass through unchanged.
- Preserve the behavior of the previous private ui-ex routes: connect depends on credential availability rather than cluster lifecycle status, and retains the existing non-ready warning after connection parameters resolve.
- Preserve the historical `cluster is still initializing, wait a bit more` result when the public default-user credentials endpoint returns a classified 404. Explicit-user 404s and other credential failures remain authoritative errors and never fall back to the private API.
- Keep flyctl's proxy path independent from credentials and lifecycle status. Endpoint availability remains authoritative: proxy proceeds when the server returns a direct endpoint and preserves the historical missing-IP error otherwise. The public path uses the advertised DNS host and port; legacy fallback uses the assigned IPv6 address on port 5432.

Public endpoint hosts are DNS names rather than the private route's bare IPv6 assignments. Proxy therefore uses its existing bounded DNS wait before listening locally when a hostname has not propagated yet.

## Validation

- `go test -count=1 ./internal/command/mpg/... ./internal/mpgutil/...`
- `go vet ./internal/command/mpg/...`
- `go build .`
- `gofmt` and `git diff --check`
- Tests cover public success, wrapped 404 fallback, authoritative non-404 errors, default and explicit credential routing, credential error ordering, endpoint host/port validation, legacy compatibility, and proxy independence from credentials for both v1 and v2.
- Staging resize validation confirmed that proxy starts through WireGuard while the cluster is non-ready when its direct endpoint is retained. Connect preserved the historical empty-password error when staging returned credentials without a password.

## Follow-up

More actionable transient-state errors should be handled separately after defining an authoritative public API contract for credential and endpoint readiness. This PR intentionally avoids inferring credential readiness from raw cluster lifecycle states.

### DIFF
--- a/internal/command/mpg/v1/run_proxy.go
+++ b/internal/command/mpg/v1/run_proxy.go
@@ -12,19 +12,11 @@ import (
 	"github.com/superfly/flyctl/internal/flag"
 	"github.com/superfly/flyctl/internal/flapsutil"
 	"github.com/superfly/flyctl/internal/flyutil"
+	"github.com/superfly/flyctl/internal/mpgutil"
 	"github.com/superfly/flyctl/internal/uiex/mpg"
 	mpgv1 "github.com/superfly/flyctl/internal/uiex/mpg/v1"
 	"github.com/superfly/flyctl/proxy"
 )
-
-// defaultMPGUser is the literal username used when resolving the default
-// fly-user on the public Machines API for Connect. Verified against the
-// orchestrator and ui-ex repos.
-const defaultMPGUser = "fly-user"
-
-// defaultMPGDatabase is the literal database name used when no --database
-// flag is given and no interactive prompt answer is available.
-const defaultMPGDatabase = "fly-db"
 
 func RunProxy(ctx context.Context, clusterID string, resolvedOrgSlug string, proxyPort string) error {
 	_, params, err := GetMpgProxyParams(ctx, proxyPort, clusterID, resolvedOrgSlug)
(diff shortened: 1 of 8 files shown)


internal/command/mpg/v1/run_proxy.go
```diff
@@ -2,15 +2,30 @@ package cmdv1
...
 	"github.com/superfly/flyctl/internal/flyutil"
+	"github.com/superfly/flyctl/internal/uiex/mpg"
 	mpgv1 "github.com/superfly/flyctl/internal/uiex/mpg/v1"
 	"github.com/superfly/flyctl/proxy"
 )
 
+// defaultMPGUser is the literal username used when resolving the default
+// fly-user on the public Machines API for Connect. Verified against the
+// orchestrator and ui-ex repos.
+const defaultMPGUser = "fly-user"
```
jwnx: jfyi: `fly-user` can be deleted in MPGv2, good thing we allow `--user`

internal/command/mpg/v1/run_proxy.go
```diff
@@ -97,32 +207,159 @@ func resolveConnectCredentials(
...
+			DBName:   defaultMPGDatabase,
+		}
+	case useLegacy:
 		credentials = response.Credentials
-	}
+	default:
+		flapsClient := flapsutil.ClientFromContext(ctx)
+		userCreds, err := flapsClient.GetManagedPostgresUserCredentials(ctx, response.Data.Id, defaultMPGUser)
+		if err != nil {
+			return nil, fmt.Errorf("failed retrieving credentials for user %s: %w", defaultMPGUser, err)
```
jwnx: suggestion: maybe help the user here by saying "you can use --user to pick a different user name" or something?

jwnx (approved):
lgtm

PR #5172, opened 2026-09-03
Run Buildpack preflight test using local Docker.

NullHypothesis:
Preflight tests share an organization-wide remote builder app. This causes a conflict for the `TestDeploy` preflight test: The "Buildpack" test wants a remote Docker/RCHAB builder whereas the "Dockerfile" test wants a BuildKit builder. Both use the "remote-docker-builder" app role despite exposing incompatible APIs and that has been causing problems when both tests run in parallel.

To fix this, use `--local-only` for the Buildpack test. This should fix the builder app race but it comes at the cost of losing some test coverage. We need to re-architect preflight tests to safely test these two different build paths.


NullHypothesis:
@mattp-fly, would you mind taking a look at this? Git history tells me that you're quite familiar with that code :)

NullHypothesis:
> This should fix the builder app race but it comes at the cost of losing some test coverage.

We can't afford to have flyctl's CI be this brittle. Some lost test coverage is a fair price to pay for a more robust CI. I'm gonna go ahead and merge this.

PR #5192, opened 2026-09-08
fix(mcp): write MCP client configuration at 0600

mjbraun:
fly mcp add with --stream or --sse records the Fly API token in the client configuration as a --bearer-token argument, and the file was written world-readable. Write it owner-only, matching flyctl's own config file, and apply the mode on rewrite as well since os.WriteFile only sets it on creation.


Lucais11 (approved):
Looks right and seems consistent with the `configureDockerJSON` pattern we already use. I also double checked that `Chmod` is load-bearing by dropping it, which then fails two of the new tests.

PR #5201, opened 2026-09-17
mpg status: show used and provisioned storage

miguel-fly:
### Issue

https://github.com/superfly/ui-ex/issues/5567

### Problem

- Status only shows allocated disk, the per-node size set at creation, so there is no way to see the storage the bill is based on

### Solution

- Show `Used Storage (GB)` and `Allocated Storage (GB)`, cluster-wide totals at one decimal
- The legacy postgresv2 fallback reads the same fields, so both paths report the same figures
- Rename the create summary's `Disk:` to `Volume Size:` and clarify the `--volume-size` help

### Notes

- Needs the fly-go release from https://github.com/superfly/fly-go/pull/307 before merging
- `--json` picks the fields up automatically through the legacy envelope
- Rows render blank when a server doesn't return the fields

### Related PRs

- https://github.com/superfly/fly-go/pull/307
- https://github.com/superfly/ui-ex/pull/5570
- https://github.com/superfly/ui-ex/pull/5571

### Screenshots
mpg create
Screenshot 2026-09-16 at 7 48 24 PM

mpg status
Screenshot 2026-09-16 at 7 52 38 PM


internal/command/mpg/v2/run_status.go
```diff
@@ -108,3 +113,13 @@ func runStatusLegacy(ctx context.Context, clusterID string) error {
 
 	return render.JSON(out, clusterDetails)
 }
+
+// One decimal, matching the dashboard's format_storage/1, so the same cluster
+// does not read 61.0 GB there and 61 here.
+func gbString(bytes *int64) string {
+	if bytes == nil {
+		return ""
```
tripledoublev: dashboard renders `N/A` here (`format_storage(nil)` on the ui-ex side), could we match that?

internal/command/mpg/v2/run_status.go
```diff
@@ -90,7 +94,8 @@ func runStatusHuman(ctx context.Context, clusterID string) error {
 		"Organization",
 		"Region",
 		"Status",
-		"Allocated Disk (GB)",
+		"Used Storage (GB)",
+		"Allocated Storage (GB)",
```
tripledoublev: I'd recommend adding `Total` on these two to mirror the dashboard, so it doesn't read as an allocation increase since both are summed across all volumes

tripledoublev (changes requested):
2 small requests so it more closely aligns with the web dashboard

internal/command/mpg/v2/run_status.go
miguel-fly: I've updated the code to match the dashboard 👌

internal/command/mpg/v2/run_status.go
miguel-fly: Good point! I've updated the code 👌

miguel-fly: Screenshot 2026-09-23 at 11 26 22 AM

internal/command/mpg/v2/run_status.go
miguel-fly: Screenshot 2026-09-23 at 11 26 10 AM

tripledoublev (approved):
Looks good. Thanks @miguel-fly!

